PR #168, opened 2014-11-23
Add programmatic introspection and discoverability for `mallctl`s

cemeyer:
In the spirit of FreeBSD sysctls, it is useful to have at least somewhat reliable programmatic access to the intended types of these sorts of tunables. To that end, this patch series:

* Adds type information to most `mallctl` nodes / oids
* Adds a special node to enumerate nodes (`introspect.next`) (like `sysctl.next` or magic MIB 0.2 on FreeBSD)
* Adds a special node to return type information (`introspect.kind`) (like `sysctl.oidfmt` or magic MIB 0.4)
* Adds a special node to return the name for a given MIB (`introspect.name`) (like `sysctl.name` or magic MIB 0.1)

Finally, I added simple program to explore the default configurations, `mallctl(1)`. [Here's a gist of some example output.](https://gist.github.com/cemeyer/dbc39bc7cef06fc7ba4a)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /*.gcov.*
 
 /bin/jemalloc.sh
+/bin/mallctl
 
 /config.stamp
 /config.log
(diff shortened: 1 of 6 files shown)


jasone:
I've come back to this patch set numerous times, and although it's cool, I still don't quite think the feature is quite worth the maintenance overhead.

Issue #163, opened 2014-11-05
Expose the amount of memory used for bookkeeping as a statistic


guilherme-pg:
As [discussed in the mailing list](http://www.canonware.com/pipermail/jemalloc-discuss/2014-November/000962.html), we need to report the memory used for jemalloc's internal data structures in Gecko.

jasone:
I've spent several hours over the past couple of days adapting your patch and addressing some of the FIXME issues.  However, it's proving to be much more involved than I expected.  The short summary is that a single statistic doesn't provide enough information, because it doesn't allow us to determine how much of the metadata are also counted as ```allocated``` and ```active```.  There are three distinct categories of metadata:

- Touched base memory.  This is easy to track.
- Per arena metadata pages (i.e. arena chunk headers).  The total number of metadata pages is straightforward to track, but this number doesn't directly translate to physical memory usage, because arena chunk headers are designed to support demand-zeroed pages, and jemalloc does not track which arena chunk header pages have actually been touched.
- Per arena internal-use allocations, which are also counted as ```allocated``` and contribute to ```active```.  The ```a0*()``` API is only one way for these to be created.  The prof code uses ```imalloc()```, and a general tracking solution would require passing a flag from ```imalloc()```-level functions all the way down to the guts of the arena code.  This is a pretty heavy impact, and one that I'm not very excited about.

As the actual complexity of this feature becomes apparent, I worry that its usefulness doesn't quite justify the cost.  I toyed with some other refactoring that might improve the situation (e.g. using a separate arena for internal allocation), but everything I've thought of so far has significant scalability and/or maintainability issues.

Thoughts?

guilherme-pg:
Thanks for working on this! I believe we pretty much [reached the same conclusions](https://bugzilla.mozilla.org/show_bug.cgi?id=899126#c12) as you did some time ago: it's hard to account for absolutely everything, but for parity with mozjemalloc, all we actually care about are the first two bullet points you mentioned.

So we ended up [importing a patch](https://dxr.mozilla.org/mozilla-central/source/memory/jemalloc/0004-Implement-stats.bookkeeping.patch) that is much smaller in scope, but still suits our use case. Would you be willing to merge it (or something like it), perhaps?

jasone:
I started writing up why I was going to put this off, and realized there was no fundamental hurdle to completion, and just finished it instead.  The new API is documented in the commit message and in the manual.  Please let me know if you see any problems with the approach.

PR #169, opened 2014-12-03
Implement stats.bookkeeping.

guilherme-pg:
This is a first and crude attempt at issue #163, but hopefully it's on the right track.

This attempts to account for:
    - All base allocations, including arena headers;
    - All chunk headers;
    - Extent nodes used for huge allocations;
    - All allocations performed via `a0{m,c}alloc`.

This leaves missing:
    - Allocations of `tcache_t`;
    - Allocations of `quarantine_t`;
    - Allocations made for heap profiling.

Thoughts?

### DIFF
--- a/include/jemalloc/internal/base.h
+++ b/include/jemalloc/internal/base.h
@@ -9,6 +9,10 @@
 /******************************************************************************/
 #ifdef JEMALLOC_H_EXTERNS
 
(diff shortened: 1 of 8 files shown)


jasone:
Obsoleted by 4581b97809e7e545c38b996870a4e7284a620bc5.

PR #179, opened 2014-12-18
Add a isblank definition for MSVC < 2013

glandium:
MSVC only got isblank in MSVC2013: http://msdn.microsoft.com/en-us/library/dn237255.aspx


jasone:
Committed as 51f86346c000aa2a44abaab08caeedcb151e6556.  Thanks!

PR #180, opened 2014-12-18
Make mixed declarations an error

glandium:
It often happens that code changes introduce mixed declarations, that then
break building with Visual Studio. Since the code style is to not use
mixed declarations anyways, we might as well enforce it with -Werror.


jasone:
Thanks!

PR #181, opened 2014-12-31
prevent allocations larger than PTRDIFF_MAX

thestinger:
In practice, this only has an impact on 32-bit processes in environments
where it's possible to allocate over half of the address space. It wipes
out a few subtle classes of undefined behavior from pointer arithmetic
without a performance hit.

The musl C library has adopted this solution to the problem already. The
alternative is pervasively checking the size against PTRDIFF_MAX in the
code asking the allocator for buffers. It seems like the burden should
be on the allocator instead because it can be solved in one place.

For example, the following C program compiled as i386 will succeed on
x86_64, but is not actually well-defined:

    #include <stdio.h>
    #include <stdlib.h>

    const size_t size = 3000000000U;

    int main(void) {
        char *ptr = malloc(size);
        if (!ptr) return 1;
        char *end = ptr + size;
        printf("%p\n", ptr);
        printf("%p\n", end);
        printf("%p\n", end - ptr);
        return 0;
    }

The C standard is not entirely clear if the `ptr + size` operation here
is undefined. It is clear that `end - ptr` is undefined because it
causes a ptrdiff_t overflow which is explicily UB.

LLVM interprets the standard as defining all pointer arithmetic in terms
of ptrdiff_t so both operations cause undefined overflows. The LLVM IR
(clang -S -emit-llvm) is clearly undefined, and the understanding of the
LLVM developers is that this is correct per the standard:

    %8 = getelementptr inbounds i8* %7, i32 -1294967296

http://llvm.org/releases/3.5.0/docs/LangRef.html#id181

Since jemalloc moved to `(uintptr_t)ptr + size`, it does not appear to
hit this issue itself. It would hit it with `(char *)ptr + size` though.

### DIFF
--- a/src/huge.c
+++ b/src/huge.c
@@ -34,6 +34,10 @@ huge_palloc(tsd_t *tsd, arena_t *arena, size_t usize, size_t alignment,
 	extent_node_t *node;
 	bool is_zeroed;
 
+	/* Prevent allocations where pointer arithmetic can be undefined. */
+	if (usize > PTRDIFF_MAX)
+		return (NULL);
+
 	/* Allocate one or more contiguous chunks for this request. */
 
 	/* Allocate an extent node with which to track the chunk. */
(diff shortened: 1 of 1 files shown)


jasone:
It's unclear to me that the allocator should explicitly limit functionality to potentially protect malformed programs from undefined behavior.  If ```PTRDIFF_MAX``` were an explicitly specified upper limit on allocation size, this change would clearly be correct, but it doesn't sound like we're that fortunate.

thestinger:
My motivation for fixing this was to implement an efficient way to fix https://github.com/rust-lang/rust/issues/18726. It would be nice to have support for something like this behind a configure flag if it's not acceptable as a default. However, I'm no longer interested in working on Rust so it will need to be someone else who figures it out.

PR #185, opened 2015-01-21
Add missing symbols to private_symbols.txt.

adk9:
Some symbols, perhaps inadvertently, were left out of private_symbols.txt. We statically link against two instances of jemalloc (a prefixed one, and a non-prefixed one that acts as malloc replacement), and run into collision issues with these symbols. Applying this patch fixes the problem.

### DIFF
--- a/include/jemalloc/internal/private_symbols.txt
+++ b/include/jemalloc/internal/private_symbols.txt
@@ -20,6 +20,7 @@ arena_cleanup
 arena_dalloc
 arena_dalloc_bin
 arena_dalloc_bin_junked_locked
+arena_dalloc_bin_junked_locked_impl
 arena_dalloc_junk_large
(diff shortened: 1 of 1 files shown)


jasone:
I try to keep ```private_symbols.txt``` current, and I do a sweep before releases to add missing entries, but there doesn't seem to be a lightweight way to automate this and catch all symbols, so it's a constant battle.  Thanks for the patch; I'll merge it momentarily.

jasone:
What version of jemalloc is this for?  It looks like some parts are relevant to the most recent dev branch, but others are perhaps related to the master branch.

jasone:
Actually, ```arena_dalloc_bin_junked_locked_impl``` was the only extra symbol in your patch, and I'm guessing that dates from the dev branch from the second half of 2014.

PR #187, opened 2015-01-25
huge_node_locked don't have to unlock huge_mtx

semarie:
in src/huge.c, after each call of huge_node_locked(), huge_mtx is
already unlocked. don't unlock it twice (it is a undefined behaviour).

under openbsd, the undefined behaviour is to call `abort()`.


jasone:
Thanks!

PR #188, opened 2015-01-25
add openbsd support

semarie:
this patch add openbsd support to jemalloc.
tested under openbsd5.7 (current developpement version).
all tests pass (with patch from PR #187).

please note that some previous versions of jemalloc need `force_tls="0"` (version currently used by rustc in particular).


jasone:
Integrated as 77d597ebb23aa47a4a0112c294ad6a68857f450c; thanks!

PR #192, opened 2015-02-03
Make opt.lg_dirty_mult work as documented

glandium:
The documentation for opt.lg_dirty_mult says:
    Per-arena minimum ratio (log base 2) of active to dirty
    pages.  Some dirty unused pages may be allowed to accumulate,
    within the limit set by the ratio (or one chunk worth of dirty
    pages, whichever is greater) (...)

The restriction in parentheses currently doesn't happen. This makes
jemalloc aggressively madvise(), which in turns increases the amount
of page faults significantly.

For instance, this resulted in several(!) hundred(!) milliseconds
startup regression on Firefox for Android.

This may require further tweaking, but starting with actually doing
what the documentation says is a good start.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -850,6 +850,7 @@ arena_maybe_purge(arena_t *arena)
 	if (opt_lg_dirty_mult < 0)
 		return;
 	threshold = (arena->nactive >> opt_lg_dirty_mult);
+	threshold = threshold < chunk_npages ? chunk_npages : threshold;
 	/*
 	 * Don't purge unless the number of purgeable pages exceeds the
 	 * threshold.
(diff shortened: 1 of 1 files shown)


jasone:
See http://www.canonware.com/pipermail/jemalloc-discuss/2015-February/001013.html for related discussion.

jasone:
Thanks!

toddlipcon:
Was there ever any progress with a time-based hysteresis here? I'm trying out jemalloc and some of our micro-benchmarks have big regressions due to increased page faults. Seting lg_dirty_mult=-1 of course works around the issues in micro-benchmarks, but a time based heuristic would probably be more generally applicable.

jasone:
@toddlipcon, yes, jemalloc 4.1.0 implements [decay-based purging](https://github.com/jemalloc/jemalloc/issues/325).  It's currently opt-in, but you can universally enable it by specifying ```--with-malloc-conf=purge:decay``` during configuration.  See the [ChangeLog](https://github.com/jemalloc/jemalloc/blob/4.1.0/ChangeLog#L18) and [manual](http://canonware.com/download/jemalloc/jemalloc-latest/doc/jemalloc.html#opt.purge) for more information.

toddlipcon:
Thanks, I'll give that a try. I am on the latest release but didn't opt-in. Will report back on the mailing list if the page fault rate doesn't improve.

PR #194, opened 2015-02-10
adding support for Bitrig

dhuseby:
This patch adds support for building jemalloc on Bitrig (https://bitrig.org)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -283,6 +283,11 @@ case "${host}" in
 	abi="elf"
 	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
 	;;
+  *-*-bitrig*)
+	CFLAGS="$CFLAGS"
+	abi="elf"
+	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -283,6 +283,11 @@ case "${host}" in
 	abi="elf"
 	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
 	;;
+  *-*-bitrig*)
+	CFLAGS="$CFLAGS"
+	abi="elf"
+	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
+	;;
   *-*-openbsd*)
```
annulen: I think you should join bitrig case with openbsd case to avoid both code duplication and readers' confusion about what is "bitrig"
```
*-*-openbsd*|*-*-bitrig*)
  ...

dhuseby:
@annulen I incorporated your feedback and merged the bitrig and openbsd cases.

jasone:
Integrated as 970fcfbca5dffda921394172c7298d274eebfd0e.  Thanks!

PR #196, opened 2015-02-11
Take into account the install suffix that jemalloc was built with in the...

adk9:
... pkg-config file.

If jemalloc is installed with an install-suffix, pkg-install fails to find it. Fix the pkg-config file to find the library correctly.

Signed-off-by: Abhishek Kulkarni <adkulkar@umail.iu.edu>


jasone:
Integrated as feaaa3df0da9972b9c5016c55b886e54853cc855.  Thanks!

PR #197, opened 2015-02-12
add missing check for new_addr chunk size

thestinger:
8ddc93293cd8370870f221225ef1e013fbff6d65 switched this to over using the
address tree in order to avoid false negatives, so it now needs to check
that the size of the free extent is large enough to satisfy the request.


jasone:
Thanks!

PR #198, opened 2015-03-04
Preserve LastError when calling TlsGetValue

glandium:
TlsGetValue has a semantic difference with pthread_getspecific, in that it
can return a non-error NULL value, so it always sets the LastError.
But allocator callers may not be expecting calling e.g. free() to change
the value of the last error, so preserve it.


jasone:
Integrated as 4d871f73af6b8310564dfcb63357dbfe8b1a1529.  Thanks!

PR #205, opened 2015-03-12
Use the error code given to buferror on Windows

glandium:
a14bce85 made buferror not take an error code, and make the Windows
code path for buferror use GetLastError, while the alternative code
paths used errno. Then 2a83ed02 made buferror take an error code
again, and while it changed the non-Windows code paths to use that
error code, the Windows code path was not changed accordingly.


jasone:
Integrated as f69e2f6fdab40c7612be5fd69960b8c7d40dba44.

PR #206, opened 2015-03-12
Make pages_purge reduce the resident set size on Windows and OSX

glandium:
On long lived processes which do many allocations and deallocations, it is
customary to end up in a situation where the amount of allocated memory is
rather low, but the number of chunks in use is rather high.

In that case, on Windows and OSX, the memory that once was used is still
committed by the OS and the resident set size ends up being closer to the
size of the chunks allocated than to the remaining allocations, creating the
impression that the process is using more memory than it actually is.

A few notes:
- This kind of mimics what we end up doing on Linux with MADV_DONTNEED.
- I don't know what MADV_FREE actually does on FreeBSD and other OSes it's used on. AFAIK, OSX doesn't use that part of FreeBSD, and delegates memory management to mach, so the behavior on OSX is possibly different.
- AFAICT from Firefox benchmarks, this doesn't seem to affect performance on either OSX or Windows, but I don't have results for the full range of benchmarks on Windows yet.
- I can understand that not all applications using jemalloc would want to pay the price for these purges. Maybe this is something that would need a pref?
- In mozjemalloc, the way we currently handle this is actually more convoluted:
  - On Windows, we MEM_DECOMMIT, but only MEM_COMMIT when the pages are going to be used again. This is actually something that would be nice to do in jemalloc3 as well, because AIUI, as long as some address space is MEM_COMMITted, its size is reserved in the page file, and it would be better to keep that low despite the long-term fragmentation. Thoughts?
  - On OSX, we use MADV_FREE, keep track of the madvised pages, and actively purge them afterwards via an explicit call to a jemalloc_purge_freed_pages function. Considering the lack of performance impact, I don't think we need something like this in jemalloc3 (although, we still do have a jemalloc_purge_freed_pages compatibility function in Firefox that uses mallctl to trigger arena.*.purge)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -262,7 +262,7 @@ case "${host}" in
   *-*-darwin* | *-*-ios*)
 	CFLAGS="$CFLAGS"
 	abi="macho"
-	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
+	AC_DEFINE([JEMALLOC_PURGE_USE_MMAP], [ ])
 	RPATH=""
 	LD_PRELOAD_VAR="DYLD_INSERT_LIBRARIES"
 	so="dylib"
--- a/include/jemalloc/internal/jemalloc_internal_defs.h.in
+++ b/include/jemalloc/internal/jemalloc_internal_defs.h.in
@@ -198,12 +198,14 @@
  *   madvise(..., MADV_DONTNEED) : On Linux, this immediately discards pages,
  *                                 such that new pages will be demand-zeroed if
  *                                 the address region is later touched.
- *   madvise(..., MADV_FREE) : On FreeBSD and Darwin, this marks pages as being
- *                             unused, such that they will be discarded rather
- *                             than swapped out.
+ *   madvise(..., MADV_FREE) : On FreeBSD, this marks pages as being unused,
+ *                             such that they will be discarded rather than
+ *                             swapped out.
+ *   mmap(..., MAP_FIXED | ...) : On Darwin, this immediately discards pages.
  */
 #undef JEMALLOC_PURGE_MADVISE_DONTNEED
 #undef JEMALLOC_PURGE_MADVISE_FREE
+#undef JEMALLOC_PURGE_USE_MMAP
 
 /* Define if operating system has alloca.h header. */
 #undef JEMALLOC_HAS_ALLOCA_H
--- a/src/chunk_mmap.c
+++ b/src/chunk_mmap.c
@@ -119,8 +119,14 @@ pages_purge(void *addr, size_t length)
 	bool unzeroed;
 
 #ifdef _WIN32
-	VirtualAlloc(addr, length, MEM_RESET, PAGE_READWRITE);
-	unzeroed = true;
+	VirtualFree(addr, length, MEM_DECOMMIT);
+	VirtualAlloc(addr, length, MEM_COMMIT, PAGE_READWRITE);
+	unzeroed = false;
+#elif defined(JEMALLOC_PURGE_USE_MMAP)
+	void *new_addr = mmap(addr, length, PROT_READ | PROT_WRITE,
+	                      MAP_FIXED | MAP_PRIVATE | MAP_ANON, -1, 0);
+	assert(new_addr == addr);
+	unzeroed = false;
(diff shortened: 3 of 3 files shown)


thestinger:
`MADV_FREE` is a lot better than `MADV_DONTNEED` because it doesn't drop the pages until there's memory pressure, so you only pay for page faults and not massive amounts of zeroing + other work. Linux is in the process of landing `MADV_FREE` (it's in linux-next right now) and I'd expect jemalloc to start using it instead of the much slower API. The measurements on the kernel mailing lists demonstrated that it results in substantial (i.e. 3x) throughout improvements for workloads causing lots of purging. I would expect other platforms to be similar.

Note that `MEM_DECOMMIT` / `MEM_COMMIT` is not specific to Windows. On Linux, it's equivalent to purging the range and setting `PROT_NONE` on it which reduces the commit charge. It's significantly more expensive than `MADV_DONTNEED` because it has to grab a global writer lock (blocking any `mmap`, `munmap`, `mremap` or `madvise` system calls along with page faults) rather than the reader lock and requires an extra system call to commit it again (turning off `PROT_NONE`) and that is itself significantly more expensive than `MADV_FREE`. Linux supports full memory accounting and full overcommit (but doesn't use either by default) and both should be equally supported.

I don't think it makes sense to do things in different ways on each platform rather than having a consistent design across them. The memory measurement issue doesn't seem like a jemalloc one as the operating system can provide accurate statistics.

glandium:
> The memory measurement issue doesn't seem like a jemalloc one as the operating system can provide accurate statistics.

The fact the operating system does *not* provide accurate statistics is why this patch exists in the first place.

thestinger:
Windows exposes separate statistics for commit charge and actual memory usage. It might not be entirely clear to users but it *can* be measured accurately. It would be nice to expose a tunable providing the ability to reduce commit charge but I don't think statistics are a reason to do it. It should be provided consistently across platforms with support for both concepts too. In glibc, the allocator picks based on whether overcommit is enabled, but users should be able to override this if they have a lot of swap space to back the commit charge.

thestinger:
I don't think there will ever be a good reason to use `MADV_DONTNEED` on Linux once `MADV_FREE` is available (perhaps by 3.20 or 3.21). The kernel will be smart enough to switch over to the `MADV_DONTNEED` code path under memory pressure. It doesn't drop the commit charge unless you also toggle `PROT_NONE` with `mprotect`, and in that case you are better off using `mmap` with `MAP_FIXED` rather than combining `madvise` and `mprotect`. I'm sure that glibc is going to switch over to it for their purging, especially since they were the ones advocating for this in the past.

glandium:
> Windows exposes separate statistics for commit charge and actual memory usage. It might not be entirely clear to users but it can be measured accurately.

If you're talking about "Working set" (which, essentially, is the same as RSS), MEM_RESET doesn't affect it. Which is exactly the problem this patch tries to solve. Presumably, VirtualUnlock could have the same effect as DECOMMIT/COMMIT.

glandium:
> Presumably, VirtualUnlock could have the same effect as DECOMMIT/COMMIT.

Err, no, it just forces a swap-out.

glandium:
> AFAICT from Firefox benchmarks, this doesn't seem to affect performance on either OSX or Windows, but I don't have results for the full range of benchmarks on Windows yet.

It turns out it does affect startup performance on Windows.

jasone:
This change is certainly a pessimization for OS X, and it sounds like you've determined that the same applies to Windows.  Is there anything else to consider before we close this pull request?

glandium:
> This change is certainly a pessimization for OS X

In practice it appeared to make no difference on Firefox benchmarks. It may or may not affect other workloads. Windows was a different story, though.

Now, with that being said, the problem of "overcommit" after freeing plenty of allocations on OSX and Windows is still real. The recent decrease in chunk size /might/ be helping here, but it seems to me there still needs to be something done. How about making the explicit mallctl(arena.*.purge) do the kind of purge I'm adding here, while keeping the current behavior for the implicit purges happening during normal operations?

jasone:
I'm going to resolve #251 shortly, and that will provide a mechanism with which to non-intrusively experiment with decommit/commit memory management policies.  If we find something that works well for Windows, we can integrate it post-4.0.0 with little trouble.

PR #207, opened 2015-03-12
Cleanup

glandium:


jasone:
I have plans to uses these any year now.  Can we just fix the nanosleep portability issue instead of removing the code?

glandium:
It may or may not be possible. I think it's not possible to do a pause under 100ns on Windows. Anyways, since that hadn't been used in more than a year, I figured whatever your plans were, they got interrupted or canceled, and that there was no point keeping those. I'm tempted to say you should merge this, and revert it when you actually start needing them, which might be well after 4.0 for all I know.

jasone:
Is there an equivalent to ```sched_yield()``` on Windows?  That would be a reasonable alternative to sleeping.

glandium:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms686352%28v=vs.85%29.aspx ?

jasone:
The portability issues caused by the mq code should be resolved now (1b0e4abbfdbcc1c1a71d1f617adb19951109bfce).

PR #209, opened 2015-03-13
fixing the configure for bitrig

dhuseby:
This fixes some issues with sbrk/tls on OpenBSD and Bitrig.

@semarie @bluerise

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -287,11 +287,9 @@ case "${host}" in
 	CFLAGS="$CFLAGS"
 	abi="elf"
 	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
-	force_tls="0"
-	sbrk_deprecated="1"
 	;;
   *-*-bitrig*)
- 	CFLAGS="$CFLAGS"
+	CFLAGS="$CFLAGS"
 	abi="elf"
 	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
 	sbrk_deprecated="1"
--- a/configure.ac
+++ b/configure.ac
@@ -283,11 +283,19 @@ case "${host}" in
 	abi="elf"
 	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
 	;;
-  *-*-openbsd*|*-*-bitrig*)
+  *-*-openbsd*)
 	CFLAGS="$CFLAGS"
 	abi="elf"
 	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
+	force_tls="0"
+	sbrk_deprecated="1"
 	;;
+  *-*-bitrig*)
+ 	CFLAGS="$CFLAGS"
+	abi="elf"
+	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
+	sbrk_deprecated="1"
+  ;;
   *-*-linux*)
 	CFLAGS="$CFLAGS"
 	CPPFLAGS="$CPPFLAGS -D_GNU_SOURCE"


dhuseby:
@marcopeereboom this PR fixes the bitrig jemalloc issues.  once it lands, I'll cherry-pick it into the rust fork of jemalloc.

Bluerise:
I'm not sure force_tls should be set to 0 for Bitrig. OpenBSD needs it, as they do not have TLS support. Bitrig has, so I don't think this is something we need.

dhuseby:
@bluerise you're right, let me fix the patch.

Bluerise:
That looks good to me!

dhuseby:
I updated the patch.  I broke out OpenBSD from Bitrig because Bitrig support thread-local storage and OpenBSD does not.

marcopeereboom:
+1 although openbsd should get that same sbrk line.  sbrk is dead zed.

Bluerise:
@marcopeereboom Which is what he did in the current patch, so +1 from me.

configure.ac
```diff
@@ -283,11 +283,19 @@ case "${host}" in
 	abi="elf"
 	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
 	;;
-  *-*-openbsd*|*-*-bitrig*)
+  *-*-openbsd*)
 	CFLAGS="$CFLAGS"
 	abi="elf"
 	AC_DEFINE([JEMALLOC_PURGE_MADVISE_FREE], [ ])
+	force_tls="0"
```
semarie: the dev version of jemalloc autodetect `force_tls=0` correctly under openbsd. it is unneeded to force it to zero.

configure.ac
```diff
@@ -283,11 +283,19 @@ case "${host}" in
...
+	sbrk_deprecated="1"
```
semarie: - autodetection found `sbrk` under openbsd
- the manpage of `sbrk` under openbsd said: 
```
The brk() and sbrk() functions are historical curiosities left over from earlier days before the advent of virtual memory management.
```

it could be seen as a form a deprecation :)

- I don't have any error message with or without `sbrk`
- the `make check` terminate without failure (with or without `sbrk_deprecated="1"`)

so having it forced to be disabled or not under openbsd isn't in my competence... I could just say that it don't introduce regression...

semarie:
@jasone to resume the configure tests under OpenBSD (I don't known about Bitrig):
- `force_tls=0` is correctly inferred
- `have_sbrk=1` is found (it is ok, has OpenBSD provide `sbrk` function)

but I don't known if `sbrk_deprecated=1` is needed or not. From code usage perspective, I don't have any issue with `make test` or jemalloc usage with rustc. I could just note that manpage for `sbrk` don't recommand using it... (full manpage under OpenBSD: http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/brk.2?query=sbrk)

configure.ac
semarie: @dhuseby `force_tls=0` is needed (for openbsd) for the **rust-lang/jemalloc** fork, but not for **jemalloc/jemalloc**. rustc don't use the latest jemalloc version, and so autodetection don't work very well for rustc version of jemalloc.

jasone:
It looks to me like ```sbrk_deprecated``` should not be specified for OpenBSD, because ```sbrk()``` actually works, unlike on OS X.

dhuseby:
I updated the patch based on the feedback.

tamird:
@dhuseby needs a rebase

dhuseby:
I'll catch up on this soon.  I need to rewrap my head around what I was trying to fix.  I think this fell by the wayside because the rust project landed my patch there.  This absolutely needs to be upstreamed though.

jasone:
Please resubmit if this gets sorted out.

PR #216, opened 2015-03-20
[wip] Per-arena 'lg_dirty_mult' not under stats

daverigby:
Fix incorrect reference of arena.<N>.lg_dirty_mult - it is not under
'stats', and the index in the mib is at index 1 not 2.


daverigby:
Proof-of-concept, I expect you'll want a cleaner solution than this.

jasone:
Similar fix integrated as 7e336e7359ec50f06ec73f29033c7807148bf476.

PR #217, opened 2015-03-21
Fix a compile error caused by mixed declarations and code.

wqfish:
Got an error when compiling the code on devserver.


jasone:
Thanks!

PR #219, opened 2015-03-22
Beware size_t wrap-around -- ok, but very unlikely().

poige:


poige:
What about other commits in this set, should I repeat pull request w/o e5e1ef859fb1dc645b278a421c173db67fa34120 ?

jasone:
Oops, I overlooked 88c31804ca749d7ed7985a0c5522a7b8faacc830, which is now integrated as ef0a0cc3283ea561a40b33f4325d54bbc351de21.  Thank you.  Re: the likely/unlikely changes, I'm not sure that I want to annotate the entire code base with them; they are only important in hot code, and I've been pretty selective about using them elsewhere only when the expected behavior isn't immediately obvious.

poige:
I see, thanks.

PR #222, opened 2015-03-25
Add creating thread's Task ID ("tid") to the arena stats.

psi-mankoski:
Knowing the arena's creating Task ID can be very valuable when analyzing program memory usage.
(Especially when the TID <==> thread / run function mapping is already known by other means.)

Comments?

### DIFF
--- a/include/jemalloc/internal/arena.h
+++ b/include/jemalloc/internal/arena.h
@@ -272,6 +272,11 @@ struct arena_s {
 	unsigned		nthreads;
 
 	/*
+	 * The Task ID for the first thread to use this arena.
+	 */
+	int				tid;
+
+	/*
 	 * There are three classes of arena operations from a locking
 	 * perspective:
 	 * 1) Thread assignment (modifies nthreads) is protected by arenas_lock.
--- a/include/jemalloc/internal/ctl.h
+++ b/include/jemalloc/internal/ctl.h
@@ -33,6 +33,7 @@ struct ctl_indexed_node_s {
 struct ctl_arena_stats_s {
 	bool			initialized;
 	unsigned		nthreads;
+	int				tid;
 	const char		*dss;
 	ssize_t			lg_dirty_mult;
 	size_t			pactive;
--- a/src/arena.c
+++ b/src/arena.c
@@ -2739,6 +2739,14 @@ arena_new(unsigned ind)
 
 	arena->ind = ind;
 	arena->nthreads = 0;
+
+#ifndef _WIN32
+	/* Save the creating thread's Task ID. */
+	arena->tid = syscall(SYS_gettid);
+#else
+	arena->tid = 0;
+#endif
+
 	if (malloc_mutex_init(&arena->lock))
 		return (NULL);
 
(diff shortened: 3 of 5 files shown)


thestinger:
This will become a *lot* less meaningful if jemalloc moves to dynamically assigned per-CPU arenas, and I think that would be a very good idea.

psi-mankoski:
Thanks for the feedback. Have you seen [my recent paper](http://highscalability.com/blog/2015/3/17/in-memory-computing-at-aerospike-scale-when-to-choose-and-ho.html) on how specifically designating certain arenas for long-term data storage (e.g., for databases) greatly helped us with server memory fragmentation at scale? I agree there are other use cases (e.g., for transient / transactional data) where different (and very likely automatic) schemes would be best. Being able to choose which to use seems most powerful to me. What do you think?

thestinger:
I haven't yet read it, but it looks interesting so I'll do that after I manage to get some sleep. It's intuitive to me that having threads reuse the same arenas would reduce fragmentation though, because threads often run different code and the allocation sizes will be more similar. It also avoids stuff like one thread making lots of memory allocations being swapped around with 3 barely allocating ones.

The great thing about per-CPU arenas is that they reduce contention even more than the current implementation with only 1 arena per core rather than 4. I think that would result in a lot less overall memory fragmentation even with threads jumping between arenas now and then.

There's an API for creating special extend arenas and it's possible to assign threads to them, so a per-CPU arena implementation as a default wouldn't prevent explicit static assignment.

thestinger:
There's no real harm in having this but it just seems a bit weird to me as in the general case there are many threads allocating from each arena over their lifetime, often with some contention. Threads may also be moving away from an arena in the future and tids can also be reused when threads die. It may help in some cases but *in general* it can't be correct as anything but a heuristic.

FWIW, storing the *last* thread id to allocate is one possible way of implementing a per-CPU load balancing algorithm, although using the thread-local storage address is a lot more portable and avoids a system call. The allocator can detect that another thread has used the arena and only do the relatively expensive `sched_getcpu` (or the even more expensive CPUID instruction if unavailable) if another thread has allocated from the arena. An alternative is using `trylock` and moving away on actual contention, but I'd *expect* the id-based method to be better. I haven't gotten conclusive results yet.

psi-mankoski:
Regarding the tab: It's all properly lined up with only <Tab> characters in the editor [Emacs]. (Think it's just how "git diff" is rendering it.)

jasone:
There are a few things that make this a less than idea approach in my opinion.

- The thread ID isn't a unique identifier, because threads can come and go.
- The potential usefulness is highly dependent on arena creation policy (and whether the arena is automatically/manually created), which is likely to change repeatedly over time.
- For an application that really cares about arena-->creator mapping, it's possible to construct the mapping via other means, specifically via "thread.arena" mallctl introspection.

psi-mankoski:
Thanks for the comments. All good points.
My objective is to analyze a program's memory use, ideally without (much) explicit instrumentation in the code, a la Valgrind. An arena keeps track of nthreads. Perhaps a useful debugging option would be a means to determine their identities more directly?
The questions we really want to answer are: who is using what right now? and what is actually in that arena? The creating thread is one hint for uncovering that mapping.

psi-mankoski:
Thanks for all the great feedback ~~ I'm closing this pull request and considering other cleaner/more general ways to achieve the same goals.

PR #226, opened 2015-04-22
Fix mallctl doc: arenas.hchunk.i.size

wqfish:


jasone:
Integrated as 897503521ddb703a1388899f79112e048c328278.  Thanks!

PR #227, opened 2015-04-29
prof: use /proc/<pid>/task/<pid>/maps on Linux

dhobsd:
In software with many thousands of maps (due to high thread counts), reading
reading from `/proc/<pid>/maps` on Linux requires `n^2` scans of the thread group
list just to annotate thread stacks. This gets very slow very quickly.

Since this information is not useful at all for heap profiling, and reading
from `/proc/<pid>/task/<pid>/maps` does not attempt to annotate maps as thread
stacks, we should simply read from this latter file when we dump maps.

An example from an actual production machine:

```
dho@xxx:~$ sudo /usr/bin/time cat /proc/177063/maps > /dev/null
0.00user 82.99system 1:33.53elapsed 88%CPU (0avgtext+0avgdata 2608maxresident)k
0inputs+0outputs (0major+204minor)pagefaults 0swaps

dho@xxx:~$ sudo /usr/bin/time cat /proc/177063/task/177063/maps > /dev/null
0.00user 0.12system 0:00.12elapsed 96%CPU (0avgtext+0avgdata 2592maxresident)k
0inputs+0outputs (0major+204minor)pagefaults 0swaps
```

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -1338,27 +1338,24 @@ label_return:
 	return (ret);
 }
 
-static bool
-prof_dump_maps(bool propagate_err)
-{
-	bool ret;
-	int mfd;
-	char filename[PATH_MAX + 1];
+enum prof_map_ret {
+	PMR_OPEN_FAIL,
+	PMR_WRITE_FAIL,
+	PMR_SUCCESS,
+};
 
-	cassert(config_prof);
-#ifdef __FreeBSD__
-	malloc_snprintf(filename, sizeof(filename), "/proc/curproc/map");
-#else
-	malloc_snprintf(filename, sizeof(filename), "/proc/%d/maps",
-	    (int)getpid());
-#endif
+static enum prof_map_ret
+prof_dump_mapfile(bool propagate_err, const char *filename)
+{
+	enum prof_map_ret ret;
+	int mfd;
+
 	mfd = open(filename, O_RDONLY);
 	if (mfd != -1) {
 		ssize_t nread;
 
-		if (prof_dump_write(propagate_err, "\nMAPPED_LIBRARIES:\n") &&
-		    propagate_err) {
-			ret = true;
+		if (prof_dump_write(propagate_err, "\nMAPPED_LIBRARIES:\n")) {
+			ret = PMR_WRITE_FAIL;
 			goto label_return;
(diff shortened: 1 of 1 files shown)


jasone:
Is ```/proc/<pid>/task/<pid>/maps``` a long-standing procfs feature?  If not, we need to protect the added code with a feature test.  Otherwise your patch can simply replace the ```/proc/<pid>/maps``` case.

dhobsd:
`proc(5)` says it's been around since since `2.6.0-test6`. Stack annotation was introduced in 3.4 though, so perhaps I should test around that. If you need 2.4 compatibility, that's probably the best route.

dhobsd:
Ok, I've added an `--enable-task-maps` option, off by default. I could add some `awk -F.` sourcery around `uname -r` to do version detection if you'd prefer going the route of just defaulting to use this file on Linux kernel versions > 3.4, but this seemed more intrusive to configure.ac (and possibly more annoying for dealing with non-Linux systems wrt autodetection doing the right thing because introducing more global state to autoconf).

jasone:
After considering for a while the best default for the task-maps option, I realized that we could instead just sequentially try opening ```/proc/<pid>/task/<pid>/maps``` (which will always work on modern systems), then ```/proc/<pid>/maps``` if that fails.  Can you think of any problems with that approach?

dhobsd:
The switch case is a little tricky, but I'm prettty sure it's correct and the tests pass. I don't have a Linux 2.4 system to test the fallback code on. Technical correctness / verifiability are the only concerns I really have with this version of the solution. It's certainly a more intrusive change, but this was the best way I could think of to keep things reasonably clean and avoid duplicating a bunch of the code.

jasone:
I have a simpler equivalent patch I'm going to push in a moment.  Thanks for your efforts on this issue, and please keep submitting patches as you discover issues.  I don't make a habit of rewriting people's submissions, so please don't let this instance deter you. =)

dhobsd:
Thanks!

dhobsd:
(Now that I have some time -- for clarity, wasn't sure whether just dropping a block in the middle was cool -- hopefully I'll have some further input to development and style will cease being a factor. Thanks again for including this!)

PR #238, opened 2015-06-24
Optimizations for Windows

mlvdm:
- Set opt_lg_chunk based on run-time OS setting
- Verify LG_PAGE is compatible with run-time OS setting
- When targeting Windows Vista or newer, use SRWLOCK instead of CRITICAL_SECTION
- When targeting Windows Vista or newer, statically initialize init_lock

### DIFF
--- a/include/jemalloc/internal/mutex.h
+++ b/include/jemalloc/internal/mutex.h
@@ -26,7 +26,11 @@ typedef struct malloc_mutex_s malloc_mutex_t;
 
 struct malloc_mutex_s {
 #ifdef _WIN32
+#  if _WIN32_WINNT >= 0x0600
+	SRWLOCK         	lock;
+#  else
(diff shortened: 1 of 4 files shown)


mlvdm:
Did this change for a local project, let me know if you agree.

Note on SRWLOCK, it is
- faster and smaller than CRITICAL_SECTION
- has built-in spincount
- cannot be locked recursively (I believe this is compatible with jemalloc lock behavior, but I could be wrong)
- can only be used on Windows Vista and newer (guarded at compile time)

jasone:
jemalloc does not lock recursively, so SRWLOCK should work fine.

mlvdm:
Fixed the problems you mentioned

jasone:
Integrated as a1aaf949a5d3b639f03dd7e33ffe1f0849b7f8df.  Thanks!

PR #246, opened 2015-07-21
Remove JEMALLOC_ALLOC_SIZE annotations on functions not returning poi…

### LINKED ISSUE #245: ae93d6bf364e9db9f9ee69c3e5f9df110d8685a4 adds a warning on OSX
```
jemalloc.h:164:5: error: alloc_size attribute only applies to functions that return a pointer [-Werror,-Wignored-attributes]
jemalloc.h:178:16: error: alloc_size attribute only applies to functions that return a pointer [-Werror,-Wignored-attributes]
```

glandium:
…nters

As per gcc documentation:
  The alloc_size attribute is used to tell the compiler that the function
  return value points to memory (...)

This resolves #245.

### DIFF
--- a/include/jemalloc/jemalloc_protos.h.in
+++ b/include/jemalloc/jemalloc_protos.h.in
@@ -13,7 +13,7 @@ JEMALLOC_EXPORT void	*@je_@calloc(size_t num, size_t size) JEMALLOC_CXX_THROW
     JEMALLOC_ATTR(malloc) JEMALLOC_ALLOC_SIZE2(1, 2) JEMALLOC_NOTHROW;
 JEMALLOC_EXPORT int	@je_@posix_memalign(void **memptr, size_t alignment,
     size_t size) JEMALLOC_CXX_THROW JEMALLOC_ATTR(nonnull(1))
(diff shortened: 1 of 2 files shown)


glandium:
Note that all the annotations in the .c files are redundant and could be removed.

jasone:
At some point during testing I ran into surprising linker issues depending on whether the header was included (related to ```throw()```?), and I decided to duplicate the attributes in order to permanently avoid that confusion.

ariccio:
Interesting. If you're going to be using annotations, maybe you ought to add some SAL-compatible macros, too?

For MSVC, you could `#define` `JEMALLOC_ALLOC_SIZE` as something like:

`#define JEMALLOC_ALLOC_SIZE(x) _Ret_opt_bytecount_(x)`

I'm fairly certain that there's a better SAL annotation, but I'm away from my computer.

glandium:
> At some point during testing I ran into surprising linker issues depending on whether the header was included (related to throw()?), and I decided to duplicate the attributes in order to permanently avoid that confusion.

AFAIK, only throw() is needed on both declarations and definitions.` __attribute__`s can be left to the declaration.

PR #248, opened 2015-07-27
MSVC compatibility changes

mlvdm:
- Decorate public function with __declspec(allocator) and __declspec(restrict), just like MSVC 1900
- Support JEMALLOC_HAS_RESTRICT by defining the restrict keyword
- Move __declspec(nothrow) between 'void' and '*' so it compiles once more


jasone:
Integrated as c1a6a51e401eb888caff5de142280754e7d99ba3.  Thanks!

PR #253, opened 2015-08-11
Make --enable-tls and --enable-lazy-lock take precedence over configu…

glandium:
…re.ac-hardcoded defaults


jasone:
Merged as ac5db02034c01357a4ce90504886046a58117921 with a small tweak to make the final configure summary output consistent.  Thanks!

PR #256, opened 2015-08-19
Fix build problems on Solaris (index_t already defined)

trondn:
I tried building jemalloc on SmartOS, but it failed to compile due to the fact that index_t is defined in sys/types.h with a different type.

in sys/types.h it is a short, and now we're trying to define it as an unsigned integer.

All unit tests pass with this change.


jasone:
I chose to rename the type in order to keep things as simple as possible.  Thanks for the report and the suggested fix!

PR #265, opened 2015-08-28
Don't purge junk filled chunks when shrinking huge allocations

### LINKED ISSUE #260: Skip purge in some cases involving opt_junk_free
In `src/huge.c`, there is this snippet in `huge_ralloc_no_move_similar`:
```c
zeroed = !chunk_purge_wrapper(arena, &chunk_hooks, ptr,
    CHUNK_CEILING(usize), usize, sdiff);
if (config_fill && unlikely(opt_junk_free)) {
        memset((void *)((uintptr_t)ptr + usize), 0x5a, sdiff);
        zeroed = false;
}
```

A similar one is in `huge_ralloc_no_move_shrink`
It seems to me the purge here is not useful considering the memset. Or maybe there shouldn't be a memset and purge should be kept here, I'm not sure what happens to those pages afterwards.
Thoughts?

glandium:
When junk filling is enabled, shrinking an allocation fills the bytes
that were previously allocated but now aren't. Purging the chunk before
doing that is just a waste of time.

This resolves #260.

### DIFF
--- a/src/huge.c
+++ b/src/huge.c
@@ -148,11 +148,12 @@ huge_ralloc_no_move_similar(void *ptr, size_t oldsize, size_t usize,
 	/* Fill if necessary (shrinking). */
 	if (oldsize > usize) {
 		size_t sdiff = oldsize - usize;
-		zeroed = !chunk_purge_wrapper(arena, &chunk_hooks, ptr,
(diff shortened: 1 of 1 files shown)


glandium:
This was applied in 4a2a3c9a6e5f8fda3536f48095d68ddaa13cd977

PR #267, opened 2015-08-30
Use 'local-dynamic' TLS model for thread variables

RCL:
Since jemalloc can be compiled as a DSO (or statically linked into another DSO),
'local-dynamic' TLS model is a better choice. 'initial-exec' usage in DSO is
generally problematic, see for example: https://sourceware.org/ml/libc-alpha/2014-10/msg00134.html

This usage caused problems for Unreal Engine, too:
https://answers.unrealengine.com/questions/229646/crash-on-ue4editor-startup.html

In general it may be better to avoid specifying TLS model in the sources and rely on
-ftls-model= switch instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -421,13 +421,13 @@ SAVED_CFLAGS="${CFLAGS}"
 JE_CFLAGS_APPEND([-Werror])
 JE_COMPILABLE([tls_model attribute], [],
               [static __thread int
-               __attribute__((tls_model("initial-exec"), unused)) foo;
+               __attribute__((tls_model("local-dynamic"), unused)) foo;
                foo = 0;],
               [je_cv_tls_model])
 CFLAGS="${SAVED_CFLAGS}"
 if test "x${je_cv_tls_model}" = "xyes" ; then
   AC_DEFINE([JEMALLOC_TLS_MODEL],
-            [__attribute__((tls_model("initial-exec")))])
+            [__attribute__((tls_model("local-dynamic")))])
 else
   AC_DEFINE([JEMALLOC_TLS_MODEL], [ ])
 fi


thestinger:
AFAIK this is important for performance. It might be less important now that there's a pointer passed around, but it's still going to be a significant cost. It's not compatible with `dlopen` and it uses up one static TLS slot, but it's not incorrect.

thestinger:
The static TLS slots are there to be used after all. A malloc implementation's thread cache is probably a lot more compelling than some of the other libraries that are filling the other 15 slots. Also, jemalloc will only use one slot vs. some libraries that probably use several due to having thread locals spread out between translation units.

RCL:
I understand that this can have performance implications (although I haven't measured it), that is why I suggested leaving this decision to the code that links to .a file (and presumably uses a proper -ftls-model switch).  A compromise change would be to force initial-exec only when compiling .a file.

Current default is somewhat treacherous.

thestinger:
What's treacherous about it? It's using one of the 16 (or 32) static TLS slots that are essentially only there for dynamic libraries. I think jemalloc probably has a more compelling use case for it than the other libraries that are exhausting the resource.

Also, I don't think `-flts-model` will have any impact once the static library is already compiled. It needs to be set when compiling jemalloc itself.

thestinger:
Can this actually be triggered without using `dlopen` on jemalloc (or a library embedding jemalloc)? That's already broken for other reasons.

thestinger:
AFAICT it allocates enough space for every library that's actually linked with 14 extra slots for `dlopen`.

RCL:
You are correct about -ftls-model, this indeed won't work.

I agree that jemalloc, in general, has a compelling use case to take up a slot. I called it treacherous because I did not expect it to be the default, given that slot shortage seems to be a problem on distributions that use stock glibc (i.e. not Ubuntu / Fedora) - and heavily modular applications like UE4 that dlopen() a bunch of plugins exacerbate it. But indeed this is not a jemalloc problem, so I will understand if jemalloc decides to keep the performance-oriented default.

PR #271, opened 2015-09-03
Force initialization of the init_lock in malloc_init_hard on Windows XP

### LINKED ISSUE #269: Initialization order issue when trying to set chunk hooks
This happened to me on Windows, but this might as well happen anywhere the compiler creates a global initialization function to fill the `init_lock` with its initial value, although that's rather unlikely.

The problem at hand is that I setup a constructor function in order to set up my custom chunk hooks, and that function ended up running before the `init_lock` initializer, so when I called `mallctl`, which calls `malloc_init`, `malloc_init` crashed while trying to lock the uninitialized `init_lock`.

The problem is that nothing guarantees in which order initializers in the same binary are going to be executing.

For now, I worked around this by forcing the `init_lock` initializer to be executed earlier by adjusting the CRT$XC* section name, but it only works on MSVC (so it would still be broken for TorBrowser, who build with mingw).

glandium:
This resolves #269.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -179,13 +179,24 @@ static bool			malloc_initializer = NO_INITIALIZER;
 static malloc_mutex_t	init_lock = SRWLOCK_INIT;
 #else
 static malloc_mutex_t	init_lock;
+static bool init_lock_initialized = false;
 
 JEMALLOC_ATTR(constructor)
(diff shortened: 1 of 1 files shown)


jasone:
This was integrated as 0a116faf95ba8541ce75448bb9b6fba0efdde69a.

PR #277, opened 2015-09-21
Put likely branch first

Dmitry-Me:
Current code reads as "if unlikely not equal", new code reads "if likely equal" which is easier to fit into brain. Also compilers that don't recognize likely/unlikely put the *if* branch code first so x86 processors default to *if* code instead of *else* code so the current code compiled with Visual C++ would always be predicted wrong.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1379,10 +1379,10 @@ imalloc_prof(tsd_t *tsd, size_t usize)
 	prof_tctx_t *tctx;
 
 	tctx = prof_alloc_prep(tsd, usize, prof_active_get_unlocked(), true);
-	if (unlikely((uintptr_t)tctx != (uintptr_t)1U))
-		p = imalloc_prof_sample(tsd, usize, tctx);
-	else
+	if (likely((uintptr_t)tctx == (uintptr_t)1U))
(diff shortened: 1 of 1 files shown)


jasone:
Does this change actually improve code generation for any of the compilers in use with jemalloc?  I'm not convinced by the cognitive load argument, so there needs to be a performance benefit for me to integrate this.

Dmitry-Me:
@jasone Swapping branches gets better code for Visual C++ which doesn't recognize likely-unlikely. Compilers that recognize likely-unlikely hints will generate the same code because the hint is still there, just the opposite one. Readability is also improved which is a nice bonus.

jasone:
After contemplating this for a while, I'm inclined not to even go so far as preferring ```likely()``` over ```unlikely()``` as a best practice, because doing so is detrimental to readability in many cases, especially early return on error.

Dmitry-Me:
@jasone Where is early return on errors in the code affected by proposed changes?

jasone:
The diff in question does not change any early return code, but as I said, I don't want to maintain a best practice preference for ```likely()``` over ```unlikely()```, and without such a preference, this diff is indistinguishable from churn that will bitrot away during future refactoring.

PR #280, opened 2015-09-21
Assert compact color bit is unused

garoose:
We are using the compact red-black tree with a custom slab allocator that doesn't align allocations with `__attribute__((packed))` structs. This could lead to the color bit being used by the node pointer and we found this assertion useful in avoiding programming errors in structure alignment.

This implementation adds an assertion to `rbt_new_node` when `RB_COMPACT` is set because it seemed part of creating a new node. An alternative would be to add the assert in `ex_insert` before calling `rbt_new_node`.


jasone:
Thanks for the rb-related patches, @garoose.  As soon as the 4.0.x stabilization work winds down I'll take a closer look and get them integrated.

PR #282, opened 2015-09-22
Add tree destruction function and test

garoose:
Iterating over the tree with post-order traversal to remove and process nodes avoids the cost of re-balancing when destroying the tree.


jasone:
Integrated as e8ab0ab9c0e395d3c09398fa981704a9be968838 and 710ca112e31e8621177d08162f60158c27dd2974.  Thanks!

PR #287, opened 2015-10-01
Add support for time-based purging

wqfish:
This allows application to specify an expiration time for dirty pages. If
dirty_exp_time is not 0, jemalloc will purge dirty pages that are older than the
specified time. The action is triggered each time a dirty run is deallocated.

If async_purge=true, jemalloc itself will not do purging automatically.
Application needs to call mallctl("arena.<i>.purge_old") periodically (probably
from an extra thread) to impose the limit on expiration time. The default
async_purge is false.


jasone:
Superseded by 243f7a0508bb014c2a7bf592c466a923911db234.  Thank you for getting the ball rolling on this, @wqfish!

Issue #278, opened 2015-09-21
Make *allocx() size class overflow behavior defined.


jasone:
Right now the ```*allocx()``` API explicitly makes behavior on size class overflow undefined.  Per the ```mallocx()``` documentation:

> Behavior is undefined if size is 0, or if request size overflows due to size class and/or alignment constraints.

Unfortunately, this is difficult to deal with with in application, because the largest size class is a function of both address space and configure-time options, thus requiring two ```mallctl()``` calls to determine the maximum size class.  In practice an application will never actually need such a large request to succeed (which in all cases would be at least half the available virtual address space), but it's not uncommon for applications to underflow ```size_t``` and accidentally ask for a size that overflows the maximum size class.  Right now this triggers assertion failures in debug builds, but for optimized builds the result is difficult-to-diagnose failures within jemalloc.

To address this, make requests for greater than ```HUGE_MAXCLASS``` result in allocation failure, e.g. ```NULL``` return for ```mallocx()```, so that a careful application can detect the error, and a careless application crashes in a more obvious way.  A similar argument could be made to return ```NULL``` for 0-byte requests, but this is an error that can (and should) be easily avoided in the application.

PR #295, opened 2015-10-30
prevent allocs larger than PTRDIFF_MAX on Android

thestinger:
Android's Bionic libc now reports an out-of-memory error for allocations
larger than PTRDIFF_MAX in order to eliminate classes of potentially
exploitable bugs caused by signed overflow. Since jemalloc merges spans
of virtual memory (both for chunk recycling and in-place huge allocation
growth), it needs to cooperate with Bionic to fully enforce this.

This can be extended to other platforms if and when they decide to do
the same thing as musl and Bionic in this case. There's no need to
handle musl in jemalloc because it doesn't support replacing the malloc
implementation either via weak symbols or hooks like glibc.

Rationale from the Android commit message:

    make mmap fail on requests larger than PTRDIFF_MAX

    Allocations larger than PTRDIFF_MAX can be successfully created on
    32-bit with a 3:1 split, or in 32-bit processes running on 64-bit.

    Allowing these allocations to succeed is dangerous, as it introduces
    overflows for `end - start` and isn't compatible with APIs (mis)using
    ssize_t to report either the size or an error. POSIX is guilty of this,
    as are many other Android APIs. LLVM even considers the `ptr + size`
    case to be undefined, as all pointer arithmetic compiles down to signed
    operations and overflow is treated as undefined for standard C pointer
    arithmetic (GNU C `void *` arithmetic works differently).

    This also prevents dlmalloc from allocating > PTRDIFF_MAX as it doesn't
    merge mappings like jemalloc. A similar check will need to be added in
    jemalloc's code path for huge allocations.

    The musl libc implementation also performs this sanity check.

https://android-review.googlesource.com/#/c/170800/

GCC is also unable to handle `ptr + size` with `size > PTRDIFF_MAX`:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67999

### DIFF
--- a/src/huge.c
+++ b/src/huge.c
@@ -61,6 +61,11 @@ huge_palloc(tsd_t *tsd, arena_t *arena, size_t size, size_t alignment,
 		return (NULL);
 	assert(usize >= chunksize);
 
+#ifdef __ANDROID__
+	if (usize > PTRDIFF_MAX)
+		return (NULL);
+#endif
+
 	/* Allocate an extent node with which to track the chunk. */
 	node = ipallocztm(tsd, CACHELINE_CEILING(sizeof(extent_node_t)),
 	    CACHELINE, false, tcache, true, arena);
@@ -290,6 +295,11 @@ huge_ralloc_no_move(void *ptr, size_t oldsize, size_t usize_min,
 	if (oldsize < chunksize || usize_max < chunksize)
 		return (true);
 
+#ifdef __ANDROID__
+	if (usize_max > PTRDIFF_MAX)
+		return (true);
+#endif
+
 	if (CHUNK_CEILING(usize_max) > CHUNK_CEILING(oldsize)) {
 		/* Attempt to expand the allocation in-place. */
 		if (!huge_ralloc_no_move_expand(ptr, oldsize, usize_max, zero))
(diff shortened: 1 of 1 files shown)


thestinger:
cc @cferris1000

jasone:
In practice this imposes a maximum size class of 1.75 GiB for 32-bit processes, and 7 PiB for 64-bit processes, right?  I'm contemplating whether this might be a reasonable limitation for all platforms.  #278 deals with a related issue, and it might make sense to address both at the same time.

thestinger:
It will reject anything above 2147483647 bytes on 32-bit and 9223372036854775807 bytes on 64-bit so whatever is the maximum size class below those. `PTRDIFF_MAX` is based on the full size of the address space, not just the portion exposed to userspace. It is effectively a no-op on 64-bit because there's a 1:1 split with the kernel and it would be impossible to allocate more than half of the address space in practice. It's needed due to the common 3:1 split on 32-bit and the fact that the full address space (4G) is available to a 32-bit process running on 64-bit. It's not really much of a limitation, especially since compilers are unable to handle these object sizes.

thestinger:
i.e. it could really just be wrapped in an ifndef 64-bit define, but it would be nice to do it in a straightforward way where someone porting jemalloc to a system without something like the traditional 1:1 split (lets say a unikernel) could overlook the issue

thestinger:
(well, the full *theoretical* size, since `PTRDIFF_MAX` is really just the max `ptrdiff_t` value, i.e. the max value before `p - q` overflows, so it's larger than the full address space on x86_64 and ARMv8 right now, but might not always be)

jasone:
Resolved by 0c516a00c4cb28cff55ce0995f756b5aae074c9e.

PR #296, opened 2015-10-31
Add uninstall hooks to Makefile.in

jayaajay:
This patch fixes make uninstall.


jasone:
Uninstall targets are dangerous, especially ones that have ```rm -rf $(MAKEVAR)```.  This is an intentional omission.  Is there some compelling use case?  Package management systems don't need these, and during testing it's easy to isolate the installed files via an otherwise unused install prefix.

PR #297, opened 2015-11-05
use correct macro definitions for clang-cl

froydnj:
clang-cl, an MSVC-compatible frontend built on top of clang, defined
_MSC_VER *and* supports __attribute__ syntax.  The ordering of the
checks in jemalloc_macros.h.in, however, do the wrong thing for
clang-cl, as we want the Windows-specific macro definitions for
clang-cl.  To support this use case, we reorder the checks so that
_MSC_VER is checked first (which includes clang-cl), and then
JEMALLOC_HAVE_ATTR) is checked.  No functionality change intended.

### DIFF
--- a/include/jemalloc/jemalloc_macros.h.in
+++ b/include/jemalloc/jemalloc_macros.h.in
@@ -36,32 +36,7 @@
 #  define JEMALLOC_CXX_THROW
 #endif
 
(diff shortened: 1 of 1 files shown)


jasone:
Integrated as 566d4c02400700b94a952eddeed34313360211d3.  Thanks!

PR #300, opened 2015-11-12
Reuse previously computed value

Dmitry-Me:
These two `ivsalloc()` calls are identical and should yield the same result. There's no need for two identical calls. It's better to cache and reuse the value instead of recomputing it. Should the assert fire it's more convenient to see the value immediately. Also the new code makes a single call both with asserts present (debug builds) and asserts removed (non-debug builds) which makes it more predictable in case of some "impossible" bugs.

### DIFF
--- a/src/zone.c
+++ b/src/zone.c
@@ -122,8 +122,10 @@ static void
 zone_free_definite_size(malloc_zone_t *zone, void *ptr, size_t size)
 {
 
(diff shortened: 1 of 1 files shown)


jasone:
Integrated as ea59ebf4d3c2a5749e170cc45c294e04129e5b49.

PR #301, opened 2015-11-13
Reduce variables scope, reduce code duplication

Dmitry-Me:
This makes code a bit easier to read.


jasone:
This is a mixture of CSE and refactoring of marginal value, with some style nits thrown in, so I'm not planning to merge.

Dmitry-Me:
@jasone Okay, no more submissions from me then.

PR #308, opened 2015-12-13
Prefer object-dir includes over source-dir includes.

yosefe:
When building out-ouf-source, we need to use object-dir include files
because they contain the generated macros, rather than the ones in
source-dir.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -23,7 +23,7 @@ abs_srcroot := @abs_srcroot@
 abs_objroot := @abs_objroot@
 
 # Build parameters.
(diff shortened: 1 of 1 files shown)


yosefe:
@shamisp @MattBBaker @miked-mellanox

shamisp:
:+1:

MattBBaker:
:+1:

jasone:
Which headers reside in both directories?  There shouldn't be any duplicates, unless a non-objdir build wasn't cleaned up.

jasone:
I can't determine whether this should be merged without an answer to my previous question.  Please reopen if you have further context that can help me understand the issue.

PR #311, opened 2015-12-22
Avoiding conditional directives that break statements.

RomeroMalaquias:


RomeroMalaquias:
Hello?

jasone:
I have a nontrivial backlog of issues to work through over the coming weeks, which is why I hadn't yet responded.  This particular pull request does not provide a rationale on which to be judged, which means I have no way to know whether it should be treated as anything more than stylistic criticism.

RomeroMalaquias:
Ok, thanks

RomeroMalaquias:
I will wait

jasone:
There is still no stated justification for this change, and stylistic preference alone is insufficient.  Closing.

PR #316, opened 2016-01-13
Jeprof curl option fixes

djwatson:
If jemalloc profiling is disabled, jeprof currently gives an incomprehensible error message


jasone:
Thanks!

PR #328, opened 2016-02-11
CMake based build for MSVC 13 and 15

yuslepukhin:
Note: this does not fixes bugs but builds tests
  so we can see where bugs are.
  None of them are fixes in this iteration. I deemed OK to suppress
  warnings that I have seen.
  The benchmark is not currently built due to linking issues
  and bugs are currently a priority.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,697 @@
+# This make file supports Windows build of Jemalloc
+#
+# Prerequisites:
+#     You must have Visual Studio 2013 Update 4 installed or Visual Studio 2015 Update 1.
+#      Start the Developer Command Prompt window that is a part of Visual Studio installation.
+#      This will provide you with the accessible toolchain commands.
+#      You must have a path git.exe in your %PATH%.
+#
+# 1. Create a build directory
+#
+# 2. Run cmake to generate project files for Windows
+#        sample command: cmake -G "Visual Studio 12 Win64"  ..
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,697 @@
...
+set(JEMALLOC_HAVE_SECURE_GETENV 0)
+
+set(JEMALLOC_HAVE_ISSETUGID 0)
+set(JEMALLOC_MALLOC_THREAD_CLEANUP 0)
+
+set(JEMALLOC_MUTEX_INIT_CB 0)
+
+# No lazy lock since there is a way in Windows
+# to bootstrap for sure
+set(JEMALLOC_LAZY_LOCK 0)
```
yuslepukhin: Now, I am missing je_init in this version which is currently invoked in our in-house version before any static data is initialized.

yuslepukhin:
@rustyx I hope this would help you iterate.

rustyx:
Wouldn't it be confusing to have both CMake and Autotools config in one project?
IMO if you want to introduce CMake, then it should be done for all targets, not just for Windows. Otherwise the risk is high that it'll become a maintenance nightmare for us Windows users.

yuslepukhin:
@rustyx  For windows users installing Cygwin and related things might be a little more effort than simply CMake. The dual approach works allright for RocksDB but it is your call. I will keep this open until tomorrow.

PR #330, opened 2016-02-17
Separate arena_avail trees

djwatson:
Summary:
Separate run trees by index, replacing the previous quantize logic.  Quantization by index is now preformed only on insertion / removal from the tree, and not on node comparison, saving some cpu.  This also means we don't have to dereference the miscelm* pointers, saving half of the memory loads from miscelms/mapbits that have fallen out of cache.  A linear scan of the indicies appears to be fast enough.

The only cost of this is an extra tree array in each arena.

It looks like part of the win here may have been a subtle quantize bug introduced in 8a03cf039c - certain run_quantize_next calls seem to return an index one or two higher than are actually necessary to support the allocation request

Test Plan:
make check

ran several canaries of our services.  Shows ~1% cpu improvement, ~1% memory savings, and ~3-4% p99 latency improvement, depending on exact service and metrics

Blame Rev: 8a03cf039c

### DIFF
--- a/include/jemalloc/internal/arena.h
+++ b/include/jemalloc/internal/arena.h
@@ -402,6 +402,7 @@ struct arena_s {
 
 	/* bins is used to store trees of free regions. */
 	arena_bin_t		bins[NBINS];
+	arena_avail_tree_t	avail[NSIZES];
 };
 #endif /* JEMALLOC_ARENA_STRUCTS_B */
 
--- a/src/arena.c
+++ b/src/arena.c
@@ -76,7 +76,8 @@ arena_miscelm_size_get(const arena_chunk_map_misc_t *miscelm)
 }
 
 JEMALLOC_INLINE_C int
-arena_run_comp(const arena_chunk_map_misc_t *a, const arena_chunk_map_misc_t *b)
+arena_run_address_comp(const arena_chunk_map_misc_t *a,
+			const arena_chunk_map_misc_t *b)
 {
 	uintptr_t a_miscelm = (uintptr_t)a;
 	uintptr_t b_miscelm = (uintptr_t)b;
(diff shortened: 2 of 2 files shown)


jasone:
Do you have any more specific info about the run quantization issues?  I'm going to write a unit test regardless, but it would help to know what exactly to look for.

djwatson:
run_quantize_next appears to be quantizing a few indexes higher than necessary.  Here is a fix just for that

https://github.com/djwatson/jemalloc/commit/139c7be3de1c94fdfbf7e8c6f6f0bbc02895e680

jasone:
Here's a [fix for run quantization](08551eee586eefa8c98f33b97679f259af50afab), as well as [tests](a9a46847925e38373e6a5da250c0cecb11a8277b) and [table-lookup-based optimization](0da8ce1e96bedff697f7133c8cfb328390b6d11d).  Regarding the memory savings you saw in your benchmarks, that suggests quantization is _increasing_ run fragmentation, so if that's a widespread result, we need to consider ripping out quantization altogether.

jasone:
Now that the run quantization issues are sorted out, I'm looking at the main part of your diff (per size class run trees).  The size class index computations can now be simplified to use ```run_quantize_floor()```, and there are some style nits, but otherwise it looks good to me.

I'm not super excited about the related (rbt-nil removal)[331], but I'll run the change through some (benchmarks)[http://canonware.com/rb/], and if there isn't a substantial impact, I'm game for these changes.

jasone:
The ```rbt_nil``` removal has no significant impact on performance (well under 1%), so I'll work on getting these patches integrated.

jasone:
Integrated as 3417a304ccde61ac1f68b436ec22c03f1d6824ec, with changes to remove more dead code, precisely size runs_avail, and clean up style nits.  Note that this also enabled ae45142adc12d39793c45ecac4dafad5674a4591.  Thanks!

PR #331, opened 2016-02-17
Remove rbt_nil

djwatson:
Summary: Since this is an intrusive tree, rbt_nil is the whole size of the node and can be quite large.  For example, miscelm is ~100 bytes.

Mostly so that https://github.com/djwatson/jemalloc/commit/ea36fe6b68ecab492d5a3903c3f1267c943904ec
doesn't cause arenas to grow in size by ~20k bytes.  WIth both patches, the memory should be roughly neutral - avail[] is added, but the bin run trees should shrink.

Test Plan:
make check
ran a canary, no crashes

### DIFF
--- a/include/jemalloc/internal/rb.h
+++ b/include/jemalloc/internal/rb.h
@@ -42,7 +42,6 @@ struct {								\
 #define	rb_tree(a_type)							\
 struct {								\
     a_type *rbt_root;							\
-    a_type rbt_nil;							\
 }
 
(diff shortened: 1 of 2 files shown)


jasone:
Integrated as 2b1fc90b7b109c5efac7974b8f9abe269ecb6daf, with the addition of some redundant code removal (```if (ret == NULL) ret = NULL;```), and some line rewrapping.  Thanks!

PR #339, opened 2016-02-24
Use linear scan for small bitmaps

djwatson:
For small bitmaps, a linear scan of the bitmap is slightly faster than
a tree search - bitmap_t is more compact, and there are fewer writes
since we don't have to propogate state transitions up the tree.
On x86_64 with the current settings, I'm seeing ~.5%-1% CPU improvement
in production canaries with this change.

The old tree code is left since 32bit sizes are much larger (and ffsl
smaller), and maybe the run sizes will change in the future.


djwatson:
Thanks for the review - 

* style nits fixed
* split entire functions by ifdef USE_TREE in bitmap.c for more clarity

jasone:
I noticed just before committing your patch that ```bitmap_info_ngroups()``` looked funny (both tree and linear versions), and I did some refactoring to clean that up, then rebased your diff.  Without that cleanup, the linear version was doing ```malloc(0)``` in tests, and therefore overflowing the bitmap buffers.

PR #340, opened 2016-02-24
Fix arena_run_first_best_fit

djwatson:
Merge of 3417a304ccde61ac1f68b436ec22c03f1d6824ec looks like a small bug:

first_best_fit doesn't scan through all the classes, since ind is offset from runs_avail_nclasses by run_avail_bias

Also it looks like we allocate one fewer tree than runs_avail_nclasses, but it looks like it is still initialized below?

thanks!

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -1079,7 +1079,7 @@ arena_run_first_best_fit(arena_t *arena, size_t size)
 	szind_t ind, i;
 
 	ind = size2index(run_quantize_ceil(size));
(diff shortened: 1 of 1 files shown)


jasone:
```arena_run_first_best_fit()``` fix integrated as cd86c1481ad7356a7bbcd14549e938769f474fd6.  Thank you thank you thank you for catching this!

jasone:
Well, you were right about the other bug, too. ;-)  Fix integrated as 38127291670af8d12a21eb78ba49201f3a5af7d1.

PR #347, opened 2016-02-29
Fix stack corruption in unit test and uninitialized var warning

yuslepukhin:
Stack corruption happens in x64 bit in hash unit test.
  `memset()` makes use of the `sizeof(ptr)` not an array since
  on Windows `VARIABLE_ARRAY` is `alloced` and the result is a ptr
  which is 8 bytes on x64 and that is bigger than 32-bit hash.
  This corrupts the stack.
  Warning in arena.c issued by MSVC 13 Release is questionable
  but breaks the build.


jasone:
Integrated with whitespace fixes, as well as type changes to silence ```-Wshorten-64-to-32``` warnings.  Thanks!

PR #348, opened 2016-02-29
Pairing heap

djwatson:
Even after recent improvements to arena's runs_avail tree, it still looks like it is extremely hot.  Since we've separated the concerns of address and size, and no longer use nsearch(), we can drop in a heap implementation instead for O(1) insertions and O(logn) deletes.

I experimented with a handful of other datastructures (skiplist, array-based heaps, the old rb.h with parent pointers, aux list or no, tagged parent pointers, etc), but a twopass pairing heap with aux list ended up being fastest so far.  

Benchmarked by hacking up a version of rb_bench to only test insert/delete/first.  Pairing heaps were ~7x faster by cycles, ~4x fewer memory loads.

Tested similarly with rb_test, and also canaried in prod - didn't see any leaks or crashes.  Canary showed 1% cpu improvement, 2% tail latency improvement.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -95,6 +95,7 @@ C_SRCS := $(srcroot)src/jemalloc.c \
 	$(srcroot)src/mutex.c \
 	$(srcroot)src/nstime.c \
 	$(srcroot)src/pages.c \
+	$(srcroot)src/ph.c \
 	$(srcroot)src/prng.c \
 	$(srcroot)src/prof.c \
 	$(srcroot)src/quarantine.c \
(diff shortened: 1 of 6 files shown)


jasone:
This looks really cool.  It's probably going to be a few days before I have time to take a close look, but I will definitely get to it.

jasone:
Merged with a couple bug fixes, build issue, and formatting cleanups, as well as a minor refactor diff (```ph_merge_ordered()```) and a diff to also convert ```bin->runs```.  See 6bafa6678fc36483e638f1c3a0a9bf79fb89bfc9, 34dca5671fec8c592f1ca80ce11dc808cf6b83ed, f8d80d62a8765c54aaa9433148fd112f7c794734, 4a0dbb5ac844830ebd7f89af20203a574ce1b3da, and 613cdc80f6b61f698b3b0c3f2d22442044473f9b.

Thanks!

PR #353, opened 2016-03-09
Hotpatch out remaining fast-path option branches

djwatson:
commit f4a0f32 sped up the fast-path substantially, but we still have
an init check (slow_path), and a prof / prof_active check.  Using
dynamic runtime code patching, we can remove all of them.

This depends on asm goto in gcc, and static global addresses, so optimize
mode is also required.  Code is provided for x86_64 only, but
the same technique should work on most arches.

Note that although we patch the option, currently running threads
may still take the old branch for a while until the cache flushes,
so this is only an optimization.

Testing:

In microbenchmarks, this is a 2% cpu improvement.  It's so small it is hard
to tell in production though.  By comparison, sized-deallocations are 10% cpu
in a microbenchmark.

linux's perf utility was used to view currently executing asm.  Used mallctl
to flip prof_active on/off, checked that it appeared to be working correctly.

Tested building in debug / opt mode, but don't have anything else besides
x86_64 to check that the compile options are correct.

### DIFF
--- a/include/jemalloc/internal/runtime_patch.h
+++ b/include/jemalloc/internal/runtime_patch.h
@@ -1,0 +1,106 @@
+#ifdef JEMALLOC_H_TYPES
+
+#if	defined(JEMALLOC_RUNTIME_PATCHING) && \
+	defined(__x86_64__) && \
+	(__GNUC__ > 4 || (__GNUC_MINOR__ >= 5 && __GNUC__ == 4))
+#  define X86_64_PATCHING
+#endif
+
+typedef struct jmp_desc_s jmp_desc_t;
+
+#endif /* JEMALLOC_H_TYPES */
+/******************************************************************************/
+#ifdef JEMALLOC_H_STRUCTS
+
+enum PATCH_OPTIONS {
+  MALLOC_SLOW = 0,
+  OPT_PROF = 1,
+};
+
+struct jmp_desc_s {
+        intptr_t option;
+	void *jump_from;
+	void *jump_to;
+} __attribute__((packed));
+
+#endif /* JEMALLOC_H_STRUCTS */
+/******************************************************************************/
+#ifdef JEMALLOC_H_EXTERNS
+
+#ifdef X86_64_PATCHING
+void malloc_patch_option(intptr_t key, bool* option);
+#endif
+
+#endif /* JEMALLOC_H_EXTERNS */
+/******************************************************************************/
+#ifdef JEMALLOC_H_INLINES
+
+#ifdef X86_64_PATCHING
+
+/* Must *always* be inlined */
+#define PATCH_FUNC_OFF(key)						\
+  JEMALLOC_ALWAYS_INLINE_C JEMALLOC_ATTR(unused) bool			\
+malloc_##key##_default_off(bool* option) {				\
+  asm goto (								\
+	    /* 8-byte noop, enough space for a 32-bit jmp later */	\
+	    "0: .byte 0x0f,0x1f,0x84,0x00,0x00,0x00,0x00,0x00;"		\
+	    /* Save key in jmp_table for later patching */		\
+	    ".pushsection jmp_table, \"a\";"				\
+	    ".quad %c0, 0b, %l1;"					\
+	    ".popsection"						\
+	    : : "i" (key) : : foo);					\
+  return false;								\
+  if (0) {								\
+  foo:									\
+    return true;							\
+  }									\
+}
+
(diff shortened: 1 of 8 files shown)


djwatson:
It may be wise to make this a configure option also?

jasone:
Re: configure option, at a minimum we need a configure test to conditionally enable this code for the supported platform (gcc+x86_64).

djwatson:
@interwq 's comments

- Attempted to add to configure.ac, could probably still be more pedantic. 

- The prof_active check does seem to work correctly in prof_alloc_prep, but we have to use the var directly and not the param.  Could probably use an additional cleanup diff there, the param is always 'prof_active_get_unlocked'

include/jemalloc/internal/prof.h
```diff
@@ -458,16 +458,16 @@ prof_sample_accum_update(tsd_t *tsd, size_t usize, bool update,
...
 {
 	prof_tctx_t *ret;
 	prof_tdata_t *tdata;
 	prof_bt_t bt;
 
 	assert(usize == s2u(usize));
 
-	if (!prof_active || likely(prof_sample_accum_update(tsd, usize, update,
-	    &tdata)))
+	if (!malloc_option_default_on(&prof_active) ||
```
jasone: This isn't safe, because there are code paths that require consistency re: ```prof_active``` across a series of function invocations, e.g. in ```irealloc_prof()```.  Should I just revert this particular change, or is there some other way we can accomplish this safely?

include/jemalloc/internal/runtime_patch.h
```diff
@@ -0,0 +1,95 @@
...
+#endif /* JEMALLOC_H_EXTERNS */
+/******************************************************************************/
+#ifdef JEMALLOC_H_INLINES
+
+#ifdef X86_64_PATCHING
+
+/* Must *always* be inlined */
+JEMALLOC_ALWAYS_INLINE_C bool
+malloc_option_default_off(bool* option) {
+	asm goto (
```
jasone: This doesn't compile, using gcc 5.3.1 on Ubuntu 16.04 LTS.

```
gcc -std=gnu99 -Wall -Werror=declaration-after-statement -Wsign-compare -pipe -g3 -fvisibility=hidden -O3 -funroll-loops -fPIC -DPIC -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -o src/jemalloc.pic.o src/jemalloc.c
In file included from include/jemalloc/internal/jemalloc_internal.h:523:0,
                 from src/jemalloc.c:2:
include/jemalloc/internal/runtime_patch.h: In function ‘free’:
include/jemalloc/internal/runtime_patch.h:39:2: warning: asm operand 0 probably doesn’t match constraints
  asm goto (
  ^
include/jemalloc/internal/runtime_patch.h:39:2: warning: asm operand 0 probably doesn’t match constraints
  asm goto (
  ^
include/jemalloc/internal/runtime_patch.h:39:2: error: impossible constraint in ‘asm’
  asm goto (
  ^
include/jemalloc/internal/runtime_patch.h:39:2: error: impossible constraint in ‘asm’
  asm goto (
  ^
```

include/jemalloc/internal/prof.h
djwatson: Hmm I see, I'll change it back to the previous approach.

include/jemalloc/internal/runtime_patch.h
djwatson: Shoot, definitely worked on 4.x gcc, will have to dive in to it.

djwatson:
* rebased
* dropped PROF_ACTIVE check, it doesn't make nearly as big of a difference as the other two (.5% vs 3-4% in microbench)
* I was never able to repro the build failure, even with the same OS/compiler.  Changed the functions to be macros instead, which seems to also work for debug builds - the macros provide a constant to emit in inline assembly, instead of relying on the inliner to provide a constant address for global values.

djwatson:
5.0 is looking big, would prefer to punt until later, possibly a minor release since this doesn't change any APIs.  

A recent paper on this technique: http://www.cs.indiana.edu/~rrnewton/papers/pldi16-crossmod.pdf
May need to add align statements to the asm, but otherwise looks like it should work fine.

May conflict with #501 , since it might remove some of the slow path branches needed to support reentrancy.  The config option branches could probably still be removed though

jasone:
Punting to a later release, per @djwatson's recommendation.

jasone:
Closing, now that #539 is tracking the hotpatching.

PR #357, opened 2016-03-15
Enable -Wsign-compare warnings.

cpeterso:
Warn about comparisons between signed and unsigned integers. Mozilla enables -Wsign-compare for all C/C++ code in Gecko, including jemalloc and other third-party libraries. jemalloc does not have any -Wsign-compare warnings with clang on OS X, gcc on Linux, or gcc for Android.

gcc's -Wall enables -Wsign-compare for C++, but not C. clang's -Wall does not enable -Wsign-compare. Both clang and gcc enable -Wsign-compare for C and C++ as part of -Wextra.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -141,6 +141,7 @@ if test "x$CFLAGS" = "x" ; then
     JE_CFLAGS_APPEND([-Wall])
     JE_CFLAGS_APPEND([-Werror=declaration-after-statement])
     JE_CFLAGS_APPEND([-Wshorten-64-to-32])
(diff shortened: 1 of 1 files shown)


jasone:
Integrated as 18903c592fdbf2384b59051bd251d234e84647af.  Note that this caused some warnings in test code, which are refactored out in 22af74e10615ce6b6898ae38a378af27757f9e16.

cpeterso:
Sorry! I don't know how I missed the warnings when I compiled and ran the tests.

PR #361, opened 2016-03-23
MSVC project fixes

rustyx:
Minor fixes for MSVC

### DIFF
--- a/include/jemalloc/internal/util.h
+++ b/include/jemalloc/internal/util.h
@@ -72,6 +72,7 @@
 #  if JEMALLOC_GNUC_PREREQ(4, 6) ||					\
       JEMALLOC_CLANG_HAS_BUILTIN(__builtin_unreachable)
 #	define unreachable() __builtin_unreachable()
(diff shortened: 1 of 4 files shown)


include/jemalloc/internal/util.h
```diff
@@ -72,6 +72,7 @@
 #  if JEMALLOC_GNUC_PREREQ(4, 6) ||					\
       JEMALLOC_CLANG_HAS_BUILTIN(__builtin_unreachable)
 #	define unreachable() __builtin_unreachable()
+#	define JEMALLOC_HAS_UNREACHABLE
 #  else
 #	define unreachable()
```
jasone: What if we instead define ```unreachable()``` as a call to abort() here?  Does that avoid the compilation warning?

rustyx: It does.

jasone:
Resolved by 232b13d86298b9eafc36b0610d7965a95bda0679 and af3184cac0e0c70045d8158b9c176696f2ca1090.  Thanks!

PR #362, opened 2016-03-26
Remove unused arenas_extend() function declaration.

cpeterso:
The `arenas_extend()` function was renamed to `arenas_init()` in commit 8bb3198f72fc7587dc93527f9f19fb5be52fa553, but its function declaration was not removed from jemalloc_internal.h.in.


jasone:
Thanks!

PR #363, opened 2016-03-26
Add JEMALLOC_ALLOC_JUNK and JEMALLOC_FREE_JUNK macros

cpeterso:
Replace hardcoded `0xa5` and `0x5a` junk values with `JEMALLOC_ALLOC_JUNK` and
`JEMALLOC_FREE_JUNK` macros, respectively.

* Or do you prefer the hardcoded `0xa5` and `0x5a` values? I think the macro names make the code clearer. I left the hardcoded values in test/unit/junk.c as a sanity check that the new junk macros have the expected values.
* Is there a more appropriate header file in which to #define these macros than jemalloc/internal/util.h?

### DIFF
--- a/include/jemalloc/internal/tcache.h
+++ b/include/jemalloc/internal/tcache.h
@@ -382,7 +382,8 @@ tcache_alloc_large(tsd_t *tsd, arena_t *arena, tcache_t *tcache, size_t size,
 		if (likely(!zero)) {
 			if (slow_path && config_fill) {
 				if (unlikely(opt_junk_alloc))
-					memset(ret, 0xa5, usize);
+					memset(ret, JEMALLOC_ALLOC_JUNK,
+					    usize);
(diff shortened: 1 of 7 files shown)


jasone:
I'd prefer that if we make this change, all instances of ```0xa5``` and ```0x5a``` are replaced by the macros.

It looks like there may be some places that this diff increases lines past the 80 character limit.

cpeterso:
Thanks. I have replaced `0xa5` and `0x5a` in test/unit/junk.c and fixed the line wrapping for 8-space tabs.

Is jemalloc/internal/util.h a reasonable header file in which to #define these JEMALLOC_*_JUNK macros?

jasone:
Yes, util.h is as good a place as any right now for the macros.  Integrated as a82070ef5fc3aa81fda43086cdcc22bfa826b894.  Thanks!

PR #364, opened 2016-03-26
Fix -Wunreachable-code warning in malloc_vsnprintf().

cpeterso:
Variables `s` and `slen` are declared inside a switch statement, but outside a case scope. clang reports these variable definitions as "unreachable", though this is not really meaningful in this case. This is the only -Wunreachable-code warning in jemalloc.

src/util.c:501:5 [-Wunreachable-code] code will never be executed

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -408,6 +408,8 @@ malloc_vsnprintf(char *str, size_t size, const char *format, va_list ap)
 			int prec = -1;
 			int width = -1;
 			unsigned char len = '?';
+			char *s;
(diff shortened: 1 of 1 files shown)


jasone:
Interesting.  I'm pretty sure the original is valid C code, and that clang is in error, since the variables are not initialized at their declaration sites (i.e. no "unreachable" code even exists).  That said, if the compiler generates bogus warnings, I'd also worry about it generating bogus code. ;-)

PR #377, opened 2016-04-18
Doc typo fixes; capitalisation of "or" to "OR"

hitstergtd:
**Changes:**
* Fix typos
* Uppercase `or` in the context of bitwise operations

### DIFF
--- a/doc/jemalloc.xml.in
+++ b/doc/jemalloc.xml.in
@@ -239,7 +239,7 @@
       <function>nallocx<parameter/></function> functions all have a
       <parameter>flags</parameter> argument that can be used to specify
       options.  The functions only check the options that are contextually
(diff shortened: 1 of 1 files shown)


jasone:
The OR change seems arbitrary, but the other fixes are good.  Can you please drop the OR part?

hitstergtd:
@jasone,
Thanks! The OR change is on the basis that when bitwise operators are usually referred to in their uppercase form and doesn't feel like it's a running part of the sentence. I checked a few sources to make extra sure.

Anyway, if you still want me to drop it, it's not a problem!

jasone:
Integrated as eb68842dafb6d86e8310b1db9ca0467d577091b1; thanks!

PR #378, opened 2016-04-18
Update private_symbols.txt

rajatgoel:
Add 4 missing symbols


jasone:
I got to wondering how I'd missed this, and discovered that the commands I was using were flawed.  That motivated ab0cfe01fa354597d28303952d3b0f87d932f6d6.  Thanks!

PR #380, opened 2016-04-23
Added stats about number of bytes cached in tcache currently.

interwq:
- Shows number of bytes currently in tcache in stats_print.
- Also removed a bit of redundent stats_merge logic in tcache_destroy.


jasone:
This isn't thread-safe, which is why I never added these stats.

interwq:
It's ture that we won't be able to get the exact number of ncached. But we hold the arena lock when going through the tcaches, so no threads will be removed from tcache_ql in this arena. I don't think it will cause any crashes. Is the concern about the inaccuracy?

jasone:
Yes, inaccuracy is my concern.  It's difficult to explain how far off the stats could be, which makes them difficult to reliably interpret.

interwq:
Got it. In that case, do you think we should report the cached bytes in the current calling thread?

jasone:
That would be reasonable.  We could also support reporting cached bytes for explicit tcaches.

interwq:
Agreed. I'll change the diff accordingly.

cmuellner:
Instead of removing the lock, you could also use atomics (e.g. via GCC's atomic API https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html or as an actual implementation for ARM http://lxr.free-electrons.com/source/arch/arm/include/asm/atomic.h).
This way the compiler creates optimised code for updating a variable atomically (i.e. thread-safe).

interwq:
Thanks for the input @cmuellner! Here atomics are too expensive for the purpose as the counter is touched on fast path.

After some further discussion with @jasone and @davidtgoldblatt, I feel the current reading w/o sync should be sufficient: we don't have any synchronization around tbin->ncached, however the underlying hardware (MESI protocol) should always help us getting the updated cacheline (and bring it into a shared state). We won't get any strong ordering guarantee but it's not required for the stats in this case. So I don't think we need to worry about inaccuracy.

interwq:
Closing this. Rebased and moved to https://github.com/jemalloc/jemalloc/pull/550

PR #389, opened 2016-05-27
Fix mixed decl in nstime.c

ronawho:
Fix mixed decl in the gettimeofday() branch of nstime_update()


ronawho:
I ran into this while trying to build on OSX (which doesn't support
clock_gettime) with gnu. I'm not sure why clang doesn't catch this as well.

As a side note, if having a monotonic clock on OSX is important, you might be
able to use mach_time.

jasone:
Integrated as 48384dc2d841bac9ab1e1ef24addbccabed43c20.

Issue #390, opened 2016-06-01
Infinite Loop in jemalloc 4.2.0


cferris1000:
The newly released version of jemalloc can cause an infinite loop which leads to crashes when the stack overflows.

The circumstances that cause the crash:

- TLS is not being used to handle thread local variables. Instead, the code is using pthread primitives.
- A new thread is started, but the program has been running for a while with other threads doing allocations.
- The new thread calls into an allocation routine, which eventually tries to allocate the memory needed for  thread local variable data.
- Another thread winds up setting runcur to NULL for the bin in which the new thread is about to allocate.
- In the function arena_malloc_small, since runcur is NULL, a call to arena_bin_malloc_hard is made instead of arena_run_reg_alloc. The arena_bin_malloc_hard call eventually leads to a call to get the tsd, which tries to allocate the memory again and infinite loop.

To verify this, I took the jemalloc 4.2.0 release, built it with JEMALLOC_TLS undefined in include/jemalloc/internal/jemalloc_internal_defs.h, and changed the number of arenas to 1 to make it more likely to trip the problem. I built this on a ubuntu derived version of linux, using --with-jemalloc_prefix=je_. I did the same with 4.1.0 and 4.1.1, but neither exhibits this problem.

I compiled my small test program like so:

g++ -o free-4.2.0 -std=c++11 free.cpp -Ijemalloc-4.2.0/include jemalloc-4.2.0/lib/libjemalloc.a -lpthread

The test code is:

```
#include <stdio.h>

#include <thread>
#include <vector>

#include <jemalloc/jemalloc.h>

int main() {
  for (size_t i = 0; i < 100; i++) {
    std::thread thread([](){
      for (size_t k = 0; k < 10000; k++) {
        std::vector<void*> vector;
        for (size_t j = 0; j < 10000; j++) {
          vector.push_back(je_malloc(120));
        }
        for (auto ptr : vector) {
          je_free(ptr);
        }
      }
    });
    thread.detach();
  }
  for (size_t i = 0; i < 1000; i++) {
    std::thread thread([](){
      void* ptr = je_malloc(120);
      je_free(ptr);
    });
    thread.join();
  }
}
```

glandium:
FWIW, I suspect something similar is happening on Windows. We have stack overflow crashes in jemalloc when running some Firefox unit tests with jemalloc 4.2.0.

jasone:
[...]

PR #397, opened 2016-06-08
Fix support for the PGI compiler

ronawho:
* Work around a weird pgi bug in test/unit/math.c

* Don't use compact red-black trees under pgi. I'm not sure if this is a
  red-black tree bug a pgi bug, but since most of the red-black trees seem to
  have been replaced with pairing-heaps I didn't investigate very hard.

* Note that this also relies on https://github.com/jemalloc/jemalloc/issues/390
  because pgi doesn't have support for TLS.

Passes `make check` with pgi 15.10.0 and 16.4.0 (-O and -O3)

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal.h.in
+++ b/include/jemalloc/internal/jemalloc_internal.h.in
@@ -155,7 +155,9 @@ static const bool config_cache_oblivious =
 #endif
 
 #include "jemalloc/internal/ph.h"
+#ifndef __PGI
 #define	RB_COMPACT
+#endif
 #include "jemalloc/internal/rb.h"
 #include "jemalloc/internal/qr.h"
 #include "jemalloc/internal/ql.h"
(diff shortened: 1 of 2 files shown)


ronawho:
Since pgi doesn't pretend to be gcc, jemalloc defaults to throwing -O instead of -O3. For pgi this is basically -O2 minus vectorization. I thought about updating the [configure test] (https://github.com/ronawho/jemalloc/blob/fix-pgi-support/configure.ac#L756) to try to add -O3 first, but I wasn't sure if any other compilers currently rely on -O being used. I don't personally care about this because we set the optimization level we want to build with, but others using the default settings could run into this.

jasone:
The ```RB_COMPACT``` optimization is mainly for space savings, which aren't particularly important in the remaining use cases, so this change is fine.  As for the underlying cause, perhaps there's a subtle aliasing issue; packing data bits into pointers skirts the edges of what's legal with out using enums for casting, and it's possible that pgi is making a valid aliasing assumption that ```RB_COMPACT``` violates.

PR #400, opened 2016-06-09
Add an AppVeyor config

glandium:
This builds jemalloc and runs all checks with:
- MSVC 2015 64-bits
- MSVC 2015 32-bits
- MINGW64 (from msys2)
- MINGW32 (from msys2)

Normally, AppVeyor configs are named appveyor.yml, but it is possible to
configure the .yml file name in the AppVeyor project settings such that
the file stays "hidden", like typical travis configs.

Considering how many times MSVC builds have broken, I'd appreciate if you could setup AppVeyor such that it runs when you push to the canonical jemalloc repository (travis-ci would be useful too, FWIW).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,28 @@
+version: '{build}'
+
+environment:
+  matrix:
+  - MSYSTEM: MINGW64
+    CPU: x86_64
+    MSVC: amd64
+  - MSYSTEM: MINGW32
+    CPU: i686
+    MSVC: x86
+  - MSYSTEM: MINGW64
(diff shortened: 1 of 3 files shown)


jasone:
I'm going to be mostly offline until September, but this helps address a burning issue, so integrating it will be one of my first tasks upon return.  Manually testing across a half dozen operating systems has been a significant strain, and MSVC has sadly slipped through the cracks until now.

jasone:
I don't know anything about AppVeyor.  Can you please create a tracking issue for getting it to run automatically?

Issue #398, opened 2016-06-09
Build failure on the dev branch with MSVC


glandium:
The errors:
```
 00:54:57     INFO -  c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc/internal/extent.h(227): error C2036: 'void *': unknown size

 00:54:57     INFO -  c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc/internal/extent.h(234): error C2036: 'void *': unknown size

 00:54:57     INFO -  c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc/internal/extent.h(235): error C2036: 'void *': unknown size

 00:54:57     INFO -  c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc/internal/extent.h(242): error C2036: 'void *': unknown size

```

Some other warnings:
```
 00:54:57     INFO -  Warning: C4013 in c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc\internal\prng.h: 'je_atomic_add_uint64' undefined; assuming extern returning int

 00:54:57     INFO -  c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc/internal/prng.h(63): warning C4013: 'je_atomic_add_uint64' undefined; assuming extern returning int

 00:54:57     INFO -  Warning: C4013 in c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc\internal\prng.h: 'je_atomic_cas_uint64' undefined; assuming extern returning int

 00:54:57     INFO -  c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc/internal/prng.h(65): warning C4013: 'je_atomic_cas_uint64' undefined; assuming extern returning int

 00:54:57     INFO -  Warning: C4146 in c:\builds\moz2_slave\try-w32-0000000000000000000000\build\src\memory\jemalloc\src\include\jemalloc\internal\extent.h: unary minus operator applied to unsigned type, result still unsigned

[...]

PR #401, opened 2016-06-09
Fix extent_*_get to build with MSVC

glandium:
This fixes the errors described in issue #398, but only actually fixes the build on MSVC 64-bits builds. 32-bits builds are still busted because of the je_atomic_*_uint64 symbols, and it seems 32-bits builds are busted on other OSes for the same reason too.

### DIFF
--- a/include/jemalloc/internal/extent.h
+++ b/include/jemalloc/internal/extent.h
@@ -224,22 +224,22 @@ JEMALLOC_INLINE void *
 extent_before_get(const extent_t *extent)
 {
 
-	return ((void *)(uintptr_t)extent->e_addr - PAGE);
+	return ((void *)((uintptr_t)extent->e_addr - PAGE));
 }
 
 JEMALLOC_INLINE void *
 extent_last_get(const extent_t *extent)
 {
 
-	return ((void *)(uintptr_t)extent->e_addr + extent_size_get(extent) -
-	    PAGE);
+	return ((void *)((uintptr_t)extent->e_addr + extent_size_get(extent) -
+	    PAGE));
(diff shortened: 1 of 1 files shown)


dcci:
I had the same problem but I fixed that in a slightly different way.

```
$ git diff
diff --git a/include/jemalloc/internal/extent.h b/include/jemalloc/internal/extent.h
index d7944c1..8843df7 100644
--- a/include/jemalloc/internal/extent.h
+++ b/include/jemalloc/internal/extent.h
@@ -224,14 +224,14 @@ JEMALLOC_INLINE void *
 extent_before_get(const extent_t *extent)
 {

-       return ((void *)(uintptr_t)extent->e_addr - PAGE);
+       return ((char *)(uintptr_t)extent->e_addr - PAGE);
 }

 JEMALLOC_INLINE void *
 extent_last_get(const extent_t *extent)
 {

-       return ((void *)(uintptr_t)extent->e_addr + extent_size_get(extent) -
+       return ((char *)(uintptr_t)extent->e_addr + extent_size_get(extent) -
            PAGE);
 }

@@ -239,7 +239,7 @@ JEMALLOC_INLINE void *
 extent_past_get(const extent_t *extent)
 {

-       return ((void *)(uintptr_t)extent->e_addr + extent_size_get(extent));
+       return ((char *)(uintptr_t)extent->e_addr + extent_size_get(extent));
 }

 JEMALLOC_INLINE bool
```

jasone:
Fixed in 0dad5b77196903ba2d6bc64a0d9fdda8e2e9ad78.

Issue #402, opened 2016-06-09
prng.h uses atomic_*_uint64 on all platforms, but it's only declared in 64-bits platforms


glandium:
in atomic.h, atomic_*_uint64 functions are behind a

```
#if (LG_SIZEOF_PTR == 3 || LG_SIZEOF_INT == 3)
```

but prng.h uses them unconditionally.

jasone:
This is roughly the third time I've made this mistake.  I wish there were a way to catch this on 64-bit systems.

jasone:
Partial fix integrated as c2942e2c0e097e7c75a3addd0b9c87758f91692e.

jasone:
In practice, c2942e2c0e097e7c75a3addd0b9c87758f91692e appears to be sufficiently general.  We have a multiplicity of implementations based on C11, gcc intrinsics, and Windows APIs that cover pretty much all hypothetically supported systems going back ~10 years.

I cherry-picked the fix to rel-4.3.0 as af33e9a59735a2ee72132d3dd6e23fae6d296e34.

PR #404, opened 2016-06-09
Define 64-bits atomics unconditionally

glandium:
They are used on all platforms in prng.h.

This fixes issue #402.


jasone:
Equivalent to c2942e2c0e097e7c75a3addd0b9c87758f91692e.

PR #407, opened 2016-06-14
Avoid doing void pointer arithmetic in extent.h

ronawho:
Void pointer arithmetic isn't legal. There's a [gcc extension][1] that allows
it, but not all compilers support that extension. It looks like the uses in
extent.h could just have been typos though. This just adds parens so that the
math is done on a uintptr_t before casting the result back to (void *).

[1]: https://gcc.gnu.org/onlinedocs/gcc/Pointer-Arith.html#Pointer-Arith


ronawho:
Whoops, just noticed that this is a duplicate of https://github.com/jemalloc/jemalloc/pull/401. Closing

PR #408, opened 2016-06-14
Add support for Cray compiler wrappers

ronawho:
This just throws `-dynamic` for the integration/stress tests and fixes librt
detection when a Cray compiler wrapper is being used.

Cray systems come with compiler wrappers to simplify building massively
parallel applications. CC is the C++ wrapper, and cc is the C wrapper.

The wrappers call the base {Cray, Intel, PGI, or GNU} compiler with vendor
specific flags. The "Programming Environment" (prgenv) that's currently loaded
determines the base compiler. e.g. compiling with gnu looks something like:

    module load PrgEnv-gnu
    cc hello.c

On most systems the wrappers default to `-static` linking, which causes them to
only look for static libraries, and not for any dynamic ones (even if the
dynamic version was explicitly listed.) This meant that the wrapper failed to
link in the jemalloc .so for the integration and stress tests. To work around
that, this just throws -dynamic for those those tests.

Additionally, `-static` usually causes the wrapper to throw `-lrt` to the base
compiler, which caused jemalloc to detect that it didn't rely on librt.
However, `-dynamic` doesn't automatically throw `-lrt`. This meant that if
jemalloc was built with `-static`, the integration and stress tests that are
running with `-dynamic` wouldn't detect that librt was needed. This just runs
the librt check a second time (ignoring the autoconf cache) with -dynamic.

With this `make check` passes for:
 - PrgEnv-gnu
 - PrgEnv-intel
 - PrgEnv-pgi

PrgEnv-cray still needs more work, which will be added in a separate patch.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -294,7 +294,7 @@ $(objroot)test/unit/%$(EXE): $(objroot)test/unit/%.$(O) $(TESTS_UNIT_LINK_OBJS) 
 
 $(objroot)test/integration/%$(EXE): $(objroot)test/integration/%.$(O) $(C_TESTLIB_INTEGRATION_OBJS) $(C_UTIL_INTEGRATION_OBJS) $(objroot)lib/$(LIBJEMALLOC).$(IMPORTLIB)
 	@mkdir -p $(@D)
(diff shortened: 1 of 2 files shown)


jasone:
Thanks, integrated as ccd64160736c6e94f84a8bf045ecbbc6a4228604 and ae3314785bf9726e5a97e5c98f70dcb12e6a7a90.

PR #413, opened 2016-06-22
Fix MSVC error C2036 in extent.h

rustyx:
MSVC 2015 throws an error: `error C2036: 'void *': unknown size`


rustyx:
Oops, duplicate of #401. Closing.

PR #419, opened 2016-07-03
Minor spell checking in comments

ottok:


jasone:
I prefer not to make non-functional changes to jeprof.in, because it makes later comparison with the upstream ```pprof``` more difficult.

PR #423, opened 2016-07-07
Add support for the Cray compiler

ronawho:
This adds support for the Cray compiler (versions 8.4.0 and newer). 8.4 was the
first version to add support for gnu style inline assembly, so there's not much
hope getting jemalloc to work with earlier versions.

There's an inlining bug in 8.4.X (that has been fixed for 8.5) so this patch
throws `-hnognu -hipa2` to work around the bug if 8.4.X is being used.

This also disables automatic dependency generation, because Cray doesn't
support the `-MM` or `-MT` dependency generation flags.

Additionally, Cray doesn't support `-Werror` so this adds the equivalent
`-herror_on_warning` everywhere that `-Werror` is thrown so that jemalloc
correctly detects when attributes don't actually work.

With this Cray 8.4.0 and 8.5.1 build cleanly and pass `make check`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -323,6 +323,10 @@ if test "x$je_cv_cray_prgenv_wrapper" = "xyes" ; then
   TEST_LD_MODE='-dynamic'
 fi
 
+if test "x${je_cv_cray}" = "xyes" ; then
+  CC_MM=
+fi
+
 AN_MAKEVAR([AR], [AC_PROG_AR])
 AN_PROGRAM([ar], [AC_PROG_AR])
 AC_DEFUN([AC_PROG_AR], [AC_CHECK_TOOL(AR, ar, :)])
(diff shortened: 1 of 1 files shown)


ronawho:
Do you have any interest in having this and the other recent portability PRs
(#397, #408, #409) in the next point release?

If so, I'd be happy to make PRs with them cherry-picked to master.

jasone:
Strictly speaking, we'd need to make it 4.3.0 rather than 4.2.2 if we include these.  I'm open to that if it's useful to you.  By the way, I'm not going to be around my computer much until September, so chances are the next 4.x release will happen mid-September.

ronawho:
It'd definitely be useful to me, and I think the [__builtin_unreachable](https://github.com/jemalloc/jemalloc/pull/409) one could be useful to others. I'll open PRs at some point before September.

Thanks!

Issue #420, opened 2016-07-05
Zone allocator registration code failing on OSX 10.12 beta 2


glandium:
Heads up: Running Firefox or the rust compiler, the former using the same zone allocator registration code, the latter using jemalloc 3, get stuck after printing the following error message:
> malloc: *** malloc_zone_unregister() failed for 0xaddress

This roots from the malloc zone registration code in zone.c. I'm not sure what's going on yet, and the sources for libmalloc in OSX 10.12 are not available on opensource.apple.com yet, so it's not going to be straightforward to fix.

glandium:
So, looking at the disassembly of libsystem_malloc.dylib, it turns out that, contrary to previous versions, the default zone is a static struct in libsystem_malloc.dylib's data segment.

malloc_default_zone always returns a pointer to that static struct, not that of the first registered zone. So https://github.com/jemalloc/jemalloc/blob/dev/src/zone.c#L249 always gets that pointer, and consequently, https://github.com/jemalloc/jemalloc/blob/dev/src/zone.c#L275 is always true, leaving us in an infinte loop.

On top of that, the default zone is used directly by malloc() and other functions, such that even if we somehow fix the loop to correctly stop when we get our zone to be the first registered one, our zone would still not be used.

It looks like unless we can get Apple to change libsystem_malloc somehow, the only option would be to modify the default zone itself, which may or may not require an mprotect (it doesn't look like so from using 10.12's libsystem_malloc.dylib on 10.11, but maybe there were dynamic linker changes in 10.12 as well...).

jasone:
I wonder if this could be the end of the line for full malloc replacement on OS X...

glandium:
Turns out it's not as bad as I thought. Explanations in the upcoming PR.

jasone:
Fixed by #427.  Thanks!

glandium:
How about cherry-picking it on master?

imajes:
Hey- 👍 for getting it merged to master, and rolling a release. This will also bite ruby installs as it's relying on jemalloc too -- so it'd be very nice to get this cherry-picked and into a release that we can push to homebrew. It's probable that the release date for sierra will be at the next event, which is sept 7th. is it viable to get this in a 4.2.1 patch release? 

Thanks for all the works!

dpereira411:
I'm using ruby 1.9.3 and getting the exact same error. Any way to fix this without upgrading? Maybe compiling with different sources?

jasone:
[...]

PR #427, opened 2016-07-08
Fix for issue #420

glandium:
This should probably go to master too.


imajes:
wasn't sure if this was obvious but it pretty much looks like this is to do with system integrity checking - dtruss showed the exact problem that @glandium fixed, but it also added a hint to the new sandboxing macOS has been introducing...

PR #431, opened 2016-07-12
Fix a bug in __builtin_unreachable configure check

ronawho:
In 1167e9e, I accidentally tested `je_cv_gcc_builtin_ffsl` instead of
`je_cv_gcc_builtin_unreachable` (copy-paste error), which meant that
`JEMALLOC_INTERNAL_UNREACHABLE` was always getting defined as `abort` even if
`__builtin_unreachable` support was detected.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1095,7 +1095,7 @@ void foo (void) {
 		foo();
 	}
 ], [je_cv_gcc_builtin_unreachable])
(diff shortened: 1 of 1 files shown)


jasone:
Integrated as c096ccfe111ca3c849a0104517a2a1f123985307.  Thanks!

PR #432, opened 2016-07-12
Cherry-pick support for the PGI compiler to master

ronawho:
Cherry-pick fbd7956 and 8a1a794 to master

* Work around a weird pgi bug in test/unit/math.c

* Don't use compact red-black trees under pgi. I'm not sure if this is a
  red-black tree bug a pgi bug, but since most of the red-black trees seem to
  have been replaced with pairing-heaps I didn't investigate very hard.

* Note that this relies on https://github.com/jemalloc/jemalloc/commit/09d7bdb
  because pgi doesn't have support for TLS.

Passes `make check` with pgi 15.10.0 and 16.5.0 (default and with -O3)

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal.h.in
+++ b/include/jemalloc/internal/jemalloc_internal.h.in
@@ -162,7 +162,9 @@ static const bool config_cache_oblivious =
 #endif
 
 #include "jemalloc/internal/ph.h"
+#ifndef __PGI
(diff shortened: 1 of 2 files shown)


jasone:
Integrated as 50a865e15adbc2c28dc92ac4e9c63b63835ac77b and 5acef864f22384064b5bfb4ad3b50114b9ef214b.  Thanks!

Issue #266, opened 2015-08-30
cannot build on Mac Yosemite with any compiler (linker error)


jeffhammond:
I attempted to build the latest version of jemalloc on my Mac.  It fails with GCC 5.2.0, Clang 3.6, and Intel 15.

This is how I configured with GCC 5:
```
jrhammon-mac01:build jrhammon$ ../configure CC=gcc-5 --prefix=/Users/jrhammon/Work/MEMORY/MEMKIND/install-jemalloc-gcc5 && make -j4
```

The error with GCC 5.2 is:
```
jrhammon-mac01:build jrhammon$ make
gcc-5 -shared -Wl,-install_name,/Users/jrhammon/Work/MEMORY/MEMKIND/install-jemalloc-gcc5/lib/libjemalloc.2.dylib  -o lib/libjemalloc.2.dylib src/jemalloc.pic.o src/arena.pic.o src/atomic.pic.o src/base.pic.o src/bitmap.pic.o src/chunk.pic.o src/chunk_dss.pic.o src/chunk_mmap.pic.o src/ckh.pic.o src/ctl.pic.o src/extent.pic.o src/hash.pic.o src/huge.pic.o src/mb.pic.o src/mutex.pic.o src/pages.pic.o src/prof.pic.o src/quarantine.pic.o src/rtree.pic.o src/stats.pic.o src/tcache.pic.o src/util.pic.o src/tsd.pic.o src/valgrind.pic.o src/zone.pic.o  -lpthread 
0  0x10b974a44  __assert_rtn + 144
1  0x10b99c5d5  mach_o::relocatable::Parser<x86_64>::parse(mach_o::relocatable::ParserOptions const&) + 3175
2  0x10b97ff81  mach_o::relocatable::Parser<x86_64>::parse(unsigned char const*, unsigned long long, char const*, long, ld::File::Ordinal, mach_o::relocatable::ParserOptions const&) + 375
0  0x10b974a44  __assert_rtn + 144
3  0x10b9c2bf8  ld::tool::InputFiles::makeFile(Options::FileInfo const&, bool) + 796
1  0x10b99c5d5  mach_o::relocatable::Parser<x86_64>::parse(mach_o::relocatable::ParserOptions const&) + 3175
2  0x10b97ff81  mach_o::relocatable::Parser<x86_64>::parse(unsigned char const*, unsigned long long, char const*, long, ld::File::Ordinal, mach_o::relocatable::ParserOptions const&) + 375
[...]

PR #433, opened 2016-07-12
Cherry-pick check for __builtin_unreachable at configure time to master

ronawho:
Cherry-pick 1167e9e (and an uncommitted bug fix in #431) to master

Add a configure check for `__builtin_unreachable` instead of basing its
availability on the `__GNUC__` version. On OS X using gcc (a real gcc, not the
bundled version that's just a gcc front-end) leads to a linker assertion:
https://github.com/jemalloc/jemalloc/issues/266

It turns out that this is caused by a gcc bug resulting from the use of
`__builtin_unreachable()`:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57438

To work around this bug, check that __builtin_unreachable() actually works at
configure time, and if it doesn't use abort() instead. The check is based on
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57438#c21.

With this `make check` passes with a homebrew installed gcc-5 and gcc-6.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1050,6 +1050,23 @@ if test "x$enable_cache_oblivious" = "x1" ; then
 fi
 AC_SUBST([enable_cache_oblivious])
 
+
+
+JE_COMPILABLE([a program using __builtin_unreachable], [
+void foo (void) {
+  __builtin_unreachable();
+}
+], [
+	{
+		foo();
+	}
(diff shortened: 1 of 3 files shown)


jasone:
Integrated as d1207f0d371eade218f7572743b5eddedc7fff94 and 38a96f07ac3b351aeb9f3d685eaf0d4f9f01195a.  Thanks!

PR #435, opened 2016-07-13
Cherry-pick support for the Cray compiler to master

ronawho:
Cherry-pick aec0753, 3ef6793, 3dee73f, and 47b34dd to master

This adds support for the Cray compiler (versions 8.4.0 and newer). 8.4 was the
first version to add support for gnu style inline assembly, so there's not much
hope getting jemalloc to work with earlier versions.

There's an inlining bug in 8.4.X (that has been fixed for 8.5) so this patch
throws `-hnognu -hipa2` to work around the bug if 8.4.X is being used.

This also disables automatic dependency generation, because Cray doesn't
support the `-MM` or `-MT` dependency generation flags.

Additionally, Cray doesn't support `-Werror` so this adds the equivalent
`-herror_on_warning` everywhere that `-Werror` is thrown so that jemalloc
correctly detects when attributes don't actually work.

With this Cray 8.4.0 and 8.5.1 build cleanly and pass `make check`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -323,6 +323,10 @@ if test "x$je_cv_cray_prgenv_wrapper" = "xyes" ; then
   TEST_LD_MODE='-dynamic'
 fi
 
+if test "x${je_cv_cray}" = "xyes" ; then
+  CC_MM=
+fi
+
 AN_MAKEVAR([AR], [AC_PROG_AR])
 AN_PROGRAM([ar], [AC_PROG_AR])
 AC_DEFUN([AC_PROG_AR], [AC_CHECK_TOOL(AR, ar, :)])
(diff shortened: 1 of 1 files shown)


ronawho:
This relies on https://github.com/jemalloc/jemalloc/pull/433 and https://github.com/jemalloc/jemalloc/pull/434. I can rebase after those go in

jasone:
Oops, wrong milestone.

jasone:
@ronawho, the dependencies have now been cherry-picked to the rel-4.3.0 branch.  Can you please rebase onto that branch?

ronawho:
@jasone rebased and retested

jasone:
Thanks!

PR #439, opened 2016-07-25
Support Debian GNU/kFreeBSD

thesam:
Handles kFreeBSD exactly like Linux since they both use GNU libc.

"make check" passes on my kFreeBSD server (64-bit).


jasone:
Integrated as 69f027b8558c30e184921ce22bb187aafcad8a02.  Thanks!

PR #444, opened 2016-08-14
Move pages_boot() call outside of init_lock; fixes #443

### LINKED ISSUE #443: Deadlock when using LD_PRELOAD libraries that hook the open() system call
In commit: c2f970c32b527660a33fa513a76d913c812dcf7c pages.c introduces a conditionally compiled _os_overcommits_proc()_ function that opens a /proc file to query the overcommit policy.  The trouble is that this is done holding the malloc init lock.  If you have an LD_PRELOAD library in place, or anything that needs to init (and expects to be able to use malloc) in order to hook the  _open()_ system call, you will end up with a recursive acquire of the malloc init lock.

Please consider acquiring the lock after querying the commit policy via /proc.

```c
#ifdef JEMALLOC_PROC_SYS_VM_OVERCOMMIT_MEMORY
static bool
os_overcommits_proc(void)
{
	int fd;
	char buf[1];
	ssize_t nread;

	fd = open("/proc/sys/vm/overcommit_memory", O_RDONLY);
	if (fd == -1)
		return (false); /* Error. */

	nread = read(fd, &buf, sizeof(buf));
	if (nread < 1)
		return (false); /* Error. */
	/*
	 * /proc/sys/vm/overcommit_memory meanings:
	 * 0: Heuristic overcommit.
	 * 1: Always overcommit.
	 * 2: Never overcommit.
	 */
	return (buf[0] == '0' || buf[0] == '1');
}
#endif
```

Example:
<pre>
__lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
135	../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S: No such file or directory.
(gdb) where
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007ffff7727f3e in __GI___pthread_mutex_lock (mutex=mutex@entry=0x1680680 <init_lock>) at ../nptl/pthread_mutex_lock.c:135
#2  0x0000000000f86940 in je_malloc_mutex_lock (tsdn=0x0, mutex=0x1680680 <init_lock>) at include/jemalloc/internal/mutex.h:97
#3  malloc_init_hard () at src/jemalloc.c:1315
<b>#4  0x0000000000f8a60d in malloc_init () at src/jemalloc.c:291</b>
#5  ialloc_body (slow_path=true, usize=<synthetic pointer>, tsdn=<synthetic pointer>, zero=true, size=<optimized out>) at src/jemalloc.c:1413
#6  calloc (num=num@entry=1, size=size@entry=32) at src/jemalloc.c:1648
#7  0x00007ffff751b627 in _dlerror_run (operate=operate@entry=0x7ffff751b020 <dlsym_doit>, args=args@entry=0x7fffffffe660) at dlerror.c:141
#8  0x00007ffff751b088 in __dlsym (handle=handle@entry=0xffffffffffffffff, name=name@entry=0x7ffff7bbbd05 "socket") at dlsym.c:70
#9  0x00007ffff7b4b877 in citp_find_all_sys_calls () at ../../../../../src/include/onload/declare_syscalls.h.tmpl:40
#10 citp_syscall_init () at ../../../../../src/lib/transport/unix/sys.c:103
#11 0x00007ffff7b4b24c in citp_do_init (max_init_level=max_init_level@entry=2) at ../../../../../src/lib/transport/unix/startup.c:650
#12 0x00007ffff7b516a2 in onload_open (pathname=pathname@entry=0x1257458 "/proc/sys/vm/overcommit_memory", flags=flags@entry=0)
    at ../../../../../src/lib/transport/unix/sockcall_intercept.c:2092
#13 0x0000000000fad5fa in open (__oflag=0, __path=0x1257458 "/proc/sys/vm/overcommit_memory") at /usr/include/x86_64-linux-gnu/bits/fcntl2.h:53
#14 os_overcommits_proc () at src/pages.c:217
#15 je_pages_boot () at src/pages.c:245
#16 0x0000000000f86852 in malloc_init_hard_a0_locked () at src/jemalloc.c:1194
#17 0x0000000000f869b4 in malloc_init_hard () at src/jemalloc.c:1322
<b>#18 0x0000000000f8a60d in malloc_init () at src/jemalloc.c:291</b>
#19 ialloc_body (slow_path=true, usize=<synthetic pointer>, tsdn=<synthetic pointer>, zero=true, size=<optimized out>) at src/jemalloc.c:1413
#20 calloc (num=num@entry=1, size=size@entry=32) at src/jemalloc.c:1648
#21 0x00007ffff751b627 in _dlerror_run (operate=operate@entry=0x7ffff751b020 <dlsym_doit>, args=args@entry=0x7fffffffe840) at dlerror.c:141
#22 0x00007ffff751b088 in __dlsym (handle=handle@entry=0xffffffffffffffff, name=name@entry=0x7ffff7bbbd05 "socket") at dlsym.c:70
#23 0x00007ffff7b4b877 in citp_find_all_sys_calls () at ../../../../../src/include/onload/declare_syscalls.h.tmpl:40
#24 citp_syscall_init () at ../../../../../src/lib/transport/unix/sys.c:103
#25 0x00007ffff7b4b24c in citp_do_init (max_init_level=max_init_level@entry=10) at ../../../../../src/lib/transport/unix/startup.c:650
#26 0x00007ffff7b4b3de in _init () at ../../../../../src/lib/transport/unix/startup.c:666
#27 0x00007ffff7de74aa in call_init (l=0x7ffff7ff7620, argc=argc@entry=1, argv=argv@entry=0x7fffffffe948, env=env@entry=0x7fffffffe958) at dl-init.c:58
#28 0x00007ffff7de75fb in call_init (env=0x7fffffffe958, argv=0x7fffffffe948, argc=1, l=<optimized out>) at dl-init.c:30
#29 _dl_init (main_map=0x7ffff7ffe168, argc=1, argv=0x7fffffffe948, env=0x7fffffffe958) at dl-init.c:120
#30 0x00007ffff7dd7cfa in _dl_start_user () from /lib64/ld-linux-x86-64.so.2
#31 0x0000000000000001 in ?? ()
#32 0x00007fffffffeb7d in ?? ()
#33 0x0000000000000000 in ?? ()
</pre>

kspinka:
To avoid a recursive acquire, while interrogating the overcommit policy
when the open() system call has been hooked by a library that expects to
be able to call malloc(), pages_boot() must be called without the
init_lock held.  The problem is apparent when running with a jailing
library, container system, etc.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1191,7 +1191,7 @@ malloc_init_hard_a0_locked()
 				abort();
 		}
 	}
-	pages_boot();
+
 	if (base_boot())
 		return (true);
 	if (chunk_boot())
@@ -1231,6 +1231,8 @@ static bool
 malloc_init_hard_a0(void)
 {
 	bool ret;
+
+	pages_boot();
 
 	malloc_mutex_lock(TSDN_NULL, &init_lock);
 	ret = malloc_init_hard_a0_locked();
@@ -1308,6 +1310,8 @@ static bool
 malloc_init_hard(void)
 {
 	tsd_t *tsd;
+
+	pages_boot();
 
 #if defined(_WIN32) && _WIN32_WINNT < 0x0600
 	_init_init_lock();


jasone:
This diff causes ```pages_boot()``` to be called after calls to ```mmap()```, which means that ```mmap_flags``` isn't initialized early enough.  Perhaps there's a way to bootstrap things in an order that would support calling ```malloc()``` from ```open()```, but in general I'm reluctant to even try supporting such use cases.  This isn't much different than an application hooking ```mmap()``` and expecting to be able to ```malloc()``` metadata; if the application hooks system calls, it needs to be careful about its own bootstrapping.

PR #452, opened 2016-09-07
use install command determined by configure

tkoecker:
The install targets in `Makefile.in` use a hardcoded `install` command, which however not always exists (like on our solaris-x86 machine). configure already has a check for which install command to use (and determines that `build-aux/install-sh` should be used on that machine).

This patch replaces the hard-coded install command with whatever configure determined should be used.


jasone:
Integrated into dev as ea68cd25b6455b2c408200c3947c0715f6a7e4a7 and cherry-picked into rel-4.3.0 as 92009b19d623748fdf8b65aea732c27ec4dd257b.  Thanks!

PR #459, opened 2016-09-22
Fix arena_bind

interwq:
when tsd is not in nominal state (e.g. during thread termination), we should not increment nthreads.


jasone:
Integrated as 1cb399b630db16892069cb37b6b0853ca318bb77; thanks!  This also needs to be applied to the next 4.x release, so I'm leaving this issue open to track that.

jasone:
Integrated into rel-4.3.0 as 57ed894f8ac154d30faf9449a76d2792cdad3850.

PR #472, opened 2016-10-14
Allow forcing an VM overcommit setting

TheCrazyLex:
Sometimes the overcommit setting from the OS
is known and static.
In this case, we can allow forcing the setting
to one value in order to avoid the lookup in
/proc/sys or sysfs.
This is useful because having to look up
in /proc/sys or sysfs creates an overheaad.

Signed-off-by: Alex Naidis alex.naidis@linux.com

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_defs.h.in
+++ b/include/jemalloc/internal/jemalloc_internal_defs.h.in
@@ -242,6 +242,16 @@
 #undef JEMALLOC_PROC_SYS_VM_OVERCOMMIT_MEMORY
 
 /*
+ * Options for forcing an overcommit setting.
+ * This can be used if the overcommit setting from
+ * the target OS is known and static.
+ * Using this is benefitial since some overhead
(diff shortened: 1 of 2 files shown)


TheCrazyLex:
@jasone I gave you write access to my forked repository so you can amend and force push my commit if required.

jasone:
What is the intended interface for utilizing this, ```CPPFLAGS=...```?  Maybe a configure option something like ```--with-overcommit=[auto|yes|no]``` to set the CPP macros?

TheCrazyLex:
My plan was actually to set this in the configure.ac file via "AC_DEFINE"

TheCrazyLex:
If you prefer that, I can convert it to "--with-overcommit=[auto|yes|no]"  like you mentioned :)

jasone:
Yes, I'd be happy to take this change with the ```--with-overcommit``` interface, since that makes it broadly accessible.

jasone:
Closing; please resubmit once ```--with-overcommit``` is implemented.

PR #476, opened 2016-10-26
jemalloc cpp new/delete bindings

djwatson:
Adds cpp bindings for jemalloc, along with necessary autoconf settings.
This is mostly to add sized deallocation support, which can't be added
from C directly.  Sized deallocation is ~10% microbench improvement.

* Import ax_cxx_compile_stdcxx.m4 from the autoconf repo, seems like the easiest
  way to get c++14 detection.
* Adds various other changes, like CXXFLAGS, to configure.ac
* Adds new rules to Makefile.in for src/jemalloc-cpp.cpp, and a basic unittest.
* Both new and delete are overridden, to ensure jemalloc is used for both.
* TODO future enhancement of avoiding extra PLT thunks for new and delete -
  sdallocx and malloc are publicly exported jemalloc symbols, using an alias
  would link them directly.  Unfortunately, was having trouble getting it
  to play nice with jemalloc's namespace support.

Testing:
Tested gcc 4.8, gcc 5, gcc 5.2, clang 4.0.  Only gcc >= 5 has sized deallocation
support, verified that the rest build correctly.

Tested mac osx and Centos.

tested --with-jemalloc-prefix, --without-export

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -9,6 +9,7 @@ vpath % .
 SHELL := /bin/sh
 
 CC := @CC@
+CXX := @CXX@
 
 # Configuration parameters.
 DESTDIR =
@@ -24,6 +25,7 @@ abs_objroot := @abs_objroot@
 
 # Build parameters.
 CPPFLAGS := @CPPFLAGS@ -I$(srcroot)include -I$(objroot)include
+CXXFLAGS := @CXXFLAGS@
 CFLAGS := @CFLAGS@
 LDFLAGS := @LDFLAGS@
 EXTRA_LDFLAGS := @EXTRA_LDFLAGS@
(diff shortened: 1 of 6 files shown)


djwatson:
#477

jasone:
I think we need to build all the tests with ```-lstdc++```.  I'm going to finish sorting this out along with #504 tomorrow.

youzp:
@djwatson I noticed that jemalloc cpp new/delete bindings added in version 5.0.0, but I am not particularly clear how to use it,  and i did not find the new/delete definition in the header file jemalloc.h,can you provide some examples? 
./configure --with-jemalloc-prefix="je_"  && make && make install

PR #481, opened 2016-10-28
Support static linking of jemalloc with glibc

### LINKED ISSUE #442: Unable to compile with -static
I'm unable to compile a simple static program with jemalloc. I'm using the jemalloc-dev package on Ubuntu and I've not compiled it from source. Here's a very simple repro:

```
#include <stdlib.h>

int main() {
  void *i = malloc(0);
  return 0;
}
```

When I compile this I get the following:

```
> g++ test.c -o test -static -ljemalloc -lpthread
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libc.a(malloc.o): In function `__libc_malloc':
(.text+0x6120): multiple definition of `malloc'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libjemalloc.a(jemalloc.o):(.text+0x27d0): first defined here
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libc.a(malloc.o): In function `__libc_free':
(.text+0x6620): multiple definition of `free'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libjemalloc.a(jemalloc.o):(.text+0xc80): first defined here
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libc.a(malloc.o): In function `__libc_realloc':
(.text+0x67e0): multiple definition of `realloc'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libjemalloc.a(jemalloc.o):(.text+0x1510): first defined here
collect2: error: ld returned 1 exit status
```

FWIW, when using tcmalloc it compiles fine:

```
> g++ test.c -o test -static -ltcmalloc_minimal -lpthread
> ./test
```

Do I need to build jemalloc with different options to enable this?

djwatson:
glibc defines its malloc implementation with several weak and strong
symbols:

strong_alias (__libc_calloc, __calloc) weak_alias (__libc_calloc, calloc)
strong_alias (__libc_free, __cfree) weak_alias (__libc_free, cfree)
strong_alias (__libc_free, __free) strong_alias (__libc_free, free)
strong_alias (__libc_malloc, __malloc) strong_alias (__libc_malloc, malloc)

The issue is not with the weak symbols, but that other parts of glibc
depend on __libc_malloc explicitly.  Defining them in terms of jemalloc
API's allows the linker to drop glibc's malloc.o completely from the link,
and static linking no longer results in symbol collisions.

Another wrinkle: jemalloc during initialization calls sysconf to
get the number of CPU's.  GLIBC allocates for the first time before
setting up isspace (and other related) tables, which are used by
sysconf.  Instead, use the pthread API to get the number of
CPUs with GLIBC, which seems to work.

Fixes #442

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -749,6 +749,15 @@ malloc_ncpus(void)
 	SYSTEM_INFO si;
 	GetSystemInfo(&si);
 	result = si.dwNumberOfProcessors;
+#elif defined(JEMALLOC_GLIBC_MALLOC_HOOK)
+	/*
+	 * GLIBC's sysconf uses isspace().  GLIBC allocates for the first
+	 * time *before* setting up the isspace tables.
+	 * Therefore we need a different method to get the number of CPUs.
+	 */
+	cpu_set_t set;
+	pthread_getaffinity_np(pthread_self(), sizeof(set), &set);
+	result = CPU_COUNT(&set);
 #else
 	result = sysconf(_SC_NPROCESSORS_ONLN);
 #endif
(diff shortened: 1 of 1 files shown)


jasone:
Using ```pthread_getaffinity_np()``` is a neat trick.

djwatson:
merged as 830938840865fe236ae2bdc0abdb0d5778146859

PR #482, opened 2016-10-28
Fix long spinning in rtree_node_init

### LINKED ISSUE #415: Long spinning in rtree_node_init under heavy contention
Currently in rtree_node_init, threads encountered CAS failure will spin and wait for the initialize thread to finish. However if the thread that's doing initialization is preempted, many other threads will be spinning and waiting for a long time. Observed in t-test1 with 128 threads on 32 cores. Practically the test is in a live lock like situation that never finishes. Perf shows > 99% CPU time spinning in rtree_node_init.

A simple spin a number of times then block (e.g. like below) fixed the issue. However we can probably avoid the sleep by fixing this differently (and better), e.g. initialize multiple nodes in rtree at once, different node size, etc.

```
unsigned n_spin = 16;
do {
  CPU_SPINWAIT;
  node = atomic_read_p((void **)elmp);
  if (n_spin-- == 0) {
    usleep(1000);
    u_spin = 16;
  }
} while (node == RTREE_NODE_INITIALIZING);
```

djwatson:
rtree_node_init basically locks the node, allocates, and then sets the node.
This is under heavy contention at the top of the tree if many threads
start to allocate at the same time.

Instead, allocate _first_ without ever locking the node, then optimistically
try to set the node.  If this fails, save the allocated node for later use.
Extra nodes could potentially be leaked on thread death, but I don't think
this will be a big issue in practice

Previous benchmark time:
./ttest1 500 100
~15s

New benchmark time:
./ttest1 500 100
.57s

fixes #415

### DIFF
--- a/include/jemalloc/internal/rtree.h
+++ b/include/jemalloc/internal/rtree.h
@@ -23,9 +23,6 @@ typedef struct rtree_s rtree_t;
 #define	RTREE_HEIGHT_MAX						\
     ((1U << (LG_SIZEOF_PTR+3)) / RTREE_BITS_PER_LEVEL)
 
-/* Used for two-stage lock-free node initialization. */
-#define	RTREE_NODE_INITIALIZING	((rtree_elm_t *)0x1)
-
 #define	RTREE_CTX_INITIALIZER	{					\
 	false,								\
 	0,								\
@@ -139,6 +136,7 @@ struct rtree_s {
 	 */
 	unsigned		start_level[RTREE_HEIGHT_MAX + 1];
 	rtree_level_t		levels[RTREE_HEIGHT_MAX];
+	malloc_mutex_t		init_lock;
 };
 
 #endif /* JEMALLOC_H_STRUCTS */
@@ -251,7 +249,7 @@ JEMALLOC_ALWAYS_INLINE bool
 rtree_node_valid(rtree_elm_t *node)
 {
 
-	return ((uintptr_t)node > (uintptr_t)RTREE_NODE_INITIALIZING);
+	return ((uintptr_t)node != (uintptr_t)0);
 }
 
 JEMALLOC_ALWAYS_INLINE rtree_elm_t *
--- a/src/rtree.c
+++ b/src/rtree.c
@@ -59,6 +59,8 @@ rtree_new(rtree_t *rtree, unsigned bits)
 	}
 	rtree->start_level[RTREE_HEIGHT_MAX] = 0;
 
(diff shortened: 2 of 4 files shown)


interwq:
How about if the CAS fails, we try using the allocated node for the next region? This way we don't need to stash it, or risk leaking.

djwatson:
I don't think that works generally since extents are mmapped at random addresses - we could still loose one node on every new extent.

Other options include a global queue/stack of saved nodes, or sleeping somehow (with a lock or adaptive)

djwatson:
Updated to a simpler mutex strategy, which seems to work almost as well, with no node leaking.

jasone:
Mmm, no leaking, I like it. :)

include/jemalloc/internal/rtree.h
```diff
@@ -251,7 +249,7 @@ JEMALLOC_ALWAYS_INLINE bool
 rtree_node_valid(rtree_elm_t *node)
 {
 
-	return ((uintptr_t)node > (uintptr_t)RTREE_NODE_INITIALIZING);
+	return ((uintptr_t)node > (uintptr_t)0);
```
jasone: Let's make this an inequality check.

```return ((uintptr_t)node != (uintptr_t)0);```

File: src/rtree.c [hunk 2]
```diff
@@ -135,25 +137,18 @@ rtree_node_init(tsdn_t *tsdn, rtree_t *rtree, unsigned level,
...
-			node = atomic_read_p((void **)elmp);
-		} while (node == RTREE_NODE_INITIALIZING);
-	} else {
+	malloc_mutex_lock(tsdn, &rtree->init_lock);
+	node = atomic_read_p((void**)elmp);
+	if (!node) {
 		node = rtree_node_alloc(tsdn, rtree, ZU(1) <<
 		    rtree->levels[level].bits);
-		if (node == NULL)
+		if (node == NULL) {
```
interwq: Seems we can get rid of this branch?

jasone: Oh, that's subtle.  If we do remove this, we need a comment that the code remains correct even if ```rtree_node_alloc()``` OOMs.

interwq: Never mind. I wasn't making sense.

interwq: Never mind. It doesn't seem readable that way.

djwatson:
* inequality check
* fixed witness lock errors

jasone:
Integrated as 25f7bbcf28f5c83b11149989b3552d87c1f3c5e9.  Thanks!

PR #486, opened 2016-11-03
Check for existance of CPU_COUNT macro 

### LINKED ISSUE #485: undefined reference to `CPU_COUNT'
got this linker error when compiling with old glibc versions.

This does not work on old linux:
83093884 src/jemalloc.c          (Dave Watson         2016-10-28 13:51:52 -0700  766) 		result = CPU_COUNT(&set);

This still works:
result = sysconf(_SC_NPROCESSORS_ONLN);

May want to check glibc version before calling? Thanks.

djwatson:
fixes #485


jasone:
Thanks!

PR #488, opened 2016-11-04
Fixes to Visual Studio Project files

mjp41:
The Visual Studio project files did not contain all the required files to build from Visual Studio.  

It also contained the ValGrind header, which I don't believe is used in that build (the .C file is not included).


jasone:
Integrated as d30b3ea51a6cb3abe88251a0fd769b0438c88a6e (dev) and 77635bf532488b180024ef50c0583e43111fee0f (rel-4.3.0).  Thanks!

PR #490, opened 2016-11-04
Fix second definition of malloc_conf

yuslepukhin:
Second definition causes the compiler to complain about malloc_conf (je_malloc_conf)
  redefinition and/or inconsistent linkage. I believe the intention was
  simply to change the existing config so it can take effect.


jasone:
The intention is for the redefinition to replace the original.  It took a lot of finagling to get it working _at all_ on Windows.  Unfortunately, I don't know of a way around the linker warning.

PR #503, opened 2016-11-16
Fix size_t -> 32 uint -> size_t convertion

maksqwe:


jasone:
Thanks, integrated as 95974c04403bbfb2fe3031e86785d354ef5e5906.

PR #504, opened 2016-11-16
Fix __builtin_unreachable configure check

ronawho:
__builtin_unreachable causes link errors with osx+gnu, but only at -O1 or higher.
If EXTRA_CFLAGS contain an optimization level it needs to be added to CFLAGS
for the __builtin_unreachable check.

EXTRA_CFLAGS used to be set for all configure tests, but that was changed with
af0e28f

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1118,6 +1118,9 @@ AC_SUBST([enable_cache_oblivious])
 
 
 
+dnl optimization level can impact whether __builtin_unreachable works
+SAVED_CFLAGS="${CFLAGS}"
+JE_CFLAGS_APPEND(["${EXTRA_CFLAGS}"])
 JE_COMPILABLE([a program using __builtin_unreachable], [
 void foo (void) {
   __builtin_unreachable();
@@ -1132,6 +1135,7 @@ if test "x${je_cv_gcc_builtin_unreachable}" = "xyes" ; then
 else
   AC_DEFINE([JEMALLOC_INTERNAL_UNREACHABLE], [abort])
 fi
+CFLAGS="${SAVED_CFLAGS}"
 
 dnl ============================================================================
 dnl Check for  __builtin_ffsl(), then ffsl(3), and fail if neither are found.


ronawho:
Do you remember the motivation for not applying EXTRA_CFLAGS to configure tests anymore? It seems like some other checks could be dependent on flags set in EXTRA_CFLAGS.

jasone:
The issue with applying ```EXTRA_CFLAGS``` to configuration tests is that some of them are designed to fail on compiler warning, so it's easy to inadvertently cause failures that shouldn't impact configuration.  If we want configuration failures due to extra flags, plain ```CFLAGS``` works, right?

ronawho:
Ah, ok.

Yeah, using plain CFLAGS seems to work, but that obviously prevents configure from setting CFLAGS that it might want to (-std=gnu11, -hnognu, etc.)

jasone:
Hmm, what do you think about adding another configure option that controls whether all the automatic ```CFLAGS``` setup occurs, so that it can be explicitly controlled rather than being gated on empty ```CFLAGS```?

ronawho:
That would work for me, although I think it makes it a little confusing for new users to figure out when they should be using CFLAGS vs. EXTRA_CFLAGS.

Would you want JE_CFLAGS_APPEND to prepend instead of append then? It seems like you would, so that a user could still override an inferred CFLAG value

Back to your "The issue with applying EXTRA_CFLAGS to configuration tests is that some of them are designed to fail on compiler warning, so it's easy to inadvertently cause failures that shouldn't impact configuration" comment:  I get the argument, but it does seem wonky for something like `EXTRA_CFLAGS=-disable-attribute-support` (a hypothetical flag). In this case you would clearly want that to be applied to the configure tests otherwise they would get the checks for attribute support wrong.

Not sure if I think this is a good idea or not, but throwing it out anyways: what do you think about reverting the behavior of EXTRA_CFLAGS and adding a new EXTRA_CFLAGS_NO_CONFIGURE flag or something instead?

jasone:
What about completely removing ```EXTRA_CFLAGS```, and replacing the ```JE_CFLAGS_APPEND``` macro with e.g. ```JE_CFLAGS_INSERT```, so that it maintains ```CFLAGS``` as the concatenation of A) flags previously added by ```configure.ac```, B) the flag to be inserted, and C) the flags passed in via the ```CFLAGS``` environment variable?  This would cause all flags to be applied to configure tests, yet the ordering would allow overriding the default flags -- assuming the typical command line semantics of later parameters overriding earlier ones.

ronawho:
I'd be happy with that, and it matches what I've seen in other projects.

Wasn't the whole point of af0e28f to avoid applying all flags to configure checks though? Wouldn't your proposed change effectively revert to the old behavior?

And to be clear, I liked the behavior prior to af0e28f, so I'd have no problem going back to that, but won't you just have the same problem that commit was trying to solve?

jasone:
I think this would be different in that the ```CFLAGS``` contents passed in to ```configure``` via the environment always comes last.

ronawho:
Ok, that works for me.

ronawho:
Just tried 194d6f9de8ff92841b67f38a2a6a06818e3240dd and it works great for me. Thanks!

Not super important to me since I have a workaround for this check in place, but do you know if they'll be anymore 4.X releases that this might make it into, or are you pretty focused on 5.0?

jasone:
I'm hoping we're done with 4.x releases.  Also, I did some other refactoring in 194d6f9de8ff92841b67f38a2a6a06818e3240dd that would require some manual effort to backport, with the expectation that it will first be released in 5.0.0.

ronawho:
@jasone Do you have any interest in backporting 194d6f9de8ff92841b67f38a2a6a06818e3240dd now that there's a 4.5.0 release planned, or do you think it's more effort than it's worth?

jasone:
I have mixed feelings on the backport.  The reason I skipped over it while creating backporting issues was the C++ support, which is irrelevant to 4.x.  I can be convinced to backport it though.

ronawho:
I'm torn. It feels like a "bug" that since 4.3 there's no good way to throw flags that you want to impact configure tests, but I have no idea if that's important for any other users. It'd be nice to get rid of the workaround I have for this, but I don't think it's worth backporting just for me if it's not affecting anybody else.

Maybe open an issue for backporting it (with the hope that others will comment on it if it matters to them) and close it if you decide it's not worthwhile?

PR #513, opened 2016-11-23
Implement a more reliable detection scheme for os_unfair_lock.

jszakmeister:
The core issue here is the weak linking of the symbol, and in certain
environments--for instance, using the latest Xcode (8.1) with the latest
SDK (10.12)--os_unfair_lock may resolve even though you're compiling on
a host that doesn't support it (10.11).

We can use the availability macros to circumvent this problem, and
detect that we're not compiling for a target that is going to support
them and error out at compile time.  The other alternative is to do a
runtime check, but that presents issues for cross-compiling.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1670,10 +1670,15 @@ dnl Check for os_unfair_lock operations as provided on Darwin.
 
 JE_COMPILABLE([Darwin os_unfair_lock_*()], [
 #include <os/lock.h>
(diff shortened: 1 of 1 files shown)


jszakmeister:
I wish we could more directly test that the function will be available, but this does the trick and avoids having to set the cache variable.

jasone:
Thanks!

Issue #522, opened 2016-12-07
Big overhead with xmalloc:true


renecannao:
While running some micro-benchmark I noticed that `malloc:true` caused unexpected overhead (+25%).
Here is how to reproduce it with a simple loop.
File `malloc_loop.c`:
```
#include <stdlib.h>
int main() {
        int i;
        char *c=NULL;
        for (i=0;i<500000000;i++) {
                c=(char *)malloc(100);
                if (__builtin_expect(c==NULL, 0)) return -1;
                free(c);
        }
        return 0;
}
```
Compile:
```
vagrant@ubuntu-14:~$ cc -O2 malloc_loop.c -o malloc_loop -Ljemalloc/jemalloc/lib -Wl,-Bstatic -ljemalloc -Wl,-Bdynamic -lpthread -lm
```
Run it:
```
$ MALLOC_CONF="xmalloc:false" time ./malloc_loop
7.54user 0.01system 0:07.56elapsed 99%CPU (0avgtext+0avgdata 908maxresident)k
0inputs+0outputs (0major+275minor)pagefaults 0swaps
```
```
$ MALLOC_CONF="xmalloc:true" time ./malloc_loop
9.47user 0.01system 0:09.50elapsed 99%CPU (0avgtext+0avgdata 916maxresident)k
0inputs+0outputs (0major+278minor)pagefaults 0swaps
```

jasone:
Closing; see discussion in #523 for additional detail.

PR #523, opened 2016-12-07
Remove xmalloc out from malloc_slow path #522

renecannao:
As reported in issue #522 , `xmalloc:true` creates a big overhead.
This simple commits takes `xmalloc` out from `malloc_slow_flags`, and now `xmalloc:true` and `xmalloc:false` have identical performance.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -869,8 +869,7 @@ malloc_slow_flag_init(void)
 	malloc_slow_flags |= (opt_junk_alloc ? flag_opt_junk_alloc : 0)
 	    | (opt_junk_free ? flag_opt_junk_free : 0)
 	    | (opt_zero ? flag_opt_zero : 0)
-	    | (opt_utrace ? flag_opt_utrace : 0)
-	    | (opt_xmalloc ? flag_opt_xmalloc : 0);
+	    | (opt_utrace ? flag_opt_utrace : 0);
 
 	malloc_slow = (malloc_slow_flags != 0);
 }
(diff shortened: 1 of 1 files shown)


jasone:
If it weren't for maintaining backward compatibility on FreeBSD, I would have removed xmalloc support years ago.  As such, it's more important to me that the fast path not be compromised even a little bit for xmalloc.  This change adds a branch back to the fast path, which IMO is a step in the wrong direction.

CC: @interwq, in case I'm missing something.

interwq:
@jasone: I also feel that fast path is more important. In this case the branch is within an unlikely branch (checking ret == NULL) so not as bad. It does increase the fast path code size a little bit (though also unlikely branch). Since xmalloc is simple to implement on top of malloc, I'd say not worth the cost?

interwq:
Also in the case of mallocx, where update_errno is set to false, that is indeed one more branch on fast path. As w/o this diff, the entire "if (unlikely(ret == NULL))" branch can be optimized away.

renecannao:
Thank you both for the feedback.
Before looking the code, indeed I thought `xmalloc` was a wrapper around `malloc`, to then realize my assumption was wrong.
I _think_ this change doesn't affect the performance of the fast path because the only branch affected is an unlikely one : `if (unlikely(ret == NULL))` .

jasone:
Thank you for the submission and clear description of the issue, @renecannao.  I'm going to close this because I don't want to add code to the fast path for xmalloc, but please do continue to submit ideas and improvements as they occur to you.

renecannao:
Thank you for the feedback!
As I am using `xmalloc:true` in ProxySQL and jemalloc is statically linked, I will use this patch.
Thanks for the great library!

PR #529, opened 2016-12-21
Use better pre-processor defines for sparc64

glaubitz:
Currently, jemalloc detects sparc64 targets by checking whether
__sparc64__ is defined. However, this definition is used on BSD
targets only. Linux targets define both __sparc__ and __arch64__
for sparc64. Since this also works on BSD, rather use __sparc__
and __arch64__ instead of __sparc64__ to detect sparc64 targets.

### DIFF
--- a/include/jemalloc/internal/mb.h
+++ b/include/jemalloc/internal/mb.h
@@ -76,7 +76,7 @@ mb_write(void)
 	    : "memory" /* Clobbers. */
 	    );
 }
-#elif defined(__sparc64__)
+#elif defined(__sparc__) && defined(__arch64__)
 JEMALLOC_INLINE void
 mb_write(void)
 {


glaubitz:
Ping. Anyone can review this?

davidtgoldblatt:
My plan is to just delete this file when I start on https://github.com/jemalloc/jemalloc/issues/533, which I expect to be done in a couple of days -- there's no mb_write calls anywhere, and it is rather confused in general.

glaubitz:
Hmm, ok. My original intention was to get this change in because Mozilla upstream asked me to, see:

> https://bugzilla.mozilla.org/show_bug.cgi?id=1275204#c42

I'm not sure when and if Mozilla is going to update their embedded version of jemalloc.

davidtgoldblatt:
They're on jemalloc 3, right?

@jasone, is this the sort of thing the stable-3 branch is for?

glaubitz:
The current copy in the git repository is 4.3.1, see:

> https://github.com/mozilla/gecko-dev/blob/master/memory/jemalloc/src/VERSION

jasone:
I'll go ahead and merge this, so that we have it available for cherry-picking when we create another 4.x branch.

jasone:
Integrated as 77de5f27d848414a6d26e86e2812339ffe1062d3.  Thank you for your patience, @glaubitz.  This should also go into the next 4.x release, or the stable-4 branch if we don't do another 4.x release.

glaubitz:
@jasone Thanks! In fact, Mozilla upstream just asked me to ask you to backport this change, see: https://bugzilla.mozilla.org/show_bug.cgi?id=1275204#c67

So, would it be possible to backport the fix? Shall I open another PR?

jasone:
@glaubitz, a tracking issue for backporting this into 4.5.0 (ne 4.4.1) would be helpful.  Thanks!

Issue #533, opened 2017-01-09
.h file refactoring


davidtgoldblatt:
The current "JEMALLOC_H_TYPES / JEMALLOC_H_STRUCTS / JEMALLOC_H_EXTERNS / JEMALLOC_H_INLINES" framework for include files is unidiomatic and makes the dependency graph cyclic. We want to have a principled way to split everything up.

davidtgoldblatt:
The plan is to do this in multiple stages:

Stage 1: Break up each internal .h file into its constituent sections (i.e. arena.h will be split into arena-types.h, arena-structs-a.h, arena-structs-b.h, arena-externs.h, arena-inlines.h). This is a no-op change; we still rely on jemalloc_internal.h to deal with any dependencies implicitly through the ordering of the includes.

Stage 2: Make inter-header dependencies explicit, moving them out of jemalloc_internal.h. We'll have each header file #include its dependencies directly.

Stage 3: Fix up the various inlining macros to use C99 inline semantics.

Stage 4: Cleanup. Make the dependency graph non-cyclic across modules where we can, breaking them up if necessary. We can choose to merge simpler modules back into a file if we choose.

jasone:
Looks good so far.  One nit: let's use underscores in file names for consistency (e.g. ```arena_structs_a.h```).

jasone:
With regard to our offline discussion about how this negatively affects ```git blame``` fidelity, I figured out the magic incantation that works well in the face of such restructuring:

```sh
git blame -M -C -C -w <path>
```

PR #543, opened 2017-01-11
Break up headers into constituent parts

davidtgoldblatt:
This is part of a broader change to make header files better represent the
dependencies between one another (see
https://github.com/jemalloc/jemalloc/issues/533). It breaks up component headers
into smaller parts that can be made to have a simpler dependency graph.

This also removes mb.h, which has no uses.

For the autogenerated headers (smoothstep.h and size_classes.h), no splitting
was necessary, so I didn't add support to emit multiple headers.


interwq:
nit: can remove the notes about mb.h in the commit message.

jasone:
@interwq, I think the diff commit message itself has been updated, so it's just the original pull request description that mentions mb.h.

interwq:
@jasone: ah you are right. New to the UI here :)

PR #546, opened 2017-01-13
Use better pre-processor defines for sparc64

jasone:
Currently, jemalloc detects sparc64 targets by checking whether
__sparc64__ is defined. However, this definition is used on BSD
targets only. Linux targets define both __sparc__ and __arch64__
for sparc64. Since this also works on BSD, rather use __sparc__
and __arch64__ instead of __sparc64__ to detect sparc64 targets.

### DIFF
--- a/include/jemalloc/internal/mb.h
+++ b/include/jemalloc/internal/mb.h
@@ -76,7 +76,7 @@ mb_write(void)
 	    : "memory" /* Clobbers. */
 	    );
 }
-#elif defined(__sparc64__)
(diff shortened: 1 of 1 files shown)


jasone:
This is a cherry-pick from dev to what will become the rc-4.5.0 branch.  It looks like I could have said as much while creating the pull request, without modifying the original commit message.  I will try that next time.

jasone:
Pushed.  Next time we create an rc branch I think manually creating the branch with no commits as a separate step from creating the first commit will work better.

PR #548, opened 2017-01-13
No leading blank lines

jasone:
This commit series removes leading blank lines from function bodies, as well as separately fixing nearby indentation issues.


jasone:
I mistakenly removed a line of code (missed due to regex fail during validation), and fixed it as an amended commit and forced push.  It looks like this does the right thing.

PR #549, opened 2017-01-13
Remove redundent stats-merging logic when destroying tcache. 

interwq:
The stats merging is already taken care of by tcache_flush.

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -357,11 +357,8 @@ tcache_destroy(tsd_t *tsd, tcache_t *tcache)
 		tcache_bin_t *tbin = &tcache->tbins[i];
 		tcache_bin_flush_small(tsd, tcache, tbin, i, 0);
 
-		if (config_stats && tbin->tstats.nrequests != 0) {
-			arena_bin_t *bin = &arena->bins[i];
-			malloc_mutex_lock(tsd_tsdn(tsd), &bin->lock);
-			bin->stats.nrequests += tbin->tstats.nrequests;
-			malloc_mutex_unlock(tsd_tsdn(tsd), &bin->lock);
+		if (config_stats) {
+			assert(tbin->tstats.nrequests == 0);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/tcache.c
```diff
@@ -357,25 +357,18 @@ tcache_destroy(tsd_t *tsd, tcache_t *tcache)
...
 
-		if (config_stats && tbin->tstats.nrequests != 0) {
-			arena_bin_t *bin = &arena->bins[i];
-			malloc_mutex_lock(tsd_tsdn(tsd), &bin->lock);
-			bin->stats.nrequests += tbin->tstats.nrequests;
-			malloc_mutex_unlock(tsd_tsdn(tsd), &bin->lock);
-		}
+		if (config_stats) {
+			assert(tbin->tstats.nrequests == 0);
+                }
```
jasone: This line's indentation is spaces rather than tabs, which is why the diff doesn't align as expected.

src/tcache.c
```diff
@@ -357,25 +357,18 @@ tcache_destroy(tsd_t *tsd, tcache_t *tcache)
...
-		if (config_stats && tbin->tstats.nrequests != 0) {
-			malloc_mutex_lock(tsd_tsdn(tsd), &arena->lock);
-			arena->stats.nrequests_large += tbin->tstats.nrequests;
-			arena->stats.lstats[i - NBINS].nrequests +=
-			    tbin->tstats.nrequests;
-			malloc_mutex_unlock(tsd_tsdn(tsd), &arena->lock);
-		}
+		if (config_stats) {
+			assert(tbin->tstats.nrequests == 0);
+                }
```
jasone: Convert spaces to tabs.

src/tcache.c
interwq: Ah good catch. Editor profile mix up again. Will fix.

jasone:
Oops, I didn't notice before that the commit message needs a tweak.  The summary needs to be a single line (ideally <= 72 characters), followed by a blank line, and then further details as necessary.

PR #550, opened 2017-01-14
Added stats about number of bytes cached in tcache currently.

interwq:
Originally discussion in https://github.com/jemalloc/jemalloc/pull/380.

### DIFF
--- a/include/jemalloc/internal/stats_structs.h
+++ b/include/jemalloc/internal/stats_structs.h
@@ -100,6 +100,9 @@ struct arena_stats_s {
 	uint64_t	ndalloc_large;
 	uint64_t	nrequests_large;
 
+	/* Number of bytes cached in tcache associated with this arena. */
+	size_t		tcache_bytes;
+
 	/* One element for each large size class. */
 	malloc_large_stats_t	lstats[NSIZES - NBINS];
 };
--- a/src/arena.c
+++ b/src/arena.c
@@ -1625,6 +1625,21 @@ arena_stats_merge(tsdn_t *tsdn, arena_t *arena, unsigned *nthreads,
 		lstats[i].nrequests += arena->stats.lstats[i].nrequests;
 		lstats[i].curlextents += arena->stats.lstats[i].curlextents;
 	}
+
+	if (config_tcache) {
+		tcache_bin_t *tbin;
+		tcache_t *tcache;
+
+		/* tcache_bytes counts currently cached bytes. */
+		astats->tcache_bytes = 0;
+		ql_foreach(tcache, &arena->tcache_ql, link) {
+			for (i = 0; i < nhbins; i++) {
+				tbin = &tcache->tbins[i];
+				astats->tcache_bytes += tbin->ncached *
+				    index2size(i);
+			}
+		}
+	}
 	malloc_mutex_unlock(tsdn, &arena->lock);
 
 	for (i = 0; i < NBINS; i++) {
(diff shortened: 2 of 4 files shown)


include/jemalloc/internal/stats_structs.h
```diff
@@ -100,6 +100,9 @@ struct arena_stats_s {
 	uint64_t	ndalloc_large;
 	uint64_t	nrequests_large;
 
+	/* Number of bytes cached in tcache assocated with this arena.*/
```
jasone: Missing space between period and comment delimiter.

s/assocated/associated/

src/arena.c
```diff
@@ -1625,6 +1625,20 @@ arena_stats_merge(tsdn_t *tsdn, arena_t *arena, unsigned *nthreads,
...
+	if (config_tcache) {
+		tcache_bin_t *tbin;
+		tcache_t *tcache;
+
+		/* tcache_bytes counts currently cached bytes. */
+		astats->tcache_bytes = 0;
+		ql_foreach(tcache, &arena->tcache_ql, link) {
+			for (i = 0; i < nhbins; i++) {
+				tbin = &tcache->tbins[i];
+				astats->tcache_bytes += tbin->ncached * index2size(i);
```
jasone: Line too long.

src/stats.c
```diff
@@ -433,6 +434,16 @@ stats_arena_print(void (*write_cb)(void *, const char *), void *cbopaque,
 		    "resident:                %12zu\n", resident);
 	}
 
+	CTL_M2_GET("stats.arenas.0.tcache_bytes", i, &tcache_bytes, size_t);
+	if (json) {
+		malloc_cprintf(write_cb, cbopaque,
+		    "\t\t\t\t\"tcache\": %zu,\n", tcache_bytes);
```
jasone: Unfortunately, the JSON formatting is super fragile.  Here we need to use the logic that's currently determining whether to print a comma after the "resident" stat, and replace that logic with a simple comma for "resident".

src/stats.c
```diff
@@ -433,6 +434,16 @@ stats_arena_print(void (*write_cb)(void *, const char *), void *cbopaque,
...
+	} else {
+		malloc_cprintf(write_cb, cbopaque,
+		    "tcache:                  %12zu\n", tcache_bytes);
+	}
+
+
```
jasone: Extraneous newline.

src/stats.c
interwq: Ah good catch. I'll fix with other stuff and update soon.

interwq: Hmm. If we put this segment behind "if (config_tcache)", the ending comma print logic becomes quite tricky (as it's possible for resident above to be the last one). What do you think about putting tcache_bytes above resident so that its not the last one?

interwq: If we don't guard this w/ config_tcache, then it outputs 0 when disable-tcache is true.

jasone: Putting ```tcache``` above ```resident``` seems like a reasonable order to me.

src/ctl.c
```diff
@@ -2107,6 +2114,8 @@ CTL_RO_CGEN(config_stats, stats_arenas_i_internal,
     stats_arenas_i(mib[2])->astats.internal, size_t)
 CTL_RO_CGEN(config_stats, stats_arenas_i_resident,
     stats_arenas_i(mib[2])->astats.resident, size_t)
+CTL_RO_CGEN(config_stats, stats_arenas_i_tcache_bytes,
```
jasone: I think the first argument to the macro should be ```config_stats && config_tcache```, no that this mallctl endpoint returns ```ENOENT``` if tcache isn't built in.

interwq: Good point. Will fix.

PR #554, opened 2017-01-16
Braces

jasone:
This diff adds braces around single-statement blocks, and does nothing else besides fixing overlapping indentation/break formatting in two or three places.


jasone:
Hmm, looks like this pulled in #552 as a result of being based on it.

jasone:
For the most part, this diff actually disrupts blame history, because it involves non-whitespace changes.

jasone:
The spurious Windows failure is MSVC corrupting its database.

jasone:
Merged as c4c2592c834d8a37beb0a0d53842095160cbf9ee.

PR #568, opened 2017-01-19
Unify the allocation paths

davidtgoldblatt:
This unifies the allocation paths for malloc, posix_memalign, aligned_alloc,
calloc, memalign, valloc, and mallocx, so that they all share common code where
they can.

There's more work that could be done here, but I think this is the smallest
discrete change in this direction.

This shows up as a net increase in total LOC, but there's two mitigating factors:
- This will allow us to straightforwardly remove many of the inline functions in jemalloc_internal.h; I'm just saving that for a later diff.
- Many of the added lines are far simpler than many of the removed lines (the allocation functions are now wrappers that do lots of straightforward parameter setting before forwarding to imalloc; the parameter setting costs many lines, but is trivial to understand).

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1352,108 +1352,405 @@ malloc_init_hard(void)
  */
 /******************************************************************************/
 /*
- * Begin malloc(3)-compatible functions.
+ * Begin allocation-path internal functions and data structures.
  */
 
-static void *
-ialloc_prof_sample(tsd_t *tsd, size_t usize, szind_t ind, bool zero,
-    prof_tctx_t *tctx, bool slow_path)
-{
-	void *p;
+/*
+ * Settings determined by the documented behavior of the allocation functions.
+ */
+typedef struct static_opts_s static_opts_t;
+struct static_opts_s {
+	/* Whether or not allocations of size 0 should be treated as size 1. */
+	bool bump_empty_alloc;
+	/*
+	 * Whether to assert that allocations are not of size 0 (after any
+	 * bumping).
+	 */
+	bool assert_nonempty_alloc;
 
-	if (tctx == NULL)
-		return (NULL);
-	if (usize <= SMALL_MAXCLASS) {
-		szind_t ind_large = size2index(LARGE_MINCLASS);
-		p = ialloc(tsd, LARGE_MINCLASS, ind_large, zero, slow_path);
-		if (p == NULL)
-			return (NULL);
-		arena_prof_promote(tsd_tsdn(tsd), iealloc(tsd_tsdn(tsd), p), p,
-		    usize);
-	} else
-		p = ialloc(tsd, usize, ind, zero, slow_path);
+	/*
+	 * Whether or not to modify the 'result' argument to malloc in case of
+	 * error.
+	 */
+	bool null_out_result_on_error;
+	/* Whether to set errno when we encounter an error condition. */
+	bool set_errno_on_error;
 
-	return (p);
-}
+	/*
+	 * The minimum valid alignment for functions requesting aligned storage.
+	 */
+	size_t min_alignment;
 
-JEMALLOC_ALWAYS_INLINE_C void *
-ialloc_prof(tsd_t *tsd, size_t usize, szind_t ind, bool zero, bool slow_path)
-{
-	void *p;
-	prof_tctx_t *tctx;
+	/* The error string to use if we oom. */
+	const char *oom_string;
+	/* The error string to use if the passed-in alignment is invalid. */
+	const char *invalid_alignment_string;
 
-	tctx = prof_alloc_prep(tsd, usize, prof_active_get_unlocked(), true);
-	if (unlikely((uintptr_t)tctx != (uintptr_t)1U))
-		p = ialloc_prof_sample(tsd, usize, ind, zero, tctx, slow_path);
-	else
-		p = ialloc(tsd, usize, ind, zero, slow_path);
-	if (unlikely(p == NULL)) {
-		prof_alloc_rollback(tsd, tctx, true);
-		return (NULL);
-	}
-	prof_malloc(tsd_tsdn(tsd), iealloc(tsd_tsdn(tsd), p), p, usize, tctx);
+	/*
+	 * False if we're configured to skip some time-consuming operations.
+	 *
+	 * This isn't really a malloc "behavior", but it acts as a useful
+	 * summary of several other static (or at least, static after program
+	 * initialization) options.
+	 */
+	bool slow;
+};
 
-	return (p);
+JEMALLOC_ALWAYS_INLINE_C void
+static_opts_init(static_opts_t *static_opts) {
+	static_opts->bump_empty_alloc = false;
+	static_opts->assert_nonempty_alloc = false;
+	static_opts->null_out_result_on_error = false;
+	static_opts->set_errno_on_error = false;
+	static_opts->min_alignment = 0;
+	static_opts->oom_string = "";
+	static_opts->invalid_alignment_string = "";
+	static_opts->slow = false;
 }
 
 /*
- * ialloc_body() is inlined so that fast and slow paths are generated separately
- * with statically known slow_path.
- *
- * This function guarantees that *tsdn is non-NULL on success.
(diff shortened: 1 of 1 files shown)


src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+ * Begin allocation-path internal functions and data structures.
  */
 
-static void *
-ialloc_prof_sample(tsd_t *tsd, size_t usize, szind_t ind, bool zero,
-    prof_tctx_t *tctx, bool slow_path)
-{
-	void *p;
+/* Settings determined by the documented behavior of the allocation functions */
+typedef struct static_opts_s static_opts_t;
```
jasone: Feel free to combine this into ```typedef struct {...} static_opts_t``` (or leave it; your choice).  The headers are consistent about using separate declarations just so that forward declarations are easy, but here we'll never have that problem.

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+	 * error.
+	 */
+	bool null_out_result_on_error;
+	/* Whether to set errno when we encounter an error condition. */
+	bool set_errno_on_error;
 
-	return (p);
-}
+	/*
+	 * The minimum valid alignment for functions requestion aligned storage.
```
jasone: s/requestion/requesting/ ?

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+	 */
+	size_t min_alignment;
 
-JEMALLOC_ALWAYS_INLINE_C void *
-ialloc_prof(tsd_t *tsd, size_t usize, szind_t ind, bool zero, bool slow_path)
-{
-	void *p;
-	prof_tctx_t *tctx;
+	/* The error string to use if we oom.*/
```
jasone: Missing space after period.

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
-		p = ialloc(tsd, usize, ind, zero, slow_path);
-	if (unlikely(p == NULL)) {
-		prof_alloc_rollback(tsd, tctx, true);
-		return (NULL);
-	}
-	prof_malloc(tsd_tsdn(tsd), iealloc(tsd_tsdn(tsd), p), p, usize, tctx);
+	/*
+	 * False if we're configured to skip some time-consuming operations.
+	 *
+	 * This isn't really a malloc "behvaior", but it acts as a useful
```
jasone: s/behvaior/behavior/

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+	 * summary of several other static (or at least, static after program
+	 * initialization) options.
+	 *
+	 * Eventually, it would be better to have a config_opts_t that captures
```
jasone: I prefer to open issues for such ideas/directions, rather than commenting on them in the code, because they have an unfortunate tendency of accumulating without action, to the point that they eventually don't make sense (todo rot, akin to bit rot?).

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+	 * the state of all options and config settings, which can be more
+	 * quickly initialized depending on the value of initialized_and_fast.
+	 */
+	bool slow;
+};
 
-	return (p);
+JEMALLOC_ALWAYS_INLINE_C void
+static_opts_init(static_opts_t *static_opts) {
```
jasone: As discussed out-of-band, we'll update the brace style to make this the One True Way.

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+ * Note that these must be kept in sync with the macros in
+ * jemalloc/jemalloc_macros.h.
  */
-JEMALLOC_ALWAYS_INLINE_C void *
-ialloc_body(size_t size, bool zero, tsdn_t **tsdn, size_t *usize,
-    bool slow_path)
-{
-	tsd_t *tsd;
-	szind_t ind;
+#define TCACHE_IND_NONE ((unsigned)-1)
```
jasone: These definitions sorta feel like they belong in jemalloc/jemalloc_macros.h, so that they are obvious targets for change if we change flags encoding.

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+ * Returns true if the allocation will overflow, and false otherwise. Sets *size
+ * to the product either way.
+ */
+JEMALLOC_ALWAYS_INLINE_C bool
+compute_size_with_overflow(dynamic_opts_t *dopts, size_t *size) {
+	/*
+	 * This function is just num_items * item_size, except that we have to
+	 * check for overflow.
+	 *
+	 * TODO(davidtgoldblatt): We ought to use built-in "multiply with
```
jasone: Make this an issue rather than a comment.

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
-		ret = ialloc_body(size, false, &tsdn, &usize, false);
-		ialloc_post_check(ret, tsdn, usize, "malloc", true, false);
-	} else {
-		ret = ialloc_body(size, false, &tsdn, &usize, true);
-		ialloc_post_check(ret, tsdn, usize, "malloc", true, true);
-		UTRACE(0, size, ret);
+	*size = dopts->item_size * dopts->num_items;
+
+	if (unlikely(*size == 0)) {
+		if (dopts->num_items == 0 || dopts->item_size == 0) {
```
jasone: Rewrite as ```return (dopts->num_items != 0 && dopts->item_size != 0);```?

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
-JEMALLOC_EXPORT JEMALLOC_ALLOCATOR JEMALLOC_RESTRICT_RETURN
-void JEMALLOC_NOTHROW *
-JEMALLOC_ATTR(malloc) JEMALLOC_ALLOC_SIZE(1)
-je_malloc(size_t size)
-{
-	void *ret;
-	tsdn_t *tsdn;
-	size_t usize JEMALLOC_CC_SILENCE_INIT(0);
+/*
+ * Returns true if the allocation will overflow, and false otherwise. Sets *size
```
jasone: Two spaces after sentence-ending periods.

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+		} else {
+			return true;
+		}
 	}
 
-	return (ret);
+	/*
+	 * We got a non-zero size, but we don't know if we overflowed to get
+	 * there.  To avoid having to do a divide, we'll be clever and note that
+	 * if both A and B can be represented in N/2 bits, than their product
```
jasone: s/than/then/

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+	 * can be represented in N bits (without the possibility of overflow).
+	 *
```
jasone: Extraneous line.

src/jemalloc.c
```diff
@@ -1352,250 +1352,466 @@ malloc_init_hard(void)
...
+}
+/******************************************************************************/
+/*
+ * Begin malloc(3)-compatible functions.
+ */
+
+JEMALLOC_EXPORT JEMALLOC_ALLOCATOR JEMALLOC_RESTRICT_RETURN
+void JEMALLOC_NOTHROW *
+JEMALLOC_ATTR(malloc) JEMALLOC_ALLOC_SIZE(1)
+je_malloc(size_t size)
```
jasone: This (and the following functions) is pure bliss to see.

src/jemalloc.c
davidtgoldblatt: This is a little bit confusingly written; the representations are actually unrelated. The only thing that needs to be kept in sync is the list of special constants, not the actual values of those constants.

Do you think it'd be simpler to read if we split these into an enum and index?

jasone: Ah, maybe just reword the comment to say they "correspond" to the other constants, or some such.

PR #569, opened 2017-01-19
Test JSON output of malloc_stats_print() and fix bugs.

jasone:
Unsurprisingly, there were bugs in the stats printing code that this uncovered.


jasone:
The main concern I have about external dependencies is that we have to deal with multiple platforms, versioning, etc.  The JSON validator would have been nice to get from somewhere else, but in retrospect, had it been more trouble to write, I would be inclined to make the JSON generation code less fragile, rather than introducing an external dependency.

PR #570, opened 2017-01-19
Fix lots of --disable-stats issues

jasone:
While testing #569, I tried ```--disable-stats``` and hit a huge set of regressions and new bugs.  This diff fixes all those revealed by the tests.

We definitely need to add some more travis-CI configurations (at least ```--enable-debug```, ```--disable-stats``` and ```--disable-tcache```).


jasone:
Hold off on reviewing this for now.  I want to see how much trouble it is to re-enable the mallctl interfaces this diff disables.

jasone:
Okay, much better now.  Ready for review.

PR #571, opened 2017-01-19
Remove -Werror=declaration-after-statement.

jasone:
This will allow us to start moving declarations to first use, e.g. in #568.


jasone:
Not all compilers support the warning.  You probably got lucky.

PR #572, opened 2017-01-20
Remove extraneous parens around return arguments

jasone:
Pull request is mainly for CI coverage.  More style update diffs still to come.

### DIFF
--- a/include/jemalloc/internal/atomic_inlines.h
+++ b/include/jemalloc/internal/atomic_inlines.h
@@ -53,8 +53,7 @@ void	atomic_write_u(unsigned *p, unsigned x);
 #if (LG_SIZEOF_PTR == 3 || LG_SIZEOF_INT == 3)
 #  if (defined(__amd64__) || defined(__x86_64__))
 JEMALLOC_INLINE uint64_t
-atomic_add_u64(uint64_t *p, uint64_t x)
-{
+atomic_add_u64(uint64_t *p, uint64_t x) {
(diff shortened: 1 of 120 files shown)


jasone:
Ready for review.

include/jemalloc/internal/atomic_inlines.h
```diff
@@ -422,16 +422,16 @@ atomic_sub_p(void **p, void *x) {
 #if (LG_SIZEOF_PTR == 3)
 	return ((void *)atomic_add_u64((uint64_t *)p, (uint64_t)-((int64_t)x)));
```
interwq: remove this one as well?

include/jemalloc/internal/rb.h
```diff
@@ -974,7 +974,7 @@ a_prefix##reverse_iter(a_rbt_type *rbtree, a_type *start,		\
 	ret = a_prefix##reverse_iter_recurse(rbtree, rbtree->rbt_root,	\
 	  cb, arg);							\
     }									\
-    return (ret);							\
+    return ret;								\
```
interwq: tab seems off?

jasone: This has the same cause as the one you pointed out in #573.

include/jemalloc/internal/atomic_inlines.h
jasone: Good catch.

PR #573, opened 2017-01-20
Replace tab after #define with space

jasone:
This resolves the most stilted aspect of old tab usage.


include/jemalloc/internal/pages_types.h
```diff
@@ -5,23 +5,23 @@
...
-#define	PAGE_MASK	((size_t)(PAGE - 1))
+#define PAGE		((size_t)(1U << LG_PAGE))
+#define PAGE_MASK	((size_t)(PAGE - 1))
 /* Return the page base address for the page containing address a. */
-#define	PAGE_ADDR2BASE(a)						\
+#define PAGE_ADDR2BASE(a)						\
 	((void *)((uintptr_t)(a) & ~PAGE_MASK))
 /* Return the smallest pagesize multiple that is >= s. */
-#define	PAGE_CEILING(s)							\
+#define PAGE_CEILING(s)							\
```
jasone: The tab just after ```PAGE_CEILING(s)``` is only occupying one column, so the ```+``` in the diff output pushes the tab to the next tab stop and causes it to then take up eight columns.

interwq: Ah I see. Thanks for explaining!

Issue #578, opened 2017-01-20
Beef up CI matrix


davidtgoldblatt:
We should also test (at least) some permutations of --enable-debug, --disable-stats and --disable-tcache.

PR #587, opened 2017-01-25
Beef up travis CI integration testing

davidtgoldblatt:
Introduces gen_travis.py, which generates .travis.yml, and updates .travis.yml
to be the generated version.

The travis build matrix approach doesn't play well with mixing and matching
various different environment settings, so we generate every build explicitly,
rather than letting them do it for us.

To avoid abusing travis resources (and save us time waiting for CI results), we
don't test every possible combination of options; we only check up to 2 unusual
settings at a time.

This will resolve https://github.com/jemalloc/jemalloc/issues/578 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,31 +3,93 @@ language: generic
 matrix:
   include:
     - os: linux
-      env: CC=gcc CXX=g++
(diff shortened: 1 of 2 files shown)


jasone:
Looks good, pre-existing test failures with ```--disable-stats``` aside.  That regression bisects to 5154ff32ee8c37bacb6afd8a07b923eb33228357.

PR #588, opened 2017-01-26
Fix a bug in which a potentially invalid usize replaced size

davidtgoldblatt:
In the refactoring that unified the allocation paths, usize was substituted for
size. This worked fine under the default test configuration, but triggered
asserts when we started beefing up our CI testing.

This change fixes the issue, and clarifies the comment describing the argument
selection that it got wrong.


djwatson:
tricky.  Worth a test case, or rely on CI?

davidtgoldblatt:
This actually did cause test failures, but in configurations I didn't test :(.

PR #590, opened 2017-01-30
Disentangle locking

jasone:
These changes were motivated by a gdump deadlock, which ended up being a lot cleaner to fix in the context of the ```extents_t``` refactor.  I did get a fix working prior to introducing ```extents_t```, but there were so many sharp edges I didn't want to propose it even as an interim state.

In general, we still have quite a bit of work to do if we want to eventually reach a point where no locks are ever held during system calls (ckh as used in the prof code is the most obvious sticking point), but at least these changes get us a lot closer.


src/jemalloc.c
```diff
@@ -1616,7 +1616,7 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts) {
 	 * some reason.  Let's grab it right away.
 	 */
 	tsd = tsd_fetch();
-	witness_assert_lockless(tsd_tsdn(tsd));
+	witness_assert_depth(tsd_tsdn(tsd), 0);
```
interwq: What do you think about keeping witness_assert_lockless API (simple wrapper), for the sake of a tiny bit easier to understand / use (maybe that's just me), but at the cost of a bit of redundant API.

src/prof.c
```diff
@@ -741,8 +753,15 @@ prof_lookup_global(tsd_t *tsd, prof_bt_t *bt, prof_tdata_t *tdata,
 		gctx.p->nlimbo++;
 		malloc_mutex_unlock(tsd_tsdn(tsd), gctx.p->lock);
 		new_gctx = false;
+
+		prof_leave(tsd, tdata);
```
interwq: move the prof_leave @ L737 to after this branch, and get rid of this one?

include/jemalloc/internal/tcache_externs.h
```diff
@@ -43,5 +43,8 @@ bool	tcaches_create(tsd_t *tsd, unsigned *r_ind);
 void	tcaches_flush(tsd_t *tsd, unsigned ind);
 void	tcaches_destroy(tsd_t *tsd, unsigned ind);
 bool	tcache_boot(tsdn_t *tsdn);
+void tcache_prefork(tsdn_t *tsdn);
```
interwq: tab vs space? Or I'm tricked by github again?

src/tcache.c
```diff
@@ -424,28 +427,37 @@ tcache_stats_merge(tsdn_t *tsdn, tcache_t *tcache, arena_t *arena) {
...
+			goto label_return;
 		}
 	}
 
 	if (tcaches_avail == NULL && tcaches_past > MALLOCX_TCACHE_MAX) {
-		return true;
+		err = true;
+		goto label_return;
 	}
 	arena = arena_ichoose(tsd, NULL);
```
interwq: I think this can trigger arena creation? If so maybe consider releasing tcaches_mtx before this. Looks like if we move the { tcaches_past++; }  branch to before this, then the code after doesn't need to be protected.

jasone: Good idea.  Same for the ```tcache_create()``` call below.

include/jemalloc/internal/witness_inlines.h
```diff
@@ -92,14 +96,28 @@ witness_assert_lockless(tsdn_t *tsdn) {
...
+	d = 0;
 	witnesses = tsd_witnessesp_get(tsd);
 	w = ql_last(witnesses, link);
 	if (w != NULL) {
-		witness_lockless_error(witnesses);
+		ql_reverse_foreach(w, witnesses, link) {
+			if (w->rank < rank_inclusive) {
+				break;
+			}
+			d++;
```
jasone: There isn't a simple way to add an assertion here, so I'm going to leave this as-is.

src/large.c
```diff
@@ -291,32 +290,39 @@ large_ralloc(tsdn_t *tsdn, arena_t *arena, extent_t *extent, size_t usize,
...
 	malloc_mutex_lock(tsdn, &arena->large_mtx);
-	ql_remove(&arena->large, extent, ql_link);
+	extent_list_remove(&arena->large, extent);
 	malloc_mutex_unlock(tsdn, &arena->large_mtx);
 	if (!junked_locked) {
 		large_dalloc_maybe_junk(extent_addr_get(extent),
 		    extent_usize_get(extent));
 	}
-	arena_extent_dalloc_large(tsdn, arena, extent, junked_locked);
+	arena_extent_dalloc_large_prep(tsdn, arena, extent, junked_locked);
```
interwq: I couldn't seem to locate this function. Is it generated in a header?

src/tcache.c
```diff
@@ -189,7 +187,17 @@ tcache_bin_flush_large(tsd_t *tsd, tcache_bin_t *tbin, szind_t binind,
 		if (config_prof) {
 			idump = false;
 		}
+
 		malloc_mutex_lock(tsd_tsdn(tsd), &locked_arena->lock);
+		for (unsigned i = 0; i < nflush; i++) {
+			void *ptr = *(tbin->avail - 1 - i);
+			assert(ptr != NULL);
+			extent = iealloc(tsd_tsdn(tsd), ptr);
+			if (extent_arena_get(extent) == locked_arena) {
```
interwq: It's probably worth saving the is_locked_arena results on stack so that we don't have to decide again in the second loop?

src/tcache.c
```diff
@@ -422,32 +434,56 @@ tcache_stats_merge(tsdn_t *tsdn, tcache_t *tcache, arena_t *arena) {
...
-	arena_t *arena;
-	tcache_t *tcache;
-	tcaches_t *elm;
+static bool
+tcaches_create_prep(tsd_t *tsd) {
+	bool err;
+
+	malloc_mutex_lock(tsd_tsdn(tsd), &tcaches_mtx);
 
 	if (tcaches == NULL) {
```
interwq: This looks to be one time. Shall we do this during init (maybe in tcache_boot)? If so we should be able to merge the (tcaches_past > MALLOCX_TCACHE_MAX) check into L487, and get rid of this prep function.

include/jemalloc/internal/witness_inlines.h
interwq: Do you think it's necessary to add some sanity check in qr_foreach etc? e.g. add a counter and assume it's smaller than uint32_max.

src/tcache.c
jasone: ```extent_arena_get()``` is an inline function that simply dereferences a field of ```extent```, so it amounts to a cached load.  I don't think that's expensive enough to justify caching, is it?

src/tcache.c
jasone: The reason this is lazy is that we don't ever call ```tcaches_create()``` unless the application uses the "tcache.create" mallctl.  Automatic per thread tcaches don't use this array.

include/jemalloc/internal/witness_inlines.h
jasone: In practice I haven't found this to be a common problem, and in fact now that I think about it the error-prone cases all involve ring mutation during iteration.  The foreach macros basically can't fail unless the ring is outright corrupt (links traverse into a cycle that does not include the element of origin).

That said, I'm open to adding sanity checks.  A good one would be assuring that linkage is initialized prior to insertion; I do continue to get burned by forgetting to call ```qr_new()```/```ql_elm_new()```.

src/tcache.c
interwq: Ah I mean we can avoid the ptr to extent translation (which is more expensive I think) above this (probably just caching extent).

src/tcache.c
interwq: I see. Thanks for explaining!

src/tcache.c
jasone: Oh!  That's an interesting optimization idea.  I will create an issue to keep track of it.

src/large.c
interwq: Hmm. I think I have seen this a couple of times. I'll see if I can report bug somewhere.

src/tcache.c
jasone: #598

jasone:
Manually tested on macOS via run_tests.sh; pushing.

PR #592, opened 2017-01-31
Handle race in stats_arena_bins_print

### LINKED ISSUE #591: Assert "assert(milli <= 1000)" in stats_arena_bins_print
We occasionally see this assert:

#7  0x000000000104ab7f in stats_arena_bins_print (write_cb=0x558de1, cbopaque=0x7fc3b7ffec60, json=false, large=true, huge=true, i=32) at jemalloc/src/stats.c:140
(gdb) info locals
availregs = 97
milli = 1020
util = "0.450"
reg_size = 4096
nregs = 1
nrequests = 3577
curruns = 97
nmalloc = 6298
ndalloc = 6199
nfills = 1303
nflushes = 1834
nruns = 6298
run_size = 4096
curregs = 99
nreruns = 0
page = 4096
config_tcache = true
in_gap = false
in_gap_prev = false
nbins = 36
j = 28

VERSION:
4.3.1-0-g0110fa8451af905affd77c3bea0d545fee2251b2

Our program periodically calls malloc_stats_print and records the data for debugging purposes.    I don't understand the code but the assertion seems to imply that 'curregs' should be equal to or less than 'availregs'.    We never see any problems in Release builds related to this code so maybe its an overzealous assert?   Thanks for your help!

interwq:
When multiple threads calling stats_print, race could happen as we read the
counters in separate mallctl calls. The removed assertion could fail when there
are other operations happening at the same time. For simplicity, assume a safe
bound when race is detected instead of retrying.

This should resolve #591.

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -133,8 +133,16 @@ stats_arena_bins_print(void (*write_cb)(void *, const char *), void *cbopaque,
 			availregs = nregs * curslabs;
 			milli = (availregs != 0) ? (1000 * curregs) / availregs
 			    : 1000;
-			assert(milli <= 1000);
-			if (milli < 10) {
+
+			if (milli > 1000) {
+				/*
+				 * Race detected: the counters were read in
+				 * separate mallctl calls and concurrent
+				 * operations happened in between. In this case
+				 * no meaningful utilization can be computed.
+				 */
+				malloc_snprintf(util, sizeof(util), " race");
+			} else if (milli < 10) {
 				malloc_snprintf(util, sizeof(util),
 				    "0.00%zu", milli);
 			} else if (milli < 100) {
@@ -144,6 +152,7 @@ stats_arena_bins_print(void (*write_cb)(void *, const char *), void *cbopaque,
 				malloc_snprintf(util, sizeof(util), "0.%zu",
 				    milli);
 			} else {
+				assert(milli == 1000);
 				malloc_snprintf(util, sizeof(util), "1");
 			}
 


interwq:
@jasone: yeah there's no really correct or meaningful number when race happened. However the concern about printing "race" is that for applications that log this data, they likely already assumed numeric data for this field. Writing text there will likely break some logging software built around this.

The other approach I considered was: 

if (race_detected) {
  j--;
  continue; // re-do the current iteration
}

and we'll also need to adjust the separator output a bit here:
https://github.com/interwq/jemalloc/blob/34d41cd2401a4678447cd55deb66791f2f2b3d8e/src/stats.c#L76

but this results in an unbounded loop potentially, unless we set a re-try bound (but again we may not have meaningful number). What do you think?

jasone:
I don't mind printing a non-numeric value here, because 1) anyone logging the data should be using the JSON output, and 2) utilization is a derived field that isn't output in JSON.

Your retry suggestion got me thinking about the general raciness in ```malloc_stats_print()```.  One way to solve the problem would be to gather all the stats before printing anything, and retry until the epoch stays the same for all stats reads.  That's a huge painful refactor though, and it has the unbounded loop problem even worse.

The manual says:
> Note that malloc_message() uses the mallctl*() functions internally, so inconsistent statistics can be reported if multiple threads use these functions simultaneously.
I kind of wish now that ```malloc_stats_print()``` would return an error if an epoch update occurred during printing, but that would break binary compatibility, and probably isn't worth the trouble.

I'm okay with this patch as long as you add a comment indicating that the computed value is bogus/arbitrary in that code path.

interwq:
Oh that's true the util isn't included in JSON. In this case I'm not against putting "race" there at all.

For a quick fix, I think maybe one of the following options should be fine:
1) output "race"
2) output an arbitrary const, e.g. "0000"
3) output an computed arbitrary like the current version

I don't really have a preference there. @jasone help me to pick one :)

I agree that the epoch based approach would be very nice, as I believe consistency is pretty valuable here. But considering the amount of effort and the unbounded latency, it's unclear if we really want to that. Would you like me to create an issue for it, or drop the thoughts for now?

jasone:
Let's go with "race", "?", or something else that can't be easily confused for a valid value.

Re: creating a tracking issue for potential ```malloc_stats_print()``` enhancement, yes, please.  We may never decide to act on it, but this is worth revisiting before making a final decision.

src/stats.c
```diff
@@ -133,8 +133,16 @@ stats_arena_bins_print(void (*write_cb)(void *, const char *), void *cbopaque,
...
-			if (milli < 10) {
+
+			if (milli > 1000) {
+				/*
+				 * Race detected: the counters were read in
+				 * separate mallctl calls and concurrent
+				 * operations happened in between. In this case
+				 * no meaningful utilization can be computed.
+				 */
+				malloc_snprintf(util, sizeof(util), " race");
```
jasone: I think the leading space in ```" race"``` will mess up alignment.

src/stats.c
```diff
@@ -133,8 +133,16 @@ stats_arena_bins_print(void (*write_cb)(void *, const char *), void *cbopaque,
 			availregs = nregs * curslabs;
 			milli = (availregs != 0) ? (1000 * curregs) / availregs
 			    : 1000;
-			assert(milli <= 1000);
-			if (milli < 10) {
+
+			if (milli > 1000) {
+				/*
+				 * Race detected: the counters were read in
```
jasone: :+1:

src/stats.c
interwq: We use %-5s for print so space should be fine. With numeric values we always output 5 chars. I tried a little bit locally and it looked a bit better with the space ("race" would be aligned with "util", otherwise it would be shifted by one char). Either case the alignment of other fields is not impacted. Any case the space would hurt?

jasone: As long as it aligns correctly in your tests, it's all good. :smile:

interwq:
Travis CI had infra failure. Earlier test was fine. Pushing.

PR #593, opened 2017-01-31
Introduce scripts to run many tests

davidtgoldblatt:
In 6e7d0890 we added better travis continuous integration tests. This is nice,
but has two problems:
- We run only a subset of interesting tests.
- The travis builds can take hours to give us back results (especially on OS X).

This adds scripts/gen_run_tests.py, and its output, run_tests.sh, which builds
and runs a larger portion of possible configurations on the local machine.

While a travis run takes several hours to complete , I can run these scripts on
my (OS X) latop and (Linux) devserve, and get a more exhaustive set of results
back in around 10 minutes.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,0 +1,1 @@
+$(dirname "$)")/scripts/gen_run_tests.py | bash
--- a/scripts/gen_run_tests.py
+++ b/scripts/gen_run_tests.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+
+from itertools import combinations
+
+def powerset(items):
+    result = []
+    for i in xrange(len(items) + 1):
+        result += combinations(items, i)
+    return result
+
+MAKE_J_VAL = 32
+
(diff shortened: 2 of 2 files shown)


davidtgoldblatt:
Convenience is all; I think being able to say ./run_tests.sh is a little nicer than writing ./scripts/gen_run_tests.py | bash . WDYT?

jasone:
Can we have the Python script generate and run the script with one invocation?

davidtgoldblatt:
You mean with (e.g.) Popen? We could; it'd just make the script more complicated. At that point I think it makes sense to do all process management from python, and not use the shell at all (which I'm not necessarily opposed to; just want to make sure it's complexity you're OK with).

jasone:
Another possibility would be to provide a shell script that wraps calling the Python script and invoking the resulting script, rather than checking in the generated output.

davidtgoldblatt:
Hah, yes. Good point.

Issue #599, opened 2017-02-03
Assert failure in extent.c


interwq:
Happens w/ t-test1, but repro rate is low (maybe 1 in 10 runs). This introduced in 
https://github.com/jemalloc/jemalloc/commit/d27f29b468ae3e9d2b1da4a9880351d76e5a1662

config: enable-debug and enable-prof.
`assert(extent_state_get(extent) == extents_state_get(extents));`
https://github.com/jemalloc/jemalloc/blob/dev/src/extent.c#L357

Core dump:
#0  0x00007fe1976a91d7 in raise () at /lib64/libc.so.6
#1  0x00007fe1976aa8c8 in abort () at /lib64/libc.so.6
#2  0x00007fe197c766b3 in extent_activate_locked (tsdn=0x7fe17d2b4578, arena=0x7fe196f5c980, extents=0x7fe196f5f078, extent=0x7fe0f7d811a0) at src/extent.c:357
#3  0x00007fe197c78b2a in extent_try_coalesce (tsdn=0x7fe17d2b4578, arena=0x7fe196f5c980, r_extent_hooks=0x7fe17d2b3980, a=0x7fe0f7da0440, b=0x7fe0f7d811a0, extents=0x7fe196f5f078) at src/extent.c:1043
#4  0x00007fe197c78dd1 in extent_record (tsdn=0x7fe17d2b4578, arena=0x7fe196f5c980, r_extent_hooks=0x7fe17d2b3980, extents=0x7fe196f5f078, extent=0x7fe0f7da0440) at src/extent.c:1091
#5  0x00007fe197c782ce in extent_grow_retained (tsdn=0x7fe17d2b4578, arena=0x7fe196f5c980, r_extent_hooks=0x7fe17d2b3980, new_addr=0x0, usize=32768, pad=4096, alignment=64, zero=0x7fe17d2b39d7, commit=0x7fe17d2b3907, slab=false) at src/extent.c:878
#6  0x00007fe197c786b0 in extent_alloc_retained (tsdn=0x7fe17d2b4578, arena=0x7fe196f5c980, r_extent_hooks=0x7fe17d2b3980, new_addr=0x0, usize=32768, pad=4096, alignment=64, zero=0x7fe17d2b39d7, commit=0x7fe17d2b3907, slab=false) at src/extent.c:952
[...]

PR #600, opened 2017-02-03
Relax extent state checking assertions.

interwq:
Attempt fix for #599


interwq:
Not sure if it the assertions are just too strict.

jasone:
Assertion failures suggest an error elsewhere; the extents heaps should *never* contain extents with mismatching states.  I don't see the cause via code inspection, but we can probably get closer to the source of the issue by adding some (temporary expensive) assertions in ```extent_state_set()```.  Do you want me to take over investigation, since I broke it and the design is fresh in my mind?

interwq:
@jasone : of course, please feel free to take over.

Issue #465, opened 2016-10-04
Implement further rtree optimizations


jasone:
The rtree code implements multiple optimizations to speed up common case lookups, especially skipping the leading 16 bits and path caching, but it could be further specialized to encode differing numbers of bits at each tree level.  Currently (assuming 4 KiB pages) we encode 64-bit addresses as {[16],16,16,4}, and 32-bit addresses as {16,4}.  We would be better served by {[16],18,18} on x64, and {10,10} or perhaps even {20} (4 MiB) on 32-bit systems.

On a related note, although `rtree_elm_lookup()` is explicitly unrolled, we dynamically initialize `rtree_t`, which forces the compiler to load run-time constants rather than encode them as compile-time immediates.  The only purpose of the current generality is testing; surely it is still possible to adequately test the code absent this pessimization.

See #360 for context.

djwatson:
Some initial experiments with this:

* For 64bit, {[16], 16, 20} seemed to work quite well, slightly better than {[16], 18, 18}. 
* If I'm following the second comment, I didn't notice any difference trying a static rtree_t in rtree_elm_lookup.
* For a two-level tree, the path cache seemed to be slightly *worse* than no path cache.

Some microbench numbers:
master: 1.22
dev: 2.02
two-level tree, no cache: 1.89

jasone:
Any idea why {[16], 16, 20} was faster than {[16], 18, 18}?  I'd expect them to be essentially the same speed, with only a minor difference in memory usage, depending on the application.

djwatson:
Sorry hit enter too early and updated the comment - probably because level[1] fits in cache better.  Might just be an artifact of the microbench.

interwq:
A couple years ago one optimization we had on a customized rtree
implementation was to mark all non-valid nodes point to a special node that
indicates lookup returns nothing (and that special node points to itself).
This can get rid of the branches when travel down though the rtree (we
always jump N times where N is the depth). The idea should also be
applicable on the rtree here?

djwatson:
@interwq - The fast path is dalloc, where the rtree nodes must exist (or free() was passed an invalid pointer), so I don't think that would help here.

djwatson:
Oh nevermind, I understand now - perf shows the cache misses as the expensive part, but it's worth trying removing some branches.

interwq:
Yep. For the hot cache case, we would be able to lookup w/o any branches.

jasone:
[...]

PR #602, opened 2017-02-04
Simplify/optimize rtree

jasone:
These changes primarily simplify the rtree lookup caching mechanism, making it more robust and slightly faster (measured ~2% speedup for test/stress/microbench).  This makes progress on #465, but does not address the bit width imbalance between levels; that will be a separate effort.

The separate changes should be pretty straightforward to review, as opposed to their squashed equivalent.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -406,6 +406,74 @@ case "${host_cpu}" in
 esac
 AC_DEFINE_UNQUOTED([CPU_SPINWAIT], [$CPU_SPINWAIT])
 
+case "${host_cpu}" in
+  aarch64)
+    AC_MSG_CHECKING([number of significant virtual address bits])
+    LG_VADDR=48
+    AC_MSG_RESULT([$LG_VADDR])
+    ;;
+  x86_64)
+    AC_CACHE_CHECK([number of significant virtual address bits],
+                   [je_cv_lg_vaddr],
+                   AC_RUN_IFELSE([AC_LANG_PROGRAM(
+[[
+#include <stdio.h>
+#ifdef _WIN32
+#include <limits.h>
+#include <intrin.h>
+typedef unsigned __int32 uint32_t;
+#else
+#include <stdint.h>
+#endif
+]], [[
+	uint32_t r[[4]];
+	uint32_t eax_in = 0x80000008U;
+#ifdef _WIN32
+	__cpuid((int *)r, (int)eax_in);
+#else
+	asm volatile ("cpuid"
+	    : "=a" (r[[0]]), "=b" (r[[1]]), "=c" (r[[2]]), "=d" (r[[3]])
+	    : "a" (eax_in), "c" (0)
+	);
+#endif
+	uint32_t eax_out = r[[0]];
+	uint32_t vaddr = ((eax_out & 0x0000ff00U) >> 8);
+	FILE *f = fopen("conftest.out", "w");
+	if (f == NULL) {
+		return 1;
+	}
+	fprintf(f, "%u", vaddr);
+	fclose(f);
+	return 0;
+]])],
+                   [je_cv_lg_vaddr=`cat conftest.out`],
+                   [je_cv_lg_vaddr=error],
+                   [je_cv_lg_vaddr=57]))
+    if test "x${je_cv_lg_vaddr}" != "x" ; then
+      LG_VADDR="${je_cv_lg_vaddr}"
+    fi
+    if test "x${LG_VADDR}" != "xerror" ; then
+      AC_DEFINE_UNQUOTED([LG_VADDR], [$LG_VADDR])
+    else
+      AC_MSG_ERROR([cannot determine number of significant virtual address bits])
+    fi
+    ;;
+  *)
+    AC_MSG_CHECKING([number of significant virtual address bits])
+    if test "x${LG_SIZEOF_PTR}" = "x3" ; then
+      LG_VADDR=64
+    elif test "x${LG_SIZEOF_PTR}" = "x2" ; then
+      LG_VADDR=32
+    elif test "x${LG_SIZEOF_PTR}" = "xLG_SIZEOF_PTR_WIN" ; then
+      LG_VADDR="(1U << (LG_SIZEOF_PTR_WIN+3))"
+    else
+      AC_MSG_ERROR([Unsupported lg(pointer size): ${LG_SIZEOF_PTR}])
+    fi
+    AC_MSG_RESULT([$LG_VADDR])
+    ;;
+esac
+AC_DEFINE_UNQUOTED([LG_VADDR], [$LG_VADDR])
+
 LD_PRELOAD_VAR="LD_PRELOAD"
 so="so"
 importlib="${so}"
--- a/include/jemalloc/internal/rtree_inlines.h
+++ b/include/jemalloc/internal/rtree_inlines.h
@@ -2,8 +2,8 @@
 #define JEMALLOC_INTERNAL_RTREE_INLINES_H
 
 #ifndef JEMALLOC_ENABLE_INLINE
-uintptr_t rtree_leafkey(rtree_t *rtree, uintptr_t key);
-uintptr_t rtree_subkey(rtree_t *rtree, uintptr_t key, unsigned level);
+uintptr_t rtree_leafkey(uintptr_t key);
+uintptr_t rtree_subkey(uintptr_t key, unsigned level);
 extent_t *rtree_elm_read(rtree_elm_t *elm, bool dependent);
 void rtree_elm_write(rtree_elm_t *elm, const extent_t *extent);
 rtree_elm_t *rtree_elm_lookup(tsdn_t *tsdn, rtree_t *rtree,
@@ -25,21 +25,21 @@ void rtree_clear(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
 
 #if (defined(JEMALLOC_ENABLE_INLINE) || defined(JEMALLOC_RTREE_C_))
 JEMALLOC_ALWAYS_INLINE uintptr_t
-rtree_leafkey(rtree_t *rtree, uintptr_t key) {
+rtree_leafkey(uintptr_t key) {
 	unsigned ptrbits = ZU(1) << (LG_SIZEOF_PTR+3);
-	unsigned cumbits = (rtree->levels[rtree->height-1].cumbits -
-	    rtree->levels[rtree->height-1].bits);
+	unsigned cumbits = (rtree_levels[RTREE_HEIGHT-1].cumbits -
+	    rtree_levels[RTREE_HEIGHT-1].bits);
 	unsigned maskbits = ptrbits - cumbits;
 	uintptr_t mask = ~((ZU(1) << maskbits) - 1);
 	return (key & mask);
 }
 
 JEMALLOC_ALWAYS_INLINE uintptr_t
-rtree_subkey(rtree_t *rtree, uintptr_t key, unsigned level) {
+rtree_subkey(uintptr_t key, unsigned level) {
 	unsigned ptrbits = ZU(1) << (LG_SIZEOF_PTR+3);
-	unsigned cumbits = rtree->levels[level].cumbits;
+	unsigned cumbits = rtree_levels[level].cumbits;
 	unsigned shiftbits = ptrbits - cumbits;
-	unsigned maskbits = rtree->levels[level].bits;
+	unsigned maskbits = rtree_levels[level].bits;
 	unsigned mask = (ZU(1) << maskbits) - 1;
 	return ((key >> shiftbits) & mask);
 }
@@ -82,7 +82,7 @@ rtree_elm_lookup(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
 	assert(!dependent || !init_missing);
 
 	if (likely(key != 0)) {
-		uintptr_t leafkey = rtree_leafkey(rtree, key);
+		uintptr_t leafkey = rtree_leafkey(key);
 #define RTREE_CACHE_CHECK(i) do {					\
 		if (likely(rtree_ctx->cache[i].leafkey == leafkey)) {	\
 			rtree_elm_t *leaf = rtree_ctx->cache[i].leaf;	\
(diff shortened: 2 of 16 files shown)


jasone:
This subsumes #604, which had the following description.

These changes simplify rtree lookup by removing the dynamic shortcut for leading zeros, and by determining the number of bits in each level at compile time rather than run time. On x64 the rtree changes from {[16], 16, 16, 4} (four levels, but the first one was skipped) to {18, 18}.

This wraps up everything I had in mind re: #465.

configure.ac
```diff
@@ -406,6 +406,26 @@ case "${host_cpu}" in
 esac
 AC_DEFINE_UNQUOTED([CPU_SPINWAIT], [$CPU_SPINWAIT])
 
+AC_MSG_CHECKING([number of significant virtual address bits])
+case "${host_cpu}" in
+  x86_64|aarch64)
+    LG_VADDR=48
```
djwatson: Just FYI this is changing soon to 57:

https://software.intel.com/sites/default/files/managed/2b/80/5-level_paging_white_paper.pdf

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -348,7 +348,7 @@ rtree_read(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx, uintptr_t key,
 	rtree_elm_t *elm;
 
 	elm = rtree_elm_lookup(tsdn, rtree, rtree_ctx, key, dependent, false);
-	if (elm == NULL) {
+	if (!dependent && elm == NULL) {
```
djwatson: What case does this change?  I don't quite follow

include/jemalloc/internal/rtree_types.h
```diff
@@ -25,11 +26,17 @@ typedef struct rtree_s rtree_t;
 #define RTREE_HEIGHT_MAX						\
     ((1U << (LG_SIZEOF_PTR+3)) / RTREE_BITS_PER_LEVEL)
 
+/*
+ * Number of leafkey/leaf pairs to cache.  Each entry supports an entire leaf,
+ * so the cache hit rate is typically high even with a small number of entries.
+ * Two entries is enough to keep the hit rate high even when an arena uses a
+ * combination of dss and mmap.
+ */
+#define RTREE_CTX_NCACHE 2
```
djwatson: Did you do benchmarks or tests for 2 vs. other numbers?

configure.ac
jasone: Do you know if this is something we can detect via a simple configure test (no need to maintain a table of CPU revisions)?

include/jemalloc/internal/rtree_inlines.h
jasone: Dependent lookups cannot result in ```NULL```, since by definition the rtree element must already exist, so this allows the ```elm == NULL``` check to be omitted at compile time.  In practice this doesn't do us any good anymore, because I recently removed the last uses of ```rtree_read()``` (5177995530557521d330486a3971469e1573d6fc).

include/jemalloc/internal/rtree_types.h
jasone: I did a bunch of benchmarks prior to converting the LRU cache from element to leaf caching, and 2 was the largest number that didn't show a substantial negative impact when suffering a high miss rate.  After converting to a leaf cache, I measured very little sensitivity to this setting, because the hit rate is very high in practice.  We could perhaps bump this up as high as ~8, but beyond that we might suffer tsd cacheline misses instead of rtree cacheline misses, negating any benefit.

configure.ac
jasone: According to that doc, the CPUID instruction exposes this info, so I'm inclined to make the configure test more robust.

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -170,103 +162,86 @@ rtree_subtree_read(tsdn_t *tsdn, rtree_t *rtree, unsigned level,
...
+				uintptr_t subkey = rtree_subkey(rtree, key,
+				    rtree->height-1);
+				return &leaf[subkey];
+			}
+		}
+		/*
+		 * Search the remaining cache elements, and on success move the
+		 * matching element to the front.
+		 */
+		for (unsigned i = 1; i < RTREE_CTX_NCACHE; i++) {
```
interwq: Would it make sense to experiment with a simple and small global cache (e.g. only 8 items), for when we have local cache miss? That way when a leaf node is newly committed or becomes hot recently, not all threads will need to walk though the rtree.

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -170,103 +162,86 @@ rtree_subtree_read(tsdn_t *tsdn, rtree_t *rtree, unsigned level,
...
-					    rtree_subtree_tryread(rtree,
-					    no_ctx_start_level, dependent);
-					rtree_ctx->start_level =
-					    no_ctx_start_level;
-					rtree_ctx->elms[no_ctx_start_level] =
-					    node;
+	/* Search the cache. */
+	uintptr_t leafkey = rtree_leafkey(rtree, key);
+	if (likely(key != 0)) {
+		if (likely(rtree_ctx->cache[0].leafkey == leafkey)) {
```
interwq: Is it true that we never update / remove a mapping from the rtree? Otherwise we would need to worry about refreshing the local cache?

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -170,103 +162,86 @@ rtree_subtree_read(tsdn_t *tsdn, rtree_t *rtree, unsigned level,
...
+		/*
+		 * Search the remaining cache elements, and on success move the
+		 * matching element to the front.
+		 */
+		for (unsigned i = 1; i < RTREE_CTX_NCACHE; i++) {
+			if (likely(rtree_ctx->cache[i].leafkey == leafkey)) {
+				rtree_elm_t *leaf = rtree_ctx->cache[i].leaf;
+				if (likely(leaf != NULL)) {
+					/* Reorder. */
+					memmove(&rtree_ctx->cache[1],
```
interwq: Probably worth having a #ifdef RTREE_CTX_NCACHE == 2, do an assignment directly instead of memmove? I'm not quite optimistic with all cases / compilers there.

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -170,103 +162,86 @@ rtree_subtree_read(tsdn_t *tsdn, rtree_t *rtree, unsigned level,
...
 		/*							\
 		 * node is a leaf, so it contains values rather than	\
 		 * child pointers.					\
 		 */							\
-		node = &node[subkey];					\
-		if (dependent || init_missing) {			\
-			rtree_ctx->elms[level - RTREE_GET_BIAS + 1] =	\
-			    node;					\
+		if (likely(key != 0)) {					\
+			if (RTREE_CTX_NCACHE > 1) {			\
```
interwq: Maybe also here about manual copy instead of memmove when RTREE_CTX_NCACHE == 2

src/rtree.c
```diff
@@ -158,6 +158,135 @@ rtree_child_read_hard(tsdn_t *tsdn, rtree_t *rtree, rtree_elm_t *elm,
...
+rtree_elm_t *
+rtree_elm_lookup_hard(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
+    uintptr_t key, bool dependent, bool init_missing) {
+	uintptr_t leafkey = rtree_leafkey(rtree, key);
+	if (likely(key != 0)) {
+		/*
+		 * Search the remaining cache elements, and on success move the
+		 * matching element to the front.
+		 */
+		for (unsigned i = 1; i < RTREE_CTX_NCACHE; i++) {
```
interwq: I'd still vote for keeping this cache searching part in the inlined function. The impact should be quite minimal but I'm not confident at all about application locality fitting well in only one leaf node.

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -130,21 +130,22 @@ rtree_read(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx, uintptr_t key,
...
+	while (true) {
+		extent_t *extent = rtree_elm_read(elm, false);
 		/* The least significant bit serves as a lock. */
-		s = (void *)((uintptr_t)extent | (uintptr_t)0x1);
-	} while (atomic_cas_p(&elm->pun, (void *)extent, s));
+		void *s = (void *)((uintptr_t)extent | (uintptr_t)0x1);
+		if (!atomic_cas_p(&elm->pun, (void *)extent, s)) {
+			break;
+		}
+		spin_adaptive(&spinner);
```
interwq: It's probably worth adding some logic to detect long spin (in which case likely the lock holder was preempted), and do sched_yield() in that case. Without that the concern is large latency spike caused by spinning.

include/jemalloc/internal/rtree_inlines.h
jasone: Indeed, once an rtree node is put in place, we never remove it.  The individual leaf node entries can change over time, but we never risk a ```NULL``` pointer dereference for dependent rtree reads.

include/jemalloc/internal/rtree_inlines.h
jasone: I suspect that our cache miss rate is so low that an additional caching layer would be almost never utilized.

include/jemalloc/internal/rtree_inlines.h
jasone: I did an experiment with manual copying prior to the elm --> leaf cache conversion, and didn't detect any performance difference.  But, I had the same concern as you.  I'll take a look at the disassembly after moving this code into ```rtree_elm_lookup()``` as you suggest, and add manual copying if the compiler isn't being smart enough.

src/rtree.c
jasone: Sounds good.

include/jemalloc/internal/rtree_inlines.h
jasone: Okay, how about I modify [```spin_adaptive()```](https://github.com/jemalloc/jemalloc/blob/3bd6d8e41d524737845949c44bd4961b0930965c/include/jemalloc/internal/spin_inlines.h#L16) to call ```sched_yield()``` instead of spinning once the iteration count reaches ~32?

interwq: I did a simple benchmark to get an idea about the cycle count:
for (i=0; i < 128k; i++) { // 32 threads running
  CPU_SPINWAIT;
}
and each iteration takes ~20 cycles on Xeon(R) CPU E5-2660. Without the pause, each empty iter takes ~6 cycles.

That makes me wonder if the upper bound is too high in [```spin_adaptive()```](https://github.com/jemalloc/jemalloc/blob/3bd6d8e41d524737845949c44bd4961b0930965c/include/jemalloc/internal/spin_inlines.h#L16). As the loop count is (KQU(1) << spin->iteration), even iteration == 6 makes it (2^6 * 20) = ~1280 cycles. Adding the previous iterations, the total will be  ~2.5k cycles and should be sufficient to conclude a long spin. (actually probably should taking it down a bit)

interwq: Yeah I believe 32 as upper bound makes sense. In which case the total spinning cycles would be ~1260 cycles, or ~0.5us on the two socket Xeon(R) CPU E5-2660 I tried.

include/jemalloc/internal/rtree_inlines.h
interwq: Sounds good.

include/jemalloc/internal/rtree_inlines.h
interwq: Got it. Thanks!

include/jemalloc/internal/rtree_inlines.h
jasone: I looked at the disassembly, and with ```RTREE_CTX_NCACHE``` set to 2, the compiler emits a short sequence of move instructions, whereas when set to 8 the compiler emits a call to ```memmove()```.

include/jemalloc/internal/rtree_inlines.h
jasone: I came to a similar conclusion, though I chose 256 as the upper bound rather less methodically. :wink:  I'll switch to 32.

include/jemalloc/internal/rtree_inlines.h
interwq: Gotcha. Great to hear :)

include/jemalloc/internal/rtree_types.h
jasone: I did some experiments with this set to 1, and in one run saw a high miss rate, presumably due to the active memory range straddling two rtree leaf nodes.  The more I think about this, the more I wonder what the appropriate setting is.  As currently configured, each entry supports 1 GiB of virtual memory, which isn't a lot these days...  I'd be tempted to set this to something big (like 128) so that the cache rarely fails even for large applications, but at some point the linear search becomes a problem.  I can come up with justifications for values in the [4..64] range, though that's based on a nebulous notion of how fast linear search (an ```memmove```) is.

Straw poll: What do you think an appropriate setting is, taking into consideration that each entry consumes 16 bytes?

djwatson: Should be based on cache misses - a tree search is potentially ~3 cacheline misses, so it can be CACHELINE_SIZE / 16 * 3 big before being worse? ~12?  Although since the cache is linear the prefetcher might help a bit.

interwq: I suggest going with 8 entries, which is 128 bytes (two cachelines). With the adjacent cacheline prefetching (always enabled I believe), getting the first cacheline always bring us the next neighbor anyway. Since even the cache miss here isn't too bad, I'm inclined to say no need going very aggressive here. 

Also worth making sure the it's cacheline aligned.

jasone: We have 8, <= 12, and 16 (mine) as poll results, which definitely helps narrow the range. :smile:  I'll make it 8 for now and we can reevaluate closer to 5.0.0 release if we see rtree performance issues.

Do we need double-cacheline alignement (128) to assure the following cacheline is pulled in automatically, or does it happen regardless of whether the cacheline is at an even multiple offset?

interwq: I don't think double cacheline alignment is necessary. From the documents I read they just say 
> two 64-byte cache lines are fetched into a 128-byte sector, regardless of whether the additional cache line has been requested or not

jasone:
AppVeyor seems to have lost the plot on one configuration.  Everything else looks good though (including the ```--enable-debug``` variant of the MIA configuration), so I'm pushing.

PR #604, opened 2017-02-05
Broaden rtrees

jasone:
[This is based on #602, hence the duplicate changesets.]

These changes simplify rtree lookup by removing the dynamic shortcut for leading zeros, and by determining the number of bits in each level at compile time rather than run time.  On x64 the rtree changes from {[16], 16, 16, 4} (four levels, but the first one was skipped) to {18, 18}.

This wraps up everything I had in mind re: #465.


jasone:
Folding into #602.

PR #605, opened 2017-02-06
Fix extent_record().

jasone:
Read adjacent rtree elements while holding element locks, since the
extents mutex only protects against relevant like-state extent mutation.

Fix management of the 'coalesced' loop state variable to merge
forward/backward results, rather than overwriting the result of forward
coalescing if attempting to coalesce backward.  In practice this caused
no correctness issues, but could cause extra iterations in rare cases.

These regressions were introduced by
d27f29b468ae3e9d2b1da4a9880351d76e5a1662 (Disentangle arena and extent
locking.).

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -1035,12 +1035,9 @@ extent_can_coalesce(const extent_t *a, const extent_t *b) {
 }
 
 static bool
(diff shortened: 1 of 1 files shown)


jasone:
Yes, ```extent->mtx``` covers relevant mutations to those extents.

PR #613, opened 2017-02-09
Fix extent_alloc_dss() regression.

jasone:
Fix extent_alloc_dss() to account for bytes that are not a multiple of
the page size.  This regression was introduced by
577d4572b0821a15e5370f9bf566d884b7cf707c (Make dss operations
lockless.), which was first released in 4.3.0.

### DIFF
--- a/src/extent_dss.c
+++ b/src/extent_dss.c
@@ -121,35 +121,45 @@ extent_alloc_dss(tsdn_t *tsdn, arena_t *arena, void *new_addr, size_t size,
 		 * malloc.
 		 */
 		while (true) {
-			void *ret, *max_cur, *gap_addr, *dss_next, *dss_prev;
-			size_t gap_size;
-			intptr_t incr;
-
-			max_cur = extent_dss_max_update(new_addr);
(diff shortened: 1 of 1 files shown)


src/extent_dss.c
```diff
@@ -162,10 +169,10 @@ extent_alloc_dss(tsdn_t *tsdn, arena_t *arena, void *new_addr, size_t size,
 			}
 
 			/* Try to allocate. */
-			dss_prev = extent_dss_sbrk(incr);
+			void *dss_prev = extent_dss_sbrk(incr);
```
interwq: nit: maybe an assertion like (incr >= size && incr <= size+page_size+alignment)

jasone: Added [assertion](https://github.com/jemalloc/jemalloc/pull/613/commits/91c2b54c7331d6d3c4b6e722b0c98ade8a0a9da0#diff-7e822d66b9b00db2d0a167d709b1e0e1R161).

jasone:
run_tests.sh succeeds on Linux and macOS.  Pushing.

Issue #466, opened 2016-10-04
Reduce contention in large [de]allocation fast path


jasone:
Large allocation currently requires holding both `arena->lock` and `arena->extents_mtx`.  We can push `extents_mtx` down to individual size classes so that different size classes can be operated on concurrently.  On top of that we can completely avoid `arena->lock` in the fast path if some stats fields are refactored such that the arena only stores per size class stats, and the ctl refresh code synthesizes the top level stats via aggregation.

Other issues:
- The `arena->large` list will need to be broken into per size class lists.
- The `arena->extents_dirty` ring will need to either become lock-free, or per size class (with e.g. per extent timestamps or serial numbers that allow approximate order reconstruction).

See #360 for context.

jasone:
We have now reduced large [de]allocation contention substantially, to the point that going any further would risk increased common-case overhead due acquiring/releasing more locks per operation.  The same applies to lockless data structures -- the heaps and LRUs are currently very fast (and protected by a single mutex lock/unlock per operation), but lockless replacements would require many atomic instructions per operation.  Current benchmarking evidence (both microbenchmarks and real-world) suggests that further optimization may not be warranted.  For the time being we are done with this issue.  If further scalability issues crop up, we can open more specific issues.

IndolentPangolin:
Just as a note from my own experience with lock-free code: lock-free code is almost always slower on any individual path. But it wins on parallelism, because while lock-free isn't always wait-free, it usually is. And so while each thread does more work individually, no thread gets stuck waiting for a thread that holds a lock but has been descheduled by the kernel.

jasone:
Note that e201e24904d53897409b1dda451d40c5d2e0dc29 was a follow-on to this issue, necessary to reduce TLB shootdown overhead.  Additionally, cdce93e4a3045bcf0d30409666d2d4c29818aec7 further streamlines fast-path large allocation by converting pairing heap operational complexity to the equivalent of LIFO operations.

PR #621, opened 2017-02-13
Refactor arena synchronization

jasone:
This series of changes refactors out arena->lock, through a combination of atomic operations (especially for stats) and field-specific independent mutexes.

This relates to #466.


include/jemalloc/internal/arena_inlines_a.h
```diff
@@ -34,44 +32,24 @@ arena_internal_get(arena_t *arena) {
...
-		ret = arena_prof_accum_impl(arena, accumbytes);
-		malloc_mutex_unlock(tsdn, &arena->lock);
-		return ret;
-	}
+	uint64_t a0, a1;
+	bool overflow;
+	do {
+		a0 = atomic_read_u64(&arena->prof_accumbytes);
+		a1 = a0 + accumbytes;
+		overflow = (a1 >= prof_interval);
```
interwq: Is it possible for accumbytes to be huge? If so this may overflow uint64, in which case the overflow calculation needs to be adjusted.

include/jemalloc/internal/arena_inlines_a.h
```diff
@@ -34,44 +32,24 @@ arena_internal_get(arena_t *arena) {
...
+		if (overflow) {
+			a1 %= prof_interval;
```
interwq: This implies that we may get 1 sample for multiple of prof_interval bytes. Probably worth mentioning in a comment?

jasone:
This is going to need some rework, because it uses 64-bit atomic ops, and we don't currently have those on 32-bit platforms.

include/jemalloc/internal/arena_inlines_a.h
jasone: ```prof_interval``` can be no larger than 2^63, so I don't think ```arena->prof_accumbytes``` can overflow in practice.

include/jemalloc/internal/arena_inlines_a.h
jasone: Will do.

jasone:
I'm going to close this pull request and create a new one once I have a workable strategy for 64-bit stats counters.

PR #623, opened 2017-02-15
Refactor arena synchronization

jasone:
[#621 redux]

This series of changes refactors out arena->lock, through a combination of atomic operations (especially for stats) and field-specific independent mutexes.

This relates to #466.

PR #624, opened 2017-02-15
Fix arena->stats.mapped accounting

jasone:
Ignore diffs that are part of #623; 956f395e301476d159ebc001a93f8889c35c17cd is the only relevant diff in this pull request.


interwq:
The AppVeyor failure on 32bit looks to be real.

jasone:
Yep, and the same failure occurs for 32-bit Linux.  I'll try to fix it tonight or tomorrow morning.

PR #625, opened 2017-02-16
Coalesce less

jasone:
Ignore diffs that are part of #623 and #624; 177dfda080d1a1bc7bcac1d3bda937ec6503e451 (Optimize extent coalescing.) and d75d00b29615063dac3f103790b53ebe5966387f (Disable coalescing of cached extents.) are the only ones relevant here.


jasone:
Note that this is related to #466, in that it reduces contended operations.

PR #637, opened 2017-02-22
Enable mutex witnesses even when !isthreaded.

jasone:
This fixes interactions with ```witness_assert_depth[_to_rank]()```, which was
added in dad74bd3c811ca2b1af1fd57b28f2456da5ba08b (Convert
```witness_assert_lockless()``` to ```witness_assert_lock_depth()```.).


jasone:
[Based on #632 to avoid merge conflicts during push.]

The previous (intermittent) test failure was due to overly restrictive witness assertions, which are fixed by 86689074c87d9fb662db547a4104a6f9d66cb5cd (Relax witness assertions related to prof_gdump().).  It turns out we have the same issue in 4.x as in the dev branch; this fix is essentially a backport.

jasone:
Ready for review.  Note that three of the six revisions belong to #632.

PR #639, opened 2017-02-22
Use MALLOC_CONF rather than malloc_conf for tests.

jasone:
This is the nature of the fix for the remaining failures in #632, but for the dev branch rather than rc-4.5.0.  I'll update #632 with a similar fix.


jasone:
Ready for review.

Issue #574, opened 2017-01-20
Mutex contention profiling


interwq:
Currently we don't have insights regarding mutex contention in jemalloc. At arena / bin level, it will be useful to add tracking for information like 
- how often a mutex is contended
- when contended, how long it takes to get the lock
- thread / CPU, e.g. # of threads / CPU contending on this lock

Should keep overhead minimal on fast path.

interwq:
Goals of mutex profiling:
1. Minimal impact on fast path (i.e. when there is no contention)
2. Minimal memory cost. We probably don't want to spend too much memory on this. A few counters per lock should be reasonable. Going histogram is probably too far.
3. Gather information to help us understand lock behavior, and guide design / performance tuning.

Current plan:
1. Instead of calling mutex_lock directly, use trylock first, enter slow path if lock is taken already;
2. on slow path, do appropriate accounting and customized adaptive spin. The customized spinning isn't required, but since we are gathering a bunch of information anyway, we can try making some better spin decisions.
3. High level output in malloc_stats, and new mallctl for control and detailed data.

Looks like all platforms / locking mechanisms we support have trylock:
``pthread_mutex_trylock``
MacOS:
``os_unfair_lock_trylock``
``OSSpinLockTry``
Windows:
``TryAcquireSRWLockExclusive``
``TryEnterCriticalSection``

Locks that we want to profile:
1. bin locks
2. extents lock
3. prof related locks (bt2gctx_mtx)
*time granularity when recording time -- ideally we want nano second granularity. Will need to check whether we have such support on each platform.

Candidate counters / events to profile --
1. average lock waiting time, when there is contention
On average how long do we have to wait (either spin or block) before we can acquire the lock.

2. longest lock waiting time, when there is contention
Max waiting time, during the last time period (likely need a mallctl to reset this). This should give information about lock related latency spikes.

3. % of locking w/ contention
i.e. (# of contention) / (total # of locking). This one adds tax on the fast path as we'll need to increment a counter after we grab the lock. It does offer quite valuable information though, and the cost should be acceptable (write to the same cacheline, and no atomic required) so should be worth doing.

4. % of locking acquired through spinning
To help us understand how effective our spinning policy is. Cost: one counter change on slow path.

5. % of lock holder different from last
[...]

PR #640, opened 2017-02-23
Get rid of witness in malloc_mutex_t when !(configured w/ debug).

interwq:
We don't touch witness at all when config_debug == false.  Let's only include it
in malloc_mutex_s when needed. Saves 44 bytes per mutex.


interwq:
Hmm got test failures (prof_accum and prof_reset) w/ enable-debug on freebsd. Couldn't repro locally. Any idea @jasone ?

jasone:
The FreeBSD failure is a configuration error on the buildbot machine.  I'll fix the configuration in a moment.

jasone:
Until a few years ago jemalloc was full of ```#ifdef``` logic similar to this, but eventually @bmaurer convinced me it was a code quality and testability liability, so I did a bunch of refactoring to minimize conditionally compiled code (e.g. ```if (config_debug) {...}``` rather than ```#ifdef JEMALLOC_DEBUG ... #endif```).  There are a few exceptions due to performance concerns, such as the size of [```opt_prof_prefix```](https://github.com/jemalloc/jemalloc/blob/de49674fbde4d124a0a7e7e97f5656e190980759/src/prof.c#L28).  I'm okay with this change if there's a performance reason to minimize the size of ```malloc_mutex_t```, but if it's only useful as a minimalism change, then we should have a high level discussion about whether we want to do more of this sort of thing.  Unlike in years past, we have adequate testing infrastructure to catch most conditional compilation mistakes, so I could potentially be convinced either way, though code readability remains a concern.

bmaurer:
You might be able to do something like this:

```
struct X {
union {
  int foo;
 #if !DEBUG
   int witness;
 #endif
};
 #if DEBUG
   int witness;
 #endif
}; 
```

So that the field has 0 size if in debug mode but is still in the structure. This is also a nice case for macros (ie you could have a macro that if in debug mode gets the witness, otherwise does an assertion).

interwq:
I agree code readability is a concern with this type of thing. It's a bit hard to say or measure the impact from smaller mutex. On one hand 44 bytes isn't much at all, but on the other hand it potentially saves a cacheline in L1. The reason that I wanted to change this was because I was working on #574 which adds more prof related stuff which blows up the mutex even more, and made me concerned about its size.

Regarding whether we should do more of this, I'd vote for doing this only for the top prio options where we have full test coverage on them (e.g. enable-debug and enable-prof).

interwq:
@jasone: I can also gives the union trick that Ben suggested a try. That way we should be able to limit the ifdefs to only in the struct definition.

jasone:
@interwq, sounds good!

interwq:
Fixing test failures. Will update soon.

interwq:
Should be ready for review. The appveyor failure is infra issue.

include/jemalloc/internal/mutex_types.h
```diff
@@ -7,25 +7,25 @@ typedef struct malloc_mutex_s malloc_mutex_t;
 #  define MALLOC_MUTEX_INITIALIZER
 #elif (defined(JEMALLOC_OS_UNFAIR_LOCK))
 #  define MALLOC_MUTEX_INITIALIZER					\
-     {OS_UNFAIR_LOCK_INIT, WITNESS_INITIALIZER("mutex", WITNESS_RANK_OMIT)}
+     {{{OS_UNFAIR_LOCK_INIT}}, WITNESS_INITIALIZER("mutex", WITNESS_RANK_OMIT)}
```
jasone: I did some reading (of a C11 draft) and found that initialization happens in declaration order unless designated initializers are used.  So, if we're going to use this union trick, I think we need to put the witness structure first so that its initialization is overwritten by the subsequent fields.

interwq: Not sure I'm following. If (!DEBUG), WITNESS_INITIALIZER is empty (has a ifdef around it) so that we only initialize once for the first member (lock); if (DEBUG), then the braces and declaration order should ensure we initialize properly. When testing locally, gcc is only happy with the two braces around {{..._INIT}}, which I believe avoids ambiguity.

The static initializer is already used here: https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L236

jasone: The ```WITNESS_INITIALIZER``` detail is what I was missing.

interwq:
Tests pass except one AppVeyor infra failure. Merging.

PR #642, opened 2017-02-24
Fix chunk_alloc_dss() regression.

jasone:
Fix chunk_alloc_dss() to account for bytes that are not a multiple of
the chunk size.  This regression was introduced by
e2bcf037d445a84a71c7997670819ebd0a893b4a (Make dss operations
lockless.), which was first released in 4.3.0.

### DIFF
--- a/src/chunk_dss.c
+++ b/src/chunk_dss.c
@@ -115,8 +115,9 @@ chunk_alloc_dss(tsdn_t *tsdn, arena_t *arena, void *new_addr, size_t size,
 		 * malloc.
 		 */
 		while (true) {
-			void *ret, *cpad, *max_cur, *dss_next, *dss_prev;
-			size_t gap_size, cpad_size;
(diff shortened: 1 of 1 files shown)


jasone:
Well, that's discouraging, because it's unrelated to this diff, and I can't reproduce it locally.

jasone:
It's likely that the test failure is due to incorrect OOM handling somewhere.  I'm going to have to test in a RAM-constrained environment.  Meanwhile, I don't think that should block this diff.

davidtgoldblatt:
Note that you can kick off a rebuild by clicking through on the "Details" link and clicking "Rebuild". I started one for the failing ubuntu build.

jasone:
I have an intermittent repro (and a core file) on a RAM-constrained Ubuntu VM.  It may be hard to diagnose without a more reliable failure, but this is a start at least.

jasone:
I investigated enough to verify that the test failure is not specific to this change, and created #647 to track fixing the issue.

PR #643, opened 2017-02-24
JSON testing and fixes.

jasone:


interwq:
Github doesn't load some files by default, e.g. when viewing https://github.com/jemalloc/jemalloc/pull/643/files, the entire test/unit/stats_print.c is not loaded by default. Any insights how to make it always load everything? This has been bothering me for a while :(

jasone:
This is all I've found on the topic.  I don't think we're anywhere near the limits described (unless that "400 lines" clause applies), but it wouldn't surprise me if the imposed limits are chosen (and changed) ad hoc as github iterates on its service.

https://help.github.com/articles/limits-for-viewing-content-and-diffs-in-a-repository/

jasone:
The [test failure](https://travis-ci.org/jemalloc/jemalloc/jobs/205087884) very likely has the same cause as the issue being tracked in #647 (quantization bugs tickled by aligned allocation).  I'm going to commit this diff under that assumption.

PR #644, opened 2017-02-25
Check nstime_monotonic (compile time const) first in nstime_update.

interwq:
Avoids calling nstime_compare in nstime_update.

### DIFF
--- a/src/nstime.c
+++ b/src/nstime.c
@@ -141,8 +141,12 @@ nstime_monotonic_t *nstime_monotonic = JEMALLOC_N(n_nstime_monotonic);
 #endif
 bool
 nstime_update(nstime_t *time) {
-	nstime_t old_time;
+	if (nstime_monotonic()) {
+		nstime_get(time);
+		return false;
+	}
 
+	nstime_t old_time;
 	nstime_copy(&old_time, time);
 	nstime_get(time);
 


interwq:
Hold on reviewing. Got to fix the tests.

src/nstime.c
```diff
@@ -141,8 +141,12 @@ nstime_monotonic_t *nstime_monotonic = JEMALLOC_N(n_nstime_monotonic);
...
 nstime_update(nstime_t *time) {
-	nstime_t old_time;
+	if (nstime_monotonic()) {
+		nstime_get(time);
+		return false;
+	}
 
+	nstime_t old_time;
 	nstime_copy(&old_time, time);
 	nstime_get(time);
```
interwq: @jasone : it looks this function is intended to ensure the *time is updated forward, regardless if clock is monotonic or not. In that case, do you think we should add an assert(!nstime_monotonic()) or fatal in the branch below, and just forget about the change here?

jasone: The input time can be set by the caller, so in principle this function needs to check for time going backward even if the clock is monotonic, unless we require the caller to never pass in a computed (possibly future) time.

Is there an issue you ran into while working on lock profiling that caused you to experiment with this change?

interwq: Oh I see. Didn't think much about the computed case. There wasn't any issue but I was thinking about calling this during lock spinning, and got worried about the cost (mainly about clock_gettime). I noticed we have the clock_monotonic const so I thought we should either trust the config and avoid a couple instructions here, or have a warning / fatal if we do detect skewed clock. 

Currently the return value can be used to detect such case. Though this may silent the failure as well. I don't think it's a real problem. So I'll just drop this commit if that sounds good to you.

PR #652, opened 2017-02-27
Handle race in stats_arena_bins_print

### LINKED ISSUE #616: Backport bbff6ca6740c27737378f6c2dec3a13053a5a150 (Handle race in stats_arena_bins_print)
Backport bbff6ca6740c27737378f6c2dec3a13053a5a150 (Handle race in stats_arena_bins_print).

jasone:
When multiple threads calling stats_print, race could happen as we read the
counters in separate mallctl calls; and the removed assertion could fail when
other operations happened in between the mallctl calls. For simplicity, output
"race" in the utilization field in this case.

This resolves #616.


jasone:
macOS tested locally; pushing.

PR #656, opened 2017-02-28
Dodge 32-bit-clang-specific backtracing failure.

jasone:
This disables run_tests.sh configurations that use the combination of
32-bit clang and heap profiling.


davidtgoldblatt:
Filed #658 to remind us to investigate.

PR #657, opened 2017-02-28
Fix/enhance THP integration.

### LINKED ISSUE #541: Fix/enhance THP integration
e98a620c59ac20b13e2de796164cc67f050ed2bf (Mark partially purged arena chunks as non-hugepage.) attempts to explicitly interact with Linux's transparent huge page (THP) functionality, but it has two shortcomings.  First, it makes the mistake of assuming new chunks are created as if ```madvise(... MADV_HUGEPAGE)``` had been applied, but that is not the case, so the chunk creation code needs to add an explicit call.  More generally, THP requests can cause serious scalability issues depending on the kernel version and configuration.  Prior to Linux 4.6 it wasn't even possible to tune the kernel to satisfy THP requests asynchronously.  We need to provide a way to opt out of explicit THP requests, so that applications can work around kernel issues as necessary.  This can just be an ```opt.thp``` option that defaults to true on relevant systems; it's hard to imagine use cases for finer-grained control.

jasone:
Detect whether chunks start off as THP-capable by default (according to
the state of /sys/kernel/mm/transparent_hugepage/enabled), and use this
as the basis for whether to call pages_nohuge() once per chunk during
first purge of any of the chunk's page runs.

Add the --disable-thp configure option, as well as the the opt.thp
mallctl.

This resolves #541.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -157,6 +157,13 @@ any of the following arguments (not a definitive list) to 'configure':
     released in bulk, thus reducing the total number of mutex operations.  See
     the "opt.tcache" option for usage details.
 
+--disable-thp
+    Disable transparent huge page (THP) integration.  On systems with THP
+    support, THPs are explicitly disabled as a side effect of unused dirty page
+    purging for chunks that back small and/or large allocations, because such
+    chunks typically comprise active, unused dirty, and untouched clean
+    pages.
+
 --disable-munmap
     Disable virtual memory deallocation via munmap(2); instead keep track of
     the virtual memory for later use.  munmap() is disabled by default (i.e.
--- a/src/arena.c
+++ b/src/arena.c
@@ -4,6 +4,8 @@
 /******************************************************************************/
 /* Data. */
 
+bool		opt_thp = true;
+static bool	thp_initially_huge;
 purge_mode_t	opt_purge = PURGE_DEFAULT;
 const char	*purge_mode_names[] = {
 	"ratio",
@@ -680,7 +682,9 @@ arena_chunk_init_hard(tsdn_t *tsdn, arena_t *arena)
 	if (chunk == NULL)
 		return (NULL);
 
-	chunk->hugepage = true;
+	if (config_thp && opt_thp) {
+		chunk->hugepage = thp_initially_huge;
+	}
 
 	/*
 	 * Initialize the map to contain one maximal free untouched run.  Mark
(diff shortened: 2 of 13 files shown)


INSTALL
```diff
@@ -157,6 +157,12 @@ any of the following arguments (not a definitive list) to 'configure':
     released in bulk, thus reducing the total number of mutex operations.  See
     the "opt.tcache" option for usage details.
 
+--disable-thp
+    Disable transparent huge page (THP) integration.  On systems with THP
+    support, THPs are explicitly avoided for chunks that back small and/or large
+    allocations, because such chunks typically comprise a active, inactive
```
interwq: nit: The "a" before active is extra? Also does this mean we explicitly disable thp unless its huge allocation?

src/arena.c
```diff
@@ -765,13 +772,16 @@ arena_chunk_discard(tsdn_t *tsdn, arena_t *arena, arena_chunk_t *chunk)
...
 		/*
-		 * Convert chunk back to the default state, so that all
-		 * subsequent chunk allocations start out with chunks that can
-		 * be backed by transparent huge pages.
+		 * Convert chunk back to initial THP state, so that all
+		 * subsequent chunk allocations start out in a consistent state.
 		 */
-		pages_huge(chunk, chunksize);
+		if (thp_initially_huge) {
+			pages_huge(chunk, chunksize);
```
interwq: What happens if chunksize is smaller than 2MB?

src/arena.c
```diff
@@ -3772,11 +3782,78 @@ bin_info_init(void)
...
 
+static void
+init_thp_initially_huge(void) {
+	int fd;
+	char buf[sizeof("[always] madvise never\n")];
+	ssize_t nread;
+	static const char *enabled_states[] = {
+		"[always] madvise never\n",
+		"always [madvise] never\n",
+		"always madvise [never]\n"
```
interwq: If this is set to never, we should probably disable it on our side completely? e.g. not doing related stuff like madvise at all.

INSTALL
jasone: Hmm, this needs updating, because we only disable THPs during purging.  Same goes for the ```opt.thp``` docs.

src/arena.c
jasone: Any neighboring chunks will end up non-hugepage unless the entire 2 MiB range has the hugepage attribute.

src/arena.c
jasone: In practice that's what the code currently does, because we start off with ```thp_initially_huge``` set to false, and therefore never call ```pages_nohuge()```.

src/arena.c
interwq: Ah I see. Makes sense.

jasone:
All tests passed for the previous version, and only docs changes have been made since.  Pushing.

PR #668, opened 2017-03-01
fix typo sytem -> system

charsyam:
just simple typo patch in configure.ac 

Thanks. I always use jemalloc well :)


jasone:
Thanks!

PR #671, opened 2017-03-01
Small style fix in ctl.c

interwq:


interwq:
Merging as only style change is involved.

PR #672, opened 2017-03-01
Fix decay_time:0

jasone:
I verified that the new test_decay_now test fails without the fix in place.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -259,7 +259,9 @@ arena_extent_cache_dalloc(tsdn_t *tsdn, arena_t *arena,
 	witness_assert_depth_to_rank(tsdn, WITNESS_RANK_CORE, 0);
 
 	extent_dalloc_cache(tsdn, arena, r_extent_hooks, extent);
-	arena_purge(tsdn, arena, false);
+	if (arena_decay_time_get(arena) == 0) {
+		arena_purge(tsdn, arena, true);
+	}
 }
 
 JEMALLOC_INLINE_C void *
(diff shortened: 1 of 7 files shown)


src/arena.c
```diff
@@ -491,6 +491,20 @@ arena_extent_ralloc_large_expand(tsdn_t *tsdn, arena_t *arena, extent_t *extent,
 	arena_nactive_add(arena, udiff >> LG_PAGE);
 }
 
+static ssize_t
+arena_decay_time_read(arena_t *arena) {
+	arena_decay_time_t dt;
+	dt.u = atomic_read_zu(&arena->decay.time.u);
+	return dt.s;
```
IndolentPangolin: Using a union like this is undefined behavior. It may compile correctly today with all common compilers, but there's no reason to believe it will continue to compile correctly.

Is there a reason the atomics can't be used on an ssize_t? Can casting be used instead to produce correct results?

jasone: Why do you say that using a union like this is undefined behavior?  In the context of type aliasing, unions are pretty much the only well defined mechanism for accessing the same bits as different types.  If anything I'd expect this union approach to be safer than casting (for which unsigned-->signed is technically implementation-defined on overflow).  The down side is that unions prevent a bunch of compiler optimizations which are actually okay, assuming the implementation-defined unsigned-->signed conversion happens to do what we want (i.e. nothing).

IndolentPangolin: This must be something that's diverged in C and C++, and been clarified in C. I have had experience in the past with C compilers optimizing away code like this.  But looking at http://en.cppreference.com/w/c/language/union now says:

"If the member used to access the contents of a union is not the same as the member last used to store a value, the object representation of the value that was stored is reinterpreted as an object representation of the new type (this is known as type punning)."

So this is fine. Thanks!

davidtgoldblatt: The C semantics are only deceptively nice, since they're too vague to be useful in interesting cases. There's a note that says `u.a = x; y = u.b;` does the obvious thing, but it's also fairly clear that union magic doesn't apply to uses that merely have possibly aliasing pointers to union members, and there's a bunch of area in between those extremes.

```
void func_a(type_a *);
void func_b(type_b *);
void func_a_b(type_a *, type_b *);
typedef union {
  type_a a;
  type_b b;
} U;
void probably_safe() {
  U u;
  func_a(&u.a);
  func_b(&u.b);
}
void risky() {
  U u;
  type_a *a_ptr = &u.a;
  type_b *b_ptr = &u.b;
  func_a(a_ptr);
  func_b(b_ptr);
}
```

So really the question has to be about acceptable levels of compiler risk instead of a binary "UB: yes/no" unless we're willing to color entirely within the lines (which I suspect will be impossible in a malloc).

PR #675, opened 2017-03-03
Perform delayed coalescing prior to purging.

### LINKED ISSUE #655: Merge before purging cached extents
2dfc5b5aac983f8f192c63c604a59fed8b39e937 (Disable coalescing of cached extents.) has the desired effect of adding hysteresis to reduce extent splitting/merging for large allocation, but it has the unfortunate side effect of increasing the number of ```madvise(2)``` system calls.  Modify the purging code to add the following steps for each extent purge:

1. Attempt to merge the LRU cached extent with its neighbors.
2. Do one of:
  a. If merging succeeds, logically reinsert the resulting extent into the cached extent LRU at the position of the MRU neighbor that was merged in (1),
  b. otherwise purge the LRU extent.

This will not result in precisely the same cached extent recycling behavior as if we were to aggressively merge cached extents, but it should continue to provide a hysteresis mechanism without needlessly increasing the number of system calls.

jasone:
Rather than purging uncoalesced extents, perform just enough incremental
coalescing to purge only fully coalesced extents.  In the absence of
cached extent reuse, the immediate versus delayed incremental purging
algorithms result in the same purge order.

This resolves #655.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -715,9 +715,9 @@ arena_stash_dirty(tsdn_t *tsdn, arena_t *arena, extent_hooks_t **r_extent_hooks,
 
 	/* Stash extents according to ndirty_limit. */
 	size_t nstashed = 0;
-	for (extent_t *extent = extents_evict(tsdn, &arena->extents_cached,
-	    ndirty_limit); extent != NULL; extent = extents_evict(tsdn,
-	    &arena->extents_cached, ndirty_limit)) {
+	extent_t *extent;
+	while ((extent = extents_evict(tsdn, arena, r_extent_hooks,
+	    &arena->extents_cached, ndirty_limit)) != NULL) {
 		extent_list_append(purge_extents, extent);
 		nstashed += extent_size_get(extent) >> LG_PAGE;
 	}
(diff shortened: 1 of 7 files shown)


src/arena.c
```diff
@@ -737,9 +737,10 @@ arena_stash_dirty(tsdn_t *tsdn, arena_t *arena, extent_hooks_t **r_extent_hooks,
 
 	/* Stash extents according to ndirty_limit. */
 	size_t nstashed = 0;
-	for (extent_t *extent = extents_evict(tsdn, &arena->extents_cached,
-	    ndirty_limit); extent != NULL; extent = extents_evict(tsdn,
-	    &arena->extents_cached, ndirty_limit)) {
+	for (extent_t *extent = extents_evict(tsdn, arena, r_extent_hooks,
```
interwq: nit: use a  ```while ((extent = extents_evict()) != NULL) { ... } ``` loop so that we don't write extents_evict twice?

src/arena.c
```diff
@@ -1707,11 +1709,11 @@ arena_new(tsdn_t *tsdn, unsigned ind, extent_hooks_t *extent_hooks) {
 	}
 
 	if (extents_init(tsdn, &arena->extents_cached, extent_state_dirty,
-	    false)) {
+	    true)) {
 		goto label_error;
 	}
 	if (extents_init(tsdn, &arena->extents_retained,
```
interwq: nit: add a bit comments about why retained extents need no delay_coalesce

jasone:
run_tests.sh passes locally on Linux and macOS; pushing.

PR #676, opened 2017-03-03
Header refactoring -- util and assert

davidtgoldblatt:
This is part of #533, the header file refactoring. It splits out two subcomponents from util.h (the stdio.h functions and bit manipulation routines), letting us make all of them as well as assert acyclic and hermetic.

In the second commit, we add in the not_reached call that we had to omit in the C11 atomics backport.


include/jemalloc/internal/bit_util.h
```diff
@@ -1,45 +1,33 @@
...
-void	set_errno(int errnum);
-int	get_errno(void);
-#endif
+#ifndef JEMALLOC_INTERNAL_BIT_UTIL_H
+#define JEMALLOC_INTERNAL_BIT_UTIL_H
+
+#include "jemalloc/internal/assert.h"
 
-#if (defined(JEMALLOC_ENABLE_INLINE) || defined(JEMALLOC_UTIL_C_))
+#define BIT_UTIL_INLINE static inline
```
jasone: You're planning to remove all of these header-specific inline definitions a late stage refactor, right?

include/jemalloc/internal/util.h
```diff
@@ -0,0 +1,72 @@
...
+#  define likely(x)   !!(x)
+#  define unlikely(x) !!(x)
+#endif
+
+#if !defined(JEMALLOC_INTERNAL_UNREACHABLE)
+#  error JEMALLOC_INTERNAL_UNREACHABLE should have been defined by configure
+#endif
+
+#define unreachable() JEMALLOC_INTERNAL_UNREACHABLE()
+
```
jasone: Extraneous newline.

src/malloc_io.c
```diff
@@ -24,22 +40,25 @@
...
+	}								\
+} while (0)
 
 /******************************************************************************/
 /* Function prototypes for non-inline static functions. */
 
-static void	wrtmessage(void *cbopaque, const char *s);
-#define U2S_BUFSIZE	((1U << (LG_SIZEOF_INTMAX_T + 3)) + 1)
-static char	*u2s(uintmax_t x, unsigned base, bool uppercase, char *s,
+static void wrtmessage(void *cbopaque, const char *s);
```
jasone: :+1:

include/jemalloc/internal/util.h
davidtgoldblatt: Done.

include/jemalloc/internal/bit_util.h
davidtgoldblatt: I figured I'd feel you guys out on it. Having these means that if we want, we can easily force all the definitions into one .o, which could be useful if we want to e.g. put a breakpoint on one.

PR #679, opened 2017-03-04
Any

jasone:
This would feel like a hack to me absent the delayed coalescing in #675, but in that context it feels natural and complementary.  The two in combination (plus #678) speed up the large allocation fast path substantially, to the point that it's probably faster than 4.x's.

@djwatson, you may find the (ab)use of the pairing heap's aux list entertaining.


jasone:
run_tests.sh passes locally on Linux and macOS; pushing.

include/jemalloc/internal/ph.h
```diff
@@ -266,15 +279,52 @@ a_prefix##remove_first(a_ph_type *ph) {					\
 									\
 	return ret;							\
 }									\
+a_attr a_type *								\
+a_prefix##remove_any(a_ph_type *ph) {					\
+	/*								\
+	 * Remove the most recently inserted aux list element, or the	\
```
djwatson: Nice hack.

Issue #680, opened 2017-03-06
Getting correct tcache associated arena


interwq:
Each tcache is associated to an arena. However current the associated arena is determined by tsd instead of by tcache itself. This means for manual tcache, if a thread is bind to a different arena (e.g. using thread.arena ctl), we don't have a way to find the associated arena for the manual tcache. We should be able to fix this by storing an pointer to the arena in tcache.

interwq:
Closed via https://github.com/jemalloc/jemalloc/pull/682

PR #682, opened 2017-03-06
Store associated arena in tcache.

interwq:
This fixes tcache_flush for manual tcaches, which wasn't able to find
the correct arena it associated with. Also changed the decay test to
cover this case (by using manually created arenas).


interwq:
This should solve #680, and make percpu arena work with unit/test/decay.

interwq:
Just rebased because of a merge conflict in test/unit/decay.c

Nothing else has changed. But will wait a bit to get some test results.

interwq:
Previous tests passed. Merging.

PR #687, opened 2017-03-08
Convert extents_t's npages field to use C11-style atomics

davidtgoldblatt:
In the process, we can do some strength reduction, changing the fetch-adds and
fetch-subs to be simple loads followed by stores, since the modifications all
occur while holding the mutex.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -186,7 +186,7 @@ extents_init(tsdn_t *tsdn, extents_t *extents, extent_state_t state,
 		extent_heap_new(&extents->heaps[i]);
 	}
 	extent_list_init(&extents->lru);
-	extents->npages = 0;
+	atomic_store_zu(&extents->npages, 0, ATOMIC_RELAXED);
 	extents->state = state;
 	extents->delay_coalesce = delay_coalesce;
 	return false;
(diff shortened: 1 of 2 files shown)


src/extent.c
```diff
@@ -216,7 +216,10 @@ extents_insert_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent,
 		extent_list_append(&extents->lru, extent);
 	}
 	size_t npages = size >> LG_PAGE;
-	atomic_add_zu(&extents->npages, npages);
+	size_t cur_extents_npages =
+	    atomic_load_zu(&extents->npages, ATOMIC_RELAXED);
+	atomic_store_zu(&extents->npages, cur_extents_npages + npages,
```
jasone: How is splitting an atomic add into an atomic load followed by an atomic store not racy?  Maybe if the load were ```ATOMIC_ACQUIRE``` and the store were ```ATOMIC_RELEASE``` this would be okay?

jasone: How about ```atomic_fetch_add_zu(&extents->npages, npages, ATOMIC_RELAXED)```?  That's still a relaxation from ```atomic_add_zu()```'s ```ATOMIC_SEQ_CST```, yet still safe.

jasone: As discussed on [gitter](https://gitter.im/jemalloc/jemalloc?at=58c0944b7ceae5376a8e6c8d), comments just above the atomic update code blocks will help to clarify why this is okay (mutex protection).

PR #692, opened 2017-03-09
Switch atomic uint64_ts in arena_stats_t to C11 atomics

davidtgoldblatt:
I expect this to be the trickiest conversion we will see, since we want atomics
on 64-bit platforms, but are also always able to piggyback on some sort of
external synchronization not under our control on non-64 bit platforms.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -78,9 +78,10 @@ arena_stats_unlock(tsdn_t *tsdn, arena_stats_t *arena_stats) {
 }
 
 static uint64_t
-arena_stats_read_u64(tsdn_t *tsdn, arena_stats_t *arena_stats, uint64_t *p) {
+arena_stats_read_u64(tsdn_t *tsdn, arena_stats_t *arena_stats,
+    arena_stats_u64_t *p) {
 #ifdef JEMALLOC_ATOMIC_U64
-	return atomic_read_u64(p);
+	return atomic_load_u64(p, ATOMIC_RELAXED);
 #else
 	malloc_mutex_assert_owner(tsdn, &arena_stats->mtx);
 	return *p;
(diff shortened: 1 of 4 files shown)


davidtgoldblatt:
Woops; forgot about an ifdef for 32-bit.

src/arena.c
```diff
@@ -78,39 +78,51 @@ arena_stats_unlock(tsdn_t *tsdn, arena_stats_t *arena_stats) {
...
 	assert(*p + x >= *p);
 #endif
 }
 
+/* Non-atomically sets *dst += src.  *dst needs external synchronization. */
+static void
+arena_stats_accum_u64(arena_stats_u64_t *dst, uint64_t src) {
+#ifdef JEMALLOC_ATOMIC_U64
+	uint64_t cur_dst = atomic_load_u64(dst, ATOMIC_RELAXED);
+	atomic_store_u64(dst, src + cur_dst, ATOMIC_RELAXED);
```
jasone: Why isn't this just ```atomic_fetch_add_u64()```?

src/ctl.c
```diff
@@ -431,6 +431,33 @@ static const ctl_named_node_t super_root_node[] = {
...
+/*
+ * Sets *dst + *src non-atomically.  This is safe, since everything is
+ * synchronized by the ctl mutex.
+ */
+static void
+accum_arena_stats_u64(arena_stats_u64_t *dst, arena_stats_u64_t *src) {
+#ifdef JEMALLOC_ATOMIC_U64
+	uint64_t cur_dst = atomic_load_u64(dst, ATOMIC_RELAXED);
+	uint64_t cur_src = atomic_load_u64(src, ATOMIC_RELAXED);
+	atomic_store_u64(dst, cur_dst + cur_src, ATOMIC_RELAXED);
```
jasone: ```atomic_fetch_add_u64()```?

src/arena.c
```diff
@@ -78,39 +78,51 @@ arena_stats_unlock(tsdn_t *tsdn, arena_stats_t *arena_stats) {
...
+	UNUSED uint64_t r = atomic_fetch_sub_u64(p, x, ATOMIC_RELAXED);
+	assert(r - x <= r);
 #else
 	malloc_mutex_assert_owner(tsdn, &arena_stats->mtx);
 	*p -= x;
 	assert(*p + x >= *p);
 #endif
 }
 
+/* Non-atomically sets *dst += src.  *dst needs external synchronization. */
```
jasone: It took me a while to realize that this function (and ```accum_arena_stats_u64()```) exists solely to deal with atomic types.  A comment to that effect would help prevent future confusion.

davidtgoldblatt: Will do.

src/arena.c
davidtgoldblatt: This is non-atomic; it lets us do a load+add+store sequence instead of an atomic fetch-add.

src/ctl.c
davidtgoldblatt: Same deal.

davidtgoldblatt:
Added the second commit converting the surrounding fields. LMK if things still look OK.

PR #693, opened 2017-03-09
Convert arena_t's purging field to bool.

jasone:
Now that boolean atomics are supported, make arena_t's purging field
bool rather than unsigned.


davidtgoldblatt:
Does this need to be atomic at all? The only places we touch arena->purging are this function (where we protect it with arena->decay.mtx), arena_reset (in which the caller guarantees no concurrency, according to the comment), and arena_new.

jasone:
I think you're right that this does not need to be atomic at all.

jasone:
Updated to be non-atomic.

PR #695, opened 2017-03-10
Zero

jasone:
The ```pages_*()``` assertions would catch an error I made in an early version of the calls to ```pages_purge_forced()```.


jasone:
Somehow this broke the ckh unit test on FreeBSD.

jasone:
```MADV_DONTNEED``` (and therefore ```pages_purge_forced()```) does not zero on FreeBSD.

PR #696, opened 2017-03-10
Convert rtree code to use C11 atomics

davidtgoldblatt:
In the process, I changed the implementation of rtree_elm_acquire so that it
won't even try to CAS if its initial read (getting the extent + lock bit)
indicates that the CAS is doomed to fail.  This can significantly improve
performance under contention.

We also remove a few unnecessary fences down the rtree_elm_write[_acquired]
paths.

### DIFF
--- a/include/jemalloc/internal/rtree_inlines.h
+++ b/include/jemalloc/internal/rtree_inlines.h
@@ -55,14 +55,16 @@ rtree_elm_read(rtree_elm_t *elm, bool dependent) {
 		 * synchronization, because the rtree update became visible in
 		 * memory before the pointer came into existence.
 		 */
-		extent = elm->extent;
+		extent = (extent_t *)atomic_load_p(&elm->child_or_extent,
+		    ATOMIC_RELAXED);
 	} else {
 		/*
 		 * An arbitrary read, e.g. on behalf of ivsalloc(), may not be
 		 * dependent on a previous rtree write, which means a stale read
 		 * could result if synchronization were omitted here.
 		 */
-		extent = (extent_t *)atomic_read_p(&elm->pun);
+		extent = (extent_t *)atomic_load_p(&elm->child_or_extent,
+		    ATOMIC_ACQUIRE);
 	}
 
 	/* Mask the lock bit. */
@@ -73,7 +75,7 @@ rtree_elm_read(rtree_elm_t *elm, bool dependent) {
 
 JEMALLOC_INLINE void
 rtree_elm_write(rtree_elm_t *elm, const extent_t *extent) {
-	atomic_write_p(&elm->pun, extent);
+	atomic_store_p(&elm->child_or_extent, (void *)extent, ATOMIC_RELEASE);
 }
 
 JEMALLOC_ALWAYS_INLINE rtree_elm_t *
@@ -161,11 +163,18 @@ rtree_elm_acquire(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
 
 	spin_t spinner = SPIN_INITIALIZER;
 	while (true) {
-		extent_t *extent = rtree_elm_read(elm, false);
 		/* The least significant bit serves as a lock. */
-		void *s = (void *)((uintptr_t)extent | (uintptr_t)0x1);
-		if (!atomic_cas_p(&elm->pun, (void *)extent, s)) {
-			break;
+		void *extent_and_lock = atomic_load_p(&elm->child_or_extent,
+		    ATOMIC_RELAXED);
+		if (likely(((uintptr_t)extent_and_lock & (uintptr_t)0x1) == 0))
+		{
+			void *locked = (void *)((uintptr_t)extent_and_lock
+			    | (uintptr_t)0x1);
+			if (likely(atomic_compare_exchange_strong_p(
+			    &elm->child_or_extent, &extent_and_lock, locked,
+			    ATOMIC_ACQUIRE, ATOMIC_RELAXED))) {
+				break;
+			}
 		}
 		spin_adaptive(&spinner);
 	}
@@ -180,9 +189,9 @@ rtree_elm_acquire(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
 JEMALLOC_INLINE extent_t *
 rtree_elm_read_acquired(tsdn_t *tsdn, const rtree_t *rtree, rtree_elm_t *elm) {
 	extent_t *extent;
-
-	assert(((uintptr_t)elm->pun & (uintptr_t)0x1) == (uintptr_t)0x1);
-	extent = (extent_t *)((uintptr_t)elm->pun & ~((uintptr_t)0x1));
+	void *ptr = atomic_load_p(&elm->child_or_extent, ATOMIC_RELAXED);
+	assert(((uintptr_t)ptr & (uintptr_t)0x1) == (uintptr_t)0x1);
+	extent = (extent_t *)((uintptr_t)ptr & ~((uintptr_t)0x1));
 	assert(((uintptr_t)extent & (uintptr_t)0x1) == (uintptr_t)0x0);
 
 	if (config_debug) {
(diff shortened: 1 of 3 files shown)


include/jemalloc/internal/rtree_inlines.h
```diff
@@ -161,11 +163,17 @@ rtree_elm_acquire(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
...
 		/* The least significant bit serves as a lock. */
-		void *s = (void *)((uintptr_t)extent | (uintptr_t)0x1);
-		if (!atomic_cas_p(&elm->pun, (void *)extent, s)) {
-			break;
+		void *extent_and_lock = atomic_load_p(&elm->child_or_extent,
+		    ATOMIC_RELAXED);
+		if (likely(((uintptr_t)extent_and_lock & 0x1) == 0)) {
+			void *locked = (void *)((uintptr_t)extent_and_lock
+			    | (uintptr_t)0x1);
+			if (likely(atomic_compare_exchange_strong_p(
```
jasone: Minor aside: I wish the new CAS functions took the "expected" by value rather than by reference.  Is there a design reason (some form of API consistency issue perhaps) that motivates this?

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -180,9 +188,9 @@ rtree_elm_acquire(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
 JEMALLOC_INLINE extent_t *
 rtree_elm_read_acquired(tsdn_t *tsdn, const rtree_t *rtree, rtree_elm_t *elm) {
 	extent_t *extent;
-
-	assert(((uintptr_t)elm->pun & (uintptr_t)0x1) == (uintptr_t)0x1);
-	extent = (extent_t *)((uintptr_t)elm->pun & ~((uintptr_t)0x1));
+	void *ptr = atomic_load_p(&elm->child_or_extent, ATOMIC_RELAXED);
```
jasone: Three questions:
1) What is the performance impact of relaxed atomic loads versus plain memory loads?
2) Does a relaxed atomic read do anything besides guard against torn reads?  I think it causes actual reads (akin to ```volatile```), but I want to be sure my mental model isn't way off.
3) Why do we need to use an atomic load here when the function is specifically used for reading when we've already acquired the elm?  Related to (2), I think it's necessary because there are no explicit dependencies that would otherwise guarantee consistency between atomic and non-atomic accesses, but I'm not sure.

src/rtree.c
```diff
@@ -66,27 +67,37 @@ rtree_delete_subtree(tsdn_t *tsdn, rtree_t *rtree, rtree_elm_t *node,
...
 rtree_node_init(tsdn_t *tsdn, rtree_t *rtree, unsigned level,
-    rtree_elm_t **elmp) {
+    atomic_p_t *elmp) {
 	rtree_elm_t *node;
 
 	malloc_mutex_lock(tsdn, &rtree->init_lock);
-	node = atomic_read_p((void**)elmp);
+	/*
+	 * If *elmp is non-null, then it was initialized with the init lock
+	 * held, so we can get by with `relaxed` here.
```
jasone: Nit: ```'relaxed'```, please; there are no other back ticks in the C code.

src/rtree.c
```diff
@@ -66,27 +67,37 @@ rtree_delete_subtree(tsdn_t *tsdn, rtree_t *rtree, rtree_elm_t *node,
...
 		if (node == NULL) {
 			malloc_mutex_unlock(tsdn, &rtree->init_lock);
 			return NULL;
 		}
-		atomic_write_p((void **)elmp, node);
+		/*
+		 * Even though we hold the lock, a later reader might not; we
+		 * need release semantics.
+		 */
+		atomic_store_p(elmp, node, ATOMIC_RELEASE);
```
jasone: Since we're releasing the mutex right after this, do I understand that the release semantics here would be unnecessary on total-store-order machines, but here you're guaranteeing correct operation on partial-store-order machines?

include/jemalloc/internal/rtree_inlines.h
davidtgoldblatt: I don't know the rationale for sure (ours is just copied from the C11 versions), but I have a theory:

The instructions a CAS gets compiled to (compxchg / lwarx + stwcx / ldxr + stxr) all spit out the previous value in the location in case of failure, so it would be nice to avoid an API that doesn't expose that. On x86, all this really does is save a load (and even then, I think it would be straightforward to teach the optimizer to figure out that a CAS failure followed by reload doesn't need the redundant load). Some places it could cost an extra fence though.

It lends itself to a sort of convenient idiom in a lot of places:

void stack_push(atomic_stack_node_p *head, int data) {
  stack_node *new_head = alloc_new_head(data);
  stack_node *cur_head = load(head, acquire);
  do {
    newHead->next = cur_head;
  } while (!compare_exchange_weak(head, &cur_head, new_head);
}

(this only reason I don't do that here is because of the adaptive spinning; the values we get back might be really stale by the time we wake up).

It is sort of an annoying API for stuff like this though. I think a convenience atomic_cas_foo that takes expected by value instead of pointer is a good idea.

include/jemalloc/internal/rtree_inlines.h
davidtgoldblatt: 1. They both compile down to a plain load instruction. Using an atomic does slightly restrict the optimizer, because it can't assume that redundant loads are guaranteed to return the same value any more. But in the cases where it's doing that, it would usually be introducing a bug:

```
void do_something(some_type *t) {
  void *alloc = NULL;
  bool b = t->should_alloc; /* Suppose this reads b = true */
  if (b) {
    alloc = do_alloc();
  }
  /*
   * Some code that doesn't change t->b. The optimizer decides not to put b in a register, but
   * does leave t in a register. To save itself from the useless stack-store, it doesn't materialize b
   * on the stack; it just reloads it from t.
   */
  /*
   * Suppose another thread sets b to false, and we pick up that modification. Because b never got
   * materialized on the stack, we'll see the updated value, and won't take the branch.
   */
  if (b) {
    /* This never runs; we leak. */
    do_free(alloc);
  }
}
```

2. Nope. The only thing it does is prevent that sort of compiler optimization. The only ordering guarantee we get is that there's a total ordering of the stores to a particular location, and our reads never go backwards in time in that ordering.

3. In this particular case, it's a type issue; atomic_u32_t and uint32_t are genuinely different types. In general it would be safe though; if the struct were:

```
struct elm_s {
  atomic_b_t lock;
  extent_t *value;
};
```

Then we can implement the lock just with elm->b, and do plain reads from elm->value safely with no races.

src/rtree.c
davidtgoldblatt: The subsequent mutex release only protects us if a later reader acquires the mutex before reading. But not all readers do (in particular, the non-dependent ones who are examining the rtree just for fun). We don't want it for a reader to come by, see the store of `node` to `*elmp`, but not see the initialization that happened in `rtree_node_alloc`.

People sometimes phrase this as: the `release` prevents previous stores from being reordered after the release store.

src/rtree.c
davidtgoldblatt: Will do, but I'll send it out on Monday.

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -161,11 +163,17 @@ rtree_elm_acquire(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
...
 	spin_t spinner = SPIN_INITIALIZER;
 	while (true) {
-		extent_t *extent = rtree_elm_read(elm, false);
 		/* The least significant bit serves as a lock. */
-		void *s = (void *)((uintptr_t)extent | (uintptr_t)0x1);
-		if (!atomic_cas_p(&elm->pun, (void *)extent, s)) {
-			break;
+		void *extent_and_lock = atomic_load_p(&elm->child_or_extent,
+		    ATOMIC_RELAXED);
+		if (likely(((uintptr_t)extent_and_lock & 0x1) == 0)) {
```
IndolentPangolin: For my own edification, why do we need to cast 0x1 to uintptr_t? Is it because Clang warns on using signed and unsigned values together? If so, wouldn't 0x1u be sufficient?

davidtgoldblatt:
run_tests.sh passes on CentOS and OS X -- pushing.

Issue #699, opened 2017-03-13
Refactor away rtree locking


jasone:
The single-bit rtree element locking makes applications vulnerable to priority inversion deadlock.  Fix this by associating a mutex with each ```extent_t```, and validating that the correct extent was locked (the ```extent_t``` could have been recycled between us reading the pointer from the rtree and acquiring the mutex).  There are some important details to making this work correctly:

- Validating that the correct extent was locked can either be achieved by validating ```e_addr``` (and ```e_size``` when locking an adjacent extent that is lower in memory), or by re-reading the rtree element to make sure it has not changed.
- Extent structures must either never be deallocated (true of the current implementation, which caches discarded extent structures), or there must be a mechanism for assuring that no threads are accessing the extent structure anymore prior to it being deallocated.
- In order to portably support forking, we need to be able to iterate over all mutexes, ideally in a reasonable amount of time.  This suggests that we use a fixed pool of mutexes (the size of which is a function of the number of CPUs), and create a stable mapping from (hashed) extent pointers to mutexes.  With such a mapping, we don't need to store mutex pointers within ```extent_t```.
- Rather than using ```extent->e_addr``` to determine extent locking order, we must use a stable mutex ordering.  Furthermore, it is possible for adjacent extents to use the same mutex, in which case care must be taken to only lock once.

davidtgoldblatt:
One other useful fact is that if we use a direct hash of extent_t * -> mutex, we don't need to worry about extent_t lifetime issues in the "read pointer, lock, re-read pointer" strategy.

jasone:
This should be done after #694, since both modify rtree.

davidtgoldblatt:
Looking into this, I think there might be a more serious data corruption bug. We pick the rtree element to lock based on the extent base address, which is unprotected at the point we read it, so that another thread might invalidate the extent<->rtree lock association in the meantime, so that both threads might think they own the locks for the extent simultaneously. (Sanity-checked this with @interwq, who thinks this might explain the intermittent rtree assertion failures he's been seeing).

I'll do the priority inversion diff now, just since it's simple and easy, but I wanted to jot this down to make sure we don't forget about it. I vaguely remember that the answer might be that we only care about this locking strategy for extents with live allocations, and the base address doesn't change for such extents, but I now can't convince myself that this is the case. Once @jasone is back, we should figure out if there's an implicit assumption about these code paths I'm missing (and document it).

PR #702, opened 2017-03-15
Get rid of rtree_elm_witness_tsd if !DEBUG.

interwq:


interwq:
Does this seem reasonable to you @jasone?

jasone:
In general I prefer to avoid ```#ifdef``` logic based on configuration options unless there's a compelling performance issue.  In this particular case there's an extra reason not to do this (yet): it's going to go away as part of #699 anyway.

interwq:
Ah I see. That makes sense. Closing this then.

Issue #705, opened 2017-03-18
Rebind i-arena when migrating between arenas?


interwq:
Currently we have an assertion to make sure internal allocation is not done with manual arenas:
https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/jemalloc_internal.h.in#L1066

However, if we manually migrate a thread to a manual arena, we will do arena_migrate which re-binds arena, including iarena:
thread_arena_ctl -- https://github.com/jemalloc/jemalloc/blob/dev/src/ctl.c#L1440
arena_migrate -- https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L450

We have been getting intermittent failure like this: https://master.jemalloc.net/#/builders/1/builds/106?

The iarena re-binding was introduced recently:
https://github.com/jemalloc/jemalloc/commit/01f47f11a67d1a2505cc1f21851c466651eba431#diff-d2be2656d37670b0c7f5f4320c0670ceR449

However I don't recall why it was added. Will test if it's necessary.

interwq:
Solved by #706

PR #706, opened 2017-03-18
Not re-binding iarena when migrate between arenas.

interwq:
Attempt to solve #705.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -447,7 +447,6 @@ arena_migrate(tsd_t *tsd, unsigned oldind, unsigned newind) {
 	arena_nthreads_dec(oldarena, false);
 	arena_nthreads_inc(newarena, false);
 	tsd_arena_set(tsd, newarena);
-	tsd_iarena_set(tsd, newarena);
 }
 
 static void


interwq:
#705 was caused by re-binding iarena in arena_migrate(), part of the change in https://github.com/jemalloc/jemalloc/commit/01f47f11a67d1a2505cc1f21851c466651eba431#diff-d2be2656d37670b0c7f5f4320c0670ceR449

The re-binding will cause internal allocations to be done via manual arenas, which triggers the assertion in #705. I don't recall why I added that part, as tests all look fine with this diff which simply removes the iarena binding.

As for now, arena_migrate has only 2 call sites,
1) thread_arena_ctl -- https://github.com/jemalloc/jemalloc/blob/dev/src/ctl.c#L1440
and 2) percpu arena -- https://github.com/jemalloc/jemalloc/blob/ee202efc79d650e16e3ecb1569efccbc5666e116/include/jemalloc/internal/arena_inlines_a.h#L59

For (1), the re-binding probably shouldn't be done, especially if the dest arena is outside of auto range.
For (2), maybe makes sense to rebind when the dest arena is in the auto range. But I don't think it will be a big deal either way.

A couple of straightforward solutions:
a) not rebind iarena in arena_migrate, i.e. this diff as is.
b) arena_migrate takes an extra bool argument migrate_internal, and pass in true if it makes sense, i.e. in percpu_arena_update (where dest arena is always in auto range), and in thread_arena_ctl when dest arena is in auto range.

Which one makes more sense to you @jasone ? Or am I missing anything?

PR #709, opened 2017-03-20
Incorporate size class and slab flag into rtree leaf elements

### LINKED ISSUE #694: Optimizing free fast path by caching extent internals in rtree cache
With the new extent based design, rtree lookup is on the critical path of free, just like chunk header access in previous chunk based design. 

As getting the extent based on the pointer-to-free is the first step on free, we are expecting a fairly large chunk of overhead from it. However, currently there is a second cache access incurs even higher overhead: after getting the ptr to the extent from rtree (almost always cached lookup), we need to dereference the extent ptr, e.g. reading if it's a slab, getting binind,  https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/arena_inlines_b.h#L138

This means, we are suffering 2 sequential cache misses, first getting extent ptr, second reading extent content. With the additional overhead from the second step, the free fast path is ~30% slower in the application benchmark I had.

To make the data access more cache friendly, we should cache more stuff in rtree_ctx_cache_elm_s, e.g. e_slab and e_slab_data->binind, so that the free fast path doesn't have to dereference *extent. I believe with this caching change we should be able to speed up free significantly to match the chunk based design.

jasone:
These revisions do the following:
1) Refactor extents to store size class index rather than usable size, because size class index has a denser representation.
2) Specialize rtree internal vs leaf elements so that extra metadata can be stored in leaves.
3) Incorporate szind and slab into rtree leaves.  This requires some refactoring to plumb szind through to all the places it needs to be set/updated.
4) Switch szind/slab lookups from reading extent contents to reading the equivalent rtree leaf contents.
5) Put off extent lookups as long as possible.

These changes resolve #694.


interwq:
I'm start to review this. Will try finishing it by end of day.

include/jemalloc/internal/rtree_structs.h
```diff
@@ -6,9 +6,26 @@ struct rtree_node_elm_s {
 };
 
 struct rtree_leaf_elm_s {
-	atomic_p_t	le_extent; /* (extent_t *) */
+#if RTREE_NHIB >= LG_CEIL_NSIZES
```
interwq: nit: ```#ifdef RTREE_LEAF_COMPACT```

include/jemalloc/internal/rtree_structs.h
```diff
@@ -6,9 +6,26 @@ struct rtree_node_elm_s {
...
+	 * For example, on a 64-bit x64 system with 48 significant virtual
+	 * memory address bits, the index, extent, and slab fields are packed as
+	 * such:
+	 *
+	 * x: index
+	 * e: extent
+	 * b: slab
+	 * k: lock
+	 *
+	 *   11111111 xxxxxxxx eeeeeeee [...] eeeeeeee eeee00bk
```
interwq: Don't we only have 3 low bits available? Or are extents always 16B aligned? Also, why making the highest 8 bits 1s instead of 0s?

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -75,6 +83,43 @@ rtree_subkey(uintptr_t key, unsigned level) {
...
+rtree_leaf_elm_bits_extent_get(uintptr_t bits) {
+	/* Restore sign-extended high bits, mask slab and lock bits. */
+	return (extent_t *)(((bits << RTREE_NHIB) >> RTREE_NHIB) &
+	    ~((uintptr_t)0x3));
+}
+
+JEMALLOC_ALWAYS_INLINE szind_t
+rtree_leaf_elm_bits_szind_get(uintptr_t bits) {
+	return (szind_t)((bits >> LG_VADDR) & (((uintptr_t)(0x1) <<
+	    LG_CEIL_NSIZES) - 1));
```
interwq: if we make high bits 0s, we can avoid this masking?

src/large.c
```diff
@@ -114,15 +114,15 @@ large_ralloc_no_move_shrink(tsdn_t *tsdn, extent_t *extent, size_t usize) {
...
+		    &extent_hooks, extent, usize + large_pad, size2index(usize),
+		    diff, NSIZES);
 		if (trail == NULL) {
 			return true;
 		}
 
 		if (config_fill && unlikely(opt_junk_free)) {
 			large_dalloc_maybe_junk(extent_addr_get(trail),
-			    extent_usize_get(trail));
+			    extent_size_get(trail));
```
jasone: ```extent_usize_get()``` can now only be called for extents that have had their szind set.  In this case we've just split off some trailing pages, and that extent has not had its szind set (nor do we want to set it for extents that won't be directly accessed by the application).  The fix is to rename the ```usize``` argument to ```size``` for ```large_dalloc_[maybe_]junk()```.

src/rtree.c
```diff
@@ -104,90 +148,186 @@ rtree_node_init(tsdn_t *tsdn, rtree_t *rtree, unsigned level,
...
 
-static rtree_elm_t *
-rtree_subtree_read(tsdn_t *tsdn, rtree_t *rtree, bool dependent) {
-	rtree_elm_t *subtree = rtree_subtree_tryread(rtree, dependent);
-	if (!dependent && unlikely(!rtree_node_valid(subtree))) {
-		subtree = rtree_node_init(tsdn, rtree, 0, &rtree->root);
+UNUSED static rtree_leaf_elm_t *
+rtree_root_leaf_read(tsdn_t *tsdn, rtree_t *rtree, bool dependent) {
+	rtree_leaf_elm_t *leaf = rtree_root_leaf_tryread(rtree, dependent);
+	if (!dependent && unlikely(!rtree_leaf_valid(leaf))) {
```
jasone: Well, this code isn't used in practice, but it needs to be there in order to support single-level rtrees.  I tried to keep this as symmetric with the used code as possible, for simplicity/maintainability reasons.  It does work (tested in 32-bit mode), but if we were to change anything here, it might be better to simply remove single-level rtree support.  Thoughts?

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -163,62 +162,110 @@ arena_vsalloc(tsdn_t *tsdn, const void *ptr) {
...
-arena_dalloc(tsdn_t *tsdn, extent_t *extent, void *ptr, tcache_t *tcache,
-    bool slow_path) {
+arena_dalloc(tsdn_t *tsdn, void *ptr, tcache_t *tcache, bool slow_path) {
 	assert(!tsdn_null(tsdn) || tcache == NULL);
 	assert(ptr != NULL);
 
-	szind_t szind = extent_szind_get(extent);
-	if (likely(extent_slab_get(extent))) {
+	rtree_ctx_t rtree_ctx_fallback;
+	rtree_ctx_t *rtree_ctx = tsdn_rtree_ctx(tsdn, &rtree_ctx_fallback);
```
jasone: Nice optimization!  I just implemented it, and will include it when updating the pull request.

include/jemalloc/internal/rtree_structs.h
jasone: We specify ```QUANTUM``` alignment when allocating ```extent_t``` structures in ```extent_alloc()```.

I'll convert the high bits to 0s and remove the unnecessary masking.

src/rtree.c
interwq: Oh in that case I feel we should keep single level support.

I was also thinking about the case ~10 lines below:
```
#if RTREE_HEIGHT > 1
node = init_missing ? rtree_root_node_read(tsdn, rtree, dependent) :
 	    rtree_root_node_tryread(rtree, dependent);
```
If we initialize the root node at startup or use static memory for it, we can avoid the runtime check there.

jasone: I just put together a prototype using static memory for the root, and it appears to work well.  I'll clean it up and integrate with this pull request later today.

jasone: This allowed some nice simplifications, and should be marginally faster too.  https://github.com/jemalloc/jemalloc/pull/709/commits/113d96bc8e97ac30231fafc278c743bfabe2237c

include/jemalloc/internal/arena_inlines_b.h
jasone: This is implemented in https://github.com/jemalloc/jemalloc/pull/709/commits/23b986944042cc9cc327f00fe2ce903753093b2e.

interwq: @jasone: nice change spliting into the _no_tcache variants! Now I can look at that function without scratching my head.

src/rtree.c
```diff
@@ -104,90 +150,137 @@ rtree_node_init(tsdn_t *tsdn, rtree_t *rtree, unsigned level,
...
 }
 
-rtree_elm_t *
-rtree_elm_lookup_hard(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
+rtree_leaf_elm_t *
+rtree_leaf_elm_lookup_hard(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
     uintptr_t key, bool dependent, bool init_missing) {
-	rtree_elm_t *node = init_missing ? rtree_subtree_read(tsdn, rtree,
-	    dependent) : rtree_subtree_tryread(rtree, dependent);
+	rtree_node_elm_t *node;
```
interwq: nit: mark node as UNUSED (when height is 1)?

jasone: I considered this, but the compiler doesn't complain about the unused variable, apparently because there is still dead code that references it.

PR #712, opened 2017-03-23
Use first fit layout policy instead of best fit.

### LINKED ISSUE #711: Experiment with "first fit" layout policy again
97c04a93838c4001688fe31bf018972b4696efe2 was a "first-fit" experiment that was subsequently reverted (aa2826621e1793db9faea31e803690ccbe36f14c), but the experiment was flawed in that jemalloc still packed to low addresses at the time, i.e. the packing criterion was unstable due to ```mmap()``` memory growing downward.  Repeat this experiment for extent heaps that do not delay coalescing.

jasone:
For extents which do not delay coalescing, use first fit layout policy
rather than first-best fit layout policy.  This packs extents toward
older virtual memory mappings, but at the cost of higher search overhead
in the common case.

This resolves #711.

I'm not sure we want to investigate this further, but feedback is welcome.  I performed some ad hoc performance tests (using t-test1) in which I disabled ```delay_coalesce``` for ```extents_dirty``` (on both the A and B sides of the experiment), and measured a ~2X slowdown for large allocation.  In other words, when this layout policy comes into play, it definitely impacts throughput negatively.  The benchmark is too simple to provide data on whether the change improves layout in practice, though in principle this should be a positive change to layout.

Since extents are essentially independent of each other, we don't suffer many negative side effects from virtual memory fragmentation (unlike for 4.x and earlier, where chunk metadata could take up lots of space as a side effect).  I think rtree metadata overhead is our biggest concern, and its cost is minuscule in comparison to chunk metadata overhead.

Thoughts?


interwq:
Hrm since there is tradeoff, maybe wait for some production test results on this one?

jasone:
Sounds reasonable to me.  I'll leave this pull request open so we have it available for performance testing.

interwq:
Sounds good. I'll test it along with a few other things tomorrow.

interwq:
I think we got another instance with a test shown in-progress here but actually finished.

jasone:
I'm going to do some profiling to check how much of the extra overhead is due to heap operations, versus iterating over the size classes.  If the iteration overhead is a major component, we can optimize it with the bitmap code (just need to add find-first-unset with min offset: ```size_t bitmap_ffu(const bitmap_t *bitmap, const bitmap_info_t *binfo, size_t min_offset)```).

jasone:
Iteration looks to be roughly 15% of the total search overhead, though I can imagine it being higher under some application loads.  I'm going to do the ```bitmap_ffu()``` optimization and see what it gains us.

jasone:
The search overhead turned out to be more like 50% of the total.  With the bitmap optimization in place, I'm convinced that this is worthy of more extensive load testing.

interwq:
@jasone: please let me know if you think this is ready for a spin.

jasone:
@interwq, I just pushed a fix for the test failure (which in practice only happens on 32-bit systems), so this should be ready for benchmarking.

jasone:
On the other hand, #715 may be related to this.  I just hit a couple other seemingly random test failures, all with this (#712) applied, so if we're lucky all the problems have this as a single cause.

jasone:
It appears that the failures were due to a subtle cpp macro expansion issue.  I refactored ```BITMAP_GROUPS_MAX``` to be:
```
#define BITMAP_GROUPS(nbits) BITMAP_BITS2GROUPS(nbits)
#define BITMAP_GROUPS_MAX BITMAP_GROUPS(BITMAP_MAXBITS)
```
However, ```BITMAP_MAXBITS``` is defined via some nested macros, and something was going wrong in some context that I didn't fully figure out (bisecting pointed to this as the cause).

jasone:
I repeated the original experiment in which I measured a ~2X slowdown for large allocation with delayed coalescing disabled -- with the bitmap optimization in place the slowdown is only ~1.15X.  With delayed coalescing enabled (i.e. default), this branch appears to actually be a bit faster than the dev branch, probably because the bitmap optimization helps the best-any fit search for ```extents_dirty``` more than first-fit search slows down the other extent searches.

jasone:
@interwq, this branch is ready for a benchmarking spin.

jasone:
The benchmarking was favorable; landing.

Issue #461, opened 2016-10-04
Optimize extent_t layout


jasone:
`extent_t` can be compacted via one or more techniques:
- Store arena index rather than a pointer.
- Combine booleans into a bitfield.
- Store usize as a size class index rather than `size_t` (assuming that meets all needs), and potentially use the same bitfield as the booleans for storage.

`extent_t` is currently 152 bytes on x64, and the above should suffice to drop the size to no more than 128 bytes.

See #360 for context.

PR #716, opened 2017-03-24
Extent layout

jasone:
These changes relate to #461, and compact ```extent_t``` to 136 bytes.  I see a few options for reaching 128 bytes, but I'm not very excited about any of them:
- Remove the ```nfree``` field from ```arena_slab_data_t```, and use inherently slower bitmap operations to detect edge conditions.
- Move the ```nfree``` field from ```arena_slab_data_t``` into ```extent_t```'s ```e_bits``` (and reduce the serial number range to 31 bits).  This feels like a layering violation.
- Overlay ```extent_t```'s ```e_size``` field with ```e_slab_data```, and use the ```szind``` bits to track size when in the slab state.  I experimented with this, and it's tricky enough to get the state transitions right that we'd need to change the extent getter/setter API to make inconsistent transitions impossible.

### DIFF
--- a/include/jemalloc/internal/extent_inlines.h
+++ b/include/jemalloc/internal/extent_inlines.h
@@ -49,7 +49,7 @@ int extent_snad_comp(const extent_t *a, const extent_t *b);
 #if (defined(JEMALLOC_ENABLE_INLINE) || defined(JEMALLOC_EXTENT_C_))
 JEMALLOC_INLINE arena_t *
 extent_arena_get(const extent_t *extent) {
-	return extent->e_arena;
+	return arenas[extent->e_arena_ind];
 }
 
 JEMALLOC_INLINE void *
@@ -151,7 +151,7 @@ extent_prof_tctx_get(const extent_t *extent) {
 
 JEMALLOC_INLINE void
 extent_arena_set(extent_t *extent, arena_t *arena) {
-	extent->e_arena = arena;
+	extent->e_arena_ind = (arena != NULL) ? arena_ind_get(arena) : UINT_MAX;
 }
 
 JEMALLOC_INLINE void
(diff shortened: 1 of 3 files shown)


include/jemalloc/internal/extent_inlines.h
```diff
@@ -49,32 +49,20 @@ int extent_snad_comp(const extent_t *a, const extent_t *b);
...
-	    !extent->e_slab);
-	return extent->e_addr;
-}
-
-JEMALLOC_INLINE size_t
-extent_size_get(const extent_t *extent) {
-	return extent->e_size;
+	unsigned arena_ind = (unsigned)((extent->e_bits &
+	    EXTENT_BITS_ARENA_MASK) >> EXTENT_BITS_ARENA_SHIFT);
+	if (arena_ind > MALLOCX_ARENA_MAX) {
```
interwq: Why would this happen?

Also maybe add cassert((1<<EXTENT_BITS_ARENA_SHIFT) > MALLOCX_ARENA_MAX) somewhwere.

jasone: We convert ```NULL``` to all 1 bits in the setter, so if anyone reads the value back out, we need to convert it back to ```NULL```.  I didn't check whether we ever actually call this getter when the field is encoding ```NULL```.

Re: the ```cassert()``` suggestion, the ```EXTENT_BITS_*``` definitions are based on ```MALLOCX_ARENA_BITS```, so the defined relationship with ```MALLOCX_ARENA_MAX``` assures we have enough bits to store the arena index.

interwq: I'd imagine reading NULLed code back should be a invalid operation? Maybe remove the brach and assert, and see if it ever happens?

jasone:
Correct, this does not impact the fast path.

interwq:
@jasone: do we have alignment constrain on extents (IIRC 16B)? I'm thinking about, if we really push it down to 128B, that's 2 cachelines on x64, and when cacheline aligned, the 2 cachelines can be fetched together with adjacent cacheline prefetching. I don't think it will make a big difference, but could potentially reduce our cache footprint a bit.

jasone:
Yes, we have 16-byte alignment, so getting down to 128 bytes would be great.  We might even want to force 64-byte alignment if we pull it off.

PR #725, opened 2017-03-27
Fix a race in rtree_szind_slab_update() for RTREE_LEAF_COMPACT.

jasone:
This fixes a race that does not currently cause problems in practice, other than to cause intermittent assertion failures when clobbering the lock bit.

### DIFF
--- a/include/jemalloc/internal/rtree_inlines.h
+++ b/include/jemalloc/internal/rtree_inlines.h
@@ -26,6 +26,8 @@ void rtree_leaf_elm_slab_write(tsdn_t *tsdn, rtree_t *rtree,
     rtree_leaf_elm_t *elm, bool acquired, bool slab);
 void rtree_leaf_elm_write(tsdn_t *tsdn, rtree_t *rtree, rtree_leaf_elm_t *elm,
     bool acquired, extent_t *extent, szind_t szind, bool slab);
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/rtree_inlines.h
```diff
@@ -465,6 +455,37 @@ rtree_leaf_elm_release(tsdn_t *tsdn, rtree_t *rtree, rtree_leaf_elm_t *elm) {
...
+	 * modified by another thread, the fact that the lock is embedded in the
+	 * same word requires that a CAS operation be used here.  It would be
+	 * possible to directly CAS here and avoid the lock/unlock operations,
+	 * but at the cost of extra special-purpose code.
+	 */
+	rtree_leaf_elm_t *elm = rtree_leaf_elm_acquire(tsdn, rtree, rtree_ctx,
+	    key, true, false);
+	rtree_leaf_elm_slab_write(tsdn, rtree, elm, true, slab);
+	rtree_leaf_elm_szind_write(tsdn, rtree, elm, true, szind);
+	rtree_leaf_elm_release(tsdn, rtree, elm);
```
interwq: I'm inclined to say that we should try optimizing this harder. With the current implementation its a CAS with multiple stores. Like you mentioned ideally this should be a single CAS. Probably not a big deal for performance as I won't expect this to be under contention often. But specialized elm_set_atomic function still makes sense to me.

PR #728, opened 2017-03-27
Move arena_slab_data_t's nfree into extent_t's e_bits.

### LINKED ISSUE #461: Optimize extent_t layout

jasone:
Compact extent_t to 128 bytes on 64-bit systems by moving
arena_slab_data_t's nfree into extent_t's e_bits.

Cacheline-align extent_t structures so that they always cross the
minimum number of cacheline boundaries.

Re-order extent_t fields such that all fields except the slab bitmap
(and overlaid heap profiling context pointer) are in the first
cacheline.

This resolves #461.

### DIFF
--- a/include/jemalloc/internal/extent_inlines.h
+++ b/include/jemalloc/internal/extent_inlines.h
@@ -11,6 +11,7 @@ extent_state_t extent_state_get(const extent_t *extent);
 bool extent_zeroed_get(const extent_t *extent);
 bool extent_committed_get(const extent_t *extent);
 bool extent_slab_get(const extent_t *extent);
+unsigned extent_nfree_get(const extent_t *extent);
 void *extent_base_get(const extent_t *extent);
 void *extent_addr_get(const extent_t *extent);
 size_t extent_size_get(const extent_t *extent);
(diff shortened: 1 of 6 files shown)


include/jemalloc/internal/extent_inlines.h
```diff
@@ -211,6 +220,12 @@ extent_szind_set(extent_t *extent, szind_t szind) {
 }
 
 JEMALLOC_INLINE void
+extent_nfree_set(extent_t *extent, unsigned nfree) {
+	extent->e_bits = (extent->e_bits & ~EXTENT_BITS_NFREE_MASK) |
```
interwq: nit: also assert is a slab?

include/jemalloc/internal/extent_inlines.h
```diff
@@ -211,6 +220,12 @@ extent_szind_set(extent_t *extent, szind_t szind) {
 }
 
 JEMALLOC_INLINE void
+extent_nfree_set(extent_t *extent, unsigned nfree) {
```
interwq: nit: instead of set, maybe provide an extent_init (which sets nfree), and extent_nfree_inc / _dec, which adds / subs 1 << EXTENT_BITS_NFREE_SHIFT. Or even making nfree the lowest bits if you feel it's worth the trouble.

include/jemalloc/internal/extent_inlines.h
jasone: I left that assertion out because it would require strict ordering between setting ```slab``` to true and initializing ```nfree```, but now that I look more closely at current setter callsites, this seems like a reasonable restriction.

include/jemalloc/internal/extent_inlines.h
jasone: The nfree field is invalid unless the extent is a slab, so the inc/dec will work okay, but I don't want to initialize nfree in ```extent_init()```, nor create a special ```extent_slab_init()```.

PR #729, opened 2017-03-28
Lookup extent once per time during tcache_flush_small / _large.

### LINKED ISSUE #598: Consider memoizing iealloc() results in tcache_bin_flush_small() / _large()
```tcache_bin_flush_large()``` potentially performs multiple passes over allocations when deallocating them, and looking up their corresponding extents via ```iealloc()``` is a somewhat costly operation.  Consider memoizing the extent lookup results for use in later passes.

### PR DESCRIPTION
**Pull request by interwq**
Caching the extents on stack to avoid redundant looking up overhead.

This resolves #598.

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -91,19 +91,23 @@ tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
 void
 tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, tcache_bin_t *tbin,
     szind_t binind, unsigned rem) {
-	arena_t *arena;
-	void *ptr;
-	unsigned i, nflush, ndeferred;
 	bool merged_stats = false;
 
 	assert(binind < NBINS);
 	assert(rem <= tbin->ncached);
 
-	arena = tcache->arena;
+	arena_t *arena = tcache->arena;
 	assert(arena != NULL);
-	for (nflush = tbin->ncached - rem; nflush > 0; nflush = ndeferred) {
+	unsigned nflush = tbin->ncached - rem;
+	VARIABLE_ARRAY(extent_t *, item_extent, nflush);
+	/* Look up extent once per item. */
+	for (unsigned i = 0 ; i < nflush; i++) {
+		item_extent[i] = iealloc(tsd_tsdn(tsd), *(tbin->avail - 1 - i));
+	}
+
+	while (nflush > 0) {
 		/* Lock the arena bin associated with the first object. */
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/tcache.c [hunk 1]
```diff
@@ -91,19 +91,23 @@ tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
 
 	assert(binind < NBINS);
 	assert(rem <= tbin->ncached);
 
-	arena = tcache->arena;
+	arena_t *arena = tcache->arena;
 	assert(arena != NULL);
-	for (nflush = tbin->ncached - rem; nflush > 0; nflush = ndeferred) {
+	unsigned nflush = tbin->ncached - rem;
+	extent_t *item_extent[nflush];
```
interwq: MSVC still doesn't like variable-length arrays. Will fix.

jasone: The approach you took to resolve this looks fine, but for future reference note that the ```VARIABLE_ARRAY()``` macro is an alternative.

src/tcache.c
```diff
@@ -140,11 +144,13 @@ tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, tcache_bin_t *tbin,
 				 * handled in a future pass.
 				 */
 				*(tbin->avail - 1 - ndeferred) = ptr;
+				item_extent[ndeferred] = item_extent[i];
```
jasone: Write the right-hand side as just ```extent``` (here and in ```tcache_bin_flush_large()```)?

src/tcache.c
```diff
@@ -91,19 +91,23 @@ tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
 
-	arena = tcache->arena;
+	arena_t *arena = tcache->arena;
 	assert(arena != NULL);
-	for (nflush = tbin->ncached - rem; nflush > 0; nflush = ndeferred) {
+	unsigned nflush = tbin->ncached - rem;
+	extent_t *item_extent[TCACHE_NSLOTS_SMALL_MAX];
+	/* Look up extent once per item. */
+	for (unsigned i = 0 ; i < nflush; i++) {
+		item_extent[i] = iealloc(tsd_tsdn(tsd), *(tbin->avail - i - 1));
```
jasone: Nit: for consistency, write ```*(tbin->avail - 1 - i)``` here (and in ```tcache_bin_flush_large()```) for consistency with the other sites that do similar math.

src/tcache.c
interwq: Ah I was looking for it and couldn't find it. Thanks for the pointer!

interwq:
Will address feedback and update again before pushing.

PR #732, opened 2017-03-29
Remove the leafkey NULL check in leaf_elm_lookup.

interwq:
This gets rid of a branch on rcache lookup. u-benchmark shows a ~2% improvement.

### DIFF
--- a/include/jemalloc/internal/rtree_types.h
+++ b/include/jemalloc/internal/rtree_types.h
@@ -39,7 +39,7 @@ typedef struct rtree_s rtree_t;
 #endif
 
 /* Needed for initialization only. */
-#define RTREE_LEAFKEY_INVALID	((uintptr_t)1)
+#define RTREE_LEAFKEY_INVALID ((uintptr_t)1)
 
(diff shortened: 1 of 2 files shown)


interwq:
This is going to need some preparation (adding non-static initializer for tsd members). Will put up that diff first and then update here.

interwq:
This is ready for review. Please only look at the top commit (the diff requires #733).

include/jemalloc/internal/rtree_types.h
```diff
@@ -51,12 +54,13 @@ typedef struct rtree_s rtree_t;
  * the tree nodes, and the cache will itself suffer cache misses if made overly
  * large, not to mention the cost of linear search.
  */
-#define RTREE_CTX_NCACHE 8
+#define RTREE_CTX_NCACHE	8
```
davidtgoldblatt: We've stopped requiring the tabs here I believe.

interwq:
Only style changes were made after previous tests (all passed). Pushing.

PR #733, opened 2017-03-29
Add init function support to tsd members.

interwq:
This will facilitate embedding tcache into tsd, which will require proper
initialization cannot be done via the static initializer.  Make tsd->rtree_ctx
to be initialized via rtree_ctx_data_init().


interwq:
This diff also changes rtree_ctx initialization to make #732 possible.

PR #735, opened 2017-03-30
Do proper cleanup for tsd_state_reincarnated.

interwq:


interwq:
Of course. I'll add a test case in test/unit/tsd

interwq:
Tests all passed previously. Pushing.

Issue #701, opened 2017-03-14
Optimizing TSD and thread cache layout


interwq:
Fast path tcache performance is critical as it's expected to be accessed very frequently. We aim to improve the fast path significantly with the following two changes:

1) Embed tcache into TSD:
Currently each tsd stores a pointer to the automatic tcache:
https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/tsd_structs.h#L19
This adds a memory dereference on the fast path, and potentially a cache miss. We can avoid this by embedding the tcache struct into tsd directly. Since we do not allow switching auto tcache, this should work fine w/o breaking APIs.

2) Optimize tcache and tsd for a more compact layout. We'd like have higher density of the tcache bins to improve cache locality. Currently tbins are 32 bytes, which will be cut down to 24. Also pack TSD tighter.

interwq:
@jasone : to embed tcache into tsd, my current plan is to split the tbins into two bins: small and large, basically:

tcache_t {
    ....
    tbin_t small_bins[NBINS];
    tbin_t *large_bins; /*Dynamically sized*/
}
So that we embed small bins for better locality, while keep the ability to set lg_tcache_max, i.e. nhbins. Does that make sense to you? Or you think we should just get rid of lg_tcache_max and make the size static?

jasone:
I think we should make ```large_bins``` have ```NSIZES - NBINS``` elements so that no dynamic sizing is necessary, and probably place tcache at the end of tsd so that the unused elements don't break up the utilized cachelines.

interwq:
That sounds reasonable. For the unused large bins (binind > nhbins), probably not allocating their avail arrays?

jasone:
Ah right, definitely avoid allocating more avail slots than actually used.

interwq:
The first half of this (embedding tcache into tsd) is implemented in #736. Tested with @davidtgoldblatt using test/stress/microbench, which shows a ~15% improvement on malloc / free / etc. Likely this made the fast path more pipeline friendly. A sample output:
With #736:
```
        100000000 iterations, malloc=2189009us, mallocx=2275009us, ratio=1:0.962
        test_malloc_vs_mallocx: pass
        100000000 iterations, free=2192009us, dallocx=2223009us, ratio=1:0.986
        test_free_vs_dallocx: pass
        100000000 iterations, dallocx=2190009us, sdallocx=2326009us, ratio=1:0.942
[...]

PR #736, opened 2017-03-30
Embed tcache into tsd

interwq:
Integrate auto tcache into TSD.
    
The embedded tcache is initialized upon tsd initialization.  The avail arrays
for the tbins will be allocated / deallocated accordingly during init / cleanup.
    
With this change, the pointer to the auto tcache will always be available, as
long as we have access to the TSD.  tcache_available() (called in tcache_get())
is provided to check if we should use tcache.

This is part of #701.


interwq:
This is ready for review. Please only look at the most recent commit "Integrate auto tcache into TSD."

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -321,13 +321,18 @@ rtree_leaf_elm_lookup(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
...
 	if (likely(rtree_ctx->cache[i].leafkey == leafkey)) {		\
 		rtree_leaf_elm_t *leaf = rtree_ctx->cache[i].leaf;	\
 		assert (leaf != NULL);					\
 		if (i > 0) {						\
 			/* Bubble up by one. */				\
-			rtree_ctx->cache[i] = rtree_ctx->cache[i - 1];	\
+			rtree_ctx->cache[i].leafkey =			\
+				rtree_ctx->cache[i - 1].leafkey;	\
+			rtree_ctx->cache[i].leaf =			\
+				rtree_ctx->cache[i - 1].leaf;		\
```
interwq: With this diff, if we don't change the struct assignment to assigning leafkey and leaf separately, it triggers a gcc internal compiler error like this one: https://travis-ci.org/jemalloc/jemalloc/jobs/217459658

include/jemalloc/internal/tsd_types.h
```diff
@@ -291,7 +291,8 @@ a_name##tsd_cleanup_wrapper(void) {					\
 	if (wrapper == NULL) {						\
 		return false;						\
 	}								\
-	if (a_cleanup != malloc_tsd_no_cleanup &&			\
+	if (/* a_cleanup != malloc_tsd_no_cleanup && */			\
+	    /* Verify compiler (mingw32+gcc) bug.*/			\
```
interwq: Confirmed compiler bug. On Windows with MinGW64, the "a_cleanup != malloc_tsd_no_cleanup" is not computed correctly and is compiled to false incorrectly, even though we never actually passed in malloc_tsd_no_cleanup. This resulted in missing cleanups. Without this line of change (and L404), the test failure is reproducible in Windows, e.g. https://ci.appveyor.com/project/jasone/jemalloc/build/638/job/ulyou1yumv1b1o6a

I'll update with a workaround.

interwq: Regarding workarounds to avoid the compiler issue:

1) I tried replacing malloc_tsd_no_cleanup with NULL, i.e. changing "a_cleanup != malloc_tsd_no_cleanup" to "a_cleanup != NULL", and that worked fine, except that a bunch of warnings about the comparison will always be as the address will never be NULL. (I suspect this was part of the reason to have malloc_tsd_no_cleanup?)
2) keep malloc_tsd_no_cleanup (change it to an empty function instead of assert not_reached), but remove the check "a_cleanup != malloc_tsd_no_cleanup". Theoretically compiler should still be able to optimize this away when malloc_tsd_no_cleanup is actually used. 

Thoughts @davidtgoldblatt?

interwq: Yeah the overhead should be minimal -- happen only once per thread and only a dull function call (if it's not optimized away).

In fact, since currently we don't use the _no_cleanup. It doesn't matter at all.

include/jemalloc/internal/tsd_structs.h
```diff
@@ -69,6 +68,7 @@ struct tsdn_s {
 };
 
 static const tsd_t tsd_initializer = TSD_INITIALIZER;
+UNUSED static const void *malloc_tsd_no_cleanup = (void *)0;
```
interwq: As mentioned previously, this change is to avoid a compiler issue, where the empty function version before this diff cannot be compared correctly in "a_cleanup != malloc_tsd_no_cleanup" under Windows MinGW64.

interwq:
Had a small initialization tweak.  To be safe will wait for the tests before pushing.

Issue #576, opened 2017-01-20
Refactor arena-tcache interactions


davidtgoldblatt:
The arena and tcache code is very tightly coupled. This complicates the logic, and makes testing in isolation difficult.

PR #739, opened 2017-04-01
Basic reentrancy support

davidtgoldblatt:
This tests that we don't get into any libc- or arena_new- related reentrancy that we can't handle. If we detect reentrancy (using tsd), we'll force ourselves into arena 0, which we hope has already been initialized.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1656,6 +1656,14 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts) {
 	szind_t ind = 0;
 	size_t usize = 0;
 
+	/*
+	 * For reentrancy checking, we get the old reentrancy level from tsd and
+	 * reset it once we're done.  In case of early bailout though, we never
+	 * bother getting the old level, so we shouldn't try to reset it.  This
+	 * is indicated by leaving the pointer as NULL.
+	 */
+	int *reentrancy_level = NULL;
+
 	/* Initialize (if we can't prove we don't have to). */
 	if (sopts->slow) {
 		if (unlikely(malloc_init())) {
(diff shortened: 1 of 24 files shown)


davidtgoldblatt:
Looks like there's some OOMing going on -- I'll dig into the CI stuff tomorrow.

src/jemalloc.c
```diff
@@ -1706,7 +1714,27 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts) {
...
+		witness_assert_lockless(tsd_tsdn(tsd));
+	}
+	if (unlikely(*reentrancy_level > 1)) {
+		/*
+		 * We should never specify particular arenas or tcaches from
+		 * within our internal allocations.
+		 */
+		assert(dopts->tcache_ind == TCACHE_IND_AUTOMATIC);
+		assert(dopts->arena_ind = ARENA_IND_AUTOMATIC);
+		dopts->tcache_ind = TCACHE_IND_NONE;
```
interwq: Any particular reason why tcache shouldn't be used for reentrancy? Is it true that only tcache initialization will cause trouble? It seems after tcache is created, using it during reentrancy should be fine.

src/jemalloc.c
```diff
@@ -1706,7 +1714,27 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts) {
...
+		/* We know that arena 0 has already been initialized. */
+		dopts->arena_ind = 0;
```
interwq: If we have the "if (*tsd_reentrancy_levelp_get(tsd) > 1) {" detection in arena_choose, do we still need this here? 

I'm thinking that, if we can make tcache working for reentrancy, and pick a0 within arena_choose, we can get rid of this branch from fast path. (arena_choose isn't called on tcache only path I believe).

src/jemalloc.c
davidtgoldblatt: It helps minimizes the number of things that we have to reason about (albeit in a vague, sort of ill-defined way).

For example: suppose we detect that the tcache is empty. We go to the arena, and become reentrant. We reentrantly do some operations that fill the tcache. We then finish being reentrant, and return to the initial allocation, having filled the tcache. Now, the reason the initial allocation went into the arena in the first place is because its tcache was empty -- I think it's pretty reasonable for us to assume at that point that the tcache is still empty.

interwq:
Right now we have ```++*reentrancy_level;``` and  ```--``` on the fast path. Would it work if we move this to be around all the external functions, so that we don't avoid the cost on fast path?

src/jemalloc.c
interwq: I see. If we have a simple way to disable tcache before calling external functions, (and restore after), would that be fine? With the tcache diffs pending right now, the tcache_enabled in tsd can be switched on/off easily to disable auto arena.

interwq:
What I'm trying to achieve eventually is, if we do the following, will it be possible to avoid the branch on fast path?
{
  // save tsd->arena & ->tcache_enabled;
  tsd->arena = a0;
  tsd->tcache_enabled = false;
  call_ext_func(...);
  // and restore them;
}

davidtgoldblatt:
I think that's the ideal end state (both for this, and also because it lets us do some other nice things, like asserting that we're not holding any big locks when doing syscalls or the like). It's going to be sort of hard to get there for a while though, because of stuff like arena/tcache coupling and the difficulty of auditing and testing for these dependencies popping up. I'd like to do another pass on this once #576 is done and those interactions have gotten more robust.

The reentrancy_level incrementing and decrementing don't actually end up hurting us (at least, according to perf).

davidtgoldblatt:
(Rebased to dev)

interwq:
@davidtgoldblatt: sounds good. Let's come back to optimization after more preparation then.

PR #741, opened 2017-04-02
detect main executable even if PIE is active

BusyJay:
ref https://github.com/gperftools/gperftools/issues/589


davidtgoldblatt:
Thanks for the diff!

Issue #303, opened 2015-11-21
Feature request: cmake support


namreeb:
CMake support would be wonderful for cross-platform support.  It would allow easy integration into existing cross-platform projects.

rustyx:
+1. Not to mention CMake is much more BSD-friendly license-wise.

yuslepukhin:
I build internally with CMake but there are 3 Windows specific places which can be fixed/augumented. However, I do not have other platforms handy to make it gneric.

jasone:
I'm open to switching over to CMake, but it's _way_ down my list of things to do right now.  If someone besides me is passionate about converting all the portability logic in ```configure.ac``` and making the new build system function reliably on at least Linux, FreeBSD, OS X, and Windows, I'm willing to do my part to learn more about CMake and help maintain it as the sole build system.

wbierman:
Is there an overview of the different build environments jemalloc currently supports?  I might be willing to undertake this.

jasone:
@wbierman, most of the supported platforms are enumerated in [```configure.ac```](https://github.com/jemalloc/jemalloc/blob/dev/configure.ac#L290).

photex:
+1 I just started to work on this for Linux and OSX. :)

yuslepukhin:
+1 I think a good idea would be to issue a fairly generic CMake and then a few people could take a platform each and fix it up. That would parallelize the effort and would not tax a specific individual trying to support everything that is in configure.ac, This is a high bar to get something out that might benefit people immediatly.

However, that would require a short transition period where CMake file would exist in parallel with the existing system. If @jasone would soften his stand on having 4 platform support at the first iteration, although those 4 make become available almost immediately due to the nature of CMake.

jasone:
I'm fine with an iterative approach to CMake transition, but I don't want to do a release with both build systems.  Incremental changes (and refactoring diffs that touch files outside the new CMake-related files) need to be self-contained so that we can cherry-pick around them for bugfix releases, and leave them out entirely if the effort isn't complete in time for the 4.2.0 release.

Re: 4.2.0 release date, pretty much all of the features we're currently considering are non-trivial, and it's likely to be at least a couple months before we have enough functionality to justify release.

rustyx:
Sounds like a good candidate for a feature branch.

jasone:
Good idea.

wbierman:
[...]

PR #742, opened 2017-04-03
Improve the build on CloudABI

EdSchouten:
[CloudABI](https://nuxi.nl/) has been using jemalloc for quite some time as part of [its C library](https://github.com/NuxiNL/cloudlibc). These are some of the local changes I had to apply to get jemalloc to build cleanly.

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_decls.h
+++ b/include/jemalloc/internal/jemalloc_internal_decls.h
@@ -11,7 +11,7 @@
 #    include <sys/param.h>
 #  endif
 #  include <sys/mman.h>
-#  if !defined(__pnacl__) && !defined(__native_client__)
+#  if !defined(__CloudABI__) && !defined(__pnacl__) && !defined(__native_client__)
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
It'll be a couple weeks wait before we get to this. In general, we're trying to reduce the number of special casing unusual build scenarios (especially when we don't have any realistic mechanism for automated testing of them). We'll try to figure out a strategy for dealing with this sort of thing at our next team meeting.

jasone:
The ```PATH_MAX``` change looks fine as-is, but the other changes should probably be implemented as configure tests, lest we end up with a brittle set of platform-specific cpp conditionals.  We're getting close to working on #303, so it might be best to revisit this just after the CMake transition is complete.

interwq:
Closing for now, and will followup after #303.

dm0-:
Could the `PATH_MAX` commit be cherry-picked from this PR?  It also solves issues with building on GNU Hurd.

PR #745, opened 2017-04-04
Make the tsd member init functions to take tsd_t * type.

interwq:


davidtgoldblatt:
What does this get us?

interwq:
This is mainly to support initialization. In the init function, we may want to access other members of the tsd (e.g. when initialize tcache, access the associated arena; not quite clean but oh well). Doing tsd_fetch() during init doesn't seem to be safe. So I figured let's pass in the tsd directly. (this also matches the tsd member cleanup function, which also takes tsd instead of member ptr).

PR #751, opened 2017-04-06
Get rid of tcache_enabled_t as we have runtime init support.

interwq:
This removes tcache_enabled_t, which was used to initialize tsd->tcache_enabled to opt_tcache.  Since now we have init function support for tsd members, we can get rid of this type.

### DIFF
--- a/include/jemalloc/internal/tcache_inlines.h
+++ b/include/jemalloc/internal/tcache_inlines.h
@@ -22,30 +22,24 @@ tcache_t	*tcaches_get(tsd_t *tsd, unsigned ind);
 #if (defined(JEMALLOC_ENABLE_INLINE) || defined(JEMALLOC_TCACHE_C_))
 JEMALLOC_INLINE bool
 tcache_enabled_get(tsd_t *tsd) {
-	tcache_enabled_t tcache_enabled;
-
 	cassert(config_tcache);
 
-	tcache_enabled = tsd_tcache_enabled_get(tsd);
-	assert(tcache_enabled != tcache_enabled_default);
-
-	return (bool)tcache_enabled;
+	return tsd_tcache_enabled_get(tsd);
 }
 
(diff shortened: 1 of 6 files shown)


interwq:
Only commit "Get rid of tcache_enabled_t as we have runtime init support." is relevant.

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -22,30 +22,24 @@ tcache_t	*tcaches_get(tsd_t *tsd, unsigned ind);
...
 
 JEMALLOC_INLINE void
 tcache_enabled_set(tsd_t *tsd, bool enabled) {
 	cassert(config_tcache);
 
-	tcache_enabled_t old = tsd_tcache_enabled_get(tsd);
+	bool old = tsd_tcache_enabled_get(tsd);
 
-	if ((old != tcache_enabled_true) && enabled) {
+	if ((old != true) && enabled) {
```
davidtgoldblatt: aka !old

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -22,30 +22,24 @@ tcache_t	*tcaches_get(tsd_t *tsd, unsigned ind);
...
 		tsd_tcache_data_init(tsd);
-	} else if ((old == tcache_enabled_true) && !enabled) {
+	} else if ((old == true) && !enabled) {
```
davidtgoldblatt: aka old

include/jemalloc/internal/tcache_inlines.h
interwq: Yep. I somehow thought it was easier to read this way. Doesn't see that anymore. Will fix before pushing.

interwq:
All tests passed previously and the commit based was merged separately. Pushing.

PR #752, opened 2017-04-07
Optimizing TSD and thread cache layout.

interwq:
1) Re-organize TSD so that frequently accessed fields are closer to the
beginning and more compact.  Assuming 64-bit, the first 3.5 cachelines now
contains everything needed on tcache fast path, expect the tcache struct itself.

2) Re-organize tcache and tbins.  Take lg_fill_div out of tbin, and reduce tbin
to 24 bytes (down from 32). Split tbins into tbins_small and tbins_large, and
place tbins_small close to the beginning.

### DIFF
--- a/include/jemalloc/internal/arena_externs.h
+++ b/include/jemalloc/internal/arena_externs.h
@@ -51,7 +51,7 @@ bool arena_muzzy_decay_time_set(tsdn_t *tsdn, arena_t *arena,
 void arena_decay(tsdn_t *tsdn, arena_t *arena, bool all);
 void	arena_reset(tsd_t *tsd, arena_t *arena);
 void	arena_destroy(tsd_t *tsd, arena_t *arena);
-void	arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena,
+void	arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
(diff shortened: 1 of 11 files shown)


interwq:
Only the top commit "Optimizing TSD and thread cache layout." is relevant.

interwq:
micro-benchmark shows no difference with this diff. However this makes tsd and tbins more compact, and will reduce cache footprint. I expect this to be helpful in production workload. 

Along with #736, this completes #701.

davidtgoldblatt:
I think this works better with a helper "Get bin N" function. This introduces enough fiddly details into every tcache access that I think we should be pretty skeptical of doing it without performance data backing it up.

interwq:
@davidtgoldblatt: yeah a get_bin helper makes sense. I'll update the diff with that.

Regarding perf data, u-benchmark confirms no regression with this. Since this is pretty tightly coupled with  #736, I'd say let's land this and benchmark together?

davidtgoldblatt:
Seems reasonable.

interwq:
Just updated with:
tcache_small_bin_get() and tcache_large_bin_get().

I have the 2 explicit versions instead of just tcache_bin_get() and checks bin index internal, because:
1) in almost all cases, we do know we want small or large bins.
2) more importantly, even with opt build at a few places the redundant binind branch inside tcache_bin_get wasn't optimized away.

So in the end I went with the two getters above.

Issue #737, opened 2017-03-30
Refactor away redundant rtree lookup on free() fast path


interwq:
We current do it twice on the free fast path:

1) if stats is enabled (pretty much always) https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L2012
and 
2) in arena_dalloc: https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/arena_inlines_b.h#L209

Let's refactor the code path to avoid the additional overhead.

interwq:
Closed via #762 and #755.

PR #755, opened 2017-04-07
Pass dealloc_ctx down free() fast path.

interwq:
This gets rid of the redundent rtree lookup down fast path.

This solves #737.


interwq:
Checked compiled code and verified that only 1 lookup per free() (was 2 before because of usize calculation). u-benchmark shows a couple of % of improvements.

PR #756, opened 2017-04-10
Port CPU_SPINWAIT to __powerpc64__

rafaelfolco:
Hyper-threaded CPUs may need a special instruction inside spin loops in
order to yield to another virtual CPU. The 'pause' instruction that is
available for x86 is not supported on Power.
Apparently the extended mnemonics like yield, mdoio, and mdoom are not
actually implemented on POWER8, although mentioned in the ISA 2.07
document. The recommended magic bits are an 'or 31,31,31'.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -398,8 +398,9 @@ case "${host_cpu}" in
 	    fi
 	fi
 	;;
-  powerpc)
+  powerpc*)
 	AC_DEFINE_UNQUOTED([HAVE_ALTIVEC], [ ])
+	CPU_SPINWAIT='__asm__ volatile("or 31,31,31")'
 	;;
   *)
 	;;


configure.ac
```diff
@@ -398,8 +398,9 @@ case "${host_cpu}" in
 	    fi
 	fi
 	;;
-  powerpc)
+  powerpc*)
 	AC_DEFINE_UNQUOTED([HAVE_ALTIVEC], [ ])
+    CPU_SPINWAIT='__asm__ volatile("or 31,31,31")'
```
tamird: Indentation looks off (I think this file uses tabs)

davidtgoldblatt:
LGTM modulo the style nit.

rafaelfolco:
Fixed indentation and squashed commits. Hope this is ok.

davidtgoldblatt:
Thanks for the diff!

kichik:
Why was this removed from the latest `configure.ac`? Is ppc64 no longer supported?

PR #762, opened 2017-04-11
Refactor sdalloc and profiling paths w/ alloc_ctx.

interwq:
This is to avoid redundant rtree lookups on these path, namely sdalloc and when opt_prof == true.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2083,17 +2083,26 @@ isfree(tsd_t *tsd, void *ptr, size_t usize, tcache_t *tcache, bool slow_path) {
 	assert(ptr != NULL);
 	assert(malloc_initialized() || IS_INITIALIZER);
 
+	dalloc_ctx_t dalloc_ctx, *ctx;
 	if (config_prof && opt_prof) {
+		rtree_ctx_t *rtree_ctx = tsd_rtree_ctx(tsd);
+		rtree_szind_slab_read(tsd_tsdn(tsd), &extents_rtree, rtree_ctx,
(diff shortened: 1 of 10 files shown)


interwq:
Verified the compiled the code that down the fast path only 1 rtree lookup is involved for both malloc and free, even when opt_prof==true.

src/jemalloc.c
```diff
@@ -1747,18 +1747,23 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts) {
 		 * Note that if we're going down this path, usize must have been
 		 * initialized in the previous if statement.
 		 */
+		bool is_large = false;
```
jasone: How about we move the ```alloc_ctx``` declaration up here, and initialize ```alloc_ctx.slab``` on all paths through the code?  That avoids repeated initialization, and the compiler will typically tell us if we miss a path.

src/jemalloc.c
```diff
@@ -1749,7 +1749,14 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts) {
 		 */
 		prof_tctx_t *tctx = prof_alloc_prep(
 		    tsd, usize, prof_active_get_unlocked(), true);
+
+		alloc_ctx_t alloc_ctx;
 		if (likely((uintptr_t)tctx == (uintptr_t)1U)) {
+			if (usize > SMALL_MAXCLASS) {
```
jasone: Equivalent, but IMO easier to read: ```alloc_ctx.slab = (usize <= SMALL_MAXCLASS);```

interwq: Will stash change this into the next diff ;)

PR #767, opened 2017-04-15
Improve rtree cache with a two-level cache design.

interwq:
Two levels of rcache is implemented: a direct mapped cache as L1, combined with
a LRU cache as L2.  The L1 cache offers low cost on cache hit, but could suffer
collision under circumstances.  This is complemented by the L2 LRU cache, which
is slower on cache access (overhead from linear search + reordering), but solves
collison of L1 rather well.

The current config is set to 16 L1 entries, and 8 L2 entires. Production benchmark shows favorable results comparing to LRU cache only (>10% cost reduction for free()). Comparing to a 8 L1+ 8 L2 setting, this showed a better cost in free() and tcache_bin_flush(). /test/stress/microbench is slightly slower with this (~2%), as comparing best case this needs a "&" operation (compute cache index).

### DIFF
--- a/include/jemalloc/internal/rtree_inlines.h
+++ b/include/jemalloc/internal/rtree_inlines.h
@@ -64,6 +64,15 @@ rtree_leafkey(uintptr_t key) {
 	return (key & mask);
 }
 
+JEMALLOC_ALWAYS_INLINE size_t
+rtree_cache_direct_map(uintptr_t key) {
+	unsigned ptrbits = ZU(1) << (LG_SIZEOF_PTR+3);
+	unsigned cumbits = (rtree_levels[RTREE_HEIGHT-1].cumbits -
+	    rtree_levels[RTREE_HEIGHT-1].bits);
+	unsigned maskbits = ptrbits - cumbits;
+	return (size_t)((key >> maskbits) & (RTREE_CTX_NCACHE - 1));
+}
+
 JEMALLOC_ALWAYS_INLINE uintptr_t
 rtree_subkey(uintptr_t key, unsigned level) {
 	unsigned ptrbits = ZU(1) << (LG_SIZEOF_PTR+3);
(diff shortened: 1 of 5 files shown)


include/jemalloc/internal/rtree_inlines.h
```diff
@@ -64,6 +64,15 @@ rtree_leafkey(uintptr_t key) {
 	return (key & mask);
 }
 
+JEMALLOC_ALWAYS_INLINE size_t
+rtree_cache_direct_map(uintptr_t key) {
```
davidtgoldblatt: Out of curiousity, did you try any hashing strategies? Should we?

interwq: I discussed with @jasone a bit on hashing. Two reasons I decided to use the simple mapping here:
1) since we are talking about a few cycles of difference here on fast path, we would have to use something very fast.
2) in normal cases, we should get continuous virtual address from kernel, in which case direct mapped hash is pretty much ideal, e.g. if we have 6GB of continuous virtual address space used,  then 6 continuous entires in the rcache will work well for them, and have a smaller footprint in CPU cache (comparing to e.g. evenly distributed hash).

src/prof.c
```diff
@@ -310,6 +310,7 @@ prof_leave(tsd_t *tsd, prof_tdata_t *tdata) {
 }
 
 #ifdef JEMALLOC_PROF_LIBUNWIND
+JEMALLOC_ALIGNED(CACHELINE)
```
jasone: Intentional change?

interwq: Oh I intended to add it in a separate diff. Stashed here for benchmarking but forgot to remove... Let's keep it since it's needed anyway?

jasone: Why do we want to cacheline-align this function?  Is there something strange going on with libunwind that this works around?

interwq: Whoops I just verified again and it's not needed. I somehow got the idea about non-aligned functions trigger slow path in libunwind. Turns out it's triggered by stack misalignment instead of function. I'll submit a diff removing this.

PR #769, opened 2017-04-16
Better support larger/multiple page sizes.

jasone:


src/pages.c
```diff
@@ -81,50 +119,80 @@ pages_unmap(void *addr, size_t size) {
...
-		if (new_addr == ret) {
-			return ret;
-		}
-		if (new_addr) {
-			pages_unmap(new_addr, size);
-		}
+static void *
+pages_map_slow(size_t size, size_t alignment, bool *commit) {
+	size_t alloc_size = size + alignment - os_page;
+	/* Beware size_t wrap-around. */
```
davidtgoldblatt: Any reason this should cause a NULL return instead of an assertion failure?

src/pages.c
```diff
@@ -56,19 +65,48 @@ pages_map(void *addr, size_t size, bool *commit) {
...
-	assert(PAGE_ADDR2BASE(addr) == addr);
-	assert(PAGE_CEILING(size) == size);
+static void *
+os_pages_trim(void *addr, size_t alloc_size, size_t leadsize, size_t size,
+    bool *commit) {
+	void *ret = (void *)((uintptr_t)addr + leadsize);
+
+	assert(alloc_size >= leadsize + size);
+#ifdef _WIN32
+	os_pages_unmap(addr, alloc_size);
```
davidtgoldblatt: Why is this necessary? My (admittedly, not too careful / informed) reading of the msdn docs implies that VirtualFree works the way we want.

src/pages.c
jasone: In principle the application can ask for ridiculously large alignment, and we need to OOM rather than crashing.

src/pages.c
jasone: Windows does not automatically split/coalesce mappings like Unix does, so we have to keep a 1:1 relationship between map and unmap operations.

PR #770, opened 2017-04-17
extent_avail

jasone:
The first two diffs are worthwhile on their own; the second reduces the number of base blocks, and therefore extent structure serial numbers, which makes it feasible to encode the extent serial number in the low bits of the extent size field, under the assumption that the extent size is a multiple of ```PAGE```.  Unfortunately those bits are actually used by the base code for tracking available base block remainders, so we need a union for ```e_size_esn``` and ```e_bsize```.  It's not very pretty, and a bit more brittle than I'd like, but the approach seems sound.  Ideally we'd get some idea as to whether the extent structure reuse policy is important, but it may be hard to get a clear signal, and additionally this optimization may help more once we explicitly use huge pages for base blocks.

### DIFF
--- a/include/jemalloc/internal/extent_structs.h
+++ b/include/jemalloc/internal/extent_structs.h
@@ -23,8 +23,8 @@ struct extent_s {
 	 * z: zeroed
 	 * t: state
 	 * i: szind
-	 * n: sn
 	 * f: nfree
+	 * n: sn
 	 *
 	 * nnnnnnnn ... nnnnnfff fffffffi iiiiiiit tzcbaaaa aaaaaaaa
 	 *
@@ -102,8 +102,20 @@ struct extent_s {
 	/* Pointer to the extent that this structure is responsible for. */
 	void			*e_addr;
 
-	/* Extent size. */
-	size_t			e_size;
+	union {
+		/*
+		 * Extent size and serial number associated with the extent
+		 * structure (different than the serial number for the extent at
+		 * e_addr).
+		 *
+		 * ssssssss [...] ssssssss ssssnnnn nnnnnnnn
+		 */
+		size_t			e_size_esn;
+	#define EXTENT_SIZE_MASK	((size_t)~(PAGE-1))
+	#define EXTENT_ESN_MASK		((size_t)PAGE-1)
+		/* Base extent size, which may not be a multiple of PAGE. */
+		size_t			e_bsize;
+	};
 
 	/*
 	 * List linkage, used by a variety of lists:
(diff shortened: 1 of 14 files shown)


interwq:
@jasone: do you think this is worth prod benchmarking? If so, benchmark the entire stack or a specific commit?

include/jemalloc/internal/extent_structs.h
```diff
@@ -102,8 +102,20 @@ struct extent_s {
...
-	size_t			e_size;
+	union {
+		/*
+		 * Extent size and serial number associated with the extent
+		 * structure (different than the serial number for the extent at
+		 * e_addr).
+		 *
+		 * ssssssss [...] ssssssss ssssnnnn nnnnnnnn
+		 */
+		size_t			e_size_esn;
```
interwq: Question: why don't we always rely on address comparison? in which case is the serial number required? Or what's the benefit of having sn?

include/jemalloc/internal/extent_structs.h
```diff
@@ -117,15 +118,18 @@ struct extent_s {
...
+		 * List linkage, used by a variety of lists:
+		 * - arena_bin_t's slabs_full
+		 * - extents_t's LRU
+		 * - stashed dirty extents
+		 * - arena's large allocations
+		 */
+		ql_elm(extent_t)	ql_link;
+		/* Red-black tree linkage, used by arena's extent_avail. */
+		rb_node(extent_t)	rb_link;
+	};
```
interwq: nit: cassert (sizeof(extent_t)) <= 128 somewhere?

include/jemalloc/internal/extent_structs.h
jasone: The serial number is the primary comparison key, and the address is secondary.  Before the introduction of serial numbers, we used address as the primary (and only) key, but if memory grew downward (as on Linux), then address-only ordering caused us to prefer newer mappings, which subverted the packing heuristic.  The serial number allows us to prefer old mappings, and therefore consistently pack.

The above describes the general situation that motivates serial numbers.  However, we need a separate serial number for the extent (sn for virtual memory that actually backs application allocations) versus the extent structure serial number (esn for metadata which came from the base allocator).

include/jemalloc/internal/extent_structs.h
jasone: Unfortunately such an assertion would not hold in general, e.g. for the PGI compiler where ```RB_COMPACT``` is disabled to avoid a compiler limitation.

jasone:
Re: benchmarking, I can't imagine this hurting performance, but I'm not sure we'll see an obvious improvement with production loads either.  If you're willing to run a benchmark comparing this entire stack of diffs versus without, it would settle a point of curiosity, but don't put a bunch of effort into it.

interwq:
I see. In that case I'll focus on the unwinder issue first. We can benchmark together after fixing that.

PR #775, opened 2017-04-18
Decommit fixes

### LINKED ISSUE #766: Intermittent failure:  src/arena.c:351: Failed assertion: "p[i] == 0"
b973ec7 on rc-4.5.0. Also  confirmed as failing on 4.4.0.

Environment is Ubuntu 16.04 LTS, kernel 4.4.0-57-generic, glibc 2.23-0ubuntu7, binutils 2.26.1-1ubuntu1~16.04.3.

--with-jemalloc-prefix=je_ --with-malloc-conf=lg_chunk:20 --enable-stats --enable-debug

The asserts only seem to occur when doing aligned allocations in an arena with custom chunk_hooks. Passing MALLOCX_TCACHE_NONE to je_mallocx (our default) makes the failures intermittent, omitting it seems to make them much more likely.

[aligntest.c.txt](https://github.com/jemalloc/jemalloc/files/921085/aligntest.c.txt)

jasone:
[master-specific fix; the first paragraph describes a problem that also exists in dev.]

When allocating runs with alignment stricter than one page, commit after
trimming the head/tail from the initial over-sized allocation, rather
than before trimming.  This avoids creating clean-but-committed runs;
such runs do not get purged (and decommitted as a side effect), so they
can cause unnecessary long-term run fragmentation.

Do not commit decommitted memory in chunk_recycle() unless asked to by
the caller.  This allows recycled arena chunks to start in the
decommitted state, and therefore increases the likelihood that purging
after run deallocation will allow the arena chunk to become a single
unused run, thus allowing the chunk as a whole to be discarded.

This resolves #766.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -403,7 +403,7 @@ arena_run_split_remove(arena_t *arena, arena_chunk_t *chunk, size_t run_ind,
 	/* Keep track of trailing unused pages for later use. */
 	if (rem_pages > 0) {
 		size_t flags = flag_dirty | flag_decommitted;
-		size_t flag_unzeroed_mask = (flags == 0) ?  CHUNK_MAP_UNZEROED :
+		size_t flag_unzeroed_mask = (flags == 0) ? CHUNK_MAP_UNZEROED :
 		    0;
 
 		arena_mapbits_unallocated_set(chunk, run_ind+need_pages,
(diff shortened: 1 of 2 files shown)


jasone:
Integrated as 807a9a3e1751ea0f350a1484a70d86b257d69f62 on the newly created [stable-4](https://github.com/jemalloc/jemalloc/commits/stable-4) branch.

PR #779, opened 2017-04-19
Use openat syscall if available

darchons:
Some architectures like AArch64 may not have the `open` syscall because it
was superseded by the `openat` syscall, so check and use `SYS_openat` if
`SYS_open` is not available.

Additionally, Android headers for AArch64 define `SYS_open` to `__NR_open`,
even though `__NR_open` is undefined. Undefine `SYS_open` in that case so
`SYS_openat` is used.


davidtgoldblatt:
Thanks for the diff!

PR #784, opened 2017-04-21
Remove --disable-tcache.

### LINKED ISSUE #580: Reduce configure options

Candidates for removal:

jasone:
Simplify configuration by removing the --disable-tcache option, but
replace the testing for that configuration with
--with-malloc-conf=tcache:false.

Fix the thread.arena and thread.tcache.flush mallctls to work correctly
if tcache is disabled.

This partially resolves #580.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -70,7 +70,6 @@ CTL_PROTO(config_prof)
 CTL_PROTO(config_prof_libgcc)
 CTL_PROTO(config_prof_libunwind)
 CTL_PROTO(config_stats)
-CTL_PROTO(config_tcache)
 CTL_PROTO(config_tls)
 CTL_PROTO(config_utrace)
 CTL_PROTO(config_xmalloc)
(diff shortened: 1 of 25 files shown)


src/ctl.c
```diff
@@ -1536,12 +1528,9 @@ thread_arena_ctl(tsd_t *tsd, const size_t *mib, size_t miblen, void *oldp,
 		}
 		/* Set new arena/tcache associations. */
 		arena_migrate(tsd, oldind, newind);
-		if (config_tcache) {
-			tcache_t *tcache = tsd_tcachep_get(tsd);
-			if (tcache != NULL) {
-				tcache_arena_reassociate(tsd_tsdn(tsd), tcache,
-				    newarena);
-			}
+		if (opt_tcache) {
```
interwq: Should this check for tcache_enabled as well? i.e. && tcache_available(tsd)

Probably the non null check before was meant to do this but failed.

src/stats.c
```diff
@@ -128,20 +128,11 @@ stats_arena_bins_print(void (*write_cb)(void *, const char *), void *cbopaque,
 	} else {
 		char *mutex_counters = "   n_lock_ops    n_waiting"
 		    "   n_spin_acq  max_wait_ns\n";
-		if (config_tcache) {
```
interwq: 👍

src/ctl.c
jasone: Ah, it looks to me like the right thing to do is *replace* this check with ```if (tcache_available(tsd)) { ... }```, here, in ```thread_tcache_flush_ctl()```, and the assert in ```tcache_flush()```.  Does that sound right to you?

interwq: Yes that's right. tcache_available incorporates opt_tcache and should be the determining factor for auto tcache. Searched through the code a bit and the changes you mentioned sound complete to me.

jasone:
For future reference: The thread.arena and thread.tcache.flush bugs do not exist in 4.x, so nothing from this change needs to be backported.

Issue #211, opened 2015-03-17
It's not possible to enable ivsalloc independently of debug anymore


glandium:
As per src/zone.c, ivsalloc is needed for Mac.

In Firefox nightlies, we have some facility around the allocator that allows to replace/hook the allocator without recompiling Firefox.
With that facility, we build jemalloc in two different ways, none of which can rely on zone.c from jemalloc. So for all that to work, we do need malloc_usable_size to call ivsalloc instead of isalloc, and we definitely don't want debug builds just for that.

glandium:
Thanks.
FWIW, I was about to send a PR with a patch limited to the changes you made to include/jemalloc/internal/jemalloc_internal.h.in and src/jemalloc.c (with config_debug || config_ivsalloc instead of config_ivsalloc alone), so that manually adding -DJEMALLOC_IVSALLOC when compiling would do it, but without exposing a configure option to do it.

PR #786, opened 2017-04-21
Remove --enable-ivsalloc.

### LINKED ISSUE #580: Reduce configure options

jasone:
Continue to use ivsalloc() when --enable-debug is specified (and add
assertions to guard against 0 size), but stop providing a documented
explicit semantics-changing band-aid to dodge undefined behavior in
sallocx() and malloc_usable_size().  ivsalloc() remains compiled in,
unlike when #211 restored --enable-ivsalloc, and if
JEMALLOC_FORCE_IVSALLOC is defined during compilation, sallocx() and
malloc_usable_size() will still use ivsalloc().

This partially resolves #580.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2678,12 +2678,14 @@ je_sallocx(const void *ptr, int flags) {
 	tsdn_t *tsdn;
 
 	assert(malloc_initialized() || IS_INITIALIZER);
+	assert(ptr != NULL);
 
 	tsdn = tsdn_fetch();
 	witness_assert_lockless(tsdn);
 
-	if (config_ivsalloc) {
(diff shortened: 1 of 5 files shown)


src/jemalloc.c
```diff
@@ -2885,10 +2887,15 @@ je_malloc_usable_size(JEMALLOC_USABLE_SIZE_CONST void *ptr) {
 	tsdn = tsdn_fetch();
 	witness_assert_lockless(tsdn);
 
-	if (config_ivsalloc) {
-		ret = ivsalloc(tsdn, ptr);
+	if (ptr == NULL) {
+		ret = 0;
```
interwq: nit: unlikely

PR #787, opened 2017-04-21
Enable -Wundef on platforms where it's supported.

davidtgoldblatt:
This can catch bugs in which one header defines a numeric constant, and another
uses it without including the defining header. Undefined preprocessor symbols
expand to '0', so that this will compile fine, silently doing the math wrong.

### DIFF
--- a/src/nstime.c
+++ b/src/nstime.c
@@ -96,7 +96,7 @@ nstime_get(nstime_t *time) {
 
 	nstime_init(time, ticks_100ns * 100);
 }
-#elif JEMALLOC_HAVE_CLOCK_MONOTONIC_COARSE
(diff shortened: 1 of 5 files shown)


src/nstime.c
```diff
@@ -114,7 +114,7 @@ nstime_get(nstime_t *time) {
 	clock_gettime(CLOCK_MONOTONIC, &ts);
 	nstime_init2(time, ts.tv_sec, ts.tv_nsec);
 }
-#elif JEMALLOC_HAVE_MACH_ABSOLUTE_TIME
+#elif defined(JEMALLOC_HAVE_MACH_ABSOLUTE_TIME)
```
interwq: Should we switch the previous few branches as well?

davidtgoldblatt: Yep, done.

PR #788, opened 2017-04-21
Remove --with-lg-size-class-group.

### LINKED ISSUE #580: Reduce configure options

jasone:
Four size classes per size doubling has proven to be a universally good
choice for the entire 4.x release series, so there's little point to
preserving this configurability.

This partially resolves #580.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -220,35 +220,6 @@ any of the following arguments (not a definitive list) to 'configure':
     when cross compiling, or when overriding the default for systems that do
     not explicitly support huge pages.
 
---with-lg-size-class-group=<lg-size-class-group>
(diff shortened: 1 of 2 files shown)


interwq:
I'm not entirely sure about removing this. I do believe the 4 sizes per doubling is a good default. However I'm pretty interested in bin sizes not multiple of cacheline size, e.g 160, but was never able to really experiment since it's a compile time option.

jasone:
This diff doesn't remove the core ability to change the size class spacing; it merely removes the explicit configure option, which we don't test, and don't currently have compelling uses for.  I want to remove the option in order to simplify the CMake transition, but I don't think it limits our options re: future experimentation.

PR #795, opened 2017-04-25
Replace --disable-munmap with opt.munmap.

### LINKED ISSUE #580: Reduce configure options

jasone:
Control use of munmap(2) via a run-time option rather than a
compile-time option (with the same per platform default).  The old
behavior of --disable-munmap can be achieved with
--with-malloc-conf=munmap:false.

This partially resolves #580.


src/jemalloc.c
```diff
@@ -958,9 +958,7 @@ malloc_conf_init(void) {
 					    "Invalid conf value",	\
 					    k, klen, v, vlen);		\
 				}					\
-				if (cont) {				\
```
interwq: 👍

PR #796, opened 2017-04-25
Further header refactoring

davidtgoldblatt:
This gets the last few "easy" headers out of the catch-all. Everything remaining requires some nontrivial refactoring to make hermetic.

Of interest is the ctl diff. It pulls out the mutex profiling data and operations into their own module. This is more or less how I plan to deal with the harder refactorings, so it's an easy case to pick out any objects you might have.


davidtgoldblatt:
There's a couple spurious failures caused by an accidental push to upstream on my part. Ignore the "appveyor/branch" and "travis-ci/push" builds.

davidtgoldblatt:
@interwq, I'm going to push so I can start working on the more substantial refactorings that will be based on this, but please do take a look at the ctl commit.

PR #797, opened 2017-04-25
Avoid prof_dump during reentrancy.

interwq:


jasone:
I'm not sure this preserves documented behavior.  I'd like to take a closer look at the effects before we commit this.

src/prof.c
```diff
@@ -1756,6 +1756,9 @@ prof_fdump(void) {
 		return;
 	}
 	tsd = tsd_fetch();
+	if (tsd->reentrancy_level > 0) {
```
jasone: I don't think it's possible to trigger this, because we only call this function via an ```atexit(3)``` callback.  An assertion should be sufficient.

src/prof.c
```diff
@@ -1818,6 +1825,9 @@ prof_mdump(tsd_t *tsd, const char *filename) {
 	if (!opt_prof || !prof_booted) {
 		return true;
 	}
+	if (tsd->reentrancy_level > 0) {
```
jasone: For this to trigger, the application would have to make a reentrant manual dump call, which is way outside supported behavior. An assertion should be sufficient.

PR #799, opened 2017-04-25
Remove redundant extent lookup in tcache_bin_flush_large.

interwq:


interwq:
Yeah this was a surprise to me too.

PR #800, opened 2017-04-25
Use trylock in tcache_flush when possible

interwq:
In a production workload, 
```free -> je_tcache_bin_flush_small -> mutex_lock``` and 
```tcache_event_hard -> tcache_bin_flush_small -> mutex_lock``` paths are ranked 2nd and 3rd hottest among all mutex related paths. This change tries to use trylock to make it non-blocking whenever possible, similar to what we did for decay_lock.

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -45,14 +45,16 @@ tcache_event_hard(tsd_t *tsd, tcache_t *tcache) {
 	} else {
 		tbin = tcache_large_bin_get(tcache, binind);
 	}
+	bool repeat_bin;
 	if (tbin->low_water > 0) {
 		/*
 		 * Flush (ceiling) 3/4 of the objects below the low water mark.
 		 */
+		unsigned nflushed;
 		if (binind < NBINS) {
-			tcache_bin_flush_small(tsd, tcache, tbin, binind,
-			    tbin->ncached - tbin->low_water + (tbin->low_water
-			    >> 2));
+			nflushed = tcache_bin_try_flush_small(tsd, tcache, tbin,
+			    binind, tbin->ncached - tbin->low_water +
+			    (tbin->low_water >> 2));
 			/*
 			 * Reduce fill count by 2X.  Limit lg_fill_div such that
 			 * the fill count is always at least 1.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/tcache.c [hunk 1]
```diff
@@ -201,21 +241,33 @@ tcache_bin_flush_large(tsd_t *tsd, tcache_bin_t *tbin, szind_t binind,
...
+			}
+		}
 
-		malloc_mutex_lock(tsd_tsdn(tsd), &locked_arena->large_mtx);
-		for (unsigned i = 0; i < nflush; i++) {
+		for (unsigned i = nskipped; i < nflush; i++) {
 			void *ptr = *(tbin->avail - 1 - i);
 			assert(ptr != NULL);
-			extent = iealloc(tsd_tsdn(tsd), ptr);
+			extent = item_extent[i];
```
interwq: This was fixed separately in #799. Will rebase past that before landing.

src/tcache.c
```diff
@@ -116,9 +122,11 @@ tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, tcache_bin_t *tbin,
 		item_extent[i] = iealloc(tsd_tsdn(tsd), *(tbin->avail - 1 - i));
 	}
 
+	bool merged_stats = false;
+	unsigned nflushed = 0, nskipped = 0;
```
jasone: Style nit: For declarations incorporating initialization, only declare one variable per statement.

src/tcache.c
```diff
@@ -169,26 +187,48 @@ tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, tcache_bin_t *tbin,
...
-		malloc_mutex_lock(tsd_tsdn(tsd), &bin->lock);
-		bin->stats.nflushes++;
-		bin->stats.nrequests += tbin->tstats.nrequests;
-		tbin->tstats.nrequests = 0;
-		malloc_mutex_unlock(tsd_tsdn(tsd), &bin->lock);
+		if (must_flush) {
+			malloc_mutex_lock(tsd_tsdn(tsd), &bin->lock);
+		}
+		if (must_flush ||
+		    !malloc_mutex_trylock(tsd_tsdn(tsd), &bin->lock)) {
```
jasone: Subtle logic here.  Add a ```malloc_mutex_assert_owner()``` to make it obvious that the mutex is owned inside the block?

interwq:
Only style changes were made. Merging.

PR #803, opened 2017-04-27
Header refactoring: tsd - cleanup and dependency breaking.

davidtgoldblatt:
This removes the tsd macros (which are used only for tsd_t in real builds).  We
break up the circular dependencies involving tsd.

We also move all tsd access through getters and setters.  This allows us to
assert that we only touch data when tsd is in a valid state.

We simplify the usages of the x macro trick, removing all the customizability
(get/set, init, cleanup), moving the lifetime logic to tsd_init and tsd_cleanup.
This lets us make initialization order independent of order within tsd_t.


include/jemalloc/internal/tsd.h
```diff
@@ -0,0 +1,305 @@
...
+#include "jemalloc/internal/tsd_win.h"
+#else
+#include "jemalloc/internal/tsd_generic.h"
+#endif
+
+/*
+ * tsd_foop_get_unsafe(tsd) returns a pointer to the thread-local instance of
+ * foo.  This omits some safety checks, and so can be used during tsd
+ * bootstrapping and initialization.
+ */
```
interwq: 👍

src/tsd.c
```diff
@@ -122,20 +144,17 @@ tsd_cleanup(void *arg) {
 	case tsd_state_nominal:
 	case tsd_state_nominal_slow:
 	case tsd_state_reincarnated:
+		prof_tdata_cleanup(tsd);
+		iarena_cleanup(tsd);
+		arena_cleanup(tsd);
+		arenas_tdata_cleanup(tsd);
+		tcache_cleanup(tsd);
+		witnesses_cleanup(tsd);
```
interwq: nit: put these into a tsd_cleanup function?

include/jemalloc/internal/tsd.h
```diff
@@ -0,0 +1,305 @@
...
+#elif (defined(_WIN32))
+#include "jemalloc/internal/tsd_win.h"
+#else
+#include "jemalloc/internal/tsd_generic.h"
+#endif
+
+/*
+ * tsd_foop_get_unsafe(tsd) returns a pointer to the thread-local instance of
+ * foo.  This omits some safety checks, and so can be used during tsd
+ * bootstrapping and initialization.
```
davidtgoldblatt: TODO: should be "initialization and cleanup"

src/tsd.c
davidtgoldblatt: That's where this is; wdym?

interwq: Oh I mean to put these *_cleanup functions in a separate static function, something like tsd_members_cleanup

include/jemalloc/internal/tsd.h
```diff
@@ -0,0 +1,299 @@
...
+ * Thread-Specific-Data layout
+ * --- data accessed on tcache fast path: state, rtree_ctx, stats, prof ---
+ * s: state
+ * e: tcache_enabled
+ * m: thread_allocated (config_stats)
+ * f: thread_deallocated (config_stats)
+ * p: prof_tdata (config_prof)
+ * c: rtree_ctx (rtree cache accessed on deallocation)
+ * t: tcache
+ * --- data not accessed on tcache fast path: arena related fields ---
```
jasone: s/arena related/arena-related/

include/jemalloc/internal/tsd.h
```diff
@@ -0,0 +1,299 @@
...
+/* tsd_foo_set(tsd, val) updates the thread-local instance of foo to be val. */
+#define O(n, t)								\
+JEMALLOC_ALWAYS_INLINE void						\
+tsd_##n##_set(tsd_t *tsd, t val) {					\
+	*tsd_##n##p_get(tsd) = val;					\
+}
+MALLOC_TSD
+#undef O
+
+JEMALLOC_ALWAYS_INLINE bool
```
jasone: This function should return ```void```.  (Pre-existing issue.)

Issue #807, opened 2017-04-28
Investigate trylock in tcache_dalloc


interwq:
The first attempt to use trylock in tcache_dalloc (part of #800) resulted in a significant CPU & memory regression under a production workload. However, stats showed that # of bytes cached in tcache hardly increased, and were far from the amount of rss regressed. @jasone pointed out that purging looked less effective, e.g. extents purged were smaller. This indicates the regression is likely caused by fragmentation, or even false sharing if it's a remote arena.

A few things to try:
1) if it's a remote arena, don't use trylock;
2) maybe in flush logic, find out items from remote arena, and flush them first;
3) upon trylock failure, instead of move on to the next item, move all items from this arena to the bottom of the bin.
4) instead of trylock once, do a spin loop (per arena);

PR #809, opened 2017-04-28
Revert "Use trylock in tcache_bin_flush when possible."

interwq:
This reverts commit 8584adc451f31adfc4ab8693d9189cf3a7e5d858.  Production
results not favorable.  Will investigate separately.


interwq:
@jasone : I did some benchmark with this diff alone, and surprisingly it's not helping as well (CPU looks flat, memory increase).  I'm going to revert this diff for now, and leave it to later (in #807).

PR #810, opened 2017-05-02
Make VS2015 project work again

rustyx:
Sync the MSVC project with the latest source file changes (mb.c, util.c, malloc_io.c)


davidtgoldblatt:
Accepting since this is a clear improvement, though hopefully it won't be useful for long, as the CMake transition will be just around the corner. Thanks for the diff!

PR #818, opened 2017-05-04
Automatically mangle private symbols

jasone:


jasone:
So much fail. :disappointed:

jasone:
No more fail!  Ready for review.

jasone:
Mmm, awk.  I experienced a strange delight while learning and using a rather ancient language which inspired others I have used for *decades*.

tamird:
Unfortunately the way "nm" is hardcoded here means that this breaks under cross compilation.

jasone:
@tamird, how can we fix it?  Add an overrideable configure variable?

tamird:
@benesch and I are working on a fix - we'll send a PR.

PR #833, opened 2017-05-12
Take extents mutex around the extent_evict loops.

interwq:
Extents mutexes are the hottest locks in a prod workload.  Instead of taking the
extents mutex multiple times for evictions, this changes it to take it only once
for the extents_evict loop.  Since the eviction during decay is guarded by the
decay mutex, we do not expect heavy contention on this mutex here.  Plus we
avoid any expensive operations (such as madvise) while holding the extents lock.
Reduce # of lock ops to lower overhead on the mutex.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -860,11 +860,15 @@ arena_stash_decayed(tsdn_t *tsdn, arena_t *arena,
 	/* Stash extents according to npages_limit. */
 	size_t nstashed = 0;
 	extent_t *extent;
+
+	malloc_mutex_lock(tsdn, &extents->mtx);
 	while ((extent = extents_evict(tsdn, arena, r_extent_hooks, extents,
 	    npages_limit)) != NULL) {
 		extent_list_append(decay_extents, extent);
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
Add a commit description? It's not obvious to me why this is necessary (are we racy in the list append?).

interwq:
Will add the description. The list append is to a local stack list during arena destroy (destroying all the extents).

interwq:
In fact, I'm not 100% convinced this is a good idea. Decay could evict a bunch of extents together, so under no contention this is better because of less lock ops. But when there is indeed contention on the extents, this will cause the extents lock to be held longer in one pass. The reason I wanted to change was because I observed very low contention rate (0.07%) under prod workload, but very high number of operations (highest among all mutex).

interwq:
Since there is tradeoff going on with this change, and we don't have a clear indication whether one way is better than the other, I'm going to close this. I don't expect this to help much anyway, so no plan to investigate further for now.

PR #836, opened 2017-05-16
Refactor rtree locking into the extent code.

### LINKED ISSUE #699: Refactor away rtree locking

davidtgoldblatt:
This hypothetically fixes a priority inversion bug, and simplifies the rtree code and the extent/rtree interactions. More details in the commit messages.

This resolves #699.


davidtgoldblatt:
I think we can use a variant of this strategy to cut down most of our contention around extents->mutex, but our locking strategy is a little ill-specified for me to be confident in doing that along with this. I'll take another look after some further code cleanup efforts.

src/extent.c
```diff
@@ -95,6 +96,59 @@ static void extent_record(tsdn_t *tsdn, arena_t *arena,
...
+
+static lock_result_t
+extent_rtree_leaf_elm_try_lock(tsdn_t *tsdn, rtree_leaf_elm_t *elm,
+    extent_t **result) {
+	*result = NULL;
+
+	extent_t *extent1 = rtree_leaf_elm_extent_read(tsdn, &extents_rtree,
+	    elm, true);
+
+	if (*result == NULL) {
```
jasone:
```diff
diff --git a/src/extent.c b/src/extent.c
index 5c8b699..e53f704 100644
--- a/src/extent.c
+++ b/src/extent.c
@@ -105,12 +105,9 @@ typedef enum {
 static lock_result_t
 extent_rtree_leaf_elm_try_lock(tsdn_t *tsdn, rtree_leaf_elm_t *elm,
     extent_t **result) {
-       *result = NULL;
-
        extent_t *extent1 = rtree_leaf_elm_extent_read(tsdn, &extents_rtree,
            elm, true);
-
-       if (*result == NULL) {
+       if (extent1 == NULL) {
                return lock_result_no_extent;
        }
        /*
```

src/extent.c
```diff
@@ -508,49 +562,36 @@ extent_activate_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents,
 }
 
 static bool
-extent_rtree_acquire(tsdn_t *tsdn, rtree_ctx_t *rtree_ctx,
+extent_rtree_leaf_elms_lookup(tsdn_t *tsdn, rtree_ctx_t *rtree_ctx,
     const extent_t *extent, bool dependent, bool init_missing,
     rtree_leaf_elm_t **r_elm_a, rtree_leaf_elm_t **r_elm_b) {
-	*r_elm_a = rtree_leaf_elm_acquire(tsdn, &extents_rtree, rtree_ctx,
+	*r_elm_a = NULL;
```
jasone: Redundant initialization is brittle in my experience, and almost never necessary.
```diff
diff --git a/src/extent.c b/src/extent.c
index 5c8b699..9b2e715 100644
--- a/src/extent.c
+++ b/src/extent.c
@@ -565,9 +565,6 @@ static bool
 extent_rtree_leaf_elms_lookup(tsdn_t *tsdn, rtree_ctx_t *rtree_ctx,
     const extent_t *extent, bool dependent, bool init_missing,
     rtree_leaf_elm_t **r_elm_a, rtree_leaf_elm_t **r_elm_b) {
-       *r_elm_a = NULL;
-       *r_elm_b = NULL;
-
        *r_elm_a = rtree_leaf_elm_lookup(tsdn, &extents_rtree, rtree_ctx,
            (uintptr_t)extent_base_get(extent), dependent, init_missing);
        if (!dependent && *r_elm_a == NULL) {
@@ -580,7 +577,7 @@ extent_rtree_leaf_elms_lookup(tsdn_t *tsdn, rtree_ctx_t *rtree_ctx,
        if (!dependent && *r_elm_b == NULL) {
                return true;
        }
-       assert(*r_elm_a != NULL);
+       assert(*r_elm_b != NULL);
 
        return false;
 }
```

davidtgoldblatt: Done.

src/extent.c
```diff
@@ -95,6 +96,59 @@ static void extent_record(tsdn_t *tsdn, arena_t *arena,
...
+typedef enum {
+	lock_result_success,
+	lock_result_failure,
+	lock_result_no_extent
+} lock_result_t;
+
+static lock_result_t
+extent_rtree_leaf_elm_try_lock(tsdn_t *tsdn, rtree_leaf_elm_t *elm,
+    extent_t **result) {
+	*result = NULL;
```
davidtgoldblatt: Urgh, I'm dumb. Thanks for the early review on this, I must have spent a couple days looking for this, and could easily have spent a couple days more.

include/jemalloc/internal/rtree_inlines.h
```diff
@@ -61,7 +56,7 @@ JEMALLOC_ALWAYS_INLINE extent_t *
 rtree_leaf_elm_bits_extent_get(uintptr_t bits) {
 	/* Restore sign-extended high bits, mask slab and lock bits. */
```
davidtgoldblatt: Done.

include/jemalloc/internal/mutex_pool_structs.h
```diff
@@ -0,0 +1,14 @@
+#ifndef JEMALLOC_INTERNAL_MUTEX_POOL_STRUCTS_H
+#define JEMALLOC_INTERNAL_MUTEX_POOL_STRUCTS_H
+
+/* This file really combines "structs" and "types", but only transitionally. */
+
+/* We do mod reductions by this value, so it should be kept a power of 2. */
+#define MUTEX_POOL_SIZE 256
```
davidtgoldblatt: I figured that for something at the same level of the stack as a mutex, having it not allocate (or at least, not allocate any more than a malloc_mutex_t) would be useful, so I just picked a number sufficiently large that I assumed no real contention ever. Are you thinking to just get the memory from ialloc?

interwq: If handling allocation is a pain, we can consider defining a large array and just initialize the first N of them depending on ncpus (assuming we don't use huge page for them).

davidtgoldblatt: I mean, no sense in *not* using them all if we've got them.

jasone: I was more thinking about a time in the future when 256 mutexes isn't enough.  I can easily imagine us having to manually bump this constant up over time.

PR #842, opened 2017-05-17
Refactor *decay_time into *decay_ms.

### LINKED ISSUE #812: Consider adding sub-second resolution for decay time options
The two-phase decay-based purging mechanism could benefit from a short initial phase, in which case whole-second resolution may be inadequate.

jasone:
Support millisecond resolution for decay times.  Among other use cases
this makes it possible to specify a short initial dirty-->muzzy decay
phase, followed by a longer muzzy-->clean decay phase.

This resolves #812.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -18,11 +18,11 @@ const char	*percpu_arena_mode_names[] = {
 const char	*opt_percpu_arena = OPT_PERCPU_ARENA_DEFAULT;
 percpu_arena_mode_t	percpu_arena_mode = PERCPU_ARENA_MODE_DEFAULT;
 
-ssize_t opt_dirty_decay_time = DIRTY_DECAY_TIME_DEFAULT;
-ssize_t opt_muzzy_decay_time = MUZZY_DECAY_TIME_DEFAULT;
+ssize_t opt_dirty_decay_ms = DIRTY_DECAY_MS_DEFAULT;
+ssize_t opt_muzzy_decay_ms = MUZZY_DECAY_MS_DEFAULT;
 
-static atomic_zd_t dirty_decay_time_default;
-static atomic_zd_t muzzy_decay_time_default;
+static atomic_zd_t dirty_decay_ms_default;
+static atomic_zd_t muzzy_decay_ms_default;
 
 const arena_bin_info_t	arena_bin_info[NBINS] = {
 #define BIN_INFO_bin_yes(reg_size, slab_size, nregs)			\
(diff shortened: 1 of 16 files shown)


jasone:
The strategy I originally had in mind would have allowed e.g. ```MALLOC_CONF=dirty_decay_time:0.123```, but that proved quite messy when it came to exposing the settings via mallctl.

src/arena.c
```diff
@@ -744,11 +743,12 @@ arena_decay_init(arena_decay_t *decay, extents_t *extents, ssize_t decay_time,
...
 static bool
-arena_decay_time_valid(ssize_t decay_time) {
-	if (decay_time < -1) {
+arena_decay_ms_valid(ssize_t decay_ms) {
+	if (decay_ms < -1) {
 		return false;
 	}
-	if (decay_time == -1 || (uint64_t)decay_time <= NSTIME_SEC_MAX) {
+	if (decay_ms == -1 || (uint64_t)decay_ms <= NSTIME_SEC_MAX *
+	    KQU(1000)) {
```
interwq: nit: make it decay_ms / 1000 to avoid overflow risks.

src/jemalloc.c
```diff
@@ -1070,10 +1070,14 @@ malloc_conf_init(void) {
 			}
 			CONF_HANDLE_UNSIGNED(opt_narenas, "narenas", 1,
 			    UINT_MAX, yes, no, false)
-			CONF_HANDLE_SSIZE_T(opt_dirty_decay_time,
-			    "dirty_decay_time", -1, NSTIME_SEC_MAX);
-			CONF_HANDLE_SSIZE_T(opt_muzzy_decay_time,
-			    "muzzy_decay_time", -1, NSTIME_SEC_MAX);
+			CONF_HANDLE_SSIZE_T(opt_dirty_decay_ms,
+			    "dirty_decay_ms", -1, NSTIME_SEC_MAX * KQU(1000) <
```
interwq: nit: also use / 1000?

src/stats.c
```diff
@@ -857,24 +855,14 @@ stats_general_print(void (*write_cb)(void *, const char *), void *cbopaque,
 		malloc_cprintf(write_cb, cbopaque, "Arenas: %u\n", uv);
 	}
 
-	CTL_GET("arenas.dirty_decay_time", &ssv, ssize_t);
 	if (json) {
+		CTL_GET("arenas.dirty_decay_ms", &ssv, ssize_t);
 		malloc_cprintf(write_cb, cbopaque,
-		    "\t\t\t\"dirty_decay_time\": %zd,\n", ssv);
-	} else {
-		malloc_cprintf(write_cb, cbopaque,
```
interwq: Why stop outputting them in non-JSON format? I know some tools only read merged stats.

interwq:
The MSVC failure is real: 
src/jemalloc.c(1076): error C2065: 'SSIZE_MAX': undeclared identifier
src/jemalloc.c(1080): error C2065: 'SSIZE_MAX': undeclared identifier
mingw32-make: *** [Makefile:339: src/jemalloc.sym.obj] Error 2

src/arena.c
jasone: That's how I originally wrote it, but this is more precise (no rounding error), and it does not risk overflow because ```NSTIME_SEC_MAX``` is guaranteed to be at least 1 billion times smaller than the full range of ```uint64_t```.

src/stats.c
jasone: This is duplicated output in the human-readable format -- we also print this stat next to the associated ```opt.dirty_decay_ms``` setting, e.g.:

```
  opt.dirty_decay_ms: 10000 (arenas.dirty_decay_ms: 10000)
  opt.muzzy_decay_ms: 10000 (arenas.muzzy_decay_ms: 10000)
```

PR #845, opened 2017-05-22
Fix # of unpurged pages in decay algorithm.

### LINKED ISSUE #813: Decay algorithm adjustment
Discussed with @jasone offline about our current decay based purging implementation. We believe there is a bug causing the decay to be slower than it should, from this line:
https://github.com/jemalloc/jemalloc/blob/c86c8f4ffbf8c118203f7327610a2ad80cf9622c/src/arena.c#L699
Here we track the # of unpurged pages at the end of each epoch, and use it to determine the # of new dirty pages in the next epoch. However, if the dirty pages are being reused quickly, e.g. if they all get reused in the next epoch, then ```decay->nunpurged``` becomes 0 in the next epoch; and then in a future epoch they will all be counted as "new" dirty pages (in other words the same dirty pages will be double counted the next time they are generated). As a result, we will allow more dirty pages to be alive then we should.

The fix we discussed was, instead of tracking the ```nunpurged```, we track the ```npages_limit``` at the end of the epoch, and rely on that to determine new # of dirty pages.

One assertion we could make based on this fix:
```npages_limit <= peak_n_dirty_pages```

Either @jasone or I will be getting this fix, probably after #714, so that test it for both cases (tick-triggered and background threads driven).

### PR DESCRIPTION
**Pull request by interwq**
When # of dirty pages move below npages_limit (e.g. they are reused), we should
not lower number of unpurged pages because that would cause the reused pages to
be double counted in the backlog (as a result, decay happen slower than it
should).  Instead, set number of unpurged to the greater of current npages and
npages_limit.
    
Added an assertion: peak # of pages should be greater than npages_limit.

This resolves #813.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -639,6 +639,17 @@ arena_decay_backlog_update_last(arena_decay_t *decay, extents_t *extents) {
 	size_t ndirty_delta = (ndirty > decay->nunpurged) ? ndirty -
 	    decay->nunpurged : 0;
 	decay->backlog[SMOOTHSTEP_NSTEPS-1] = ndirty_delta;
+
+	if (config_debug) {
+		if (ndirty > decay->ceil_npages) {
+			decay->ceil_npages = ndirty;
+		}
+		size_t npages_limit = arena_decay_backlog_npages_limit(decay);
+		assert(decay->ceil_npages >= npages_limit);
+		if (decay->ceil_npages > npages_limit) {
+			decay->ceil_npages = npages_limit;
+		}
+	}
 }
 
 static void
@@ -664,11 +675,9 @@ arena_decay_backlog_update(arena_decay_t *decay, extents_t *extents,
 }
 
 static void
-arena_decay_try_purge(tsdn_t *tsdn, arena_t *arena,
-    arena_decay_t *decay, extents_t *extents) {
-	size_t npages_limit = arena_decay_backlog_npages_limit(decay);
-
-	if (extents_npages_get(extents) > npages_limit) {
+arena_decay_try_purge(tsdn_t *tsdn, arena_t *arena, arena_decay_t *decay,
+    extents_t *extents, size_t current_npages, size_t npages_limit) {
+	if (current_npages > npages_limit) {
 		arena_decay_to_limit(tsdn, arena, decay, extents, false,
 		    npages_limit);
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/background_thread.c [hunk 1]
```diff
@@ -264,7 +264,7 @@ background_work(tsdn_t *tsdn, unsigned ind) {
 		gettimeofday(&tv, NULL);
 		nstime_t before_sleep;
 		nstime_init2(&before_sleep, tv.tv_sec, tv.tv_usec * 1000);
-		
+
```
interwq: Will rebase to fix this.

interwq:
Only last commit is relevant here. This is on top of #714.

src/arena.c
```diff
@@ -637,6 +637,14 @@ arena_decay_backlog_update_last(arena_decay_t *decay, extents_t *extents) {
 	size_t ndirty_delta = (ndirty > decay->nunpurged) ? ndirty -
 	    decay->nunpurged : 0;
 	decay->backlog[SMOOTHSTEP_NSTEPS-1] = ndirty_delta;
+
+	if (config_debug) {
+		if (ndirty > decay->peak_npages) {
+			decay->peak_npages = ndirty;
+		}
+		assert(decay->peak_npages >=
```
jasone: This seems like an unnecessarily weak assertion.  Can we change ```peak_npages``` to be a ceiling, so that the assertion is relevant more of the time?

```C
if (config_debug) {
        if (ndirty > decay->ceil_npages) {
                decay->ceil_npages = ndirty;
        }
        size_t npages_limit =
            arena_decay_backlog_npages_limit(decay);
        assert(decay->ceil_npages >= npages_limit);
        if (decay->ceil_npages > npages_limit) {
                decay->ceil_npages = npages_limit;
        }
}
```

src/arena.c
```diff
@@ -700,16 +706,20 @@ static void
...
+		    current_npages, npages_limit);
 	}
-
 	/*
 	 * There may be concurrent ndirty fluctuation between the purge above
 	 * and the nunpurged update below, but this is inconsequential to decay
 	 * machinery correctness.
 	 */
-	decay->nunpurged = extents_npages_get(extents);
+	decay->nunpurged = npages_limit > current_npages ? npages_limit :
```
jasone: Parentheses around the conditional help readability.

src/arena.c
interwq: Good point. Will change.

interwq:
All tests passed previously. Landing.

PR #846, opened 2017-05-23
dss

jasone:


src/extent_dss.c
```diff
@@ -65,37 +67,41 @@ extent_dss_prec_set(dss_prec_t dss_prec) {
 	return false;
 }
 
+static void
+extent_dss_extending_start(void) {
+	spin_t spinner = SPIN_INITIALIZER;
```
interwq: Why not use mutex here?

jasone: There were bootstrapping issues in this code prior to switching away from mutexes.

davidtgoldblatt:
I don't have any specific objection to this, but I guess I question our sbrk strategy in general. Is the level of difficulty we're going through here just to avoid wasting virtual address space on highly aligned allocations?

PR #848, opened 2017-05-23
Disable percpu_arena by default.

jasone:
This is based on #845; only the last commit is relevant to this pull request.


interwq:
LGTM! I'll create an issue or reuse one to track the further investigation separately.

PR #849, opened 2017-05-23
Break up tsd->mutex->witness cycle

davidtgoldblatt:
This is one of the "tough" ones to do, in the sense that the dependency cycle here is real. It's broken by breaking the witness->tsd dependency; witness now doesn't need to see any structs or function definitions from the tsd module (it takes a thread-local, but witness-specific context arg instead of getting to see all of tsd, which I think is a nice thing to have in general).

It "cheats" a little bit, in that the tsd->mutex dependency is pushed down into tsd.c (so, e.g., we couldn't compile and test each module seperately, even after these commits), but that's a battle I'm willing to fight another day.


davidtgoldblatt:
Going to close and reopen this to try to debug the CI screwiness.

PR #855, opened 2017-05-24
Refactor run_tests to increase parallelism.

jasone:
Rather than relying on parallel make to build individual configurations
one at a time, use xargs to build multiple configurations in parallel.
This allows the configure scripts to run in parallel.  On a 14-core
system (28 hyperthreads), this increases average CPU utilization from
~20% to ~95%.

### DIFF
--- a/scripts/gen_run_tests.py
+++ b/scripts/gen_run_tests.py
@@ -1,14 +1,18 @@
 #!/usr/bin/env python
 
 from itertools import combinations
+from os import uname
+from multiprocessing import cpu_count
+
+nparallel = cpu_count() * 2
+
+uname = uname()[0]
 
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
Love this.

Can we tweak it so that we'll explicitly print an error message at the end if xargs exits with an error? Depending on how log spew race conditions work out, we might miss the error output message in a way that makes it look like we succeeded.

jasone:
Good idea; I use zsh and have it configured to print a message if the exit status is non-zero, so I notice errors.  Even better would be a message that says which config failed.

jasone:
A bit uglier code now, but a lot better behavior.

PR #856, opened 2017-05-24
Fix stats.mapped during deallocation.

interwq:
This partially solves #857 (the mapped part).


jasone:
The regression was caused by 64e458f5cdd64f9b67cb495f177ef96bf3ce4e0e.

PR #860, opened 2017-05-25
Abort on invalid config options when opt_abort is set.

### LINKED ISSUE #853: Add abort on invalid conf options.
Sometimes people incorrectly set the conf string w/o verifying or checking stdout warnings. Let's add the option to abort on invalid configs.

interwq:
This resolves #853.


jasone:
This needs to be protected by e.g. ```opt.abort_conf```, because as-is it will cause unwanted aborts for FreeBSD-CURRENT's libc.

interwq:
Updated w/ ```opt.abort_conf```.

PR #861, opened 2017-05-26
Extent grow next

jasone:


include/jemalloc/internal/witness.h
```diff
@@ -29,6 +29,7 @@
 #define WITNESS_RANK_PROF_TDATA		8U
 #define WITNESS_RANK_PROF_GCTX		9U
 
+#define WITNESS_RANK_EXTENT_GROW_NEXT	10U
```
interwq: Why not make it exclusive?

src/extent.c
```diff
@@ -1150,45 +1170,37 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
...
 extent_alloc_retained(tsdn_t *tsdn, arena_t *arena,
     extent_hooks_t **r_extent_hooks, void *new_addr, size_t size, size_t pad,
     size_t alignment, bool slab, szind_t szind, bool *zero, bool *commit) {
-	extent_t *extent;
-
 	assert(size != 0);
 	assert(alignment != 0);
 
-	extent = extent_recycle(tsdn, arena, r_extent_hooks,
+	malloc_mutex_lock(tsdn, &arena->extent_grow_next_mtx);
```
interwq: Is it true that this doesn't cause much additional contention because we need the extents lock in extent_recycle anyway (and grow rarely happen)?

src/extent.c
```diff
@@ -1150,45 +1170,37 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
...
 	    &arena->extents_retained, new_addr, size, pad, alignment, slab,
 	    szind, zero, commit);
 	if (extent != NULL) {
+		malloc_mutex_unlock(tsdn, &arena->extent_grow_next_mtx);
 		if (config_prof) {
 			extent_gdump_add(tsdn, extent);
 		}
-	}
-	if (opt_retain && extent == NULL) {
+	} else if (opt_retain) {
```
interwq: nit: document extent_grow_retained() does unlocking.

include/jemalloc/internal/witness.h
jasone: It seemed an arbitrary choice, so I went for diff minimization.

interwq: I think make it different could be helpful, especially given that background thread triggers all sorts of reentrancy behaviors.

jasone: I just refactored it to be a non-core mutex, which addressed this as a side effect.

src/extent.c
jasone: Good point.  The mutex covered a much larger code region than what ```arena->extents_retained.mtx``` covered, so I refactored the code to drop (renamed) ```arena->extent_grow_mtx``` as soon as all non-OOM failure possibilities are past.  This leaves the distant possibility of excessive retained memory on OOM, but if we're OOMing, we probably can't create more retained memory.

Also, ```arena->extent_grow_mtx``` should be a non-core mutex, so I fixed that.

src/extent.c
jasone: Done.

jasone:
There's a low-probability (~1/200 runs) failure on my dev machine with the following config:

```CC=gcc  CXX=g++  ./autogen.sh --enable-debug --enable-prof --with-malloc-conf=tcache:false --with-malloc-conf=dss:primary,percpu_arena:percpu```

interwq:
Any chance the test failure is caused by the dss code path?

jasone:
This failure appears to be a flaw in the failing assertion, rather than in the fix itself.  Each time we grow retained memory, there is the potential to create a bit of retained memory that cannot service subsequent 1-MiB requests.  The accumulation of this fragmentation across multiple retained memory growth iterations needs to be taken into account by the assertion.

As for how the fragmentation happens only with some probability, that is due to races with newly created threads to initialize auto arenas.  If the auto arenas' memory is sufficiently interspersed with the test arena's memory, enough fragmentation accumulates to push the assertion into incorrect behavior.

jasone:
The good news here is that the unit test is now much more precise than before.  The bad news is that I was wrong about being able to release ```extent_grow_mtx``` so early.  We have to hold the lock until all splitting and recording of remainder lead/trail extents is complete; otherwise we race with other allocating threads that don't know about the still-available memory in those lead/trail extents.

I refactored the code to pass a ```bool growing_retained``` parameter to all functions which may care about lock depth.  This maintains locking hygiene at the expense of extra verbosity.  We could avoid this verbosity by weakening various ```witness_assert_depth_to_rank()``` assertions, but my preference is maintain the higher precision.

PR #868, opened 2017-05-30
Explicitly say so when aborting on opt_abort_conf.

interwq:


src/jemalloc.c
```diff
@@ -878,13 +878,21 @@ malloc_conf_next(char const **opts_p, char const **k_p, size_t *klen_p,
 }
 
 static void
+malloc_abort_invalid_conf(void) {
+	assert(opt_abort_conf);
+	malloc_printf("<jemalloc>: Abort (opt_abort_conf:true) on invalid "
```
jasone: ```s/opt_abort_conf/abort_conf/```

PR #872, opened 2017-05-30
Pass the O_CLOEXEC flag to open(2).

### LINKED ISSUE #528: jemalloc opens /proc/sys/vm/overcommit_memory without O_CLOEXEC
src/pages.c has the following lines:

```
  #if defined(JEMALLOC_HAVE_SYSCALL) && defined(SYS_open)
          fd = (int)syscall(SYS_open, "/proc/sys/vm/overcommit_memory", O_RDONLY);
  #else
          fd = open("/proc/sys/vm/overcommit_memory", O_RDONLY);
  #endif
          if (fd == -1)
                  return (false); /* Error. */
```
To avoid file descriptor leakage within multithreaded programs, it's recommended that all open calls also include the O_CLOEXEC flag, eg:

`  fd = open("/proc/sys/vm/overcommit_memory", O_RDONLY | O_CLOEXEC);
`

Even if jemalloc doesn't have this particular threading issue, O_CLOEXEC is still recommended from a code cleanliness point of view.

jasone:
This resolves #528.

### DIFF
--- a/src/pages.c
+++ b/src/pages.c
@@ -353,12 +353,13 @@ os_overcommits_proc(void) {
 	ssize_t nread;
 
 #if defined(JEMALLOC_USE_SYSCALL) && defined(SYS_open)
-	fd = (int)syscall(SYS_open, "/proc/sys/vm/overcommit_memory", O_RDONLY);
+	fd = (int)syscall(SYS_open, "/proc/sys/vm/overcommit_memory", O_RDONLY |
+	    O_CLOEXEC);
 #elif defined(JEMALLOC_USE_SYSCALL) && defined(SYS_openat)
 	fd = (int)syscall(SYS_openat,
-	    AT_FDCWD, "/proc/sys/vm/overcommit_memory", O_RDONLY);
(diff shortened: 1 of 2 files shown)


rudreshahk:
This is causing the build to fail with the below error:
[user@host jemalloc-5.0.0]$ **make**
gcc -std=gnu99 -Wall -Wsign-compare -Wundef -pipe -g3 -fvisibility=hidden -O3 -funroll-loops -fPIC -DPIC -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -Iinclude -DJEMALLOC_NO_PRIVATE_NAMESPACE -o src/pages.sym.o src/pages.c
src/pages.c: In function ‘os_overcommits_proc’:
src/pages.c:357: error: ‘O_CLOEXEC’ undeclared (first use in this function)
src/pages.c:357: error: (Each undeclared identifier is reported only once
src/pages.c:357: error: for each function it appears in.)
make: *** [src/pages.sym.o] Error 1

PR #873, opened 2017-05-30
Output total_wait_ns for bin mutexes.

interwq:


interwq:
I found myself wondering about the average waiting time of bins locks. This adds the total wait time to the output. Now we only omit "# of owner switch" and "max # of thds" for bin mutexes.

interwq:
@jasone : are you referring to the code removed L224-L232, and L238-L245? These were dead code cleaned up. Do you want me to make them a separate commit?

PR #874, opened 2017-05-31
Track background thread status separately at fork.

interwq:
Use a separate boolean to track the enabled status, instead of leaving the
global background thread status inconsistent.

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -41,6 +41,8 @@ bool background_thread_stats_read(tsdn_t *tsdn,
 #undef NOT_REACHED
 #else
 
+static bool background_thread_enabled_at_fork;
+
 static void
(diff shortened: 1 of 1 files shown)


src/background_thread.c
```diff
@@ -16,6 +16,7 @@ bool opt_background_thread = BACKGROUND_THREAD_DEFAULT;
 malloc_mutex_t background_thread_lock;
 /* Indicates global state.  Atomic because decay reads this w/o locking. */
 atomic_b_t background_thread_enabled_state;
+UNUSED static bool background_thread_enabled_at_fork;
```
jasone: Maybe all the variables that are only used when ```JEMALLOC_BACKGROUND_THREAD``` is defined should just be in a ```#ifdef``` block.

PR #875, opened 2017-05-31
Use real pthread_create for creating background threads.

interwq:


src/mutex.c
```diff
@@ -30,7 +30,7 @@ static malloc_mutex_t	*postponed_mutexes = NULL;
 static void
 pthread_create_once(void) {
 	pthread_create_fptr = load_pthread_create_fptr();
-	isthreaded = true;
+	assert(isthreaded == true);
```
jasone: Simplify to ```assert(isthreaded);```.

jasone: Simplify to ```assert(isthreaded);```.

src/background_thread.c
```diff
@@ -640,9 +647,12 @@ load_pthread_create_fptr(void) {
 	if (pthread_create_fptr) {
 		return pthread_create_fptr;
 	}
-
+#ifdef JEMALLOC_LAZY_LOCK
+	isthreaded = true;
+#endif
 	pthread_create_fptr = dlsym(RTLD_NEXT, "pthread_create");
-	if (pthread_create_fptr == NULL) {
+	real_pthread_create = dlsym(RTLD_DEFAULT, "pthread_create");
```
jasone: This naming seems backwards as compared to the meaning used for "real" in the ```dlsym(3)``` manual page.  ```RTLD_DEFAULT``` finds the top symbol in the stack, i.e. the override, whereas ```RTLD_NEXT``` finds the "real" symbol that was overridden.  Given the naming mixup, I can't tell what the actual intention is here.

interwq: Yeah I realized that wasn't what I needed. Update soon.

src/background_thread.c
```diff
@@ -364,7 +364,11 @@ background_thread_create(tsd_t *tsd, unsigned arena_ind) {
 	pre_reentrancy(tsd);
 	int err;
 	load_pthread_create_fptr();
-	if ((err = pthread_create(&info->thread, NULL,
+	/*
+	 * To avoid complications (besides reentrancy), create internal
```
jasone: After having done some refactoring around this code, it's not obvious to me what the complications are that we're trying to avoid.  What problems did you hit that motivated this change?

interwq: Oh if the app defines it's own pthread_create, and we call it here, then we have a path this way: app => malloc => app (pthread_create) => malloc (called by pthread lib). I had this with HHVM and felt we should just cut the step in the middle. Does this make sense to you?

jasone: Ah, I think I understand.  This workaround only helps us if we don't have ```JEMALLOC_LAZY_LOCK``` defined, right?

interwq: Right. That's why I was trying to get "the bottom of the stack" symbol, but didn't find a way to do so. __real_pthread_create doesn't look widely available. Then I didn't find a way with dlopen / dlsym there.

PR #880, opened 2017-05-31
Pass *arena to ckh_new().

interwq:
Avoid using arena_choose / _ichoose() since it's used by prof_boot2().


interwq:
We have better (more general) solution so closing this one.

PR #878, opened 2017-05-31
Test with background_thread:true.

jasone:
Add testing for background_thread:true, and condition a xallocx() -->
rallocx() escalation assertion to allow for spurious in-place rallocx()
following xallocx() failure.

PR #881, opened 2017-05-31
Set reentrancy level to 1 during init.

interwq:
This makes sure we go down slow path w/ a0 in init.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1397,6 +1397,18 @@ malloc_init_hard_finish(void) {
 	return false;
 }
 
+static void
+malloc_init_hard_cleanup(tsdn_t *tsdn, bool reentrancy_set) {
+	malloc_mutex_assert_owner(tsdn, &init_lock);
+	malloc_mutex_unlock(tsdn, &init_lock);
+	if (reentrancy_set) {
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
I'm not opposed to this kind of thing necessarily, but we sort of deliberately decided not to make the reentrancy stuff a bootstrapping mechanism. If we're going to intermingle them, I think we should go all the way, and make bootstrapping happen just enough so that we can reentrantly allocate, and then do all the rest of our initialization relying only on the reentrancy allocator.

So I guess I think this is fine if it's a blocker for background threads, but if so we should try to do this "for real" next half, and if it's doable just by reordering a0's background thread's creation, we should at least consider it.

interwq:
Yeah this can be done more throughly. However as a simple first step this will be helpful at a very low cost & complexity. In particular this helps percpu_arena (which will feel free to pick arenas w/o setting reentrancy). The percpu_arena + background_thread issue @jasone found could be simply avoided this way, and we thought combining this w/ https://github.com/jemalloc/jemalloc/pull/878/commits/5d59a4e76b42478264af53e447fdaab773123a15 would be a useful thing to do.

Since there is no real down side with the simple approach, I'd say let's do this first, and only go further if we still get complications after this.

PR #882, opened 2017-06-01
Fix background thread creation.

interwq:
The state initialization should be done before pthread_create.


interwq:
Only the top commit is relevant. This is based on #878.

I initially had the pthread_create covered by the thread mutex, but then took it out because of lock ranking w/ reentrancy issues. That part was left incorrect and could cause the newly created thread exits immediately.

jasone:
Committed as c84ec3e9da66162943ee33afe73c7f898fa134e2 to avoid the merge commit that would have resulted from using the GitHub GUI.

PR #884, opened 2017-06-01
Set isthreaded when enabling background_thread.

interwq:


interwq:
@jasone : by the time we reach ```pthread_create_wrapper()```, we are holding background_thread_mutex already. If the ```isthreaded``` is set after we took the lock (but before we unlock), we will be unlocking a lock never taken.

jasone:
Okay, can we refactor this code a bit to make it less brittle?  Perhaps with some minor changes ```background_thread_enabled_set()``` can do the work of setting ```isthreaded```, using the same ```pthread_once()``` machinery as already exists.

interwq:
Well, we will have to set ```isthreaded``` w/o taking background_thread_lock. While the first thing we do in background_thread_ctl is to take that mutex. That's why I ended up setting it there, as we only have two ways of enable it (default with malloc_conf, or toggle with mallctl).

Adopting pthread_once in ctl is probably not best. Alternatively, we can have a separate background_thread interface to do the check, if you think that may look better.

jasone:
My main concern with this patch is that code motion during future refactoring will be fraught with peril.  The dependency is not currently documented, nor is it coupled by any API.  Ideally we'd structure the code such that it's not too easy to break accidentally.

interwq:
@jasone : I refactored the code to move the logic into ```background_thread_ctl_init()```. Does this look alright to you?

PR #890, opened 2017-06-07
Make tsd no-cleanup during tsd reincarnation.

interwq:
Since tsd cleanup isn't guaranteed when reincarnated, we set up tsd in a way
that needs no cleanup, by making it going through slow path instead.

### DIFF
--- a/src/tsd.c
+++ b/src/tsd.c
@@ -63,6 +63,45 @@ tsd_slow_update(tsd_t *tsd) {
 	}
 }
 
+static bool
+tsd_data_init(tsd_t *tsd) {
+	/*
+	 * We initialize the rtree context first (before the tcache), since the
+	 * tcache initialization depends on it.
+	 */
+	rtree_ctx_data_init(tsd_rtree_ctxp_get_unsafe(tsd));
+
+	return tsd_tcache_enabled_data_init(tsd);
+}
+
+static void
+assert_tsd_data_cleanup_done(tsd_t *tsd) {
+	assert(!tsd_nominal(tsd));
+	assert(*tsd_arenap_get_unsafe(tsd) == NULL);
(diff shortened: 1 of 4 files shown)


src/tsd.c
```diff
@@ -63,6 +63,48 @@ tsd_slow_update(tsd_t *tsd) {
...
+	if (tsd_tcache_enabled_data_init(tsd)) {
+		return true;
+	}
+	return false;
+}
+
+static void
+assert_tsd_data_cleanup_done(tsd_t *tsd) {
+	assert(!tsd_nominal(tsd));
+	assert(*tsd_arenap_get_unsafe(tsd) == NULL);
```
jasone: It sorta feels like these assertions should be generated via X-macros, but we don't currently have the necessary column in ```MALLOC_TSD``` to support it.

src/tsd.c
```diff
@@ -63,6 +63,48 @@ tsd_slow_update(tsd_t *tsd) {
...
 
+static bool
+tsd_data_init(tsd_t *tsd) {
+	/*
+	 * We initialize the rtree context first (before the tcache), since the
+	 * tcache initialization depends on it.
+	 */
+	rtree_ctx_data_init(tsd_rtree_ctxp_get_unsafe(tsd));
+
+	if (tsd_tcache_enabled_data_init(tsd)) {
```
jasone: ```return tsd_tcache_enabled_data_init(tsd);```

src/tsd.c
interwq: Yeah it's possible to use x-macros, but might not be pretty. We need 2 columns: ```have_cleanup_val``` (some members like embedded tcache don't have a cleanup state), and ```cleanup_val```.  Also here we may want to assert something more intrusive, e.g. ```tcache.bins[0].avail == NULL```

interwq:
Tests passed previously. Merging.

PR #897, opened 2017-06-07
Fix background_thread locking and tsd.

interwq:


src/arena.c
```diff
@@ -2054,9 +2054,11 @@ arena_new(tsdn_t *tsdn, unsigned ind, extent_hooks_t *extent_hooks) {
 		/* background_thread_create() handles reentrancy internally. */
 		if (have_background_thread) {
 			bool err;
+			malloc_mutex_unlock(tsdn, &arenas_lock);
```
jasone: This doesn't look safe to me.  If we really need to avoid owning this lock during background thread creation, we need to do some refactoring that still allows ```arenas_lock``` to protect ```arenas``` writes.

interwq: Actually ```arena_set(ind, arena);``` is done before this (I changed that as part of a previous refactor). After we return from this function, we pretty much unlock immediately.

jasone: Okay, can we dramatically reduce the range of code executed with ```arenas_lock``` held, so that we don't hold it by the time we create the background thread, and we therefore don't need to reacquire it?

interwq: I checked through the code path however it didn't seem quite possible. We need to protect ```arena_choose_hard()``` which walks all arenas to make the decision. How about I move the ```background_thread_create()``` around so that it's outside of arenas_lock's scope?

interwq:
@jasone : just updated -- now background_thread_create is called after we release arenas_lock.

PR #906, opened 2017-06-13
Fix extent_hooks in extent_grow_retained().

interwq:
Fix extent_hooks in extent_grow_retained().
    
This issue caused the default extent alloc function to be incorrectly
used even when ```arena.<i>.extent_hooks``` is set.  This bug was introduced
by 411697adcda2fd75e135cdcdafb95f2bd295dc7f (Use exponential series to
size extents.), which was first released in 5.0.0.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -1066,9 +1066,18 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
 	}
 	bool zeroed = false;
 	bool committed = false;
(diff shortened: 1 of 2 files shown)


interwq:
@jasone: yes just updated (the message looks a bit strange on github because it hyperlinks the hash).

interwq:
Only commit message was updated after tests passed. Merging.

Issue #907, opened 2017-06-14
5.0.0: Error in dlsym(RTLD_NEXT, "pthread_create")


Safari77:
jemalloc 4.5.0 worked, but not 5.0.0.

```
$ LD_DEBUG=files ldd mpv 2>&1|grep libjemalloc
      3778:	file=libjemalloc.so.2 [0];  needed by /lib64/libavcodec.so.57 [0]
      3778:	file=libjemalloc.so.2 [0];  generating link map
	libjemalloc.so.2 => /lib64/libjemalloc.so.2 (0x00007f5664908000)
```
```
$ ./mpv
<jemalloc>: Error in dlsym(RTLD_NEXT, "pthread_create")
Aborted
```

Fedora 25 x86_64, glibc-2.24-6.fc25.

Safari77:
Hello World program linked with `-ljemalloc` works OK, though.

aidermoi:
I am looking at a similar issue. Static link my program with jemalloc seems to be broken for me. It works fine without -static.
/med/code3/shchen/jemalloc/src/background_thread.c:786: undefined reference to `dlsym'

davidtgoldblatt:
Assigning to @interwq on the assumption that this is related to the background thread stuff.

Earnestly:
Perhaps related: https://bugs.archlinux.org/task/54447

davidtgoldblatt:
People hitting this bug: could you clarify if you're linking with `-pthread`?

Safari77:
`-pthread` is used at least when linking `mpv`, and
```
     29877:     file=libpthread.so.0 [0];  needed by ./mpv [0]
```

Earnestly:
The commit https://github.com/jemalloc/jemalloc/commit/a4d6fe73cf07b3be3af6b7811cfc5950320bb37f has solved the issue with bulding neovim and running it as discussed here: https://github.com/neovim/neovim/issues/6895

Safari77:
a4d6fe73cf07b3be3af6b7811cfc5950320bb37f and 52fc887b49c768a9cee61d9bda9c885efb10fe95 crash with `vlc` in `malloc_mutex_trylock_final`

```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
[...]

PR #908, opened 2017-06-14
Only abort on dlsym when necessary.

interwq:
If neither background_thread nor lazy_lock is in use, do not abort on dlsym
errors.

Part of the solution to #907.


src/ctl.c
```diff
@@ -1522,6 +1522,13 @@ background_thread_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
 
 		background_thread_enabled_set(tsd_tsdn(tsd), newval);
 		if (newval) {
+			if (!can_enable_background_thread) {
+				malloc_write("<jemalloc>: Error in dlsym("
```
jasone: We shouldn't print to stderr here IMO.

PR #914, opened 2017-06-16
[Test only] Experiment pthread_mutex on MacOS.

interwq:


interwq:
Will get back to this later.

PR #917, opened 2017-06-20
Mask signals during background thread creation.

jasone:
This prevents signals from being inadvertently delivered to background
threads.

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_decls.h
+++ b/include/jemalloc/internal/jemalloc_internal_decls.h
@@ -22,6 +22,7 @@
 #    include <sys/uio.h>
 #  endif
 #  include <pthread.h>
+#  include <signal.h>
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
Paraphrasing the internal discussion here:

- Disabling signals can cause problems for programs that assume they can stop the world via their signal handler.
- Leaving them enabled can cause problems for programs that assume that they'll take signals on an application thread.
- Either way, we can break people who sigprocmask after they malloc.

I think this makes sense by default, since I'm guessing STW-ing is less common than expecting precise signal delivery.

Another option would be making this configurable via malloc conf (e.g. background_thread=yes means we pick, =no_signals means we block them, =yes_signals means we allow them).

interwq:
Agree on blocking by default should have less damage. After all this is opt in, and it can be stopped via mallctl.

PR #918, opened 2017-06-20
Clear tcache_ql after fork in child.

interwq:


davidtgoldblatt:
Possibly a stupid question: why don't we need to special-case the main thread after fork? Is there a way to flush tcaches of children after fork in the child (we can read their reentrancy_levels after all).

interwq:
Flushing other threads tcache sounds possible (but risky), also not sure if there's much reward for doing that. But probably an optimization we can consider later on.

davidtgoldblatt:
We still need to special case the main thread though, right? Otherwise its tcache/tsd won't be in the global list, right?

davidtgoldblatt:
Oh wait, I'm dumb. You do.

davidtgoldblatt:
Stamping for now, but I'm also gonna assign you an issue in a minute.

jasone:
For future reference, note that this change is difficult to write a ChangeLog entry for.  It isn't clear what failure mode this change fixes, because the log message is overly brief, and because there are no accompanying tests.

interwq:
Yeah it was missing the context. I wasn't able to repro it locally and the issue only happened in production with a specific python version. My theory was that the python runtime tried to reclaim pthread context like stacks in the child process after fork. Since the tcache was inlined in TLS, it was zeroed out and thus caused tcache_ql corruption. This change fixed the issue with python.

PR #924, opened 2017-06-22
Avoid inactivity_check within background threads.

interwq:
Passing is_background_thread down the decay path, so that background thread
itself won't attempt inactivity_check.  This fixes an issue with background
thread doing trylock on a mutex it already owns.


jasone:
What are the repercussions of doing the inactivity check for background threads?  It's not immediately clear whether this is an optimization or a behavior fix.

interwq:
Oh this is to fix the lock owner assertion reported on gitter by @aorenste. If background thread does the inactivity check for itself, it will do the trylock on a mutex it owns, thus the lock order assertion.

interwq:
Only commit msg was updated since last time (all tests passed). Merging.

PR #925, opened 2017-06-22
Set reentrancy when invoking customized extent hooks.

### LINKED ISSUE #923: Reentrancy support in extent_hook functions
If we have non-trivial operations in hook functions, e.g. need malloc / free, we should set up reentrancy properly.

interwq:
Customized extent hooks may malloc / free thus trigger reentry.  Support this
behavior by adding reentrancy on hook functions.

This resolves #923.

### DIFF
--- a/src/base.c
+++ b/src/base.c
@@ -15,7 +15,7 @@ static base_t	*b0;
 /******************************************************************************/
 
 static void *
-base_map(extent_hooks_t *extent_hooks, unsigned ind, size_t size) {
+base_map(tsdn_t *tsdn, extent_hooks_t *extent_hooks, unsigned ind, size_t size) {
 	void *addr;
 	bool zero = true;
 	bool commit = true;
@@ -25,15 +25,18 @@ base_map(extent_hooks_t *extent_hooks, unsigned ind, size_t size) {
 	if (extent_hooks == &extent_hooks_default) {
 		addr = extent_alloc_mmap(NULL, size, PAGE, &zero, &commit);
 	} else {
+		assert(!tsdn_null(tsdn));
+		pre_reentrancy(tsdn_tsd(tsdn));
 		addr = extent_hooks->alloc(extent_hooks, NULL, size, PAGE,
 		    &zero, &commit, ind);
+		post_reentrancy(tsdn_tsd(tsdn));
 	}
 
 	return addr;
 }
 
 static void
-base_unmap(extent_hooks_t *extent_hooks, unsigned ind, void *addr,
+base_unmap(tsdn_t *tsdn, extent_hooks_t *extent_hooks, unsigned ind, void *addr,
     size_t size) {
 	/*
 	 * Cascade through dalloc, decommit, purge_forced, and purge_lazy,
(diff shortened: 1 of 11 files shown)


src/base.c
```diff
@@ -25,15 +25,17 @@ base_map(extent_hooks_t *extent_hooks, unsigned ind, size_t size) {
 	if (extent_hooks == &extent_hooks_default) {
 		addr = extent_alloc_mmap(NULL, size, PAGE, &zero, &commit);
 	} else {
+		pre_reentrancy(tsdn_tsd(tsdn));
```
davidtgoldblatt: The tsdn_tsd makes me a little nervous here - how do we known tsd has been initialized here? And if we do, then why not have the base code take a raw tsd instead?

interwq: I did check this part. The only place we pass NULL tsdn to base is when initializing a0. After that we always have tsd.

davidtgoldblatt: Let's add an assert then?

I think this convinces me that we indeed shouldn't allow setting extent hooks on the shared arenas (unless we do so from the very beginning).

interwq:
Updated with tsdn_null assertions, also added a new commit to check arena != a0 in pre_reentrancy().

src/base.c
```diff
@@ -25,15 +25,18 @@ base_map(extent_hooks_t *extent_hooks, unsigned ind, size_t size) {
 	if (extent_hooks == &extent_hooks_default) {
 		addr = extent_alloc_mmap(NULL, size, PAGE, &zero, &commit);
 	} else {
+		assert(!tsdn_null(tsdn));
```
davidtgoldblatt: Nit: Comment on why this assert should be safe.

Sorry for the multiple rounds on just a style thing, I'd like to raise our level of explanation for this sort of thing in general.

interwq: Will add after the tests finish.

interwq:
@davidtgoldblatt: this is ready for review. As discussed offline, I added checks for tsdn_null just to be safe. We'll need a significant refactor to make base_alloc() path tsd-friendly.

src/base.c
davidtgoldblatt: I think this might have gotten lost in the shuffle.

src/base.c
```diff
@@ -64,8 +65,8 @@ base_unmap(tsdn_t *tsdn, extent_hooks_t *extent_hooks, unsigned ind, void *addr,
 		/* Nothing worked.  This should never happen. */
 		not_reached();
 	} else {
-		assert(!tsdn_null(tsdn));
-		pre_reentrancy(tsdn_tsd(tsdn));
+		tsd_t *tsd = tsdn_null(tsdn) ? tsd_fetch() : tsdn_tsd(tsdn);
```
davidtgoldblatt: tsd_fetch can fail, returning NULL, right? We should error out instead of segfaulting.

src/extent.c
```diff
@@ -1025,6 +1025,18 @@ extent_alloc_default(extent_hooks_t *extent_hooks, void *new_addr, size_t size,
 	    alignment, zero, commit);
 }
 
+static void
+extent_hook_pre_reentrancy(tsdn_t *tsdn, arena_t *arena) {
+	tsd_t *tsd = tsdn_null(tsdn) ? tsd_fetch() : tsdn_tsd(tsdn);
```
davidtgoldblatt: Here too (not mentioning for the rest of the comments).

interwq:
@davidtgoldblatt : the first comment looks to be a previous update. Not in the current change set. 

Regarding tsd_fetch(), it's not allowed to fail. We pass init == true here: https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/tsd.h#L257

PR #928, opened 2017-06-23
Fix assertion typos.

jasone:
Reported by Conrad Meyer.


src/ctl.c
```diff
@@ -1970,7 +1970,7 @@ arena_reset_finish_background_thread(tsd_t *tsd, unsigned arena_ind) {
 			unsigned ind = arena_ind % ncpus;
 			background_thread_info_t *info =
 			    &background_thread_info[ind];
-			assert(info->state = background_thread_paused);
+			assert(info->state == background_thread_paused);
```
interwq: Whooops. Good catch.

interwq:
A grep suggested the only 2 instances of this silly assertion pattern are both fixed here.

PR #930, opened 2017-06-23
Make sure LG_PAGE <= LG_HUGEPAGE.

### LINKED ISSUE #883: Unit Tests fail when configured with --with-lg-page=22
When I do the following, the unit tests fail. They pass with --with-lg-page=13

./configure --with-lg-page=22
make
make check

jasone:
This resolves #883.


interwq:
The freebsd failure looks scary. Is this the trylock bug you fixed @jasone ?

jasone:
Ah, yes, we're finally being bitten by not having updated the CI machine's OS to include the fix.

PR #933, opened 2017-06-24
Move thread signal mask setting to be done inline.

interwq:
Change the signal mask setting to be done by background threads directly,
instead of by creator thread.  This fixes an issue where a thread has
sigprocmask() before creating background thread, the mask setting / restoring
will break the previous signal mask.

In order to make sure the signal masks are effective from the very beginning,
thread0 creation will wait for thread0 to finish initialization before
returning.  Other background threads will inherit from thread0 so they are good
from the beginning.

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -66,6 +66,8 @@ void background_thread_ctl_init(tsdn_t *tsdn) NOT_REACHED
 #else
 
 static bool background_thread_enabled_at_fork;
+static bool t0_initialized;
+static pthread_cond_t t0_init_cond;
 
 static void
 background_thread_info_init(tsdn_t *tsdn, background_thread_info_t *info) {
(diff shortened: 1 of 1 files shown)


interwq:
So, apparently the pthread_sigmask() operations could break sigprocmask() done by app threads. To fix the issue, move the sigmask calls to background threads themselves. However this requires a bit of additional synchronization with thread0, to make sure the signal is set from the beginning.

jasone:
This change opens up a race window during which the background thread may not have signals masked, so in effect it reverts us to the original problem we were trying to solve by masking signals.  I'm not sure we should do this.

interwq:
@jasone: yes for thread0 we have a window without signal masking. However if we don't do this, it breaks sigprocmask () programs completely. Plus, in the existing approach we have a window that the application thread doesn't get it's signals, which is also bad in slightly different ways.

Issue #887, opened 2017-06-05
Experiment with address-based size class (or other metadata) lookup


davidtgoldblatt:
For a 64-bit server workload, we can burn a lot of virtual address space without anyone complaining. We can use this to encode the size class in the pointers we return, for allocations that will end up in the tcache:

Allocate a 4GB-aligned, 128GB-region of virtual address space. Reserve the lowest 4GB of the region for allocations from size class 0, the next 4GB for allocations of size class 1, etc.

Then, to compute the size class of an allocation, we range-check it (to see if it's in our 128GB "fast lookup" space), shift off the low 32 bits, and then mask off all but the low 5.

(The numbers above are back-of-the-envelope; there's no particular magic to them or reason they shouldn't be tweaked).

This lets us avoid any metadata lookups for allocations that are going to end up in the tcache during `free()`.

Things to note:
- This is orthogonal to the size of extents within those ranges of memory. We don't run into the usual slab allocator fragmentation problem in which a small number of live allocations keep the slab alive, and we don't need to change purging logic for this.
- We have talked about tweaking the caching logic (especially w.r.t. per-cpu arenas), so that a small allocation from a remote CPU won't be reused by the current CPU if it might induce false sharing (so that CPUs "own" parts of their address space). Encoding the CPU number by having one of these regions per cpu could be an efficient way to check this (though, at some point we have to start taking seriously the risk of taking up too much of the virtual address space).

jasone:
If we 128-GiB-align the memory (i.e. whatever the natural alignment needs to be for the amount of virtual memory we use), then we can do a simple equality check of the upper bits to do the "range check".

PR #936, opened 2017-06-26
Implement address-based size-class computations.

davidtgoldblatt:
This implements the ideas in https://github.com/jemalloc/jemalloc/issues/887.
We split a large portion of virtual address space into regions reserved for
particular size classes.  We can then compute the size class of a particular
allocation by using only its address and a small amount of global metadata,
avoiding any cache-unfriendly lookups.


davidtgoldblatt:
Not ready for review actually - I realized I forgot to prevent these extents from changing size classes.

PR #942, opened 2017-06-29
Add extent_grow_mtx in pre_ / post_fork handlers.

interwq:
This fixed the issue that could cause the child process to stuck after fork.


interwq:
One of the Travis job didn't get scheduled for a long time (https://travis-ci.org/jemalloc/jemalloc/jobs/248577441). Merging since this fix is needed in production soon.

jasone:
Yeah, I even manually canceled and restarted that delayed job, with no positive effect.

PR #949, opened 2017-07-06
Make symbol listing work when cross-compiling

tamird:
@jasone see https://github.com/jemalloc/jemalloc/pull/818#issuecomment-313485737

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,7 +8,7 @@ dnl ============================================================================
 dnl Custom macro definitions.
 
 dnl JE_CONCAT_VVV(r, a, b)
-dnl 
+dnl
 dnl Set $r to the concatenation of $a and $b, with a space separating them iff
 dnl both $a and $b are non-emty.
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
Thanks for the diff!

configure.ac
```diff
@@ -8,7 +8,7 @@ dnl ============================================================================
 dnl Custom macro definitions.
 
 dnl JE_CONCAT_VVV(r, a, b)
-dnl
```
interwq: @jasone: I think you mentioned the space after dnl was required, but couldn't find exactly why. Is this going to be an issue?

jasone: ```dnl ``` is an m4 comment delimiter.  ```dnla``` is not a comment delimiter, and IIRC best practice is to always follow ```dnl``` by a space.

interwq:
@davidtgoldblatt: I'll level the dnl space part to you then (link #950 as a reminder).

Issue #932, opened 2017-06-24
Think about hugepages


davidtgoldblatt:
We know we should do *something* a little more hugepage-aware. @jasone has some thoughts on it jotted down.

jasone:
There's some good initial food for thought in this Go-related conversation: https://groups.google.com/forum/#!msg/golang-codereviews/VXgv__c6GBE/595799wKBwAJ.
The underlying issue we need to consider is that Linux only supports 65530 distinct virtual memory areas (VMAs) by default, and if we're careless with ```MADV_[NO]HUGEPAGE``` it's possible to run out of VMAs long before we run out of RAM.  Assuming huge pages are 2 MiB, the obvious strategy is to keep the ```MADV_[NO]HUGEPAGE``` attributes homogeneous within each naturally aligned 2 MiB VM span.

In the context of jemalloc's extents, this means we have to take special care to manage ```MADV_[NO]HUGEPAGE``` state transitions any time we want to purge unused dirty pages within a 2 MiB VM span that also contains active extents.  Following is a viable strategy, where all alignment mentions regard huge page (2 MiB) alignment.

For each extent, maintain three state bits, one each for 1) unaligned **leader**, 2) aligned **body**, and 3) unaligned **trailer**.  For the body, the bit accurately indicates the ```MADV_[NO]HUGEPAGE``` state, but for the leader and trailer, we compute the ```MADV_[NO]HUGEPAGE``` states as the intersection of all extents' segments which fit within the 2 MiB VM span.  Ideally start out in the ```MADV_HUGEPAGE``` state.  In the following (probably non-exhaustive) examples, ```L``` is a leader, ```B``` is a body, ```T``` is a trailer, ```[  ]``` bound the relevant span, and lower case letters are incidental adjacent context:
- ```lb[T L]```
- ```lb[T L L]```
- ```lb[T L ... L]```
- ```lb[T L]bt```
- ```[L L]```
- ```[L L]bt```

[...]

PR #951, opened 2017-07-06
Address-based size-class computations

davidtgoldblatt:
Let's try this again.

The performance wins vary based on machine from "no difference" to "substantial", but I'm hopeful that we'll see wins everywhere once we actually have to share the cache with real application workloads.

### DIFF
--- a/include/jemalloc/internal/arena_structs_b.h
+++ b/include/jemalloc/internal/arena_structs_b.h
@@ -122,13 +122,17 @@ struct arena_bin_s {
 	extent_t		*slabcur;
 
 	/*
-	 * Heap of non-full slabs.  This heap is used to assure that new
-	 * allocations come from the non-full slab that is oldest/lowest in
+	 * Heaps of non-empty, non-full slabs.  This heap is used to ensure that
+	 * new allocations come from the non-full slab that is oldest/lowest in
 	 * memory.
+	 * When picking an extent to allocate from, we prefer slabs from the
+	 * sized alloc region, but always prefer active slabs to non-active
+	 * ones.
 	 */
-	extent_heap_t		slabs_nonfull;
+	extent_heap_t		slabs_sized_alloc_region_nonfull;
+	extent_heap_t		slabs_other_nonfull;
 
-	/* List used to track full slabs. */
+	/* List used to track full slabs (regardless of their origin). */
 	extent_list_t		slabs_full;
 
 	/* Bin statistics. */
@@ -273,12 +277,6 @@ struct arena_s {
 /* Used in conjunction with tsd for fast arena-related context lookup. */
 struct arena_tdata_s {
 	ticker_t		decay_ticker;
-};
-
-/* Used to pass rtree lookup context down the path. */
-struct alloc_ctx_s {
-	szind_t szind;
-	bool slab;
 };
 
 #endif /* JEMALLOC_INTERNAL_ARENA_STRUCTS_B_H */
--- a/include/jemalloc/internal/sized_alloc_region.h
+++ b/include/jemalloc/internal/sized_alloc_region.h
@@ -1,0 +1,134 @@
+#ifndef JEMALLOC_INTERNAL_SIZED_ALLOC_REGION_H
+#define JEMALLOC_INTERNAL_SIZED_ALLOC_REGION_H
+
+#include "jemalloc/internal/alloc_ctx.h"
+#include "jemalloc/internal/assert.h"
+#include "jemalloc/internal/atomic.h"
+#include "jemalloc/internal/pages.h"
+
+/*
+ * This file defines sized alloc regions.  These regions have portions of memory
+ * dedicated to a particular size class.  When allocations come out of these
+ * regions, we can quickly determine their size based only on their address and
+ * a little bit of global metadata, thus avoiding any rtree lookups or (if we're
+ * lucky) any other cache-unfriendly lookups at all.
+ *
+ * We orchestrate it so that zero-initialization is a valid state for the
+ * sized_alloc_region_t.  This lets us avoid branching on
+ * opt_sized_alloc_region.
+ */
+
+/* The number of size classes to put in the sized_alloc_region. */
+#define SIZED_ALLOC_REGION_NUM_SCS NBINS
+
+typedef struct sized_alloc_region_s sized_alloc_region_t;
+struct sized_alloc_region_s {
+	JEMALLOC_ALIGNED(CACHELINE)
+	uintptr_t region_start;
+	/*
+	 * This is always either 0 (if uninitialized, or if initialization
+	 * failed), or SIZED_ALLOC_REGION_SIZE.  Doing it this way, instead of
+	 * having "bool initialized;", allows us to fold the initialization and
+	 * range checking branches into one.
+	 */
+	size_t region_size;
+	bool committed;
+
+	/* How much space is used in each size class. */
+	JEMALLOC_ALIGNED(CACHELINE)
+	atomic_zu_t size_class_space_used[SIZED_ALLOC_REGION_NUM_SCS];
+};
+
+/* Whether this feature is enabled. */
+extern bool opt_sized_alloc_region;
+/* The lg of the number of bytes reserved for each size class in the region. */
+extern ssize_t opt_sized_alloc_region_lg_sc_size;
+/*
+ * The region we actually allocate out of.  There should be only one if these in
+ * an actualy program.
+ */
+extern sized_alloc_region_t sized_alloc_region_global;
+
+void sized_alloc_region_init(sized_alloc_region_t *region);
+#ifdef JEMALLOC_JET
+/*
+ * We need this only during testing, when we might spin up multiple of these,
+ * and should be considerate of virtual address space.
+ */
+void sized_alloc_region_destroy(sized_alloc_region_t *region);
+#endif
+
+/*
+ * Returns true if we found the address.  If alloc_ctx is non-null, we fill it
+ * in.  Safe to call on zero-initialized regions.
+ */
+static inline bool
+sized_alloc_region_lookup(sized_alloc_region_t *region, void *addrp,
+    alloc_ctx_t *alloc_ctx) {
+	uintptr_t addr = (uintptr_t)addrp;
+
+	/* Note: wraparound possible if addr < region. */
+	uintptr_t addr_offset = addr - region->region_start;
+
+	if (unlikely(addr_offset >= region->region_size)) {
+		return false;
+	}
+	if (alloc_ctx != NULL) {
+		alloc_ctx->szind = (szind_t)(
+		    addr_offset >> opt_sized_alloc_region_lg_sc_size);
+		alloc_ctx->slab = true;
+	}
+	return true;
+}
+
+/*
+ * This is safe to call on zero-initialized region, or one for whom
+ * initialization failed.
+ */
+static inline void *
+sized_alloc_region_bump_alloc(sized_alloc_region_t *region, size_t size,
+    szind_t szind, bool slab, bool *zero, bool *commit) {
+	assert(PAGE_CEILING(size) == size);
+	/* Only slab allocations allowed in the sized-alloc region. */
+	assert(slab);
+
+	if (region->region_size == 0 || szind >= SIZED_ALLOC_REGION_NUM_SCS) {
+		return NULL;
+	}
+
+	size_t sc_size = (ZU(1) << opt_sized_alloc_region_lg_sc_size);
+
+	/*
+	 * Overflow should be at most a theoretical problem here, since
+	 * presumably if the caller can't get memory from us, they'll get it
+	 * from elsewhere, so we don't hit problems until we exhaust the address
+	 * space.  For extra safety (and simplicity while debugging), we use a
+	 * CAS loop instead of a fetch-add.
+	 */
+	size_t cur = atomic_load_zu(&region->size_class_space_used[szind],
+	    ATOMIC_RELAXED);
+	size_t new_size;
+	do {
+		new_size = cur + size;
+		if (new_size > sc_size) {
+			return NULL;
+		}
+	} while (!atomic_compare_exchange_weak_zu(
+	    &region->size_class_space_used[szind], &cur, new_size,
+	    ATOMIC_RELAXED, ATOMIC_RELAXED));
+
+	void *ret = (void *)(region->region_start + (szind * sc_size) + cur);
+	if (*commit && !region->committed) {
+		if (pages_commit(ret, size)) {
+			/* Just leak the virtual address space here. */
+			return NULL;
+		}
+	} else {
+		*commit = region->committed;
+	}
+	/* We get our memory straight from the OS, and never reuse it. */
+	*zero = true;
+	return ret;
+}
+
+#endif /* JEMALLOC_INTERNAL_SIZED_ALLOC_REGION_H */
--- a/src/arena.c
+++ b/src/arena.c
@@ -8,6 +8,7 @@
 #include "jemalloc/internal/mutex.h"
 #include "jemalloc/internal/rtree.h"
 #include "jemalloc/internal/size_classes.h"
+#include "jemalloc/internal/sized_alloc_region.h"
 #include "jemalloc/internal/util.h"
 
 /******************************************************************************/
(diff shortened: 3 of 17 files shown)


src/extent.c
```diff
@@ -965,10 +1016,15 @@ extent_recycle(tsdn_t *tsdn, arena_t *arena, extent_hooks_t **r_extent_hooks,
  * memory, in which case *zero is toggled to true.  arena_extent_alloc() takes
  * advantage of this to avoid demanding zeroed extents, but taking advantage of
  * them if they are returned.
+ *
+ * The caller can specify szind == NSIZES to indicate "don't know" (forgoing the
```
djwatson: Where do we need this?  Just extent_alloc_core?

include/jemalloc/internal/sized_alloc_region.h
```diff
@@ -0,0 +1,144 @@
...
+
+/*
+ * SIZED_ALLOC_REGION_LG_NUM_SCS:
+ *     The log of the number of size classes that can come out of the
+ *     sized-alloc region.
+ * SIZED_ALLOC_REGION_LG_SC_SIZE:
+ *     The log of the size of virtual address space that we allocate to each
+ *     size class within the sized-alloc region.
+ */
+#if LG_SIZEOF_PTR == 2
```
djwatson: It doesn't look like the region size is configerable, how hard would it be to make it a malloc_conf option?  We've got widely varying working set sizes

include/jemalloc/internal/sized_alloc_region.h
```diff
@@ -0,0 +1,144 @@
...
+	assert(slab);
+
+	if (region->region_size == 0 || szind >= SIZED_ALLOC_REGION_NUM_SCS) {
+		return NULL;
+	}
+
+	size_t prev = atomic_fetch_add_zu(&region->size_class_space_used[szind],
+	    size, ATOMIC_RELAXED);
+
+	if (unlikely(prev + size >= SIZED_ALLOC_REGION_SC_SIZE)) {
```
djwatson: Do we have any size classes where this will leave holes in the address space?  Should we try to make SIZED_ALLOC_REGION_SC_SIZE per-sizeclass?

djwatson:
Overall it's what I expected for return-of-the-chunk, much easier to understand.  

I'm still concerned that not being able to change size class may have a negative perf impact, it's hard to tell how that will balance out vs. saving the memory lookup.  Benchmarks? :)

include/jemalloc/internal/sized_alloc_region.h
davidtgoldblatt: It wouldn't be crazy. It needs to be a power of two so we can get by with just a shift. It would move the shift amount from being embdedded in the instruction into being a load + dynamic shift; not sure if we would notice.

These settings allow for 4GB of allocations from each size class (in 64 bit mode) though; this exceeds anything I've seen even from big services.

include/jemalloc/internal/sized_alloc_region.h
davidtgoldblatt: We talked about it, but it would make the szind computation in free more expensive. I don't think there should be address-space hole problems, since we never care about alignment here and all allocations are of the same size.

src/extent.c
davidtgoldblatt: Yeah. We care about szind for sized-alloc-region extents, but for all others the extent code is mostly agnostic about it.

davidtgoldblatt:
The microbenchmark on my laptop (which has a Xeon that Lenovo has started cramming into a small space, but is still just a laptop) showed something like a 20% win. On my older but beefier devserver Xeon, it doesn't seem to make any difference.

I'll try to get some benchmarks from production machines.

davidtgoldblatt:
... and the reason for that is, something changed to make us never use the sized alloc region. I'm digging now.

davidtgoldblatt:
Summarizing some in-person discussion:

- This should get compiled-in by default; not a config option (note that this costs an extra fast-path branch in the case where this is disabled).
- Number of size-classes we do with this should just be set to NBINS.
- Size per size class should stay a compile-time constant.
- Disabled by default.

src/arena.c
```diff
@@ -1177,14 +1196,21 @@ arena_reset(tsd_t *tsd, arena_t *arena) {
...
-		while ((slab = extent_heap_remove_first(&bin->slabs_nonfull)) !=
-		    NULL) {
+		while (config_sized_alloc_region
+		    && (slab = extent_heap_remove_first(
+		    &bin->sized_alloc_region_nonfull_slabs)) != NULL) {
+			malloc_mutex_unlock(tsd_tsdn(tsd), &bin->lock);
+			arena_slab_dalloc(tsd_tsdn(tsd), arena, slab);
+			malloc_mutex_lock(tsd_tsdn(tsd), &bin->lock);
+		}
+		while ((slab = extent_heap_remove_first(
```
interwq: This doesn't look right to me. Aren't we leaking the slab allocated above?

interwq: Oh NVM. This is reset.

src/extent.c
```diff
@@ -312,11 +316,17 @@ extents_insert_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent,
...
-	if (extent_heap_empty(&extents->heaps[pind])) {
-		bitmap_unset(extents->bitmap, &extents_bitmap_info,
-		    (size_t)pind);
+	if (config_sized_alloc_region && extent_in_sized_alloc_region(extent)) {
+		szind_t szind = extent_szind_get(extent);
+		extent_heap_insert(&extents->sized_alloc_region_heaps[szind],
+		    extent);
+	} else {
+		if (extent_heap_empty(&extents->other_heaps[pind])) {
+			bitmap_unset(extents->bitmap, &extents_bitmap_info,
```
jasone: Why is this logic omitted in the ```config_sized_alloc_region``` branch?  The same question applies to ```extents_remove_locked()```.  My guess is that you're keeping slabs in the per bin heaps forever once allocated, even when they're completely empty.  I think this strategy is insufficient in the context of arena reset, because there's no mechanism for reusing the slabs that were in the heaps at reset time.

src/extent.c
```diff
@@ -812,7 +851,19 @@ extent_recycle_extract(tsdn_t *tsdn, arena_t *arena,
 			extent_unlock(tsdn, unlock_extent);
 		}
 	} else {
-		extent = extents_fit_locked(tsdn, arena, extents, alloc_size);
+		extent = NULL;
```
jasone: Nit: Initialize only once.

src/extent.c
```diff
@@ -1243,9 +1299,17 @@ extent_alloc_wrapper_hard(tsdn_t *tsdn, arena_t *arena,
 	}
 	void *addr;
 	if (*r_extent_hooks == &extent_hooks_default) {
-		/* Call directly to propagate tsdn. */
-		addr = extent_alloc_default_impl(tsdn, arena, new_addr, esize,
-		    alignment, zero, commit);
+		if (config_sized_alloc_region && alignment == PAGE
```
jasone: It may be clearer here to ```assert(!slab || alignment == PAGE)```.

test/unit/sized_alloc_region.c
```diff
@@ -0,0 +1,98 @@
+#include "test/jemalloc_test.h"
+
+#include "jemalloc/internal/sized_alloc_region.h"
+
+#define ALLOCS_PER_SC 10
+
+TEST_BEGIN(test_lookup) {
+	sized_alloc_region_t region;
+	sized_alloc_region_init(&region);
+
```
jasone: Extra line.

test/unit/sized_alloc_region.c
```diff
@@ -0,0 +1,98 @@
...
+				assert_d_eq(sc, (int)alloc_ctx.szind,
+				    "Lookup found incorrect size class.");
+				assert_true(alloc_ctx.slab,
+				    "Lookup found non-slab alloc.");
+			}
+		}
+	}
+
+	sized_alloc_region_destroy(&region);
+
```
jasone: Extra line.

interwq:
Any efficient way to get stats around this? Or maybe do some lazy updates when reading stats.

include/jemalloc/internal/arena_structs_b.h
```diff
@@ -122,14 +122,18 @@ struct arena_bin_s {
...
+	 * ones.
 	 */
-	extent_heap_t		slabs_nonfull;
+	extent_heap_t		sized_alloc_region_nonfull_slabs;
+	extent_heap_t		other_nonfull_slabs;
 
-	/* List used to track full slabs. */
-	extent_list_t		slabs_full;
+	/* List used to track full slabs (regardless of their origin). */
+	extent_list_t		full_slabs;
```
davidtgoldblatt: Yeah, sounds reasonable. I originally changed them because, once there were four of them, having it be adjective-noun made it feel more readable. Once we're used to it I suppose it doesn't matter.

include/jemalloc/internal/sized_alloc_region.h
```diff
@@ -0,0 +1,144 @@
...
+ * a little bit of global metadata, thus avoiding any rtree lookups or (if we're
+ * lucky) any other cache-unfriendly lookups at all.
+ *
+ * We orchestrate it so that zero-initialization is a valid state for the
+ * sized_alloc_region_t.  This lets us avoid branching on
+ * opt_sized_alloc_region.
+ */
+
+/*
+ * SIZED_ALLOC_REGION_LG_NUM_SCS:
```
davidtgoldblatt: Yeah, there isn't (this was an artifact of an old computation mechanism, where I did the range check via a bit mask). I think making it be NBINS is the right call.

include/jemalloc/internal/sized_alloc_region.h
```diff
@@ -0,0 +1,144 @@
...
+ */
+static inline bool
+sized_alloc_region_lookup(sized_alloc_region_t *region, void *addrp,
+    alloc_ctx_t *alloc_ctx) {
+	uintptr_t addr = (uintptr_t)addrp;
+
+	/* Note: wraparound possible if addr < region. */
+	uintptr_t addr_offset = addr - region->region_start;
+
+	if (unlikely(addr_offset >= region->region_size)) {
```
davidtgoldblatt: Doing it this way allows us to fold the "is sized-alloc-region enabled" branch into this comparison, since in that case the region is all zeros.

include/jemalloc/internal/sized_alloc_region.h
```diff
@@ -0,0 +1,144 @@
...
+
+	if (region->region_size == 0 || szind >= SIZED_ALLOC_REGION_NUM_SCS) {
+		return NULL;
+	}
+
+	size_t prev = atomic_fetch_add_zu(&region->size_class_space_used[szind],
+	    size, ATOMIC_RELAXED);
+
+	if (unlikely(prev + size >= SIZED_ALLOC_REGION_SC_SIZE)) {
+		return NULL;
```
davidtgoldblatt: The way I figured it was, if the sized_alloc_region alloc fails, the memory will get mapped in some other way, so for this to overflow we will have to have already exhausted the address space. Doesn't hurt to do, though, I suppose.

include/jemalloc/internal/sized_alloc_region.h
IndolentPangolin: If I'm understanding these comments, the following may be relevant.

We use jemalloc at work, with custom allocation hooks (currently we're on 4.2.0 partly because I haven't had time to experiment with other versions). In our use we have about 54GB of address space for the application. We've seen up to 20GB allocated as individual 2MB allocations, and I've seen 12GB from a 655kB allocation. Additionally, the flash page we're caching is a 16kB size class and we can have many GB of that allocation as well.

I.e. we definitely have a use-case for more than 4GB for an individual size class.

davidtgoldblatt: Oh, interesting.

Yeah, I think this will end up being runtime configurable. Though, note that if you're using custom extent/chunk hooks, we'll still have to do the current metadata lookup pathways for those allocations, so you won't see any benefit.

IndolentPangolin: Would it be possible for my custom hooks to also present the same address offsets to allow the cheaper lookup? I suspect the performance for us would be worse compared to the lookups, but it would be nice to experiment.

We use custom hooks really only for two reasons: first, to use MAP_HUGETLB to use preallocated 2MB pages, and second to limit the release of memory back to the system via munmap() (we instead hold on to the allocation and hand chunks back when the alloc hook is called again). Both of those net us measurable performance wins, but probably aren't suitable for general-purpose code.

davidtgoldblatt: I don't think it's feasible to allow using this with custom hooks, since this is internally at the same "level" -- both the sized alloc region and the extent hooks are just "one source of memory among several". Making this use custom hooks would be hard.

The MAP_HUGETLB thing is tricky, and something we've been talking about a lot (both trying to use hugepages behind the scenes and also to figure out how to work with them in the extent infrastructure -- see #932 and #952). I don't think we have a great story for letting the user enforce a hugepage requirement from a hook in 5.0 (and, we only accidentally have such a story for 4.x).

With respect to limiting the release of memory back to the OS -- have you looked at the decay settings in malloc_conf? You should be able to limit usage pretty effectively by tweaking those settings (or even disabling releasing back to the OS in general, if that's what you'd prefer). If you're implementing an alternate decay policy via the hooks, I'd be curious to hear about it (it could end up as a useful feature for other users, too).

IndolentPangolin: I'm not sure I really understand how decaying works. (Not sure this review is the best place to discuss this either, so if you have a better thought, let me know).

I ended up with a moderately complicated implementation of our hooks, but basically trying to solve the problem that we want to use 2MB TLB entries when we can, and we want to limit syscalls for performance. chnk_alloc pulls from a recycling list, if there's memory present (we cache 2MB and 4MB chunks, and our system rarely allocates larger than this). If there's nothing in the recycler, we mmap in 64MB blocks and push the unused part into the recycler. dalloc just pushes the chunk into our recycling lists.

commit just returns false; decommit just returns true; purge just returns true; split just returns false; merge just returns false. As best I could tell from the 4.2 documentation this is correct.

Our alloc is more complex than I described, because we also want to release some of our recycler's memory when we get a alloc() request over 4MB, if we're near the process's virtual memory limit. There's also some fiddly bits so we can track how much memory we've mmap'd and whether it was HUGETLB or 4k mappings, so we can report on this occasionally in the logs.

We had to tweak Linux to disable transparent hugepages, because it caused a significant performance issue where our mmap() calls would hang for a while; we can't afford to have threads hanging for 100ms or more in a syscall. So we manually take about 2/3rds of the DRAM on the box and pre-reserve hugepages, and then our chunk hooks will take those if available and fall back to normal 4k mappings. My investigation into MADV_HUGETLB showed I had no way to tell if I really got a 2MB mapping or not, and I really want to be able to report in the application how much data is backed by 2MB TLBs. I didn't spend a ton of time testing the performance of this versus the pre-reserved MAP_HUGETLB.

I agree MAP_HUGETLB is really hard -- how would jemalloc know if the pre-allocated large pages were meant for jemalloc or for something else in the system?

IndolentPangolin: Oh, and our application memory usage is probably a bit atypical. We cap the VMEM of the process because if we have a memory leak we'd rather crash, report on it, and restart than page. We can't afford to page. Memory is often allocated by one CPU and free'd on another. We probably have windows of time where one CPU does lots of allocations and no frees, and another CPU is doing lots of frees and no allocations.

I experimented with turning off tcache and I did measure a noticeable drop in performance, so we're definitely getting something from that.

We would prefer everything to be lock free since we start up one thread per CPU core and then have an internal task system. Anytime a thread takes a lock (or enters the kernel which often means a lock) we can end up both starving the system of work (multiple threads are blocked on a lock and not running our tasks), and it also can cause a latency spike for the operation that was unfortunate enough to need to wait on a lock or call a syscall.

davidtgoldblatt: Opened #953 for this, it's a little out of scope for this PR.

configure.ac
```diff
@@ -1120,6 +1120,22 @@ if test "x$enable_prof" = "x1" ; then
 fi
 AC_SUBST([enable_prof])
 
+dnl Do not enable the sized alloc region by default.
+AC_ARG_ENABLE([sized-alloc-region],
```
davidtgoldblatt: Done.

include/jemalloc/internal/arena_structs_b.h
davidtgoldblatt: Done.

include/jemalloc/internal/sized_alloc_region.h
davidtgoldblatt: Done.

include/jemalloc/internal/sized_alloc_region.h
```diff
@@ -0,0 +1,144 @@
...
+ */
+void sized_alloc_region_destroy(sized_alloc_region_t *region);
+#endif
+
+/*
+ * Returns true if we found the address.  If alloc_ctx is non-null, we fill it
+ * in.  Safe to call on zero-initialized regions.
+ */
+static inline bool
+sized_alloc_region_lookup(sized_alloc_region_t *region, void *addrp,
```
davidtgoldblatt: Fits the general pattern of modulename_fn; I like the consistency enough that I'll leave it barring someone's strong objection.

include/jemalloc/internal/sized_alloc_region.h
davidtgoldblatt: Done.

src/arena.c
```diff
@@ -1079,17 +1080,35 @@ arena_slab_dalloc(tsdn_t *tsdn, arena_t *arena, extent_t *slab) {
...
+		    slab);
+	} else {
+		extent_heap_remove(&bin->other_nonfull_slabs, slab);
+	}
 }
 
 static extent_t *
 arena_bin_slabs_nonfull_tryget(arena_bin_t *bin) {
-	extent_t *slab = extent_heap_remove_first(&bin->slabs_nonfull);
+	extent_t *slab = NULL;
```
davidtgoldblatt: Done.

src/arena.c
```diff
@@ -1708,7 +1734,8 @@ arena_dalloc_bin_locked_impl(tsdn_t *tsdn, arena_t *arena, extent_t *slab,
 
 	arena_slab_reg_dalloc(tsdn, slab, slab_data, ptr);
 	unsigned nfree = extent_nfree_get(slab);
-	if (nfree == bin_info->nregs) {
+	if (nfree == bin_info->nregs && !sized_alloc_region_lookup(
```
davidtgoldblatt: Yeah. All this logic got pushed down into helper functions, which means we don't care about slab location here. Fixed.

src/extent.c
davidtgoldblatt: Oh, that's tricky. I think I'm OK just declaring that the sized-alloc-region is reserved for auto arenas, letting us sidestep the issue.

If this turns out to be a problem for some reason, we can always add a fallback global extents_t.

davidtgoldblatt: Done.

src/extent.c
davidtgoldblatt: This one actually makes the logic much trickier to follow, since both !slab and sized alloc region failure can cause the fallback to be necessary.

src/extent.c
davidtgoldblatt: Done.

test/unit/sized_alloc_region.c
davidtgoldblatt: Done.

test/unit/sized_alloc_region.c
davidtgoldblatt: Done.

davidtgoldblatt:
Still has some bugs (--enable-debug --with-malloc-conf=dss:primary fails), but, fewer bugs!

davidtgoldblatt:
Superseded by #1058.

PR #958, opened 2017-07-20
Logging functionality

davidtgoldblatt:
This allows us to enable fine-grained log information, for use in debugging.

### DIFF
--- a/include/jemalloc/internal/log.h
+++ b/include/jemalloc/internal/log.h
@@ -1,0 +1,89 @@
+#ifndef JEMALLOC_INTERNAL_LOG_H
+#define JEMALLOC_INTERNAL_LOG_H
+
+#include "jemalloc/internal/atomic.h"
+#include "jemalloc/internal/mutex.h"
+
+#ifdef JEMALLOC_LOG
+#  define JEMALLOC_LOG_BUFSIZE 1000
+#else
+#  define JEMALLOC_LOG_BUFSIZE 1
+#endif
+
+/*
+ * The log_vars malloc_conf option is a '|'-delimited list of log_var name
+ * segments to log.  The log_var names are themselves hierarchical, with '.' as
+ * the delimiter (a "segment" is just a prefix in the log namespace).  So, if
+ * you have:
+ *
+ * static log_var_t log_arena = LOG_VAR_INIT("arena"); // 1
+ * static log_var_t log_arena_a = LOG_VAR_INIT("arena.a"); // 2
+ * static log_var_t log_arena_b = LOG_VAR_INIT("arena.b"); // 3
+ * static log_var_t log_arena_a_a = LOG_VAR_INIT("arena.a.a"); // 4
+ * static_log_var_t log_extent_a = LOG_VAR_INIT("extent.a"); // 5
+ * static_log_var_t log_extent_b = LOG_VAR_INIT("extent.b"); // 6
+ *
+ * And your malloc_conf option is "log_vars=arena.a|extent", then log_vars 2, 4,
+ * 5, and 6 will be enabled.  You can enable logging from all log vars by
+ * writing "log_vars=.".
+ *
+ * You can then log by writing:
+ *   log(log_var, "format string -- my int is %d", my_int);
+ *
+ * None of this should be regarded as a stable API for right now.  It's intended
+ * as a debugging interface, to let us keep around some of our printf-debugging
+ * statements.
+ */
+
+extern char log_var_names[JEMALLOC_LOG_BUFSIZE];
+extern atomic_b_t log_init_done;
+
+typedef struct log_var_s log_var_t;
+struct log_var_s {
+	/*
+	 * Lowest bit is "inited", second lowest is "enabled".  Putting them in
+	 * a single word lets us avoid any fences on weak architectures.
+	 */
+	atomic_u_t state;
+	const char *name;
+};
+
+#define LOG_NOT_INITIALIZED 0U
+#define LOG_INITIALIZED_NOT_ENABLED 1U
+#define LOG_ENABLED 2U
+
+#define LOG_VAR_INIT(name_str) {ATOMIC_INIT(LOG_NOT_INITIALIZED), name_str}
+
+/*
+ * Returns the value we should assume for state (which is not necessarily
+ * accurate; if logging is done before logging has finished initializing, then
+ * we default to doing the safe thing by logging everything).
+ */
+unsigned log_var_update_state(log_var_t *log_var);
+
+/* We factor out the metadata management to allow us to test more easily. */
+#define log_do_begin(log_var)						\
+if (config_log) {							\
+	unsigned log_state = atomic_load_u(&(log_var).state,		\
+	    ATOMIC_RELAXED);						\
+	if (unlikely(log_state == LOG_NOT_INITIALIZED)) {		\
+		log_state = log_var_update_state(&(log_var));		\
+		assert(log_state != LOG_NOT_INITIALIZED);		\
+	}								\
+	if (log_state == LOG_ENABLED) {					\
+		{
+			/* User code executes here. */
+#define log_do_end(log_var)						\
+		}							\
+	}								\
+}
+
+#define log(log_var, format, ...)					\
+do {									\
+	log_do_begin(log_var)						\
+		malloc_printf("%s: " format "\n",			\
+		    (log_var).name, __VA_ARGS__);			\
+	log_do_end(log_var)						\
+} while (0)
+
+#endif /* JEMALLOC_INTERNAL_LOG_H */
--- a/include/jemalloc/internal/log.h
+++ b/include/jemalloc/internal/log.h
@@ -30,6 +30,12 @@
  * You can then log by writing:
  *   log(log_var, "format string -- my int is %d", my_int);
  *
+ * The namespaces currently in use:
+ *   core.[malloc|free|posix_memalign|...].[entry|exit]:
+ *       The entry/exit points of the functions publicly exposed by jemalloc.
+ *       The "entry" variants try to log arguments to the functions, and the
+ *       "exit" ones try to log return values.
+ *
  * None of this should be regarded as a stable API for right now.  It's intended
  * as a debugging interface, to let us keep around some of our printf-debugging
  * statements.
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1974,6 +1974,13 @@ je_malloc(size_t size) {
 	static_opts_t sopts;
 	dynamic_opts_t dopts;
 
+	static log_var_t log_core_malloc_entry = LOG_VAR_INIT(
+	    "core.malloc.entry");
+	static log_var_t log_core_malloc_exit = LOG_VAR_INIT(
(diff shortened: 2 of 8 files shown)


src/jemalloc.c
```diff
@@ -1173,6 +1174,16 @@ malloc_conf_init(void) {
 				CONF_HANDLE_BOOL(opt_prof_final, "prof_final")
 				CONF_HANDLE_BOOL(opt_prof_leak, "prof_leak")
 			}
+			if (config_log) {
+				if (CONF_MATCH("log_vars")) {
+					size_t cpylen = (
+					    vlen <= sizeof(log_var_names) ?
+					    vlen : sizeof(log_var_names) - 1);
+					strncpy(log_var_names, v, cpylen);
+					log_var_names[cpylen] = '\0';
```
interwq: If we have ```log_vars``` set through multiple ways (e.g. with env_var + source code), we should try merging them instead of overwriting.

Do we need to check if log_vars is valid? Or it always can be parsed correctly?

src/log.c
```diff
@@ -0,0 +1,78 @@
...
+	assert(log_var_begin < log_var_end);
+
+	ptrdiff_t segment_len = segment_end - segment_begin;
+	ptrdiff_t log_var_len = log_var_end - log_var_begin;
+	/* The special '.' segment matches everything. */
+	if (segment_len == 1 && *segment_begin == '.') {
+		return true;
+	}
+	if (segment_len < log_var_len) {
+                return strncmp(segment_begin, log_var_begin, segment_len) == 0
```
interwq: nit: switch spaces to tabs.

src/log.c
```diff
@@ -0,0 +1,78 @@
...
+		    && log_var_begin[segment_len] == '.';
```
interwq: nit: swap the two conditions, as the second one is likely an easier bail out.

include/jemalloc/internal/log.h
```diff
@@ -0,0 +1,83 @@
...
+#define LOG_ENABLED 2U
+
+#define LOG_VAR_INIT(name_str) {ATOMIC_INIT(LOG_NOT_INITIALIZED), name_str}
+
+/*
+ * Returns the value we should assume for state (which is not necessarily
+ * accurate; if logging is done before logging has finished initializing, then
+ * we default to doing the safe thing by logging everything).
+ */
+unsigned log_var_init(log_var_t *log_var);
```
interwq: Probably should name this function differently (now it's too similar to the macro above). Maybe log_var_state_init?

src/log.c
```diff
@@ -0,0 +1,78 @@
...
+	 * instructed to.
+	 */
+	if (!atomic_load_b(&log_init_done, ATOMIC_ACQUIRE)) {
+		return LOG_ENABLED;
+	}
+
+	while (true) {
+		const char *segment_end = log_var_extract_segment(
+		    segment_begin);
+		if (log_var_matches_segment(segment_begin, segment_end,
```
interwq: nit: ```assert(segment_end < log_var_names + BUF_SIZE);```

test/unit/log.c
```diff
@@ -0,0 +1,181 @@
...
+}
+
+TEST_BEGIN(test_log_disabled) {
+	test_skip_if(!config_log);
+	atomic_store_b(&log_init_done, true, ATOMIC_RELAXED);
+	expect_no_logging("");
+	expect_no_logging("abc");
+	expect_no_logging("a.b.c");
+	expect_no_logging("l12");
+	expect_no_logging("l123|a456|b789");
```
interwq: nit: check for ```|||```?

include/jemalloc/internal/log.h
```diff
@@ -0,0 +1,83 @@
...
+ * you have:
+ *
+ * static log_var_t log_arena = LOG_VAR_INIT("arena"); // 1
+ * static log_var_t log_arena_a = LOG_VAR_INIT("arena.a"); // 2
+ * static log_var_t log_arena_b = LOG_VAR_INIT("arena.b"); // 3
+ * static log_var_t log_arena_a_a = LOG_VAR_INIT("arena.a.a"); // 4
+ * static_log_var_t log_extent_a = LOG_VAR_INIT("extent.a"); // 5
+ * static_log_var_t log_extent_b = LOG_VAR_INIT("extent.b"); // 6
+ *
+ * And your malloc_conf option is "log_vars=arena.a|extent", then log_vars 2, 4,
```
interwq: Is ```arena.``` valid? Say I want to log all vars under that namespace.

include/jemalloc/internal/log.h
```diff
@@ -0,0 +1,83 @@
...
+			/* User code executes here. */
+#define log_do_end(log_var)						\
+		}							\
+	}								\
+}
+
+#define log(log_var, format, ...)					\
+do {									\
+	log_do_begin(log_var)						\
+		malloc_printf("%s: " format "\n",			\
```
interwq: Also, thread id could be useful here, or maybe even timestamp.

src/jemalloc.c
```diff
@@ -1974,6 +1974,13 @@ je_malloc(size_t size) {
 	static_opts_t sopts;
 	dynamic_opts_t dopts;
 
+	static log_var_t log_core_malloc_entry = LOG_VAR_INIT(
+	    "core.malloc.entry");
```
interwq: Let's add a section in ```log.h``` documenting the namespaces as we add them.

include/jemalloc/internal/log.h
davidtgoldblatt: I'm inclined to say no -- to log everything under arena, I think "arena" should be the only allowed mechanism.

include/jemalloc/internal/log.h
davidtgoldblatt: Done. I changed it to `log_var_compute_state`, which I also don't love. Suggestions welcome.

include/jemalloc/internal/log.h
```diff
@@ -0,0 +1,83 @@
...
+ * Returns the value we should assume for state (which is not necessarily
+ * accurate; if logging is done before logging has finished initializing, then
+ * we default to doing the safe thing by logging everything).
+ */
+unsigned log_var_init(log_var_t *log_var);
+
+/* We factor out the metadata management to allow us to test more easily. */
+#define log_do_begin(log_var)						\
+if (config_log) {							\
+	unsigned state = atomic_load_u(&(log_var).state,		\
```
davidtgoldblatt: Done.

include/jemalloc/internal/log.h
```diff
@@ -0,0 +1,83 @@
...
+	    ATOMIC_RELAXED);						\
+	if (state == LOG_NOT_INITIALIZED) {				\
```
davidtgoldblatt: Done.

Really, I think this ought to be done guarded with the fast-path/slow-path trickery we've used elsewhere, except that we could plausibly want to log in places where we don't necessarily have tsd available. I think once we have some experience with it we should take another look and see if we can make this always on.

include/jemalloc/internal/log.h
```diff
@@ -0,0 +1,83 @@
...
+		state = log_var_init(&(log_var));			\
```
davidtgoldblatt: We do lazy init on a per logvar basis, so malloc_initialized doesn't necessarily predict any values we get here. I added an assert that log_state doesn't end up as "don't know" after initialization though. We test the pre-init behavior in the unit tests, also.

include/jemalloc/internal/log.h
davidtgoldblatt: I agree about about all this. I didn't add any bells in whistles here just because that's its own whole thing because of the cross-platform issues. I think once we have a specific need we should totally do it. (Another issue we'll likely want to deal with eventually: I don't think malloc_printf does anything to prevent multiple log messages being interleaved).

src/jemalloc.c
davidtgoldblatt: Merging is a little hard just because of C string irritation stuff. For a purely-debugging thing it feels like it can be deferred until we need it?

I guess log_vars is always valid. Like, if someone says "turn on any logging for log vars named asgs$&^@(.\t\v\a" then it won't match any of the logvars we set, but I don't see the harm in letting them.

src/log.c
davidtgoldblatt: Good catch. Not sure how that sneaked in.

src/log.c
davidtgoldblatt: Done.

src/log.c
davidtgoldblatt: Done.

src/jemalloc.c
interwq: Sounds reasonable to do merging when we need.

For validation, does "arena.|" match anything?

test/unit/log.c
davidtgoldblatt: Done.

src/jemalloc.c
davidtgoldblatt: Good idea, done.

src/jemalloc.c
davidtgoldblatt: No; the "arena." doesn't because of the spurious '.', and the second one is empty.

I suppose we might get an accidental match if someone wrote a log var that starts with a '.'.

davidtgoldblatt:
So, this logs some junk during initialization, because we e.g. log in malloc before knowing if we're initialized. There's two solutions we could choose:
- Never log during initialization.
- Keep the current strategy (default to logging more), and don't worry about the log-spew, since it's a config option.

I'm inclined to change the diff again and go with the first of these (under the assumption that we eventually want this always-on, and so we should care about log-spew), but I could be persuaded that the second is better.

interwq:
I don't have strong preference whether over-log during init; but considering the always on scenario I think you are making the right call.

Last question: do you want to add documentation to manual now, or you are thinking to add more stuff and do it together? I feel the convention is to add to the manual at the same time. In this case maybe keep the description short since we have more features in mind.

davidtgoldblatt:
Hmm, OK, I'll make it default to "off". (In the long run, it'd be nice for init checks to get moved earlier though, to solve the issue for us).

Re: the manual: I mostly regard this as just for development for right now. Maybe once we have a better idea of what we find useful or don't, we could start publicizing it, or maybe committing to log names,

interwq:
> Re: the manual: I mostly regard this as just for development for right now. Maybe once we have a better idea of what we find useful or don't, we could start publicizing it, or maybe committing to log names

Sounds reasonable. Let's add some simple notes regarding this not being in the manual?

davidtgoldblatt:
Where are you thinking? Just in log.h?

interwq:
Yeah either in log.h, or around the definition ```static const bool config_log```. We have similar undocumented option there as well: https://github.com/jemalloc/jemalloc/blob/13685ab1b767091d817cb4959d24a42447a6fb78/include/jemalloc/internal/jemalloc_preamble.h.in#L161

davidtgoldblatt:
OK, added a warning.

davidtgoldblatt:
Thanks for the quick reviews, I'll rebase once tests come back clean. I opened #960 for possible improvements we could make.

Issue #946, opened 2017-07-02
O_CLOEXEC on linux under 2.6.23


wangpeiwen:
I notice that jemalloc now use `open` with `O_CLOEXEC` which is supported since linux 2.6.23
Can we support the old linux or not?

davidtgoldblatt:
In general we'll be sort of vague to support levels for older kernels, since we don't test on them and can't really promise not to break them in refactorings and such.

That said, this is sufficiently local and easy to reason about that, if you don't mind submitting a pull request falling back to fcntl'ing FD_CLOEXEC when O_CLOEXEC isn't defined, I'll accept.

zonyitoo:
It should be fixed right now. Please @davidtgoldblatt check https://github.com/jemalloc/jemalloc/pull/961 this pr. We should not call `fcntl` if `open` fails.

davidtgoldblatt:
Resolved by #959 and #961.

PR #959, opened 2017-07-20
Compatible with old FD_CLOEXEC

zonyitoo:
Ref: #946

### DIFF
--- a/src/pages.c
+++ b/src/pages.c
@@ -353,14 +353,31 @@ os_overcommits_proc(void) {
 	ssize_t nread;
 
 #if defined(JEMALLOC_USE_SYSCALL) && defined(SYS_open)
-	fd = (int)syscall(SYS_open, "/proc/sys/vm/overcommit_memory", O_RDONLY |
-	    O_CLOEXEC);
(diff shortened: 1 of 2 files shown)


src/pages.c
```diff
@@ -353,14 +353,32 @@ os_overcommits_proc(void) {
...
-	fd = open("/proc/sys/vm/overcommit_memory", O_RDONLY | O_CLOEXEC);
+	#if defined(O_CLOEXEC)
+		fd = open("/proc/sys/vm/overcommit_memory", O_RDONLY | O_CLOEXEC);
+	#else
+		fd = open("/proc/sys/vm/overcommit_memory", O_RDONLY);
+	#endif
 #endif
+
+#if !defined(O_CLOEXEC)
+	fcntl(fd, F_SETFD, fcntl(fd, F_GETFD) | FD_CLOEXEC);
```
davidtgoldblatt: Let's put an fcntl right next to each open/syscall call instead of right at the end -- that will make it easier to see the logic that needs to be retained as this gets refactored.

davidtgoldblatt:
And, thanks for the patch!

PR #962, opened 2017-07-21
Logging: allow logging with empty varargs.

davidtgoldblatt:
Currently, the log macro requires at least one argument after the format string,
because of the way the preprocessor handles varargs macros.  I realized that we
can eliminate this requirement by splitting up some of our formatting calls.

### DIFF
--- a/include/jemalloc/internal/log.h
+++ b/include/jemalloc/internal/log.h
@@ -2,13 +2,16 @@
 #define JEMALLOC_INTERNAL_LOG_H
 
 #include "jemalloc/internal/atomic.h"
+#include "jemalloc/internal/malloc_io.h"
 #include "jemalloc/internal/mutex.h"
 
 #ifdef JEMALLOC_LOG
-#  define JEMALLOC_LOG_BUFSIZE 1000
+#  define JEMALLOC_LOG_VAR_BUFSIZE 1000
 #else
-#  define JEMALLOC_LOG_BUFSIZE 1
+#  define JEMALLOC_LOG_VAR_BUFSIZE 1
 #endif
+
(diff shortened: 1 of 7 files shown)


include/jemalloc/internal/log.h
```diff
@@ -84,12 +84,15 @@ if (config_log) {							\
...
 do {									\
-	log_do_begin(log_var)						\
-		malloc_printf("%s: " format "\n",			\
-		    (log_var).name, __VA_ARGS__);			\
-	log_do_end(log_var)						\
+	log_do_begin(JEMALLOC_VA_ARGS_HEAD(__VA_ARGS__))		\
+		malloc_printf("%s: ",					\
+		    (JEMALLOC_VA_ARGS_HEAD(__VA_ARGS__)).name);		\
+		malloc_printf(JEMALLOC_VA_ARGS_TAIL(__VA_ARGS__));	\
+		malloc_printf("\n");					\
```
interwq: any chance we can fold one of the printfs? snprinf doesn't seem too pretty here also.

src/jemalloc.c
```diff
@@ -2364,7 +2364,7 @@ je_free(void *ptr) {
 		}
 		check_entry_exit_locking(tsd_tsdn(tsd));
 	}
-	log(log_core_free_exit, "%s", "");
```
interwq: or maybe we should consider simply have a macro for ```"%s", ""```

include/jemalloc/internal/log.h
davidtgoldblatt: It's hard, because we'd have to insert some of our things in the middle of some of theirs, and we wouldn't know if we need a comma or not. It's hard to fix that portably and without a bunch of macro hacks.

PR #964, opened 2017-07-24
Logging: log using the log var names directly.

davidtgoldblatt:
Currently we have to log by writing something like:

  static log_var_t log_a_b_c = LOG_VAR_INIT("a.b.c");
  log (log_a_b_c, "msg");

This is sort of annoying.  Let's just write:

  log("a.b.c", "msg");


davidtgoldblatt:
Note that this isn't an improvement in every dimension -- it gets easier to typo a log var name string. I think the syntactic convenience of this way overwhelms that though.

PR #976, opened 2017-08-04
Don't dump memory obtained via extent_grow_retained until we touch it.

davidtgoldblatt:
This can reduce the size of core dumps depending on kernel configuration.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -171,6 +171,29 @@ extent_lock_from_addr(tsdn_t *tsdn, rtree_ctx_t *rtree_ctx, void *addr) {
 	return ret;
 }
 
+/*
+ * The effect of this function is slightly weaker than the name implies.  In
+ * fact, it only makes the extent's memory dumpable if we were the ones to make
+ * it undumpable (as opposed to a user extent hook).  See the note on the
+ * "dumpable" bit in extent_t.
+ */
+static void
+extent_ensure_dumpable(extent_t *extent) {
+	if (extent != NULL && !extent_dumpable_get(extent)) {
+		pages_dodump(extent_base_get(extent), extent_size_get(extent));
+		extent_dumpable_set(extent, true);
+	}
+}
+
+static void
+extent_prevent_dumpable(extent_t *extent) {
+	if (extent != NULL && extent_dumpable_get(extent)) {
+		pages_dontdump(extent_base_get(extent),
+		    extent_size_get(extent));
+		extent_dumpable_set(extent, false);
+	}
+}
+
 extent_t *
 extent_alloc(tsdn_t *tsdn, arena_t *arena) {
 	malloc_mutex_lock(tsdn, &arena->extent_avail_mtx);
@@ -449,8 +472,11 @@ extents_alloc(tsdn_t *tsdn, arena_t *arena, extent_hooks_t **r_extent_hooks,
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, 0);
 
-	return extent_recycle(tsdn, arena, r_extent_hooks, extents, new_addr,
-	    size, pad, alignment, slab, szind, zero, commit, false);
+	extent_t *extent = extent_recycle(tsdn, arena, r_extent_hooks, extents,
+	    new_addr, size, pad, alignment, slab, szind, zero, commit, false);
+	extent_ensure_dumpable(extent);
+
+	return extent;
 }
 
 void
@@ -1094,7 +1120,8 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
 
 	extent_init(extent, arena, ptr, alloc_size, false, NSIZES,
 	    arena_extent_sn_next(arena), extent_state_active, zeroed,
-	    committed);
+	    committed, true);
+
 	if (ptr == NULL) {
 		extent_dalloc(tsdn, arena, extent);
 		goto label_err;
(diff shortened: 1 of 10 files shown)


src/extent.c
```diff
@@ -1930,6 +1967,11 @@ extent_merge_impl(tsdn_t *tsdn, arena_t *arena,
 		return true;
 	}
 
+	if (extent_dumpable_get(a) || extent_dumpable_get(b)) {
```
interwq: Tricky part like this makes me wonder if we should go the other direction:
1) each arena keeps a special "no_dump" extent, which is the left over from ```grow_retained()```.
2) in grow_retained(), check the no_dump_extent first, and split / enable_dump as appropriate.

That way we sort of limit the scope of this dump / nodump thing in grow_retained itself. However the logic also feels more specialized.

The other thing is, this adds a branch on extent_alloc() which may translate to a tiny bit of cost. It won't be anything noticeable but makes this core dump thing a bit more irritating...

Thoughts?

interwq:
Hmm looks like the freebsd bots are down?

src/extent.c
davidtgoldblatt: We actually still need the logic in a lot of places, because of the possibility of aligned allocations (which can cause splits even inside of already split regions). We also have the merging issue either way, since we could have a no-dump extent next to a dumpable extent by chance or because of growth patterns.

I think the way we'd have to do it is by making "undumpable" its own extent state, and giving it its own extents_t in the arena. This is nice conceptually in some ways (if retained extents are ones we try to avoid using, retained, nondumpable extents are the ones we *really* try to avoid using), but OTOH I don't like the idea of adding another set of stats with different semantics, and shifting the dumping functionality to the arena code.

davidtgoldblatt:
Bots are back up.

interwq:
I just realized we'd have to specialize when extent_hooks is set: in the hooks the user may set DONT_DUMP (e.g. the NoDumpAllocator in folly does this), which maybe reverted by us with this change.

davidtgoldblatt:
Ugh, you're right, we should only do this for the default extent hooks[1]. Why don't we try this out to verify our theory that it's useless, and I'll only bother fixing it if it turns out it helps.

[1] I don't dare to think of what will happen if the user changes things mid-use, though.

src/extent.c
```diff
@@ -1105,6 +1126,13 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
 		goto label_err;
 	}
 
+	/*
+	 * There are situations in which the Linux kernel will include
```
djwatson: Be more specific here, especially with steps to repro (I hate when I find comments like this years later and don't know if I can remove or modify it...)

Linux *always* includes the mapped memory in dumps (except with MADV_DONTDUMP), it just happens to be sparsified with modern filesystems, and not with piped core_patterns

djwatson:
Is it too crazy to turn this on, for, say, all unallocated extents > x pages?  What's the perf overhead of this?

davidtgoldblatt:
The perf overhead of this as is is relatively small (once we reach steady state we stop doing the extra madvises).

I'm hesitant to turn this on as part of the purging process though, since someone looking at the core of a program with use-after-free bugs might quite reasonably want to know what values their program is getting when it touches that invalid memory.

src/extent.c
davidtgoldblatt: How do you feel about:

When core-dumping, the Linux kernel will include the whole address space in its core dumps, even pages we haven't touched yet. This can increase the size of core dumps, and the amount of time it takes to create them. To avoid these costs, we'll set the memory as undumpable until we're ready to use it.

davidtgoldblatt: Really, I guess just losing "there are situations", since the kernel always does it.

djwatson: sounds good to me, thanks

davidtgoldblatt:
OK, the newest iteration makes it so that we won't fiddle with dumpability on memory allocated with extent hooks. I also updated the comments of the extent_t struct, which I realized I had forgotten to do.

davidtgoldblatt:
One thing I'm thinking of is also setting undumpable pages as PROT_NONE, so that the error scenarios in which a program is accidentally using unallocated memory but can't see that memory in the core dump will go away.

src/extent.c
```diff
@@ -1930,6 +1980,11 @@ extent_merge_impl(tsdn_t *tsdn, arena_t *arena,
 		return true;
 	}
 
+	if (extent_dumpable_get(a) || extent_dumpable_get(b)) {
+		extent_ensure_dumpable(a);
+		extent_ensure_dumpable(b);
```
interwq: Just realized, this could make no dump state go away, if we happen to release and merge with the large remaining undumpable extent from grow_retained. Is that true? If that's the case, I'm start to leaning toward having a specific extent in the arena (and never merge with it, only alloc from it when trying to grow_retain).

davidtgoldblatt: <discussed offline>: we'll try making grow_retained its own state (which at any given point only 0 or 1 extents per arena can be in), and upgrade it to just retained before actually growing retained. This avoids some RSS explosion / VM fragmentation issues.

davidtgoldblatt: That should be \<discussed offline\>: at the beginning.

djwatson: I'm probably just missing something, but shouldn't we just not merge if the dumbable states differ?

davidtgoldblatt: It's not crazy. We didn't want to create yet another source of fragmentation caused by adjacent extents in different states.

davidtgoldblatt:
Superseded by #1023.

PR #980, opened 2017-08-08
Remove external linkage for spin_adaptive

rlibby:
The external linkage for spin_adaptive was not used, and the inline
declaration of spin_adaptive that was used caused a probem on FreeBSD
where CPU_SPINWAIT is implemented as a call to a static procedure for
x86 architectures.

A little more explanation:

On FreeBSD, CPU_SPINWAIT is defined as cpu_spinwait(), which for amd64
and i386 is in turn defined as ia32_pause, which is declared as static
inline.

The problem is, the C standard says that inline non-static procedures
(e.g. spin_adaptive) are not allowed to call static procedures (e.g.
ia32_pause) (ISO/IEC 9899:201x 6.7.4,
http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf).

At some point gcc started emitting a warning for this.  FreeBSD is now
using clang by default but I have been trying to get the amd64/gcc build
functional again.  This warning in jemalloc is one of the last
obstacles.

This could perhaps be solved differently (FreeBSD could maybe change
ia32_pause to a macro?) but since the external linkage of spin_adaptive
doesn't actually seem to be used and since this seems to be the only
use of "extern inline" in jemalloc, it seems to make sense to me just to
remove the extern definition.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -112,7 +112,6 @@ C_SRCS := $(srcroot)src/jemalloc.c \
 	$(srcroot)src/prof.c \
 	$(srcroot)src/rtree.c \
 	$(srcroot)src/stats.c \
-	$(srcroot)src/spin.c \
 	$(srcroot)src/sz.c \
 	$(srcroot)src/tcache.c \
 	$(srcroot)src/ticker.c \
--- a/include/jemalloc/internal/spin.h
+++ b/include/jemalloc/internal/spin.h
@@ -1,19 +1,13 @@
 #ifndef JEMALLOC_INTERNAL_SPIN_H
 #define JEMALLOC_INTERNAL_SPIN_H
 
-#ifdef JEMALLOC_SPIN_C_
-#  define SPIN_INLINE extern inline
-#else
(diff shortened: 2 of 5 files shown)


davidtgoldblatt:
Seems reasonable to me. I'll merge once tests come back green.

Is clang the official future of freebsd these days? Fears of having to be compatible with gcc 4.2 have stopped us from pulling the trigger on C++.

rlibby:
> Seems reasonable to me. I'll merge once tests come back green.

Great, thanks!

> Is clang the official future of freebsd these days? Fears of having to be compatible with gcc 4.2 have stopped us from pulling the trigger on C++.

I can't speak for the official messaging, but yes my peronsal
understanding is:
 - Clang is the future for FreeBSD's built-in compiler,
 - FreeBSD will still try to be compiler agnostic,
 - Gcc may still be needed for some time to build architectures for
   which clang/llvm support is not yet mature (but perhaps via an
   external toolchain with gcc more modern than 4.2).

PR #985, opened 2017-08-11
Couple of portability fixes

paravoid:
- GNU/kFreeBSD support is broken as far as I can tell, by assuming that because the system runs glibc, it also has some features that are Linux-only. Workaround that by creating a new kFreeBSD entry, that enables the combination of features that glibc + kFreeBSD have. The right fix would probably be to split the `$host` case to two, one of which matching the kernel, and another one the libc, but that's way over my head :)
- A few Debian porters have submitted bug reports for (still unofficial) ports they're building, with the lg quantum size that is required for their architecture, and this just incorporates them upstream.

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_types.h
+++ b/include/jemalloc/internal/jemalloc_internal_types.h
@@ -79,7 +79,13 @@ typedef int malloc_cpuid_t;
 #  ifdef __hppa__
 #    define LG_QUANTUM		4
 #  endif
+#  ifdef __m68k__
+#    define LG_QUANTUM		3
+#  endif
 #  ifdef __mips__
+#    define LG_QUANTUM		3
+#  endif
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/jemalloc_internal_types.h
```diff
@@ -94,7 +100,7 @@ typedef int malloc_cpuid_t;
 #  ifdef __s390__
 #    define LG_QUANTUM		4
 #  endif
-#  ifdef __SH4__
+#  if (defined (__SH3E__) || defined(__SH4_SINGLE__) || defined(__SH4__) || defined(__SH4_SINGLE_ONLY__))
```
davidtgoldblatt: Nit: 80-character line limit.

paravoid:
Fixed. Thanks for the super quick reviews folks!

PR #986, opened 2017-08-11
Fix test/unit/pages

interwq:
As part of the metadata_thp support, We now have a separate swtich
(JEMALLOC_HAVE_MADVISE_HUGE) for MADV_HUGEPAGE availability.  Use that instead
of JEMALLOC_THP (which doesn't guard pages_huge anymore) in tests.


davidtgoldblatt:
Better commit message though please.

interwq:
This is blocking a production test run. Landing now since the risk is rather low.

PR #989, opened 2017-08-14
Begin arena/tcache disentanglement

davidtgoldblatt:
The first two commits in this series pull out most of the arena/tcache interactions into a cache_bin_t helper object.

What remains is to move more of the allocation path logic to sit on top of the tcache directly, instead of indirecting through the arena code. Helpful in that transition will be the third commit, which documents the somewhat terse naming conventions of the ialloc functions.


include/jemalloc/internal/tcache_structs.h
```diff
@@ -7,21 +7,46 @@
...
-	ql_elm(tcache_t) link;		/* Used for aggregating stats. */
-	arena_t		*arena;		/* Associated arena. */
-	szind_t		next_gc_bin;	/* Next bin to GC. */
+
+	/*
+	 * This data is less hot; we can be a little less careful with our
+	 * footprint here.
+	 */
+	/* Lets us track all the tcaches in an arena. */
+	ql_elm(tcache_t) link;
```
interwq: Shall we kill this? I guess I don't see it being that useful now. Or let's kill it separately if you'd like.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -5,6 +5,24 @@
 #include "jemalloc/internal/sz.h"
 #include "jemalloc/internal/witness.h"
 
+/*
+ * Translating the names of the 'i' functions:
```
interwq: 👍

src/tcache.c
```diff
@@ -180,18 +180,18 @@ tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, tcache_bin_t *tbin,
 	memmove(tbin->avail - rem, tbin->avail - tbin->ncached, rem *
 	    sizeof(void *));
 	tbin->ncached = rem;
-	if ((low_water_t)tbin->ncached < tbin->low_water) {
+	if ((cache_bin_sz_t)tbin->ncached < tbin->low_water) {
```
davidtgoldblatt: Good point, will fix.

include/jemalloc/internal/tcache_structs.h
davidtgoldblatt: Eventually we can and should kill it. In the short term we still use the tcache links in e.g. jemalloc.c. None of it is fundamental any more though.

davidtgoldblatt:
Updated

include/jemalloc/internal/cache_bin.h
```diff
@@ -0,0 +1,114 @@
...
+ * cache_bin_array_descriptor_t objects and reading out per-bin stats it
+ * contains.  This makes it so that the arena need not know about the existence
+ * of the tcache at all.
+ */
+
+
+/*
+ * The count of the number of cached allocations in a bin.  We make this signed
+ * so that negative numbers can encode "invalid" states (e.g. a low water mark
+ * for a bin that has never been filled).
```
interwq: nit: the -1 value of low_water means the cache has been depleted once.

include/jemalloc/internal/cache_bin.h
```diff
@@ -0,0 +1,114 @@
...
+	/* Pointers to the tcache bins. */
+	cache_bin_t *bins_small;
+	cache_bin_t *bins_large;
+};
+
+static inline void
+cache_bin_array_descriptor_init(cache_bin_array_descriptor_t *descriptor,
+    cache_bin_t *bins_small, cache_bin_t *bins_large) {
+	ql_elm_new(descriptor, link);
+	descriptor->bins_small = bins_small;
```
interwq: nit: can we use some offsetof/macro magic here so that we avoid the two pointers?

include/jemalloc/internal/cache_bin.h
davidtgoldblatt: Fixed, thanks.

include/jemalloc/internal/cache_bin.h
davidtgoldblatt: Not without making the cache bin aware of the layout of the tcache itself, which reintroduces a circular dependency. We could save a word by encoding pointer differences relative to the link, but it doesn't seem worth it given that space in this area of the tcache struct is pretty cheap.

davidtgoldblatt:
microbenchmarks look good.

Issue #937, opened 2017-06-26
static TLS errors from jemalloc 5.0.0 built on CentOS 6


wesm:
I help maintain packages on conda-forge which has become fairly popular in the Python community. We recently added jemalloc 5.0.0 to the package manager, built on CentOS 6 with devtoolset-2 from this base Docker image (glibc 2.12 I think)

https://github.com/conda-forge/docker-images/blob/master/linux-anvil/Dockerfile

On some platforms, like Ubuntu 14.04 (glibc 2.19), using dlopen on the produced shared library leads to errors like

```
libjemalloc.so: cannot allocate memory in static TLS block
```

What is the recommended workaround given that we need to compile on a glibc 2.12 system and deploy the binaries on systems with newer glibc?

this may be related to https://sourceware.org/bugzilla/show_bug.cgi?id=14898

cc @xhochy

davidtgoldblatt:
Trying to dlopen jemalloc is bad news bears.

Thread local storage in shared libraries will, by default, require a malloc on first use of the thread-local data by a thread. This can trigger some reentrancy scenarios, and even if you don't, it's easy to accidentally mix malloc-from-jemalloc and malloc-from-glibc. We also don't try to reclaim global data structures on shutdown, so dlopen-ing and dlclose-ing libjemalloc.so repeatedly will leak.

To avoid the reentrancy scenarios, we mark our TLS so that its storage gets allocated with new threads. This works fine for LD_PRELOAD'd (or whatever) shared libraries, but has some problems with dlopen'd ones (where, obviously, the loader doesn't know that it will need that space allocated in advance). glibc tries to save some extra space in advance for any dlopen that happens to come along, but not a ton. jemalloc 5.0 increased the amount of TLS it uses, so it plausibly pushed itself over the limit and can no longer fit in glibc's spare capacity.

We could potentially add a config option to move our TLS out of line, but given the pitfalls here I'm pretty nervous. Could you describe a little more why people want to be able to dlopen libjemalloc.so?

davidtgoldblatt:
(also, in any case, a bugfix 5.0.1 release will be coming out relatively soon that may be worth waiting for before increasing the blast radius of the 5.0 change).

xhochy:
[...]

PR #1005, opened 2017-08-29
Do not assume dss never decreases.

### LINKED ISSUE #802: Do not assume dss break never decreases
```{chunk,extent}_dss_max_update()``` assumes that the dss break will never decrease, but the application could conceivably call ```sbrk()``` twice, the second time with a negative value, and this could cause an infinite loop in jemalloc.  This is an improbable (almost implausible) failure mode, but should be fixed nonetheless.

This issue should be fixed on both the stable-4 and dev branches.

### PR DESCRIPTION
**Pull request by xhochy**
An sbrk() caller outside jemalloc can decrease the dss, so add a
separate atomic boolean to explicitly track whether jemalloc is
concurrently calling sbrk(), rather than depending on state outside
jemalloc's full control.

Fixes #802 for stable-4

This is a backport of https://github.com/jemalloc/jemalloc/commit/196a53c2aec9fd9fbd17af6803d75c7f631c1ae3g

### DIFF
--- a/src/chunk_dss.c
+++ b/src/chunk_dss.c
@@ -20,6 +20,8 @@ static unsigned		dss_prec_default = (unsigned)DSS_PREC_DEFAULT;
 
 /* Base address of the DSS. */
 static void		*dss_base;
+/* Atomic boolean indicating whether a thread is currently extending DSS. */
+static unsigned     dss_extending;
 /* Atomic boolean indicating whether the DSS is exhausted. */
 static unsigned		dss_exhausted;
 /* Atomic current upper limit on DSS addresses. */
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/chunk_dss.c [hunk 1]
```diff
@@ -63,36 +65,40 @@ chunk_dss_prec_set(dss_prec_t dss_prec)
...
 }
 
+static void
+chunk_dss_extending_start(void) {
+	spin_t spinner;
+
+	spin_init(&spinner);
+	while (true) {
+		unsigned expected = 0;
+		if (atomic_cas_u(&dss_extending, expected, 1)) {
```
interwq: The atomic_cas API here returns true on failure. So we would want to flip the condition. Looks this is causing the Travis test failure.

davidtgoldblatt:
This looks good, but I'm going to delay merging until after the weekend so that we can fix our travis build (they updated some of their packages in a way that introduced a clang/libstdc++ incompatibility it seems).

If, as I suspect, you're interested in this as a fix for #937, I'd regard this commit as safe enough for your purposes, even if we don't cut a release specifically for it.

xhochy:
@davidtgoldblatt this is only a workaround for us for #937. With the 4.x series we don't run into #937 but into #802. With this change, we can for now release Apache Arrow with jemalloc as its main allocator, just keeping it on the 4.x series for now.

In the next weeks; I can have a more in depth look into #937 but I first need to get myself acquainted with the matter.

davidtgoldblatt:
Yes, apologies, that's what I meant. I pinged the thread there with an update.

interwq:
@xhochy: just a nitpick, can you please squash the two commits into one? Thanks.

xhochy:
I squashed the commits all into one. Anything that needs to be done to get this merged?

interwq:
Thanks for the patch @xhochy ! LGTM. Merging.

PR #1015, opened 2017-09-07
[Testing] Travis CI fix

interwq:


interwq:
@davidtgoldblatt : the current hacky change seems working https://travis-ci.org/jemalloc/jemalloc/jobs/274747194.

If you are not against avoiding the mutex in the oom handler, I'll do a spin lock there.

davidtgoldblatt:
Why not just use a malloc_mutex_t? We've managed to purge almost all spinlock-esque constructs elsewhere; I'm hesitant to let that number increase.

interwq:
Static initialization for the mutex could be a bit tricky. But should work. Let me try that.

davidtgoldblatt:
I think we can just do static malloc_mutex_t mtx = MALLOC_MUTEX_INITIALIZER, no?

interwq:
IIRC we'll need to do some dance for Windows like this one: https://github.com/jemalloc/jemalloc/blob/e55c3ca26758bcb7f6f1621fd690caa245f16942/src/jemalloc.c#L128-L131

It looks MALLOC_MUTEX_INITIALIZER wasn't doing anything for Windows: https://github.com/jemalloc/jemalloc/blob/18ecbfa89e7dd39a802f52bcd461184b8065e97e/include/jemalloc/internal/mutex.h#L101

interwq:
Well I guess I can stop trying the mutex thing since this one fails: https://travis-ci.org/jemalloc/jemalloc/jobs/274747211

interwq:
@davidtgoldblatt: I guess the tricks here are done. Do you mind taking it over with enforcing the platform version? I haven't looked at that part much.

davidtgoldblatt:
Can do.

Issue #1019, opened 2017-09-12
-fsanitize=undefined errors


marxin:
Hello. I've just run test-suite on x86_64-linux-gnu with -fsanitize=undefined and there are various errors:

[jemalloc.report.txt](https://github.com/jemalloc/jemalloc/files/1295668/jemalloc.report.txt)

marxin:
Adding -Wextra -Wno-unused-arguments shows:

      3 : missing initializer for field ‘l2_cache’ of ‘rtree_ctx_t {aka struct rtree_ctx_s}’ [-Wmissing-field-initializers]
      3 : ‘static’ is not at beginning of declaration [-Wold-style-declaration]
     84 : this statement may fall through [-Wimplicit-fallthrough=]
    156 : left shift of negative value [-Wshift-negative-value]
[jemalloc.report.warnings.txt](https://github.com/jemalloc/jemalloc/files/1295690/jemalloc.report.warnings.txt)

davidtgoldblatt:
Looks like most of these boil down to two missing casts to unsigned (defining delta_inverse_mask and in the mallocx flags arena computation). All in all, I'm surprised at how clean we end up, tbh.

Thanks for checking on this, our lack of sanitization-testing has been making me nervous for a while.

marxin:
Thanks for taking a look, I was just curious whether these cannot lead to a real issue.

interwq:
Thanks for reporting these @marxin! I'll fix these up soon.

interwq:
The warnings reported should be fixed in dev. We may want to add the flags later. Closing.

PR #1022, opened 2017-09-18
Fix warnings 

interwq:
Fix warnings discovered in #1019.

### DIFF
--- a/include/jemalloc/internal/rtree_tsd.h
+++ b/include/jemalloc/internal/rtree_tsd.h
@@ -26,7 +26,7 @@
  * Zero initializer required for tsd initialization only.  Proper initialization
  * done via rtree_ctx_data_init().
  */
-#define RTREE_CTX_ZERO_INITIALIZER {{{0}}}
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
Does the fallthrough thing actually fix the fallthrough warnings? I thought you needed a special attribute.

(We should still probably leave #1019 open after this though, there's a couple other things it found).

interwq:
Comments can also be used mentioned here: https://gcc.gnu.org/gcc-7/changes.html

Also since we have many other places using fall through comments not reported, so I assumed this fixed it.

interwq:
Yeah the left shift by negative thing has been fixed yet.

davidtgoldblatt:
Huh, legit-seeming travis failures?

https://travis-ci.org/jemalloc/jemalloc/jobs/277047620

interwq:
Tried for a while but couldn't repro or relate to the changes here... I'll drop the rtree initializer diff for now and retry.

interwq:
@davidtgoldblatt : these 2 diffs work fine. Let's land them and I'll try other fixes on top?

PR #1023, opened 2017-09-19
Make the grow-retained region non-dumpable

davidtgoldblatt:
This had some surprising subtleties. More comments inline, but the high bit is that we separate out the grow-retained regions into their own extents ("eden" extents), which are moved into an extents_t little by little as we carve new allocations out of it, rather than all at once. The region stays undumpable until we actually decide to let the user touch it.


davidtgoldblatt:
All tests passing now.

Fixing the bug the Windows breakage uncovered actually led me to a nice cleanup; we can be agnostic as to the location of the extent hook failure.

src/extent.c
```diff
@@ -1225,72 +1222,126 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
...
+ * we can split memory for newly requested extents.  This strategy limits the
+ * total number of disjoint virtual memory ranges retained by each arena, which
+ * both helps merging to be effective, and also helps us avoid any kernel limits
+ * on the number of mappings allowed.
+ */
+static extent_t *
+extent_eden_alloc(tsdn_t *tsdn, arena_t *arena,
+    extent_hooks_t **r_extent_hooks, size_t size, size_t pad, size_t alignment,
+    bool slab, szind_t szind, bool *zero, bool *commit) {
+	malloc_mutex_assert_owner(tsdn, &arena->extent_eden_mtx);
```
interwq: nit: document this function unlocks on return.

src/extent.c
```diff
@@ -1225,72 +1222,126 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
...
-			extent_record(tsdn, arena, r_extent_hooks,
-			    &arena->extents_retained, lead, true);
-		}
-		if (trail != NULL) {
-			extent_record(tsdn, arena, r_extent_hooks,
-			    &arena->extents_retained, trail, true);
-		}
-	} else {
+	/*
+	 * We couldn't allocate; either because the grow region was NULL, or
```
interwq: nit: the grow region cannot be NULL anymore when reaching here.

src/pages.c
```diff
@@ -312,6 +312,29 @@ pages_nohuge(void *addr, size_t size) {
 #endif
 }
 
+bool
+pages_dontdump(void *addr, size_t size) {
+	assert(PAGE_ADDR2BASE(addr) == addr);
+	assert(PAGE_CEILING(size) == size);
+#ifdef JEMALLOC_DONTDUMP
```
interwq: I might be missing something. Where is this defined?

src/extent.c
```diff
@@ -1225,72 +1222,126 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
...
+}
+
+/*
+ * If virtual memory is retained, we allocate from the "eden" region of the
+ * arena.  This creates mappings of exponentially increasing size from which
+ * we can split memory for newly requested extents.  This strategy limits the
+ * total number of disjoint virtual memory ranges retained by each arena, which
+ * both helps merging to be effective, and also helps us avoid any kernel limits
+ * on the number of mappings allowed.
+ */
```
interwq: nit: explicitly say eden_alloc is the last step and the slow path.

src/extent.c
davidtgoldblatt: Done.

src/extent.c
davidtgoldblatt: Done.

src/pages.c
davidtgoldblatt: Woops, good catch. Should be JEMALLOC_MADVISE_DONTDUMP, from jemalloc_internal_defs.h.

interwq: It looks this one should be changed as well.

davidtgoldblatt: Ugh. Fixed that one too. Sorry.

davidtgoldblatt:
Performance stress testing indicated some issues here. Nothing that seems unresolvable, but we think state fragmentation is a more important cause of vaddr expansion.

The useful portions of the refactor and metadata support for dumpability have been moved to #1052.

PR #1029, opened 2017-09-22
Add "falls through" comment explicitly.

interwq:
Fix warnings by -Wimplicit-fallthrough.

### DIFF
--- a/include/jemalloc/internal/sz.h
+++ b/include/jemalloc/internal/sz.h
@@ -61,7 +61,7 @@ sz_psz2ind(size_t psz) {
 		pszind_t lg_delta = (x < LG_SIZE_CLASS_GROUP + LG_PAGE + 1) ?
 		    LG_PAGE : x - LG_SIZE_CLASS_GROUP - 1;
 
-		size_t delta_inverse_mask = ZD(-1) << lg_delta;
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/sz.h
```diff
@@ -61,7 +61,7 @@ sz_psz2ind(size_t psz) {
 		pszind_t lg_delta = (x < LG_SIZE_CLASS_GROUP + LG_PAGE + 1) ?
 		    LG_PAGE : x - LG_SIZE_CLASS_GROUP - 1;
 
-		size_t delta_inverse_mask = ZD(-1) << lg_delta;
+		size_t delta_inverse_mask = ZD(ZU(-1) << lg_delta);
```
davidtgoldblatt: I think we don't want a ZD at all, right? No sense in casting an unsigned to a signed back to an unsigned.

interwq: Yeah this doesn't really do anything. But I think we could get warnings from -Wsign-conversion w/o it?

davidtgoldblatt: Will we? It's unsigned shifted by unsigned, right? Or a ZU(lg_delta).

interwq:
@davidtgoldblatt : removed the extra cast.

PR #1033, opened 2017-09-26
Fixing return code error check.

haribert:
Previously the returncode of the if statement was returned rather then
the returncode of the command. Removed unnecessary for loop.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -4,13 +4,13 @@ which autoconf
 returncode=$?
 if [ $returncode -ne 0 ]; then
     echo "Error $returncode in autoconf"
-    exit $returncode 
+    exit 1
 fi
 
 echo "./configure --enable-autogen $@"
 ./configure --enable-autogen $@
 returncode=$?
 if [ $returncode -ne 0 ]; then
-    echo "Error $returncode in ./confinure"
+    echo "Error $returncode in ./configure"
     exit $returncode 
 fi
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,17 +1,16 @@
 #!/bin/sh
 
-for i in autoconf; do
-    echo "$i"
-    $i
-    if [ $? -ne 0 ]; then
-	echo "Error $? in $i"
-	exit 1
-    fi
-done
+which autoconf
+returncode=$?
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
This never actually runs autoconf, right?

I don't actually think we gain anything from returning the error code of the failing command; a script calling autogen.sh shouldn't try to guess the failing subcommand; it just needs a binary success/fail signal.

haribert:
`autogen.sh` does not execute autoconf, but returns the path where it is installed. Or the path where it could not find it.

The return code is only additional information, can be useful in some situation. Otherwise it can just be ignored, and check if it is greater then zero (fail signal).

davidtgoldblatt:
What I mean is, this script will always execute incorrectly, since it never generates the configure script that it wants to execute.

Is there a "root cause" issue you're trying to resolve (e.g. a bug or a desired feature or something)?

haribert:
My commit is perhaps a little bit misleading.

My root cause was the line

> echo "Error $? in ./configure"

which returned "`Error 0 in ./configure`" (because the $? references to the if statement above) and is confusing as an error message. Therefore I introduced the variable `$returncode` and decided to tidy up the rest of the script. But if you want to leave `exit 1` for consistency reasons, feel free to change it back.

autogen.sh
```diff
@@ -1,17 +1,16 @@
...
+fi
 
 echo "./configure --enable-autogen $@"
 ./configure --enable-autogen $@
-if [ $? -ne 0 ]; then
-    echo "Error $? in ./configure"
-    exit 1
+returncode=$?
+if [ $returncode -ne 0 ]; then
+    echo "Error $returncode in ./confinure"
```
djwatson: spelling

interwq:
@haribert: can you please squash the 2 commits into 1? It makes the history a bit cleaner.

davidtgoldblatt:
Two things:
- As Qi mentioned but it seems deleted, the squash
- Let's actually run autoconf instead of just printing the path (i.e. no behavior changes for this script).

interwq:
Closing as no recent update. Please feel free to reopen if needed.

PR #1034, opened 2017-09-26
Add runtime detection of lazy purging support.

interwq:
It's possible to build with lazy purge enabled but depoly to systems without
such support.  In this case, rely on the boot time detection instead of keep
making unnecessary madvise calls (which all returns EINVAL).

### DIFF
--- a/src/pages.c
+++ b/src/pages.c
@@ -27,6 +27,9 @@ static bool	os_overcommits;
 
 bool thp_state_madvise;
 
+/* Runtime support for lazy purge. Irrelevant when !pages_can_purge_lazy. */
+static bool pages_can_purge_lazy_runtime = true;
+
 /******************************************************************************/
 /*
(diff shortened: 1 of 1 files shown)


src/pages.c
```diff
@@ -491,5 +501,19 @@ pages_boot(void) {
...
 	init_thp_state();
 
+	/* Detect lazy purge runtime support. */
+	if (pages_can_purge_lazy) {
+		bool committed = false;
+		void *madv_free_page = os_pages_map(NULL, PAGE, PAGE, &committed);
+		if (madv_free_page == NULL) {
+			return true;
+		}
+		assert(pages_can_purge_lazy_runtime);
```
djwatson: It seems like this just asserts that pages_boot() is only called once?

interwq: Yeah. Added it mostly to make the code easier to follow.

PR #1035, opened 2017-09-29
ARM: Don't extend bit LG_VADDR to compute high address bits.

davidtgoldblatt:
In userspace ARM on Linux, zero-ing the high bits is the correct way to do this.
This doesn't fix the fact that we currently set LG_VADDR to 48 on ARM, when in
fact larger virtual address sizes are coming soon.  We'll cross that bridge when
we come to it.


davidtgoldblatt:
I fixed the sign-extension issue, but didn't tweak our LG_VADDR settings (so, we still don't accomodate 52-bit virtual address spaces). In my testing, this results in a clean build on ARM.

@paravoid, could you confirm that jemalloc built at this diff fixes things on ARM on your end as well?

Once we have this in, I'll add testing on ARM to our CI builds, so that we avoid similar breakages in the future.

marxin:
Thanks, I can confirm it works for me.

PR #1039, opened 2017-10-02
Use ph instead of rb tree for extents_avail_

### LINKED ISSUE #888: Change extent_avail to a heap
We don't need any of the operations the rb tree gives us over the heap, and we like our heap code better. There are some lifetime-y issues in play, and we need to be careful not to expand the size of the extent struct, but it should be doable.

djwatson:
As far as I can tell, there does not seem to be any overlap between usage of
extent_avail and extent_heap, so we can use the same hook.

The only remaining usage of rb trees is in the profiling code,
which has some 'interesting' iteration constraints.

Fixes #888

### DIFF
--- a/include/jemalloc/internal/extent_structs.h
+++ b/include/jemalloc/internal/extent_structs.h
@@ -5,7 +5,6 @@
 #include "jemalloc/internal/bitmap.h"
 #include "jemalloc/internal/mutex.h"
 #include "jemalloc/internal/ql.h"
-#include "jemalloc/internal/rb.h"
 #include "jemalloc/internal/ph.h"
 #include "jemalloc/internal/size_classes.h"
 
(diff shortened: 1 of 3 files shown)


include/jemalloc/internal/extent_structs.h
```diff
@@ -120,20 +119,17 @@ struct extent_s {
...
+	 * - arena_bin_t's slabs_full
+	 * - extents_t's LRU
+	 * - stashed dirty extents
+	 * - arena's large allocations
+	 */
+	ql_elm(extent_t)	ql_link;
 
-	/* Linkage for per size class sn/address-ordered heaps. */
+	/* Linkage for per size class sn/address-ordered heaps, and for
+           extent_avail */
```
interwq: nit: multi-line comment format.

PR #1041, opened 2017-10-03
Logging: capitalize log macro.

davidtgoldblatt:
Dodge a name-conflict with the math.h logarithm function. D'oh.


davidtgoldblatt:
The arm failure is in the --enable-prof configuration, which isn't supported. I'll get rid of that CI config tomorrow, but merge in the meantime to minimize any rebasing pain.

PR #1046, opened 2017-10-05
Enable a0 metadata thp on the 3rd base block.

interwq:
Since we allocate rtree nodes from a0's base, it's pushed to over 1 block on
initialization right away, which makes the auto thp mode less effective on a0.
We change a0 to make the switch on the 3rd block instead.

### DIFF
--- a/src/base.c
+++ b/src/base.c
@@ -126,10 +126,58 @@ base_extent_init(size_t *extent_sn_next, extent_t *extent, void *addr,
 }
 
 static bool
-base_is_single_block(base_t *base) {
-	assert(base->blocks != NULL &&
-	    (base->blocks->size & HUGEPAGE_MASK) == 0);
-	return (base->blocks->next == NULL);
+base_auto_thp_triggered(base_t *base, bool with_new_block) {
+	assert(opt_metadata_thp == metadata_thp_auto);
+	base_block_t *b1 = base->blocks;
+	assert(b1 != NULL);
+
+	base_block_t *b2 = b1->next;
+	if (base_ind_get(base) != 0) {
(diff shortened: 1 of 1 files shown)


interwq:
As discussed offline, we should think about moving rtree node allocation out of base, or find a way to specialize them (e.g. using a dedicated base allocator instead of a0's base).

src/base.c
```diff
@@ -233,20 +258,36 @@ base_block_alloc(tsdn_t *tsdn, base_t *base, extent_hooks_t *extent_hooks,
...
-		if (base != NULL && base_is_single_block(base) &&
-		    opt_metadata_thp == metadata_thp_auto) {
-			/* Make the first block THP lazily. */
-			base_block_t *first_block = base->blocks;
-			assert((first_block->size & HUGEPAGE_MASK) == 0);
-			pages_huge(first_block, first_block->size);
-			if (config_stats) {
-				assert(base->n_thp == 0);
-				base->n_thp += first_block->size >> LG_HUGEPAGE;
+		} else if (opt_metadata_thp == metadata_thp_auto &&
```
davidtgoldblatt: It feels like a lot of this logic belongs in base_auto_thp_switch (or, that logic belongs here or something); the decision making about what and where thps should be turned on are better if unified. Right now the fact that a0 is special is sort of threaded through implicitly in a lot of places rather than having one place where we say "here's the a0 logic and there's the non-a0 logic".

interwq: Good point. I'll consolidate the two places.

interwq: Just updated. Now the logic is contained in base_auto_thp_triggered and base_auto_thp_switch.

PR #1053, opened 2017-10-16
Fix MSVC 2015 project and add a VS 2017 solution

rustyx:
This fixes the MSVC 2015 build and adds an MSVC 2017 solution.


jasone:
Thanks!

PR #1056, opened 2017-10-20
Use statically defined PAGE_SIZE under FreeBSD.

trasz:
Doing this using sysctl(2) serves no purpose - the sysctl always
returns the same constant value.  This shaves off one syscall on
binary startup.

Signed-off-by: Edward Tomasz Napierala <trasz@FreeBSD.org>

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -131,7 +131,8 @@ stats_arena_bins_print(void (*write_cb)(void *, const char *), void *cbopaque,
 		    "\t\t\t\t\"bins\": [\n");
 	} else {
 		char *mutex_counters = "   n_lock_ops    n_waiting"
-		    "   n_spin_acq  total_wait_ns  max_wait_ns\n";
+		    "   n_spin_acq n_owner_switch  total_wait_ns"
+		    "  max_wait_ns max_n_thds\n";
 		malloc_cprintf(write_cb, cbopaque,
 		    "bins:           size ind    allocated      nmalloc"
 		    "      ndalloc    nrequests      curregs     curslabs regs"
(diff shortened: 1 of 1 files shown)


src/pages.c
```diff
@@ -357,6 +357,8 @@ os_page_detect(void) {
 	SYSTEM_INFO si;
 	GetSystemInfo(&si);
 	return si.dwPageSize;
+#elif defined(__FreeBSD__)
+	return (PAGE_SIZE);
```
davidtgoldblatt: Nit: jemalloc doesn't do the trailing return type thing. Otherwise LGTM.

davidtgoldblatt: That should be "parenthesized return value". Brain isn't totally turned on yet.

interwq: @trasz: this function is for sanity check of (compile time page size == runtime page size). Does the macro always reflect the runtime value on FreeBSD? I have to be honest I'm no expert there.

trasz: Shouldn't this sanity check be disabled when built with MALLOC_PRODUCTION?  I've just tried that, and it seems it isn't.

trasz: David, thanks for the hint.  I've amended the commit.

interwq: We do need the check on the fly, in order to make sure the runtime value matches compile time setting (a number of internal constants like size classes are determined based on page size). This is needed because we may build on and deploy to different machines, potentially with different page sizes.

trasz: Okay.  Still, on FreeBSD this check is not needed - page size is compile-time constant, and the sysctl simply returns that constant.  Since jemalloc and the kernel are two parts of the same source tree, changing the constant and rebuilding would still keep those two in sync.

interwq: Well, for the built-in jemalloc, yes. However jemalloc can still be built and deployed separately, e.g. for updating to a newer version or static linking to a different version. So for strict correctness, I’m afraid the runtime check is still necessary.

PR #1061, opened 2017-10-26
FreeBSD binary startup improvements

trasz:
The purpose of those commits is to do one sysctl(2) syscall during binary startup instead of current three.

### DIFF
--- a/src/pages.c
+++ b/src/pages.c
@@ -357,6 +357,8 @@ os_page_detect(void) {
 	SYSTEM_INFO si;
 	GetSystemInfo(&si);
 	return si.dwPageSize;
+#elif defined(__FreeBSD__)
+	return getpagesize();
 #else
 	long result = sysconf(_SC_PAGESIZE);
 	if (result == -1) {
--- a/src/pages.c
+++ b/src/pages.c
@@ -10,6 +10,9 @@
 
 #ifdef JEMALLOC_SYSCTL_VM_OVERCOMMIT
 #include <sys/sysctl.h>
+#ifdef __FreeBSD__
+#include <vm/vm_param.h>
+#endif
 #endif
 
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
Looks reasonable to me, but I'll wait for @interwq since he had concerns with the previous diff, too.

Is jemalloc startup time something you/FreeBSD more generally are trying to optimize for, or are you just taking a minor win where you can get one? I bet if we sat down and thought about it we could find some more places.

src/pages.c
```diff
@@ -373,9 +378,19 @@ os_overcommits_sysctl(void) {
 	size_t sz;
 
 	sz = sizeof(vm_overcommit);
+#if defined(__FreeBSD__) && defined(VM_OVERCOMMIT)
+	int mib[2];
+
+	mib[0] = CTL_VM;
+	mib[1] = VM_OVERCOMMIT;
+	if (sysctl(mib, 2, &vm_overcommit, &sz, NULL, 0) == -1) {
```
interwq: nit: use ```!= 0``` to match the other branch

davidtgoldblatt:
I *think* any update to the commit should trigger another run (I added @trasz to the whitelist).

I opened https://github.com/jemalloc/jemalloc-ci/issues/10; I think we should kill the whitelist.

trasz:
Hm, so what's the preferred course of action here?  Should I just amend the commit to change that comparison, and then "push -f"?

As for the startup time - yeah, it kind of turned out to be a miniproject of sorts, to get rid of all the unnecessary syscalls during binary startup.  I've already got rid of unnecessary path lookups; sysctl-s are pretty much done with this PR.

davidtgoldblatt:
Yeah, that should do it (and should also trigger CI).

I'll try and ping you on an issue if we think of a startup time improvement idea.

interwq:
It looks our CI wasn't triggered @davidtgoldblatt -- do we have to close and re-open the pull request?

davidtgoldblatt:
Looks like the commit wasn't amended yet?

interwq:
I think it has been changed. The previous comment is now covered by "show outdated".

davidtgoldblatt:
Ugggh, left my laptop at work and don't have an appropriate key at home. Sorry, @trasz, I'll try to fix things first thing Monday (or test manually over the weekend).

trasz:
Ping?

davidtgoldblatt:
Hi @trasz, apologies again for the amount of dumb stuff you've had to wade through for this. I killed the whitelist entirely for our CI; could you try one more time to tweak the commit somehow and amend?

trasz:
Okay.  Is it ok now?

davidtgoldblatt:
Not sure why CI still isn't triggering, but I tested manually. Merging.

PR #1062, opened 2017-10-31
metadata_thp: auto mode adjustment for a0.

interwq:
We observed that arena 0 can have much more metadata allocated comparing to
other arenas.  Tune the auto mode to only switch to huge page on the 5th block
(instead of 3 previously) for a0.


interwq:
Discussed offline with @davidtgoldblatt about this. Given that a0 usually experience higher metadata usage, while we want to avoid triggering auto THP switch for short-lived programs (regardless of narenas setting), we want to tune a0 to make the switch on the 4th block.

Issue #1009, opened 2017-08-30
Extent hook wishlist


davidtgoldblatt:
Tracking issue for some extent hook things we've come across

- #952 (let hooks overallocate if they want)
- An indication of whether or not an allocation is for metadata or not (example application: allocation within memory sharing mappings with hardware. Metadata can be CPU-only, but allocations need to be within the memory shared with the hardware).

interwq:
An additional use case came up with 1G huge page:
- grow retained has an increasingly larger sequence for expanding VM
- the sequence has no limit by default; however with 1G huge pages, physical memory is reserved already. grow_retained will obtain real committed physical memory
- in addition, when the grow sequence becomes high, it could cause bad fragmentation, e.g. asking for 800MB when grow_retained, while we have 0.5G available from a huge page.

#1064 is a workaround the issue with the current API. To solve this properly, we would want the API to incorporate more info, e.g. size_required and size_desired so that the extent hooks can determine how much memory to return.

davidtgoldblatt:
A way to segregate thread-caches for different sets of extent hooks. (e.g. we really don't want no-dump memory leaking into non-nodump allocations).

davidtgoldblatt:
Let users specify up-front whether or not their set of hooks should be considered to allow overcommit.

davidtgoldblatt:
A "zero" option. Right now, we may call pages_purge_forced in the middle of a custom extent, which is more or less never desired.

interwq:
Support for extent head / interior state, or to indicate a mergeable / splittable state -- as mentioned in #1545, on Windows we need to track if an extent is the head of a VirtualAlloc region (to avoid cross-region merge). Currently this information is tracked internally with a `is_head` bit, and we check the state before calling user hooks (instead of forcing the user to track this information).

davidtgoldblatt:
Two:
- The ability to mark a specific extent as "eager", in that it gets decayed / dalloc'd as soon as possible. This would let extent hooks from different arenas coordinate behind the scenes to share address space (say, that lives in a huge page somewhere).
- The ability to set retain settings on a per-arena basis (different extent hooks may have different opinions there).

davidtgoldblatt:
Information about whether or not an allocation is metadata can be useful on dealloc, too.

davidtgoldblatt:
Letting the user squirrel away a pointer inside an extent (and get it back during extent operations) could be handy.

davidtgoldblatt:
[...]

PR #1064, opened 2017-11-03
Add arena.i.retain_grow_limit

interwq:
This option controls the max size when grow_retained.  This is useful when we
have customized extent hooks reserving physical memory (e.g. 1G huge pages).
Without this feature, the default increasing sequence could result in fragmented
and wasted physical memory.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -1886,6 +1886,33 @@ arena_muzzy_decay_ms_default_set(ssize_t decay_ms) {
 	return false;
 }
 
+bool
+arena_retain_grow_limit_get_set(tsd_t *tsd, arena_t *arena, size_t *old_limit,
+    size_t *new_limit) {
+	assert(opt_retain);
+
+	pszind_t new_ind JEMALLOC_CC_SILENCE_INIT(0);
+	if (new_limit != NULL) {
+		size_t limit = *new_limit;
+		/* Grow no more than the new limit. */
+		if ((new_ind = sz_psz2ind(limit + 1) - 1) >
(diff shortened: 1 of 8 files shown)


interwq:
As discussed offline with @davidtgoldblatt and @djwatson, this solves an immediate issue with 1G hugepages. Long term we'd need to solve this at the extent hook API level (#1009).

src/arena.c
```diff
@@ -1886,6 +1886,33 @@ arena_muzzy_decay_ms_default_set(ssize_t decay_ms) {
 	return false;
 }
 
+bool
+arena_retain_grow_limit_get_set(tsd_t *tsd, arena_t *arena, size_t *old_limit,
+    size_t *new_limit) {
+	assert(opt_retain);
+
+	pszind_t new_ind JEMALLOC_CC_SILENCE_INIT(0);
```
interwq: Previous CI test failures were complaining about this being unused uninitialized.

src/arena.c
```diff
@@ -1886,6 +1886,33 @@ arena_muzzy_decay_ms_default_set(ssize_t decay_ms) {
...
+	if (new_limit != NULL) {
```
davidtgoldblatt: It seems like this should either be an error, or, if we want a null new-limit to mean "unlimited", we should document that.

davidtgoldblatt: Oh wait I'm dumb, there should be a way to get the current limit. nvm.

interwq: Yep. NULL new_limit means reading the value.

interwq:
OSX tests passed previously. Merging.

PR #1065, opened 2017-11-03
Use hugepage alignment for base allocator.

interwq:
This gives us an easier way to tell if the allocation is for metadata in the
extent hooks.


interwq:
This is also a workaround for the items in #1009, specifically checking is_metadata.

Issue #852, opened 2017-05-24
Reduce extent mutex locking


davidtgoldblatt:
Right now we hold a lock on extents->mtx for a number of complicated operations. After the switch to optimistic locking for extents, it's much more feasible to do fine-grained locking for these sorts of operations. We need to hold the mutex pool locks for the involved extents, and a per-size-bucket mutex on the involved heaps in the extents struct.

This gives us more mutex acquisitions in total, but it is still likely a win, since the extents mutex is one of our most contended.

djwatson:
Still in progress trying to get canaries to give reasonable results, but an update:

* Made a mutex pool instead of extents_mtx.  A couple places have to use optimistic locking, and a few things combined (like finding and removing the extent becomes a single operation).
* the rtree mutex pool locks aren't necessary for anything more than they already are: The same number of total mutex operations happens, we just shard up the extents_mtx.
* the prng for extent address randomization is atomic, and becomes a point of contention.  I'll put up a diff to make it threadlocal shortly
* the LRU list is split up, similar to how we shard the extent_mtx lock.
* in extent_coalesce, there are some lru operations that need to be changed, or grab two locks - currently we try to replace the inner with the outer lru list

Currently canaries show we can reduce the number of default arenas by 2-4x after this change, however I'm seeing a memory regression of ~7%.  Suspecting that the lru list changes are affecting purging, but haven't nailed it down yet.

djwatson:
Still in progress trying to get canaries to give reasonable results, but an update:

* Made a mutex pool instead of extents_mtx.  A couple places have to use optimistic locking, and a few things combined (like finding and removing the extent becomes a single operation).
* the rtree mutex pool locks aren't necessary for anything more than they already are: The same number of total mutex operations happens, we just shard up the extents_mtx.
* the prng for extent address randomization is atomic, and becomes a point of contention.  I'll put up a diff to make it threadlocal shortly
* the LRU list is split up, similar to how we shard the extent_mtx lock.
* in extent_coalesce, there are some lru operations that need to be changed, or grab two locks - currently we try to replace the inner with the outer lru list

Currently canaries show we can reduce the number of default arenas by 2-4x after this change, however I'm seeing a memory regression of ~7%.  Suspecting that the lru list changes are affecting purging, but haven't nailed it down yet.

PR #1070, opened 2017-11-09
Use tsd offset_state instead of atomic

djwatson:
While working on #852, I noticed the prng state is atomic.  This is the only
atomic use of prng in all of jemalloc.  Instead, use a threadlocal prng
state if possible to avoid unnecessary cache line contention.


davidtgoldblatt:
Let's seed the tsd state somehow (say, global atomic counter) so that we don't end up having every thread generate the same sequence in lock-step?

PR #1075, opened 2017-11-16
Avoid setting zero and commit if split fails in extent_recycle.

interwq:
Avoid setting zero and commit in extent_recycle_extract.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -791,7 +791,7 @@ static extent_t *
 extent_recycle_extract(tsdn_t *tsdn, arena_t *arena,
     extent_hooks_t **r_extent_hooks, rtree_ctx_t *rtree_ctx, extents_t *extents,
     void *new_addr, size_t size, size_t pad, size_t alignment, bool slab,
-    bool *zero, bool *commit, bool growing_retained) {
+    bool growing_retained) {
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, growing_retained ? 1 : 0);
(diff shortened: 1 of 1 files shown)


interwq:
@davidtgoldblatt: updated the diff after the offline discussion.

davidtgoldblatt:
Why should `extent_recycle_extract` be setting zero at all? It seems to me that we should assign to *zero at most once, once we're assured of success in the main `extent_recycle`.

interwq:
That's valid. Didn't realized extent_recycle_extract was only used in extent_recycle. I'll refactor that part then.

interwq:
@davidtgoldblatt: refactored both zero and commit to be not included in extent_recycle_extract.

src/extent.c
```diff
@@ -1021,23 +1014,25 @@ extent_recycle(tsdn_t *tsdn, arena_t *arena, extent_hooks_t **r_extent_hooks,
...
-		*commit = true;
-	}
 
 	extent = extent_recycle_split(tsdn, arena, r_extent_hooks, rtree_ctx,
 	    extents, new_addr, size, pad, alignment, slab, szind, extent,
 	    growing_retained);
 	if (extent == NULL) {
 		return NULL;
 	}
+	if (extent_committed_get(extent)) {
```
davidtgoldblatt: We can still fail here, no?

davidtgoldblatt: It seems line 1046 (which, weirdly, I don't seem able to comment on) is the earliest place we should set *commit and *zero

interwq: Ah good catch. Just fixed.

PR #1076, opened 2017-11-20
Add missing deregister before extents_leak.

interwq:
This fixes an regression introduced by 211b1f3 (refactor extent split).


aorenste:
No test?

interwq:
I'll try to add a test case there. We probably need more coverage on custom extent hooks' failure paths.

interwq:
Test added in #1079.

PR #1081, opened 2017-12-03
Correct the spelling of __riscv.

EdSchouten:
According to the RISC-V toolchain conventions, `__riscv__` is the old spelling of this definition. `__riscv` should be used going forward.

https://github.com/riscv/riscv-toolchain-conventions#cc-preprocessor-definitions

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_types.h
+++ b/include/jemalloc/internal/jemalloc_internal_types.h
@@ -94,7 +94,7 @@ typedef int malloc_cpuid_t;
 #  ifdef __powerpc__
 #    define LG_QUANTUM		4
 #  endif
-#  ifdef __riscv__
+#  if defined(__riscv) || defined(__riscv__)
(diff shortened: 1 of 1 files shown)


include/jemalloc/internal/jemalloc_internal_types.h
```diff
@@ -94,7 +94,7 @@ typedef int malloc_cpuid_t;
 #  ifdef __powerpc__
 #    define LG_QUANTUM		4
 #  endif
-#  ifdef __riscv__
+#  ifdef __riscv
```
interwq: To make it compatible in both cases, let’s change it to:
```#if defined(__riscv) || defined(__riscv__)```

Also please update the commit message accordingly.

EdSchouten: Done!

interwq:
Thanks for the patch!

PR #1084, opened 2017-12-08
Fix test/integration/extent.

interwq:
Should only run the hook tests without background threads.  This was introduced
in 6e841f6.


interwq:
@davidtgoldblatt : can you take another look? I added a new commit to fix an issue on the extent leak path.

davidtgoldblatt:
Still LGTM. I wish we had some more systematic way of handling gdump stuff, but I don't think it needs to be a precondition to this PR.

PR #1085, opened 2017-12-11
jemalloc_mangle.sh: set sh in strict mode

nicolov:
Set sh in "strict mode" in `jemalloc_mangle.sh`.

I've had a [TensorFlow](https://github.com/tensorflow/tensorflow/blob/master/third_party/jemalloc.BUILD) build fail in weird ways because the sysroot did not have `awk` installed. It would have been nice if the script failed right away instead of producing a malformed header.

### DIFF
--- a/include/jemalloc/jemalloc_mangle.sh
+++ b/include/jemalloc/jemalloc_mangle.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -eu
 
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
Good idea, thanks for the patch. ARM CI is running a little behind at the moment; I'll wait a bit before merging to let it catch up, to make sure we don't have any weird environment issues there.

PR #1091, opened 2017-12-14
Correct miss spelled words

stanman29:
My first pull request to correct miss spelled words

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,8 +9,8 @@ dnl Custom macro definitions.
 
 dnl JE_CONCAT_VVV(r, a, b)
 dnl
-dnl Set $r to the concatenation of $a and $b, with a space separating them iff
-dnl both $a and $b are non-emty.
+dnl Set $r to the concatenation of $a and $b, with a space separating them if
+dnl both $a and $b are non-empty.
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -9,8 +9,8 @@ dnl Custom macro definitions.
 
 dnl JE_CONCAT_VVV(r, a, b)
 dnl
-dnl Set $r to the concatenation of $a and $b, with a space separating them iff
```
davidtgoldblatt: `iff` here is actually used in a technical sense; it's an abbreviation for "if and only if". The "emty" fix looks good, though.

davidtgoldblatt:
(And, although I think it's kind of funny, we should probably do "miss spelled" -> "misspelled" in the commit message.

Good catch!

interwq:
@stanman29: Please also address the other feedback (i.e. not changing iff).

davidtgoldblatt:
And, one last nit: could you make the sequence a single commit without the spelling issue in the title? We disable the "squash and merge" button on our end to prevent accidental squashing, but that also means that however the commit history shows up in the PR will be how it shows up in the repo.

(Sorry for all the back and forth for just style stuff).

PR #1092, opened 2017-12-16
Over purge by 1 extent always.

interwq:
When purging, large allocations are usually the ones that cross the npages_limit
threshold, simply because they are "large".  This means we often leave the large
extent around for a while, which has the downsides of: 1) high RSS and 2) more
chance of them getting fragmented.  Given that they are not likely to be reused
very soon (LRU), let's over purge by 1 extent (which is often large and not
reused frequently).


interwq:
Production benchmark results for this looks surprisingly good, almost too good to be true. I'll re-run the benchmark to be sure.

interwq:
Production results showed this change improve memory usage significantly (multiple percent process wide), and reduce VM size by >30% (which saves metadata usage).

PR #1093, opened 2017-12-17
Pull bin management into its own module

davidtgoldblatt:
There shouldn't be functionality changes here; just pulling out functions and cutting a line between internal and external bin operations. My intent over the longer term is try to split off what discrete chunks of tcache and arena functionality I can, until we get to a point where a more principled disentangling is feasible.


interwq:
Travis has been reporting failures like this pretty often recently: https://travis-ci.org/jemalloc/jemalloc/jobs/317812196

include/jemalloc/internal/bin.h
```diff
@@ -0,0 +1,81 @@
...
+	uint32_t		nregs;
+
+	/*
+	 * Metadata used to manipulate bitmaps for slabs associated with this
+	 * bin.
+	 */
+	bitmap_info_t		bitmap_info;
+};
+
+extern const bin_info_t bin_infos[NBINS];
```
interwq: nit: not sure about ```infos```

davidtgoldblatt: (Discussed offline; bin_infos is OK, since it means we don't have a name collision with any bin_info_t *bin_info variables we might want to create locally).

PR #1095, opened 2017-12-21
Don't hardcode regind divisors.

davidtgoldblatt:
These commits implement a dynamic version of the fast division-by-constants trick. Thanks to some knowledge we have that the compiler doesn't, we can actually get a slightly faster implementation (no dynamic shifting, no subtractions, and only ever one shift), avoiding an indirect jump and a couple of arithmetic operations, and cutting down the size of `arena_slab_regind` from around 860 bytes to around 31 (on x86).

A slow path test of running test/unit/stress with MALLOC_CONF=tcache:false showed about a 2% improvement on my machine.

### DIFF
--- a/include/jemalloc/internal/div.h
+++ b/include/jemalloc/internal/div.h
@@ -1,0 +1,41 @@
+#ifndef JEMALLOC_INTERNAL_DIV_H
+#define JEMALLOC_INTERNAL_DIV_H
+
+#include "jemalloc/internal/assert.h"
+
+/*
+ * This module does the division that computes the index of a region in a slab,
+ * given its offset relative to the base.
+ * That is, given a divisor d, an n = i * d (all integers), we'll return i.
+ * We do some pre-computation to do this more quickly than a CPU division
+ * instruction.
+ * We bound n < 2^32, and don't support dividing by one.
+ */
+
+typedef struct div_info_s div_info_t;
+struct div_info_s {
+	uint32_t magic;
+#ifdef JEMALLOC_DEBUG
+	size_t d;
+#endif
+};
+
+void div_init(div_info_t *div_info, size_t divisor);
+
+static inline size_t
+div_compute(div_info_t *div_info, size_t n) {
+	assert(n <= (uint32_t)-1);
+	/*
+	 * This generates, e.g. mov; imul; shr on x86-64. On a 32-bit machine,
+	 * the compilers I tried were all smart enough to turn this into the
+	 * appropriate "get the high 32 bits of the result of a multiply" (e.g.
+	 * mul; mov edx eax; on x86, umull on arm, etc.).
+	 */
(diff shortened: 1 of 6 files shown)


davidtgoldblatt:
(Fixed a failure due to an implicit cast warning).

include/jemalloc/internal/div.h
```diff
@@ -0,0 +1,35 @@
...
+ * That is, given a divisor d, an n = i * d (all integers), we'll return i.
+ * We do some pre-computation to do this more quickly than a CPU division
+ * instruction.
+ * We bound n < 2^32, and don't support dividing by one.
+ */
+
+typedef struct div_info_s div_info_t;
+struct div_info_s {
+	uint32_t magic;
+#ifndef NDEBUG
```
interwq: why use NDEBUG here? JEMALLOC_DEBUG?

include/jemalloc/internal/div.h
```diff
@@ -0,0 +1,35 @@
...
+	size_t d;
+#endif
+};
+
+void div_init(div_info_t *div_info, size_t divisor);
+
+static inline size_t
+div_compute(div_info_t *div_info, size_t n) {
```
interwq: nit: although straightforward, I'd say still include the expected asm in some comments here.

include/jemalloc/internal/div.h
```diff
@@ -0,0 +1,35 @@
...
+	assert(n <= (uint32_t)-1);
+	size_t i = ((uint64_t)n * (uint64_t)div_info->magic) >> 32;
```
interwq: will the mult instruction be very slow on 32-bit arch in this case (If so should we consider not doing it for them)?

src/arena.c
```diff
@@ -39,6 +40,8 @@ const uint64_t h_steps[SMOOTHSTEP_NSTEPS] = {
 #undef STEP
 };
 
+static div_info_t arena_binind_div_info[NBINS];
```
interwq: maybe create a beginner-friendly task to make this compile time const?

include/jemalloc/internal/div.h
davidtgoldblatt: Will do.

include/jemalloc/internal/div.h
davidtgoldblatt: Will do.

include/jemalloc/internal/div.h
davidtgoldblatt: In the places where I tried, the compiler is smart enough to generate a "multiply high-unsigned" instruction on 32-bit. I added a comment.

src/arena.c
davidtgoldblatt: I figure it doesn't matter so much in a world of dynamic size classes, which we expect to come shortly after this.

davidtgoldblatt:
Only comments changed from last green build; merging.

Issue #994, opened 2017-08-21
Musings about state-based fragmentation


davidtgoldblatt:
I've accumulated enough thoughts about this to want to jot some down. I'll probably edit this over the next few weeks as I clarify my thoughts. State-induced fragmentation can cause use to use extra memory unnecessarily, if scenarios like the following happen (each letter represents the state of one page in a sequential region of memory; A=Active, D=Dirty, M=Muzzy, R=Retained):

`DDMMRR`

Suppose we have a 3-page allocation request come in. We'll have to allocate extra memory for it, getting:

`DDMMRRAAA`

These leads to us having 5 or 7 pages contributing to RSS (depending on the muzzy pages' OS status). Better would have been to upgrade the first muzzy page's status to dirty:

`DDDMRR`

and then satisfy the allocation request that way:

`AAAMRR`

This would leave us 3 or 4 pages contributing to RSS (depending on the muzzy page's OS status).

Here's a way I think we could keep around enough metadata to do a better job:
- Introduce the notion of "big" extents. A big extent of a particular type contains pages of that purge level and higher, and has a "head" extent.
- Each extent keeps track of whether or not it's a head of each type of big extent, and the size of the big extent if so.
- Each inactive extent that's not a head keeps a pointer to the head extent of the big extent it's part of (this is necessary to allow big extent merging).
- We keep big extents in heaps of the same sort we keep regular extents in.
- When the arena fails to allocate an extent from the free extents in a given state, it can instead try to allocate from the free big extents in that state.

Some potential pitfalls:
- extent_t struct bloat: The extra metadata is for the free extents, and so can share space with the slab bitmap.
- Extra big extent heap overhead: I think this is small relative to other costs down these pathways, since, most of the time, we either A. save an OS-level allocation, B. Just did an OS-level page state change (e.g. madvise), or C. Are probably going to take a soft fault when we try to touch memory from the next purging level.

Other random thoughts:
- When doing an allocation out of a set of big extents, we could try to err towards a big extent with more active pages, by changing our heap-ordering of extents.
- When purging, can try to purge pages in a fragmented big extent, to reclaim as much metadata fragmentation as possible.
- We don't have the statistics to track how often this sort of fragmentation really happens. To try to get a better guess, we can try to track the average number of contiguous pages in a single state relative to the avg number of contiguous pages in non-active states (unfortunately, the latter is costly to compute without implementing all the metadata changes. We could do a heavyweight version fairly easily.).
[...]

PR #1096, opened 2017-12-22
Improve the fit for aligned allocation.

interwq:
We compute the max size required to satisfy an alignment.  However this can be
quite pessimistic, especially with frequent reuse (and combined with state-based
fragmentation).  This commit adds one more fit step specific to aligned
allocations, searching in all potential fit size classes.

w/o this diff, looping ```{ ptr = mallocx(2M, align(2M); free(ptr); }``` results in no reuse of any previously freed memory:
`
0x7f917f000000
0x7f917dc00000
0x7f917d600000
0x7f917ce00000
0x7f917d200000
0x7f917c600000
0x7f917ca00000
0x7f917bc00000
0x7f917c000000
0x7f917b000000
`
with this diff:
`
0x7fa53f600000
0x7fa53f600000
0x7fa53f600000
0x7fa53f600000
0x7fa53f600000
0x7fa53f600000
0x7fa53f600000
0x7fa53f600000
0x7fa53f600000
0x7fa53f600000
`

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -363,6 +363,43 @@ extents_remove_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent,
 	    cur_extents_npages - (size >> LG_PAGE), ATOMIC_RELAXED);
 }
 
+/*
+ * Find an extent with size [min_size, max_size) to satisfy the alignment
+ * requirement.  For each size, try only the first extent in the heap.
+ */
+static extent_t *
+extents_fit_alignment(extents_t *extents, size_t min_size, size_t max_size,
+    size_t alignment) {
+        pszind_t pind = sz_psz2ind(extent_size_quantize_ceil(min_size));
+        pszind_t pind_max = sz_psz2ind(extent_size_quantize_ceil(max_size));
+
+	for (pszind_t i = (pszind_t)bitmap_ffu(extents->bitmap,
+	    &extents_bitmap_info, (size_t)pind); i < pind_max; i =
+	    (pszind_t)bitmap_ffu(extents->bitmap, &extents_bitmap_info,
+	    (size_t)i+1)) {
+		assert(i < NPSIZES);
+		assert(!extent_heap_empty(&extents->heaps[i]));
+		extent_t *extent = extent_heap_first(&extents->heaps[i]);
+		uintptr_t base = (uintptr_t)extent_base_get(extent);
+		size_t candidate_size = extent_size_get(extent);
+		assert(candidate_size >= min_size);
+
+		uintptr_t next_align = ALIGNMENT_CEILING((uintptr_t)base,
+		    PAGE_CEILING(alignment));
+		if (base > next_align || base + candidate_size <= next_align) {
+			/* Overflow or not crossing the next alignment. */
+			continue;
+		}
+
+		size_t leadsize = next_align - base;
+		if (candidate_size - leadsize >= min_size) {
+			return extent;
+		}
+	}
+
+	return NULL;
+}
+
 /* Do any-best-fit extent selection, i.e. select any extent that best fits. */
 static extent_t *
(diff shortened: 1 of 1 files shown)


interwq:
Testing in progress. Not ready for review yet.

interwq:
Production testing showed this diff fixed a regression with aligned allocations.

Fundamentally, two factors contributed to the issue: 1) state based fragmentation (#994) because we cannot coalesce dirty and retained extents to satisfy the alignment; and 2) in the alignment path, asking for pessimistic size which is fixed in this PR. Although testing results showed the regression being fixed with (2) alone, we probably should get down to #994 sooner.

Ready for review now.

src/extent.c
```diff
@@ -363,6 +363,43 @@ extents_remove_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent,
...
+	for (pszind_t i = (pszind_t)bitmap_ffu(extents->bitmap,
+	    &extents_bitmap_info, (size_t)pind); i < pind_max; i =
+	    (pszind_t)bitmap_ffu(extents->bitmap, &extents_bitmap_info,
+	    (size_t)i+1)) {
+		assert(i < NPSIZES);
+		assert(!extent_heap_empty(&extents->heaps[i]));
+		extent_t *extent = extent_heap_first(&extents->heaps[i]);
+		assert(extent_size_get(extent) >= min_size);
+		uintptr_t base = (uintptr_t)extent_base_get(extent);
+		size_t esize = extent_size_get(extent);
```
davidtgoldblatt: Let's call this something other than esize? We already use that in the rest of the file to mean "effective size" (i.e. size + pad).

src/extent.c
```diff
@@ -363,6 +363,43 @@ extents_remove_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent,
...
+
+		uintptr_t next_align = ALIGNMENT_CEILING((uintptr_t)base,
```
davidtgoldblatt: It looks like we don't do any overflow checking -- right?

src/extent.c
interwq: Good point. Will fix.

src/extent.c
interwq: Right. Doesn't seem to be needed (if this overflows, we won't and can't use the current extent).

davidtgoldblatt: Can you add a comment with some math? I'm definitely not following.

Suppose there's a very large alignment. Then next_align will be less then base, so we'll pass the first check, so leadsize will be very big, so esize - leadsize will be very big, so we'll return an improper extent, no?

interwq: Will add comments. However if the next_align is less than base, we will skip this extent in the first check, right?

interwq: Hold on. That wasn’t right. Let me fix this.

interwq:
Feedback addressed. @davidtgoldblatt you wanna take another look?

interwq:
Only a style change was made after the last green build. Merging.

edwinsmith:
For large size classes, I understand what went wrong and how this fixes it. For small size classes, managed in slabs with a bitmap, is it [still*] the case that sizes that are powers of 2 are going to be naturally aligned? e.g. IIUC, 4K is a small size, so it would come from a slab containing only same-sized (4K) naturally aligned objects?

* I'm saying "still" because I think with jemalloc 4 and earlier, this was the case... but correct me otherwise.

caveat: I know the APIs don't guarantee this alignment, I'm thinking about data structures that would profit from naturally aligned nodes, and what other pitfalls there might be when asking for aligned blocks.

interwq:
For power of 2 size classes equal to or smaller than page size (default 4K), yes they will be naturally aligned of the same size, e.g 1k / 2k.

For size classes greater than page size such as 8k, no such alignment will be guaranteed (they are still small size classes, but the slab contains only 1 item and is only page aligned).

PR #1100, opened 2018-01-01
handle 32 bit mutex counters

### LINKED ISSUE #1063: Properly handle mutex stats output
For convenience in the stats outputting code, we treat all mutex profiling statistics as 64-bit, which is not actually the case. This can cause uninitialized data to be output, since the ctls the stats boil down to will only write 4 bytes out for 4-byte fields. For the stuff I'm doing in #1058 I can get away with memsetting to 0, but this only happens to work because I'm on a little-endian machine. We should do it the "right" way, and properly size the fields.

rkmisra:
This fixes #1063 , where we are separating 64 bit and 32 bit mutex stats counters.  During read, we read them in their respective types.

### DIFF
--- a/include/jemalloc/internal/mutex_prof.h
+++ b/include/jemalloc/internal/mutex_prof.h
@@ -35,21 +35,34 @@ typedef enum {
 	mutex_prof_num_arena_mutexes
 } mutex_prof_arena_ind_t;
 
-#define MUTEX_PROF_COUNTERS						\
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
That being said, the fact that these sorts of tricks are necessary reconfirms my belief that we should do most of this manually, where we have names and types explicit and configurable.

PR #1101, opened 2018-01-01
Make sure JE_CXXFLAGS_ADD uses CPP compiler

nehaljwani:
All the invocations of `AC_COMPILE_IFELSE` inside `JE_CXXFLAGS_ADD` were
running 'the compiler and compilation flags of the current language'
which was always the `C` compiler and the `CXXFLAGS` were never being tested
against a `C++` compiler. This patch fixes this issue by temporarily
changing the chosen compiler to `C++` by pushing it over the stack and
popping it immediately after the compilation check.


davidtgoldblatt:
Thanks!

Issue #1045, opened 2017-10-04
test_pages_huge:test/unit/pages.c fails on ARM (armv7l)


marxin:
Testing version 5.0.1:

```
./configure --host=armv7hl-suse-linux-gnueabi --build=armv7hl-suse-linux-gnueabi --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib --libexecdir=/usr/lib --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --disable-dependency-tracking --enable-cc-silence --enable-prof
```

I spot:

```
[  266s] === test/unit/pages ===
[  266s] test_pages_huge:test/unit/pages.c:15: Failed assertion: (pages_huge(hugepage, HUGEPAGE)) != (config_thp) --> true == true: Unexpected pages_huge() result
[  266s] test_pages_huge:test/unit/pages.c:17: Failed assertion: (jet_pages_nohuge(hugepage, ((size_t)(1U << 21)))) == (false) --> true != false: Unexpected pages_nohuge() result
[  266s] test_pages_huge (non-reentrant): fail
[  266s] test_pages_huge:test/unit/pages.c:15: Failed assertion: (pages_huge(hugepage, HUGEPAGE)) != (config_thp) --> true == true: Unexpected pages_huge() result
[  266s] test_pages_huge:test/unit/pages.c:17: Failed assertion: (jet_pages_nohuge(hugepage, ((size_t)(1U << 21)))) == (false) --> true != false: Unexpected pages_nohuge() result
[  266s] test_pages_huge (libc-reentrant): fail
[  266s] test_pages_huge:test/unit/pages.c:15: Failed assertion: (pages_huge(hugepage, HUGEPAGE)) != (config_thp) --> true == true: Unexpected pages_huge() result
[  266s] test_pages_huge:test/unit/pages.c:17: Failed assertion: (jet_pages_nohuge(hugepage, ((size_t)(1U << 21)))) == (false) --> true != false: Unexpected pages_nohuge() result
[  266s] test_pages_huge (arena_new-reentrant): fail
[  266s] --- pass: 0/3, skip: 0/3, fail: 3/3 ---
```

lheckemann:
[...]

PR #1102, opened 2018-01-02
Disable JEMALLOC_HAVE_MADVISE_HUGE for arm* CPUs.

marxin:
As mentioned in #1045, the macro should be disabled for arm*.
Unfortunately the patch as is can't be backported to 5.0.x release.


davidtgoldblatt:
Looks great, thanks for the fix.

marxin:
Thanks for review. May I ask you to add the commit to a release you're planning (5.0.2 I guess).
It's small and enables one arch, thus useful for distributions like openSUSE.

PR #1103, opened 2018-01-03
Refactored extent_t to fix #975

### LINKED ISSUE #975: Refactor extent_t bitpacking logic
The bitpacking for e_bits is a little bit brittle, in the sense that adding a new field requires tweaking the math for all the subsequent fields' shifts and masks. Instead, we could add definitions of the number of bits of each field, at which point the mask and shift of a given field can be calculated purely in terms of its definitions and those of the immediately preceding field.

rkmisra:
This commit fixes #975 

New cpp define where each has width and shift of bit fields which is calculated from proceeding field's shift and width and finally mask which is function of current bit field's width and shift.


davidtgoldblatt:
Thanks for taking care of these, @rkmisra. Sorry it took so long to review (I jumped from a vacation right into our big end-of-year-planning season).

PR #1104, opened 2018-01-05
Dynamic slab sizes

davidtgoldblatt:
This makes the slab size selection a runtime option rather than a configure-time one. It also does most of the work to allow the introduction of new size classes at startup time.


davidtgoldblatt:
This passes all tests and is ready for review; +cc @interwq.

src/sc.c
```diff
@@ -0,0 +1,62 @@
...
+#if SC_NTINY != 0
+	data->lg_tiny_maxclass = LG_TINY_MAXCLASS;
+#else
+	data->lg_tiny_maxclass = -1;
+#endif
+	data->lookup_maxclass = LOOKUP_MAXCLASS;
+	data->small_maxclass = SMALL_MAXCLASS;
+	data->lg_large_minclass = LG_LARGE_MINCLASS;
+	data->large_minclass = LARGE_MINCLASS;
+	data->large_maxclass = LARGE_MAXCLASS;
```
interwq: I'm a bit concerned regarding the change from the macros to the global data here. Though maybe a very small difference, I'd still try to use static and immediate values as much as possible, unless we want it to be mutable for the dynamic slabs. Maybe keep LARGE_MAXCLASS etc unchanged?

include/jemalloc/internal/sc.h
```diff
@@ -0,0 +1,302 @@
+#ifndef JEMALLOC_INTERNAL_SC_H
+#define JEMALLOC_INTERNAL_SC_H
+
+#include "jemalloc/internal/jemalloc_internal_types.h"
+
+/*
+ * Size class computations:
```
interwq: 👍

include/jemalloc/internal/sc.h
```diff
@@ -0,0 +1,302 @@
...
+#  error "Too many small size classes"
+#endif
+
+/* The largest size class in the lookup table. */
+#define SC_LOOKUP_MAXCLASS ((size_t)1 << 12)
+
+typedef struct sc_s sc_t;
+struct sc_s {
+	/* Size class index, or -1 if not a valid size class. */
+	int index;
```
interwq: Why we need the index here? I'd imagine we read this with index known already.

include/jemalloc/internal/sc.h
```diff
@@ -0,0 +1,302 @@
...
+	/* Lg group base size (no deltas added). */
+	int lg_base;
+	/* Lg delta to previous size class. */
+	int lg_delta;
+	/* Delta multiplier.  size == 1<<lg_base + ndelta<<lg_delta */
+	int ndelta;
```
interwq: It looks we always use ndelta and lg_delta together. Can we store size directly here instead?

interwq:
Review still in progress.

include/jemalloc/internal/sc.h
davidtgoldblatt: Not any particular reason (this is a dead store in fact). It's mostly there because this started as a direct translation of the bash, and so every bash variable used in sc scope got its own place in the sc_t.

I figured it might come in handy some day; a use case where some module wants an sc_t, and wants to know its szind_t but nothing else seemed reasonable to me.

davidtgoldblatt: (I don't really have any strong opinions though).

include/jemalloc/internal/sc.h
davidtgoldblatt: This I think is more useful to keep; if we want to do nontrivial tweaking of the size class selection algorithm later on we'll probably end up reinserting these.

src/sc.c
davidtgoldblatt: Fair enough; especially for the things on the fast path, I think being risk-averse with these makes sense. I made them dynamic here in anticipation of dynamic size class selection (i.e. if we want to make a 24-byte size class a runtime option), where we will probably need this to be truly dynamic (or else insert some other sort of bound), but we can cross that bridge when we come to it.

davidtgoldblatt:
(Added back the SMALL_MAXCLASS etc. macros; leaving in the "unneeded, but useful to have around" data for now unless you have a strong opinion to the contrary).

src/sc.c
```diff
@@ -1,62 +1,291 @@
...
+static int
+slab_size(int lg_page, int lg_base, int lg_delta, int ndelta) {
+	size_t page = (ZU(1) << lg_page);
+	size_t reg_size = reg_size_compute(lg_base, lg_delta, ndelta);
+
+	size_t try_slab_size = page;
+	size_t try_nregs = try_slab_size / reg_size;
+	size_t perfect_slab_size = 0;
+	bool perfect = false;
+	while (!perfect) {
```
interwq: nit: add some comments about this being okay since in practice the common case is a few pages.

src/jemalloc.c
```diff
@@ -1192,6 +1235,33 @@ malloc_conf_init(void) {
...
+						    &sc_data_global, slab_start,
+						    slab_end, (int)pgs);
+					} else {
+						malloc_conf_error(
+						    "Invalid settings for "
+						    "slab_sizes", k, klen, v,
+						    vlen);
+					}
+				} while (!err && vlen_left > 0);
+				if (!err) {
```
interwq: nit: No need to check `!err`?

test/integration/slab_sizes.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# Some screwy-looking slab sizes.
+export MALLOC_CONF="slab_sizes:1-4096:17|100-200:1|128-128:2"
```
interwq: What do you think about adding another set of default "aggressive" parameters? i.e. instead of using minimal number of pages, that option will use max or high number of pages, and we can enable it through a single malloc_conf option. It can make the feature easier to use.

configure.ac
```diff
@@ -1430,11 +1430,6 @@ if test "x${LG_PAGE}" != "xundefined" -a \
 fi
 AC_DEFINE_UNQUOTED([LG_HUGEPAGE], [${je_cv_lg_hugepage}])
 
-AC_ARG_WITH([lg_page_sizes],
-  [AS_HELP_STRING([--with-lg-page-sizes=<lg-page-sizes>],
-   [Base 2 logs of system page sizes to support])],
-  [LG_PAGE_SIZES="$with_lg_page_sizes"], [LG_PAGE_SIZES="$LG_PAGE"])
-
```
interwq: Also update `INSTALL.md`.

cc: @jasone that this option is no longer needed with this PR.

src/sc.c
davidtgoldblatt: Added a comment; the strict bound here is 2 * ngroup - 1.

src/jemalloc.c

include/jemalloc/internal/sc.h
interwq: Sorry forgot to mention this: I still prefer removing this for now, since the data here is pretty hot, and only pay the cost when we have some direct use of it.

davidtgoldblatt: Discussed offline -- the data here isn't hot.

davidtgoldblatt:
@jasone, any reason we'd want to keep --with-lg-page-sizes if we do all the computation directly instead of through the shell script?

jasone:
I haven't had a chance to review the diff, but the only use for
--with-lg-page-sizes is to specify the full set of page sizes used across
all FreeBSD targets.  If that use is obsoleted, feel free to rip it out;
otherwise preserve it to allow subsequent imports into FreeBSD's libc.

davidtgoldblatt:
Yep, should be unnecessary for importing after the diff.

Issue #1004, opened 2017-08-29
Cache extent serial numbers in the extents_t


davidtgoldblatt:
When doing first-fit, we touch every pointed-to extent in every extents_t heap for size-classes that could satisfy our request, while holding extents->mtx. We could cache the serial numbers of the current oldest extent in the extents_t bins themselves, allowing us to avoid cache misses reading information from extents that we won't end up using.

PR #1111, opened 2018-01-13
Extents cache issue 1004

rkmisra:
Fixing issue #1004 , basically a cached sn is used during first fit search.  This cached sn is updated during insert and remove operations in paired heap.  If sn is same during comparison, we load extent_t to compare e_addr.


davidtgoldblatt:
Hi, sorry for the long review delay; I spent the past few weeks on vacation.

This mostly looks good, I'll try to do a more in-depth review shortly. I want to do some test runs on production services to make sure we see some improvements, as well.

A couple git nits:
- It looks like this has some commits unrelated to the issue (maybe an accidental rebase off of master instead of dev?).
- We try to maintain a merge-free history for dev, with each commit staying mostly self-contained. Could you patch just commit 0ac520f onto dev, and update the commit message to be a little more self-contained? (Referencing the github issue is fine, but scanning through commit titles and summaries should still let the reader know approximately the commit contents).

rkmisra:
Thanks David and sorry about the git mess up as I am new to git.  I have created a new pull request #1121 which only has the relevant changes.  We can discard this current pull request.

davidtgoldblatt:
Thanks for the new PR! I'll close this one for now to move discussion there.

PR #1121, opened 2018-01-31
Cache the serial number of oldest/first extent.

rkmisra:
Fixing issue #1004. Cache the serial number of oldest/first extent in extent_t.  serial number is used during search of first fit extent. Caching sn should avoid filling cache lines with unnecessary extent data which is discarded soon as only sn info is needed during search. This cached sn is updated during insert and remove operations in paired heap. If sn is same during comparison, we load extent_t to compare e_addr.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -352,12 +352,20 @@ extents_remove_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent,
 	size_t psz = extent_size_quantize_floor(size);
 	pszind_t pind = sz_psz2ind(psz);
 	size_t extent_sn = extent_sn_get(extent);
+	bool first = extent_heap_first_after_remove(
+												&extents->heaps[pind], extent);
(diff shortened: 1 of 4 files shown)


src/extent.c
```diff
@@ -319,6 +319,8 @@ extents_insert_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent,
 		    (size_t)pind);
 	}
 	extent_heap_insert(&extents->heaps[pind], extent);
+	extent_t *first_extent = extent_heap_first(&extents->heaps[pind]);
```
djwatson: The current pairing heap code currently does a merge for the first() operation, changing the complexity here from O(1) to O(log n).  Can we just do a comparison against the current lowest_sn_heaps instead? (or change the pairing heap code...)

davidtgoldblatt:
I still think this is a good idea, but this PR has bitrotted from (our) neglect. We have a slew of pending optimization ideas for this code, ones that we can approach holistically.

Issue #1122, opened 2018-01-31
Segfault issue with multi-thread


JacobHeAMD:
When jemalloc is integrated into another library, for example, mylib.so,
the issue could be reproduced with following simple code,
   void\* hd = NULL;
   int vulkan_draw()
   \{
       hd = dlopen("mylib.so", RTLD_LAZY);
       sleep(1);
       dlclose(hd);
   \}

   int main(int argc, char \*\*argv) \{
       pthread_t ntid;
       int err = pthread_create(&ntid, NULL, vulkan_draw, NULL);
       pthread_join(ntid,NULL);
   \}

   Root Cause:
   A thread key is created and set by jemalloc in global constructor, however it's not set NULL / deleted before dl-closed. As a consequence, there is something left in thread local storage even after dl-closing, and it will be referenced during thread exiting (pthreads trying to free those resources). That causes segfault.

davidtgoldblatt:
We don't really support being dlclose'd (for example, note that even if we removed the indicated pthread key, we'd still leak everything in the thread caches). I'd like to one day live in a world where a clean shutdown feature isn't too hard to implement, but we're a long ways off from that world. Closing for now.

PR #1124, opened 2018-01-31
To fix the issue: https://github.com/jemalloc/jemalloc/issues/1122

JacobHeAMD:
Solution: Set the thread key NULL explicitly in global destructor.


davidtgoldblatt:
Closing for now -- I don't think we want to half-support being unloaded from a process, per the comment in #1122.

PR #1131, opened 2018-02-12
Tweak the ticker paths to help GCC generate better code.

davidtgoldblatt:
GCC on its own isn't quite able to turn the ticker subtract into a memory
operation followed by a js.

It's hard to reliably measure things that small, but it looks to be about a 0.2% end-to-end improvement.

### DIFF
--- a/include/jemalloc/internal/ticker.h
+++ b/include/jemalloc/internal/ticker.h
@@ -32,14 +32,42 @@ ticker_read(const ticker_t *ticker) {
 	return ticker->tick;
 }
 
+/*
+ * Not intended to be a public API.  Unfortunately, on x86, neither gcc nor
+ * clang seems smart enough to turn
+ *   ticker->tick -= nticks;
+ *   if (unlikely(ticker->tick < 0)) {
+ *     fixup ticker
+ *     return true;
+ *   }
+ *   return false;
+ * into
+ *   subq %nticks_reg, (%ticker_reg)
+ *   js fixup ticker
+ *
+ * unless we force "fixup ticker" out of line.  In that case, gcc gets it right,
+ * but clang now does worse than before.  So, on x86 with gcc, we force it out
+ * of line, but otherwise let the inlining occur.  Ordinarily this wouldn't be
+ * worth the hassle, but this is on the fast path of both malloc and free (via
+ * tcache_event).
+ */
+#if defined(__GNUC__) && !defined(__clang__)				\
+    && (defined(__x86_64__) || defined(__i386__))
+JEMALLOC_NOINLINE
+#endif
+static bool
+ticker_fixup(ticker_t *ticker) {
+	ticker->tick = ticker->nticks;
+	return true;
+}
+
 static inline bool
 ticker_ticks(ticker_t *ticker, int32_t nticks) {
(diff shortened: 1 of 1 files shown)


include/jemalloc/internal/ticker.h
```diff
@@ -32,14 +32,50 @@ ticker_read(const ticker_t *ticker) {
...
+#if defined(__GNUC__) && !defined(__clang__)				\
+    && (defined(__x86_64__) || defined(__i386__))
+/*
+ * -Wattributes complains about noinline inline functions, even though that's
+ *  not always incorrect (noinline being about optimization, and "inline" being
+ *  about symbol conflicts).  We'll disable -Wattributes temporarily.  Ideally,
+ *  we'd pragma push and pragma pop around the definition, but we need to
+ *  compile on GCC versions that don't support it.
+ */
+#pragma GCC diagnostic ignored "-Wattributes"
```
interwq: Maybe something like  
```
static JEMALLOC_NOINLINE

#else 

static inline bool

#endif 
```
to avoid the Wattributes hassle?

include/jemalloc/internal/ticker.h
```diff
@@ -32,14 +32,50 @@ ticker_read(const ticker_t *ticker) {
...
 static inline bool
 ticker_ticks(ticker_t *ticker, int32_t nticks) {
-	if (unlikely(ticker->tick < nticks)) {
-		ticker->tick = ticker->nticks;
-		return true;
-	}
 	ticker->tick -= nticks;
-	return(false);
+	if (unlikely(ticker->tick < 0)) {
+		return ticker_fixup(ticker);
```
interwq: Just curious, will ```ticker_fixup(); return true;``` make any difference in the output?

include/jemalloc/internal/ticker.h
davidtgoldblatt: Hah. Ordinarily we don't do just plain static because the compiler will *also* complain about an unused static function. Here it's fine, but muscle memory is strong. I'll change it.

include/jemalloc/internal/ticker.h
davidtgoldblatt: Discussed offline: it probably shouldn't matter, but because of further compiler screwiness it does.

I'm inclined to say we should leave it for now. In the long run, I'd like to do a better job of figuring out how to handle bad compiler optimization more systematically.

interwq: I’d suggest still switch to the other way even if it’s only a tiny bit better? Also because the two representations both make sense (we can think of ticker_fixup as no failure allowed).

davidtgoldblatt: Oh yeah, I think we're saying the same thing. You're suggesting to keep the code as it is in the diff, right?

interwq: Which version had better output? I was thinking keeping that one.

davidtgoldblatt: Yeah, the current version has better output / a faster fast path.

interwq:
Probably rebase and test AppVeyor again to be safe?

davidtgoldblatt:
Tests pass after the appveyor fix.

PR #1132, opened 2018-02-14
CI: Remove "catgets" dependency on appveyor.

davidtgoldblatt:
This seems to cause a configuration error with msys2.


davidtgoldblatt:
I'm not sure if this will work or not actually; I have no idea what catgets is or why there's a conflict.

davidtgoldblatt:
Yep, it works.

PR #1134, opened 2018-02-16
Add opt.thp which allows explicit hugepage usage.

### LINKED ISSUE #1127: jemalloc 5 and transparent hugepages
What is the expected behavior/interactions with jemalloc 5 and transparent hugepages? I saw there were some changes made in #243 but from my cursory browsing of the code it seems like there are no actual calls to `pages_nohuge` in jemalloc 5. I see #932 is open but I wanted to understand what the expected behavior and motivation for **not** calling MADV_NOHUGEPAGE in jemalloc 5 is.

To give some additional context, we have an application with lots of threads (~1000 is typical) and with jemalloc 5 we've seen instances where the application's RSS is 50% higher than what jemalloc itself reports (based on malloc_stats_print's `mapped` stat). We've turned off THP for now, but of course we don't want to be forced to do so just because of jemalloc. Also, would you recommend switching to jemalloc 4 (we haven't tried this yet)?

interwq:
"always" marks all user mappings as MADV_HUGEPAGE; while "never" marks all
mappings as MADV_NOHUGEPAGE. The default setting "default" does not change any
settings.  Note that all the madvise calls are part of the default extent hooks
by design, so that customized extent hooks have complete control over the
mappings including hugepage settings.

This resolves #1127

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1876,27 +1876,6 @@ case "${host_cpu}" in
 esac
 fi
 
-dnl Enable transparent huge page support by default.
-AC_ARG_ENABLE([thp],
-  [AS_HELP_STRING([--disable-thp],
-                  [Disable transparent huge page support])],
-[if test "x$enable_thp" = "xno" -o "x${je_cv_thp}" != "xyes" ; then
-  enable_thp="0"
-else
-  enable_thp="1"
-fi
-],
-[if test "x${je_cv_thp}" = "xyes" ; then
-  enable_thp="1"
-else
-  enable_thp="0"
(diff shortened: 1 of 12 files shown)


davidtgoldblatt:
I might just be dumb, but I can't actually tell what the appveyor failure is. It looks like all tests passed?

interwq:
@davidtgoldblatt : no idea what went wrong with AppVeyor there. It looks fine now.

I also removed config.thp in this PR, as far as I could tell it wasn't doing anything.

PR #1136, opened 2018-02-22
RFC: Remove extents->mtx

### LINKED ISSUE #852: Reduce extent mutex locking

djwatson:
Arena concurrency seems to be bounded by the extents->mtx in real workloads.  The pairing heap is a somewhat expensive datastructure to modify (much more expensive than lists), and we hold a single lock while modifying any of the heaps.

This diff introduces an arena mutex pool to lock the heaps on a per-size-class basis.   Other things protected by extents->mtx (bitmap, npages, lru), use separate mutexes. 

In a single-arena-many-thread microbench, this improves the performance of arena by 2-5x (depending on workload).

I have tested previous versions of this set in production, with slight latency improvements and slight memory regressions.  Needs another round of testing.

Fixes #852


davidtgoldblatt:
(CI machine down for a reboot; test coverage should be coming soon).

src/extent.c
```diff
@@ -541,11 +541,15 @@ extents_evict(tsdn_t *tsdn, arena_t *arena, extent_hooks_t **r_extent_hooks,
 	 * the loop will iterate until the LRU extent is fully coalesced.
 	 */
 	extent_t *extent;
+	int lists = NPSIZES;
```
davidtgoldblatt: Doesn't this mean we'll (effectively) never end up purging small extents so long as we can find a large one? I would think better would be to shard randomly N-ways, and cycle through the lru list we use to purge (i.e. move `lists` into the `extents_t`).

This will also reduce contention on the LRU lists if lots of threads are using a single size class often.

src/extent.c
```diff
@@ -357,10 +363,11 @@ extents_remove_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent) {
  * requirement.  For each size, try only the first extent in the heap.
  */
 static extent_t *
-extents_fit_alignment(extents_t *extents, size_t min_size, size_t max_size,
-    size_t alignment) {
-        pszind_t pind = sz_psz2ind(extent_size_quantize_ceil(min_size));
-        pszind_t pind_max = sz_psz2ind(extent_size_quantize_ceil(max_size));
+extents_fit_alignment(tsdn_t *tsdn, arena_t * arena, extents_t *extents,
+		      size_t min_size, size_t max_size,
```
davidtgoldblatt: Nit: formatting is screwy

src/extent.c
```diff
@@ -474,8 +487,8 @@ extents_fit_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents,
 		 * pessimistic.  Next we try to satisfy the aligned allocation
 		 * with sizes in [esize, max_size).
 		 */
-		extent = extents_fit_alignment(extents, esize, max_size,
-		    alignment);
+		extent = extents_fit_alignment(tsdn, arena, extents, esize, 
+				max_size, alignment);
```
davidtgoldblatt: formatting again.

src/extent.c
```diff
@@ -902,16 +915,14 @@ extent_recycle_extract(tsdn_t *tsdn, arena_t *arena,
 			}
 			extent_unlock(tsdn, unlock_extent);
 		}
+		if (extent) {
+			extent_activate_locked(tsdn, arena, extents, extent);
```
davidtgoldblatt: Is this a no-op? It seems like everywhere we *get* an extent from ought to be activating it for us.

include/jemalloc/internal/mutex_pool.h
```diff
@@ -60,27 +60,26 @@ mutex_pool_postfork_child(tsdn_t *tsdn, mutex_pool_t *pool) {
 
 static inline void
 mutex_pool_lock(tsdn_t *tsdn, mutex_pool_t *pool, uintptr_t key) {
-	mutex_pool_assert_not_held(tsdn, pool);
```
davidtgoldblatt: What about instead having each thread keep a few spare pointers to mutex_pool_ts (and a lock count too maybe) in its witness data, and we can just assert that the pool we're locking isn't already there.

Locking pool mutexes incrementally is a fairly easy-to-write but subtle bug that will usually just never show up in testing. If we're expanding pool uses, it'd be nice to know we can't introduce those sorts of bugs.

src/arena.c
```diff
@@ -1795,6 +1798,11 @@ arena_new(tsdn_t *tsdn, unsigned ind, extent_hooks_t *extent_hooks) {
 		}
 	}
 
+        if (mutex_pool_init(&arena->extents_mutex_pool, "extents_mutex_pool",
+                            WITNESS_RANK_EXTENT_HEAPS)) {
```
davidtgoldblatt: Formatting.

src/extent.c
```diff
@@ -317,28 +317,31 @@ extents_npages_get(extents_t *extents) {
...
 static void
-extents_insert_locked(tsdn_t *tsdn, extents_t *extents, extent_t *extent) {
+extents_insert_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents, extent_t *extent) {
 	malloc_mutex_assert_owner(tsdn, &extents->mtx);
-	assert(extent_state_get(extent) == extents->state);
 
 	size_t size = extent_size_get(extent);
 	size_t psz = extent_size_quantize_floor(size);
 	pszind_t pind = sz_psz2ind(psz);
+	mutex_pool_lock(tsdn, &arena->extents_mutex_pool, pind);
```
davidtgoldblatt: Why not just have NPSIZES mutexes, instead of a whole pool? Pools are pretty big, and doing a 1:1 bin:pool mapping seems strictly better.

Actually, now that I mention it, how do you prevent extent operations from racing with coalescing ones? It seems like you need some per-size-class locks *and* some locks in the existing extent pool.

src/extent.c
```diff
@@ -478,16 +499,16 @@ extents_fit_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents,
 	}
 
 	extent_t *extent = extents->delay_coalesce ?
-	    extents_best_fit_locked(tsdn, arena, extents, max_size) :
-	    extents_first_fit_locked(tsdn, arena, extents, max_size);
+		extents_best_fit_locked(tsdn, arena, extents, max_size) :
```
davidtgoldblatt: formatting

src/extent.c
```diff
@@ -311,8 +311,7 @@ extents_npages_get(extents_t *extents) {
 }
```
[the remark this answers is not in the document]
djwatson: super!

include/jemalloc/internal/mutex_pool.h
djwatson: sure, sounds reasonable

src/extent.c
djwatson: No, we get it by addr here, the one place it isn't already activated for us.

src/extent.c
djwatson: We could, but I didn't want to duplicate the pool code.  We could make the pool size configurable?  

Extent operations racing - The coalesce *does* race, but the can_coalesce function is enough to bail out if something changed.

The activate/deactivate don't really race, because you don't coalesce active extents, and we check for heap empty after locking, so we'd just try the next heap.

The only place I had race issues was activating by address (realloc), where we have to double check after locking nothing has changed.

src/extent.c
```diff
@@ -364,8 +373,8 @@ extents_fit_alignment(extents_t *extents, size_t min_size, size_t max_size,
 	    (pszind_t)bitmap_ffu(extents->bitmap, &extents_bitmap_info,
 	    (size_t)i+1)) {
 		assert(i < NPSIZES);
-		assert(!extent_heap_empty(&extents->heaps[i]));
 		extent_t *extent = extent_heap_first(&extents->heaps[i]);
```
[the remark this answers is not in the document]
djwatson: This assert doesn't need to be removed till a subsequent diff I guess, but we don't lock the heap until after bitmap_ffu, so someone may have come along and changed the heap.  bitmap is locked for updates, and the heap is locked for updates, but they are not locked w.r.t. each other.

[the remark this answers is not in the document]
davidtgoldblatt: Is the heap locked here? I guess my question is, what's protecting you during all these heap operations.

More generally, could you add a file-level comment explaining the locking strategy and which mutexes protect which operations?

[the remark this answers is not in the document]
djwatson: I think the confusion is looking halfway through the stack, this diff is still protected by extents->mtx, the heap pool lock is introduced in a later diff, and yes it is held here, but after the bitmap lookup.

See https://github.com/jemalloc/jemalloc/pull/1136/files#diff-7b26b977303fe92c093a2245b0eaf255R374

src/extent.c
```diff
@@ -306,80 +313,70 @@ extents_npages_get(extents_t *extents) {
...
 	for (pszind_t i = (pszind_t)bitmap_ffu(extents->bitmap,
 	    &extents_bitmap_info, (size_t)pind); i < pind_max; i =
 	    (pszind_t)bitmap_ffu(extents->bitmap, &extents_bitmap_info,
 	    (size_t)i+1)) {
 		assert(i < NPSIZES);
-		assert(!extent_heap_empty(&extents->heaps[i]));
+		mutex_pool_lock(tsdn, &arena->extents_mutex_pool, i);
 		extent_t *extent = extent_heap_first(&extents->heaps[i]);
+		if (!extent) {
+			mutex_pool_unlock(tsdn, &arena->extents_mutex_pool, i);
```
interwq: nit: change the two unlock+continue to a goto the unlock at the end of the loop?

also, change to `if (extent != NULL) `, and a few other places with `if (!extent)` or `if (extent)`.

src/extent.c
```diff
@@ -407,17 +408,26 @@ extents_best_fit_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents,
 	pszind_t pind = sz_psz2ind(extent_size_quantize_ceil(size));
 	pszind_t i = (pszind_t)bitmap_ffu(extents->bitmap, &extents_bitmap_info,
 	    (size_t)pind);
-	if (i < NPSIZES+1) {
+	while (i < NPSIZES+1) {
```
interwq: nit: the function probably shouldn't be called _locked

src/extent.c
```diff
@@ -438,11 +448,22 @@ extents_first_fit_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents,
...
 		extent_t *extent = extent_heap_first(&extents->heaps[i]);
+		if (!extent) {
+			mutex_pool_unlock(tsdn, &arena->extents_mutex_pool, i);
+			continue;
+		}
 		assert(extent_size_get(extent) >= size);
 		if (ret == NULL || extent_snad_comp(extent, ret) < 0) {
+			extent_activate_locked(tsdn, arena, extents, extent);
+			mutex_pool_unlock(tsdn, &arena->extents_mutex_pool, i);
+			if (ret) {
```
interwq: nit: if (ret != NULL)

also maybe some comments about why activate + deactivate is needed?

src/extent.c
```diff
@@ -896,17 +911,26 @@ extent_recycle_extract(tsdn_t *tsdn, arena_t *arena,
 			}
 			extent_unlock(tsdn, unlock_extent);
 		}
+		if (extent) {
+			size_t size = extent_size_get(extent);
+			size_t psz = extent_size_quantize_floor(size);
+			pszind_t pind = sz_psz2ind(psz);
+			mutex_pool_lock(tsdn, &arena->extents_mutex_pool, pind);
+			/* Can race with other recycle_extract, double check
```
interwq: nit: formatting

interwq:
I think @djwatson mentioned there was a slight memory increase with this. Is that because the additional mutex pool?

davidtgoldblatt:
Even absent the mutex pools, I'd expect there to be an increase in "true" memory use; with this change, it's more likely that allocation requests will "miss" free (but being touched by another thread, and hence not appearing free) extents, right?

djwatson:
Canary testing shows there is still some concurrency bug in the extent merging code.  Debugging as time allows.

Breaking out individual diffs if they make sense (preserve lru, mutex pool, possibly atomic npages / atomic bitmap diffs)

djwatson:
After discussion, we will try flat combining as an alternative - this only provided ~2x improvement to lock contention, which is roughly what previous flat combining experiments provided.

djwatson:
Sure, we can try cache first - although flat combining could be used anywhere locks are used, and would not cause more fragmentation like a cache would, so it may be worth trying both anyway.

PR #1139, opened 2018-02-27
Add deterministic testing support

djwatson:
This diff adds a deterministic scheduler to jemalloc.  In debug mode,
the deterministic scheduler *may* be activated for testing purposes.
It is extremely useful in testing concurrency changes, to ensure
correctness.  See notes at top of deterministic.c for usage.

Jemalloc's mutex and atomic classes are instrumented to support
determinsitic scheduling.  Currently only gcc_atomic is instrumented,
but others are similar.

In non-debug mode, it should compile out completely, they are inlined empty functions.  In debug mode with the scheduler off, it is an extra branch or two for each mutex or atomic call.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -97,6 +97,7 @@ C_SRCS := $(srcroot)src/jemalloc.c \
 	$(srcroot)src/bitmap.c \
 	$(srcroot)src/ckh.c \
 	$(srcroot)src/ctl.c \
+	$(srcroot)src/deterministic.c \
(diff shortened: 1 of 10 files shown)


interwq:
Sorry for the delay here Dave. This looks good to me, however let's merge this after 5.1, given the scope of change is fairly broad?

davidtgoldblatt:
Bitrotted. Still probably a good idea, though; we can grab what we want at some point in the future.

PR #1144, opened 2018-03-07
Convert stats printing to use a structured text emitter

davidtgoldblatt:
We introduce the emitter module, which mostly transparently handles the differences of the human-readable and json output modes. Gone are the days of manually indenting every json field, or manually aligning table rows with their fields, or any of the other reasons why it's super annoying to deal with the stats code.

The actual LOC-level savings in stats.c understate the improvement, since many of the newly added lines are much easier to read than their pre-conversion equivalents. The emitter itself is tested as well, so that it's harder to accidentally introduce incorrect json or misaligned tables.

### DIFF
--- a/include/jemalloc/internal/emitter.h
+++ b/include/jemalloc/internal/emitter.h
@@ -1,0 +1,381 @@
+#ifndef JEMALLOC_INTERNAL_EMITTER_H
+#define JEMALLOC_INTERNAL_EMITTER_H
+
+typedef enum emitter_output_e emitter_output_t;
+enum emitter_output_e {
+	emitter_output_json,
+	emitter_output_table
+};
+
+typedef enum emitter_justify_e emitter_justify_t;
+enum emitter_justify_e {
+	emitter_justify_left,
+	emitter_justify_right,
+	/* Not for users; just to pass to internal functions. */
+	emitter_justify_none
+};
+
+typedef enum emitter_type_e emitter_type_t;
+enum emitter_type_e {
+	emitter_type_bool,
+	emitter_type_int,
+	emitter_type_unsigned,
+	emitter_type_uint32,
+	emitter_type_uint64,
+	emitter_type_size,
+	emitter_type_ssize,
+	emitter_type_string,
+};
+
+typedef struct emitter_s emitter_t;
+struct emitter_s {
+	emitter_output_t output;
+	/* The output information. */
+	void (*write_cb)(void *, const char *);
+	void *cbopaque;
+	int nesting_depth;
+	/* True if we've already emitted a value at the given depth. */
+	bool item_at_depth;
+};
+
+static inline void
+emitter_init(emitter_t *emitter, emitter_output_t emitter_output,
+    void (*write_cb)(void *, const char *), void *cbopaque) {
+	emitter->output = emitter_output;
+	emitter->write_cb = write_cb;
+	emitter->cbopaque = cbopaque;
+	emitter->item_at_depth = false;
+	emitter->nesting_depth = 0;
+}
+
+/* Internal convenience function.  Write to the emitter the given string. */
+JEMALLOC_FORMAT_PRINTF(2, 3)
+static inline void
+emitter_printf(emitter_t *emitter, const char *format, ...) {
+	va_list ap;
+
+	va_start(ap, format);
+	malloc_vcprintf(emitter->write_cb, emitter->cbopaque, format, ap);
+	va_end(ap);
+}
+
+/* Write to the emitter the given string, but only in table mode. */
+JEMALLOC_FORMAT_PRINTF(2, 3)
+static inline void
+emitter_table_printf(emitter_t *emitter, const char *format, ...) {
+	if (emitter->output == emitter_output_table) {
+		va_list ap;
+		va_start(ap, format);
+		malloc_vcprintf(emitter->write_cb, emitter->cbopaque, format, ap);
+		va_end(ap);
+	}
+}
+
+static inline void
+emitter_gen_fmt(char *out_fmt, size_t out_size, const char *fmt_specifier,
+    emitter_justify_t justify, int width) {
+	size_t written;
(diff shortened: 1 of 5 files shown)


jasone:
Sweet!

interwq:
Cool! Maybe post the new output here (I remember you mentioned some changes)?

davidtgoldblatt:
Modulo bugs, output should be the same in json mode. [Example table output](https://github.com/jemalloc/jemalloc/files/1790760/table.txt), The only changes that came to mind are that I print out couple extra lines (config and prof settings), and a couple strings are now quoted (e.g. the version).

I mostly tried to keep things the same; I think we could do some cleanup to cut down on duplication some, but that can be more incremental.

davidtgoldblatt:
(minor fix, I was accidentally emitting config.prof twice).

include/jemalloc/internal/emitter.h
```diff
@@ -0,0 +1,400 @@
...
+	} else {
+		written = malloc_snprintf(out_fmt, out_size,
+		    "%%%d%s", width, fmt_specifier);
+	}
+	/* Only happens in case of bad format string, which *we* choose. */
+	assert(written <  out_size);
+}
+
+/*
+ * Internal.  Emit the given given value type in the relevant encoding (so that
```
interwq: nit: the extra given

include/jemalloc/internal/emitter.h
```diff
@@ -0,0 +1,400 @@
...
+ * the bool true gets mapped to json "true", but the string "true" gets mapped
+ * to json "\"true\"", for instance.
+ *
+ * Width is ignored if justify is emitter_justify_none.
+ */
+static inline void
+emitter_print_value(emitter_t *emitter, emitter_justify_t justify, int width,
+    emitter_type_t value_type, const void *value) {
+	const bool *boolp;
```
interwq: nit: define these when they are used?

include/jemalloc/internal/emitter.h
```diff
@@ -0,0 +1,400 @@
...
+	 * cases.
+	 */
+	char fmt[FMT_SIZE];
+	char buf[BUF_SIZE];
+
+	switch (value_type) {
+	case emitter_type_bool:
+		boolp = (const bool *)value;
+		emitter_gen_fmt(fmt, FMT_SIZE, "s", justify, width);
+		emitter_printf(emitter, fmt, *boolp ? "true" : "false");
```
interwq: maybe a macro for the statements within each case?

include/jemalloc/internal/emitter.h
```diff
@@ -0,0 +1,400 @@
...
+	const char *const *charpp;
+	const int *intp;
+	const unsigned *unsignedp;
+	const ssize_t *ssizep;
+	const size_t *sizep;
+	const uint32_t *uint32p;
+	const uint64_t *uint64p;
+
+	size_t str_written;
+#define BUF_SIZE 1000
```
interwq: 256? The recent arm libunwind stack overflow made me a bit worry.

include/jemalloc/internal/emitter.h
```diff
@@ -0,0 +1,400 @@
...
+	va_start(ap, format);
+	malloc_vcprintf(emitter->write_cb, emitter->cbopaque, format, ap);
+	va_end(ap);
+}
+
+/* Write to the emitter the given string, but only in table mode. */
+JEMALLOC_FORMAT_PRINTF(2, 3)
+static inline void
+emitter_table_printf(emitter_t *emitter, const char *format, ...) {
+	va_list ap;
```
interwq: nit: move ap into the branch

src/stats.c
```diff
@@ -286,90 +328,82 @@ stats_arena_bins_print(void (*write_cb)(void *, const char *), void *cbopaque,
...
-				} else {
-					not_reached();
-				}
-			}
-			uint64_t mutex_stats_64[mutex_prof_num_uint64_t_counters];
-			uint32_t mutex_stats_32[mutex_prof_num_uint32_t_counters];
-			if (mutex) {
-				read_arena_bin_mutex_stats(i, j, mutex_stats_64, mutex_stats_32);
-			}
+		if (mutex) {
```
interwq: Should this be combined with the branch at L353?

interwq:
Question: in table mode, if the counter has more digits than the width, does it still output the all the digits? In a couple of cases where the binary is long-lived, I have seen the nrequests gone too high. Or we should just reserve more space there.

include/jemalloc/internal/emitter.h
davidtgoldblatt: Fixed.

include/jemalloc/internal/emitter.h
davidtgoldblatt: Not actually syntactically allowed. I switched to using macros though, so it ends up being easy to remove anyways.

davidtgoldblatt: (Which I definitely had forgotten until I hit the compiler error)

include/jemalloc/internal/emitter.h
davidtgoldblatt: Done (where possible).

include/jemalloc/internal/emitter.h
davidtgoldblatt: Done.

src/stats.c
davidtgoldblatt: I sort of like the separation of having all the mallctlls happening in one place, followed by all the emitting happening in one place (as opposed to mallctl -> json emitting -> maybe some more mallctl -> table emitting). WDYT?

interwq: That sounds good too. Keep it as is then.

davidtgoldblatt:
If there's more digits than width, we'll emit the whole number (and the row will be misaligned from then on). The emitter itself doesn't really use the width parameter on its own; it just forwards it snprintf in a format string.

PR #1146, opened 2018-03-15
Stats fix

interwq:
Change mutable option output in stats to avoid stringify issues.

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -868,19 +868,19 @@ stats_general_print(emitter_t *emitter) {
 
 	/* opt. */
 #define OPT_WRITE(name, var, size, emitter_type)			\
-	if (je_mallctl("opt."#name, (void *)&var, &size, NULL, 0) ==	\
+	if (je_mallctl("opt."name, (void *)&var, &size, NULL, 0) ==	\
 	    0) {							\
-		emitter_kv(emitter, #name, "opt."#name, emitter_type,	\
+		emitter_kv(emitter, name, "opt."name, emitter_type,	\
 		    &var);						\
 	}
 
 #define OPT_WRITE_MUTABLE(name, var1, var2, size, emitter_type,		\
     altname)								\
-	if (je_mallctl("opt."#name, (void *)&var1, &size, NULL, 0) ==	\
-	    0 && je_mallctl(#altname, (void *)&var2, &size, NULL, 0)	\
+	if (je_mallctl("opt."name, (void *)&var1, &size, NULL, 0) ==	\
+	    0 && je_mallctl(altname, (void *)&var2, &size, NULL, 0)	\
 	    == 0) {							\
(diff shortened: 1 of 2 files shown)


src/stats.c
```diff
@@ -876,26 +876,26 @@ stats_general_print(emitter_t *emitter) {
...
 #define OPT_WRITE_UNSIGNED(name)					\
 	OPT_WRITE(name, uv, usz, emitter_type_unsigned)
 
 #define OPT_WRITE_SSIZE_T(name)						\
 	OPT_WRITE(name, ssv, sssz, emitter_type_ssize)
 #define OPT_WRITE_SSIZE_T_MUTABLE(name, altname)			\
-	OPT_WRITE_MUTABLE(name, ssv, ssv2, sssz, emitter_type_ssize,	\
-	    altname)
+	OPT_WRITE_MUTABLE(#name, ssv, ssv2, sssz, emitter_type_ssize,	\
+	    #altname)
```
interwq: I'm a bit rusted on macro expansion. Somehow if we pass name / altname and stringify later, we got strings with prefix, e.g. je_arenas.dirty_decay_ms which obviously isn't what we want. Because of this, stats is missing a few options currently. 

Tried a few rounds but only this fix the issue, and not sure how name prefix gets involved here. Any ideas @davidtgoldblatt?

davidtgoldblatt: My guess is name mangling; we have the "arenas" definition in jemalloc.c, so the name-mangling changes "arenas" to "je_arenas" somewhere if we pass it as non-stringified.

interwq: Yeah that's probably the case -- however what's confusing is that it only happens if we use two level macros and stringify in the inner macro; if we use OPT_WRITE_MUTABLE directly, it's also fine. I'm probably missing something about the macro expansion? Which way would you prefer for the fix?

davidtgoldblatt: I think I understand the macro expansion problem, but it's not interesting. I think the right fix is just to pass in the arguments to the macro as strings themselves (so that we write `OPT_WRITE_BOOL_MUTABLE("prof_active", "prof.active")`, since the only thing we do with them is stringify right away, and that's their "moral" status (i.e. they "are" really strings, not C tokens).

interwq: That sounds reasonable. Will update.

interwq:
@davidtgoldblatt: just updated the stringify as we discussed.

Also AppVeyor is having trouble like this since yesterday: https://ci.appveyor.com/project/jasone/jemalloc/build/1406/job/354rs657at0f3g2j  did we change anything there?

davidtgoldblatt:
AppVeyor is complaining that we remove "catgets" in our setup. We added that a few weeks ago, because AppVeyor was breaking because we *didn't* remove it. Reverting https://github.com/jemalloc/jemalloc/commit/ae0f5d5c3f29beb9977148dedb58575757139586 should fix it.

PR #1147, opened 2018-03-15
Revert "CI: Remove "catgets" dependency on appveyor."

interwq:
This reverts commit ae0f5d5c3f29beb9977148dedb58575757139586.


interwq:
power builds failing arena_reset tests. OOMed?

davidtgoldblatt:
Hard to say about OOMing; we should follow up. In the meantime, so long as we aren't introducing *new* breakages, it's not the worst thing in the world if they fail.

Issue #943, opened 2017-06-30
Consider tracking all mutexes in a rank-ordered tree for per- / post-fork


interwq:
Quote @jasone :
re: forking and mutex acquisition. If we were to insert mutexes into e.g. a red-black tree ordered by lock rank, we could iterate over the ordered mutexes and lock/unlock/reinit them in a centralized way, with way less pre/post-fork infra than we currently have.
And it would be less brittle, because we wouldn't accidentally miss mutexes.

This would require a bit more space in mutex struct, however since we plan to add padding, it may not take extra space. Also, we'll need to deregrister from the tracking tree when destroying mutexes, e.g. during arena_destroy.

davidtgoldblatt:
We could also use this to make mutex stats collection more automatic.

PR #1153, opened 2018-03-29
RFC : Mutexes rank ordered tree pre post fork 943

rkmisra:
Trying to implement suggestion mentioned in #943.  Not sure if this is right approach or not.

Used 2 tracking tree instead of single tree.  First tree (named top) keeps all mutexes smaller than rank 3 (arean_lock's rank) and second (named bottom) contains initialized mutexes with rank greater than 3.  Reason being,  when we start iteration in prefork,  arena_new call may already be in progress (with arenas_lock already acquired by other thread),  and that thread may keep on adding more mutexes (such as bin mutex) into our tracking tree and we can't allow modification of tree while iterating it.  Keeping all mutexes larger than arenas_lock ensures that we only start iteration of tree (bottom) in prefork after we lock arenas_lock which ensures no other thread is simultaneously doing arena_new (basically arenas_lock protect our bottom red black tree during iteration from being modified).  Top tree (with rank < 3) does not need explicit protection as it contains ctl & tcache lock which in my understanding are only added during bootstrap of jemalloc.


davidtgoldblatt:
Probably still a good idea, but some bitrotted.

PR #1156, opened 2018-04-02
background_thread: add max thread count config

djwatson:
Looking at the thread counts in our services, jemalloc's background thread is useful, but mostly idle.  Add a config option to tune down the number of threads.

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -11,12 +11,14 @@
 #define BACKGROUND_THREAD_DEFAULT false
 /* Read-only after initialization. */
 bool opt_background_thread = BACKGROUND_THREAD_DEFAULT;
+size_t opt_max_background_threads = MAX_BACKGROUND_THREAD_LIMIT;
 
 /* Used for thread creation, termination and stats. */
 malloc_mutex_t background_thread_lock;
 /* Indicates global state.  Atomic because decay reads this w/o locking. */
 atomic_b_t background_thread_enabled_state;
 size_t n_background_threads;
+size_t max_background_threads;
 /* Thread info per-index. */
 background_thread_info_t *background_thread_info;
 
@@ -287,7 +289,7 @@ background_work_sleep_once(tsdn_t *tsdn, background_thread_info_t *info, unsigne
 	uint64_t min_interval = BACKGROUND_THREAD_INDEFINITE_SLEEP;
 	unsigned narenas = narenas_total_get();
 
-	for (unsigned i = ind; i < narenas; i += ncpus) {
+	for (unsigned i = ind; i < narenas; i += max_background_threads) {
 		arena_t *arena = arena_get(tsdn, i, false);
 		if (!arena) {
 			continue;
@@ -390,7 +392,7 @@ check_background_thread_creation(tsd_t *tsd, unsigned *n_created,
 
 	tsdn_t *tsdn = tsd_tsdn(tsd);
 	malloc_mutex_unlock(tsdn, &background_thread_info[0].mtx);
-	for (unsigned i = 1; i < ncpus; i++) {
+	for (unsigned i = 1; i < max_background_threads; i++) {
 		if (created_threads[i]) {
 			continue;
 		}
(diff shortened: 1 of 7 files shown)


src/background_thread.c
```diff
@@ -11,12 +11,14 @@
 #define BACKGROUND_THREAD_DEFAULT false
 /* Read-only after initialization. */
 bool opt_background_thread = BACKGROUND_THREAD_DEFAULT;
+size_t opt_max_background_threads = SIZE_T_MAX;
```
interwq: nit: add a `#define MAX_BACKGROUND_THREAD_LIMIT MALLOCX_ARENA_LIMIT`? It's probably safe to assume we will never have more than 1 thread per arena.

src/background_thread.c
```diff
@@ -857,12 +864,12 @@ background_thread_boot1(tsdn_t *tsdn) {
 	}
 
 	background_thread_info = (background_thread_info_t *)base_alloc(tsdn,
-	    b0get(), ncpus * sizeof(background_thread_info_t), CACHELINE);
+	    b0get(), max_background_threads * sizeof(background_thread_info_t), CACHELINE);
```
interwq: since this is allocated at boot, I don't think we can increase the number beyond this?

src/ctl.c
```diff
@@ -1564,6 +1568,67 @@ background_thread_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+		}
+
+		if (background_thread_enabled()) {
+			background_thread_enabled_set(tsd_tsdn(tsd), false);
+			if (background_threads_disable(tsd)) {
+				ret = EFAULT;
+				goto label_return;
+			}
+			max_background_threads = newval;
+			background_thread_enabled_set(tsd_tsdn(tsd), true);
```
interwq: move this line to after the can_enable check.

src/background_thread.c
```diff
@@ -857,12 +865,12 @@ background_thread_boot1(tsdn_t *tsdn) {
 	}
 
 	background_thread_info = (background_thread_info_t *)base_alloc(tsdn,
-	    b0get(), ncpus * sizeof(background_thread_info_t), CACHELINE);
+	    b0get(), MAX_BACKGROUND_THREAD_LIMIT * sizeof(background_thread_info_t), CACHELINE);
```
interwq: sizeof(background_thread_info_t) is actually quite a bit (256) and when times 4096 it's 1MB. This overhead becomes real when we use huge pages for metadata (metadata_thp is enabled by default in our setup). How about we don't allow changing the number of threads on the fly?

src/jemalloc.c
```diff
@@ -1175,6 +1175,8 @@ malloc_conf_init(void) {
 			}
 			CONF_HANDLE_BOOL(opt_background_thread,
 			    "background_thread");
+			CONF_HANDLE_SIZE_T(opt_max_background_threads,
+                                           "max_background_threads", 0, 1024, no, yes, true);
```
interwq: Maybe not allow value 0? The max should be MALLOCX_ARENA_LIMIT

src/background_thread.c
djwatson: I'd like to at least be allowed to dynamically set it lower to make it easier to test, I'll make this opt_max_background_threads (default to ncpus), since it shouldn't change after startup

interwq: Set it lower sounds good.

src/ctl.c
```diff
@@ -1564,6 +1568,71 @@ background_thread_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	if (!have_background_thread) {
+		return ENOENT;
+	}
+	background_thread_ctl_init(tsd_tsdn(tsd));
+
+	malloc_mutex_lock(tsd_tsdn(tsd), &ctl_mtx);
+	malloc_mutex_lock(tsd_tsdn(tsd), &background_thread_lock);
+	if (newp == NULL) {
+          oldval = max_background_threads;
+		READ(oldval, size_t);
```
interwq: nit: indent seems off.

src/jemalloc.c
```diff
@@ -1175,6 +1175,10 @@ malloc_conf_init(void) {
 			}
 			CONF_HANDLE_BOOL(opt_background_thread,
 			    "background_thread");
+			CONF_HANDLE_SIZE_T(opt_max_background_threads,
+					   "max_background_threads", 1,
+					   opt_max_background_threads, no, yes,
```
interwq: should be both "yes" for check_min and check_max

PR #1157, opened 2018-04-05
Consolidate the two memory loads in rtree_szind_slab_read().

interwq:
@davidtgoldblatt noticed that two memory loads of the same location was generated on the free() fast path. Let's do a bit of manual work to avoid that (u-benchmark shows ~1% improvement).

Commit message:
szind and slab bits are read on fast path, where compiler generated two memory
loads separately for them before this diff.  Manually operate on the bits to
avoid the extra memory load.

### DIFF
--- a/include/jemalloc/internal/rtree.h
+++ b/include/jemalloc/internal/rtree.h
@@ -460,8 +460,14 @@ rtree_szind_slab_read(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
 	if (!dependent && elm == NULL) {
 		return true;
 	}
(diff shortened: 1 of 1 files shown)


interwq:
Before:
```
add    0x30(%r13),%rax 

mov    (%rax),%r14 

mov    (%rax),%rsi       
```

after:
```
add    0x30(%r12),%rax

mov    (%rax),%rsi

mov    %rsi,%r10
```

davidtgoldblatt:
Nit: opearte -> operate in the commit message. Otherwise looks good.

interwq:
> Nit: opearte -> operate in the commit message. Otherwise looks good.

Good catch. Just fixed.

PR #1160, opened 2018-04-06
Control idump and gdump with prof_active.

### LINKED ISSUE #1114: Feature request : To be able to reset the value of "lg_prof_interval"
I would like enable and disable heap profiling
dynamically. When I enabled profiling               
I would like to have multiple dumps based on the allocation activity.
So I am setting the value of "lg_prof_interval" in configuration.
Example: "prof:true,prof_active:false,prof_prefix:jeprof.out, lg_prof_interval:10"

The blocker I am having is once I enabled profiling
(Example: mallctl("prof.active", nullptr, nullptr, &enable_profiling, val_sz))
,with the above configuration, I am not able to disable profiling.

When my configuration doesn't contain "lg_prof_interval", I am able to enable
and disable profiling using mallctl. 
 
Currently Jemalloc can reset the value of
"lg_prof_sample" using "prof.reset" but not "lg_prof_interval".

My request is to be able to set the value of "lg_prof_interval" dynamically.
 
Thanks, 
haritha

interwq:
This resolves #1114.

### DIFF
--- a/include/jemalloc/internal/prof_inlines_a.h
+++ b/include/jemalloc/internal/prof_inlines_a.h
@@ -69,4 +69,15 @@ prof_accum_cancel(tsdn_t *tsdn, prof_accum_t *prof_accum, size_t usize) {
 #endif
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+prof_active_get_unlocked(void) {
+	/*
+	 * Even if opt_prof is true, sampling can be temporarily disabled by
+	 * setting prof_active to false.  No locking is used when reading
+	 * prof_active in the fast path, so there are no guarantees regarding
+	 * how long it will take for all threads to notice state changes.
+	 */
+	return prof_active;
+}
+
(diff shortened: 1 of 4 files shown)


include/jemalloc/internal/prof_inlines_a.h
```diff
@@ -69,4 +69,15 @@ prof_accum_cancel(tsdn_t *tsdn, prof_accum_t *prof_accum, size_t usize) {
...
 
+JEMALLOC_ALWAYS_INLINE bool
+prof_active_get_unlocked(void) {
+	/*
+	 * Even if opt_prof is true, sampling can be temporarily disabled by
+	 * setting prof_active to false.  No locking is used when reading
+	 * prof_active in the fast path, so there are no guarantees regarding
+	 * how long it will take for all threads to notice state changes.
+	 */
+	return prof_active;
```
davidtgoldblatt: prof_active should be atomic in this case. Not a blocker to just move it, but I filed #1162 so we don't forget.

PR #1163, opened 2018-04-10
Introduce allocator hooks

davidtgoldblatt:
This introduces a (highly experimental, user-hostile, undocumented) set of hooking primitives that can be used to introduce new and exciting sorts of profiling.


include/jemalloc/internal/hook.h
```diff
@@ -0,0 +1,125 @@
...
+
+typedef struct hooks_s hooks_t;
+struct hooks_s {
+	hook_alloc alloc_hook;
+	hook_dalloc dalloc_hook;
+	hook_expand expand_hook;
+};
+
+/*
+ * Returns an opaque handle to be used when removing the hook.  NULL means that
```
davidtgoldblatt: Nit: move to go with hook_install, not hook_boot.

src/hook.c
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+	return NULL;
+}
+
+void *
+hook_install(tsdn_t *tsdn, hooks_t *to_install, void *extra) {
+	malloc_mutex_lock(tsdn, &hooks_mu);
+	void *ret = hook_install_locked(to_install, extra);
+	tsd_global_slow_inc(tsdn);
```
interwq: Let's define a global_slow max (should be the same as max # of hooks?), and assert here?

test/unit/hook.c
```diff
@@ -0,0 +1,180 @@
...
+	/* Null hooks should be ignored, not crash. */
+	hooks_t hooks1 = {NULL, NULL, NULL};
+	hooks_t hooks2 = {&test_alloc_hook, NULL, NULL};
+	hooks_t hooks3 = {NULL, &test_dalloc_hook, NULL};
+	hooks_t hooks4 = {NULL, NULL, &test_expand_hook};
+
+	void *handle1 = hook_install(TSDN_NULL, &hooks1, NULL);
+	void *handle2 = hook_install(TSDN_NULL, &hooks2, NULL);
+	void *handle3 = hook_install(TSDN_NULL, &hooks3, NULL);
+	void *handle4 = hook_install(TSDN_NULL, &hooks4, NULL);
```
interwq: nit: test install failure as well (by installing more than HOOKS_MAX)

src/jemalloc.c
```diff
@@ -2038,6 +2038,14 @@ je_malloc(size_t size) {
 	dopts.item_size = size;
 
 	imalloc(&sopts, &dopts);
+	/*
+	 * Note that this branch gets optimized away -- it immediately follows
+	 * the check on tsd_fast that sets sopts.slow.
+	 */
+	if (sopts.slow) {
+		uintptr_t args[3] = {size};
```
interwq: nit: add a HOOK_ARGS macro?

src/ctl.c
```diff
@@ -2849,3 +2861,48 @@ stats_arenas_i_index(tsdn_t *tsdn, const size_t *mib, size_t miblen, size_t i) {
...
+    void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret;
+	if (oldp == NULL || oldlenp == NULL|| newp == NULL) {
+		ret = EINVAL;
+		goto label_return;
+	}
+	/*
+	 * Note: this is a *private* struct.  This is an experimental interface;
+	 * forcing the user to know the jemalloc internals well enough to
+	 * extract the ABI hopefully ensures nobody gets to comfortable with
```
interwq: nit: do you mean "too comfortable"?

src/ctl.c
```diff
@@ -2849,3 +2861,48 @@ stats_arenas_i_index(tsdn_t *tsdn, const size_t *mib, size_t miblen, size_t i) {
...
+    void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret;
+	WRITEONLY();
+	void *handle = NULL;
+	WRITE(handle, void *);
+	if (handle == NULL) {
+		ret = EINVAL;
+		goto label_return;
+	}
+	hook_remove(tsd_tsdn(tsd), handle);
```
interwq: remove should still be able to fail? Say if the hook is not in_use, caused like a double remove?

test/unit/hook.c
interwq: Never mind. I see it's tested in a later diff.

src/tsd.c
```diff
@@ -51,14 +51,157 @@ bool tsd_booted = false;
...
+}
+
+static void
+tsd_add_nominal(tsd_t *tsd) {
+	assert(!tsd_in_nominal_list(tsd));
+	assert(tsd_state_get(tsd) <= tsd_state_nominal_max);
+	ql_elm_new(tsd, use_a_getter_or_setter_instead_link);
+	malloc_mutex_lock(tsd_tsdn(tsd), &tsd_nominal_tsds_lock);
+	ql_tail_insert(&tsd_nominal_tsds, tsd,
+	    use_a_getter_or_setter_instead_link);
```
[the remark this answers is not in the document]
davidtgoldblatt: Added an explicit TSD_MANGLE macro; this also simplifies the rest of the mangling. Seem reasonable?

src/tsd.c
```diff
@@ -51,14 +51,157 @@ bool tsd_booted = false;
...
+	 * other thread, and then have that other thread not have picked up the
+	 * update yet (since the original incrementing thread might still be
+	 * making its way through the tsd list).
+	 */
+	tsd_force_recompute(tsdn);
+}
+
+void tsd_global_slow_dec(tsdn_t *tsdn) {
+	atomic_fetch_sub_u32(&tsd_global_slow_count, 1, ATOMIC_RELAXED);
+	/* Se the note in ..._inc(). */
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

src/tsd.c
```diff
@@ -51,14 +51,157 @@ bool tsd_booted = false;
...
+/******************************************************************************/
+
+static uint8_t
+tsd_state_compute(tsd_t *tsd) {
+	if (!tsd_nominal(tsd)) {
+		return tsd_state_get(tsd);
+	}
+	/* We're in *a* nominal state; but which one? */
+	if (malloc_slow || !tsd_tcache_enabled_get(tsd) ||
+	    tsd_reentrancy_level_get(tsd) > 0 || tsd_global_slow()) {
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

include/jemalloc/internal/seq.h
```diff
@@ -0,0 +1,55 @@
...
+#include "jemalloc/internal/atomic.h"
+
+/*
+ * A simple seqlock implementation.
+ */
+
+#define seq_define(type, short_type)					\
+typedef struct {							\
+	atomic_zu_t seq;						\
+	atomic_zu_t data[						\
```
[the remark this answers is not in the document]
davidtgoldblatt: Yeah -- it's UB to have races on non-atomics. (Well, OK, we ignore that sometimes because of optimizer bugs, but we should probably try not to do so if it's not necessary).

include/jemalloc/internal/hook.h
```diff
@@ -0,0 +1,125 @@
...
+ * called before the free is done, and expand hooks are called after the
+ * allocation is expanded.
+ *
+ * For realloc and rallocx, if the expansion happens in place, the expansion
+ * hook is called.  If it is moved, then the alloc hook is called on the new
+ * location, and then the free hook is called on the old location.
+ *
+ * (We omit no-ops, like free(NULL), etc.).
+ *
+ * Reentrancy:
```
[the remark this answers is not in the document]
davidtgoldblatt: Discussed offline -- we'll support reentrancy a little more strongly, making this no longer apply.

src/hook.c
davidtgoldblatt: I don't think the assert is right (any value is a valid return type here), but now that you mention it, we shouldn't be incrementing slowness if hook installation fails. Will fix.

davidtgoldblatt: Discussed offline. Will do.

src/jemalloc.c
davidtgoldblatt: Discussed offline, I hated this.

src/ctl.c
davidtgoldblatt: Fixed.

src/ctl.c
davidtgoldblatt: Yeah, but that's a usage error, a la arena destruction. I'd say that should be an assertion failure, not a mallctl error return.

davidtgoldblatt:
As per offline feedback, moved reentrancy handling responsibility from the user to our hook implementation.

src/hook.c
```diff
@@ -99,41 +99,74 @@ for (int for_each_hook_counter = 0;					\
...
+	 * that it might secretly return a pointer to some global data if we
+	 * know that the caller will take the early-exit path.
+	 * If we return a bool that indicates that we are reentrant, then the
+	 * caller will go down the early exit path, leaving the global
+	 * untouched.
+	 */
+	static bool in_hook_global = true;
+	tsdn_t *tsdn = tsdn_fetch();
+	bool *in_hook = tsdn_in_hookp_get(tsdn);
+	if (in_hook != NULL) {
```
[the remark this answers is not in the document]
davidtgoldblatt: Discussed offline -- added a comment.

PR #1165, opened 2018-04-10
Stack address should not be used for ordering mutexes

### LINKED ISSUE #1161: Stack address comparison in witness comp function
In mutex.c we initialize mutex comparison function as below
```
static bool
malloc_mutex_init(malloc_mutex_t *mutex, const char *name,
    witness_rank_t rank, malloc_mutex_lock_order_t lock_order) {
.
.
    witness_init(&mutex->witness, name, rank,
					mutex_addr_comp, &mutex);
.
.
}
```
In witness_init we are passing address from stack of mutex argument (see &mutex in witness_init) that was passed in function malloc_mutex_init.  I think intent was to provide some kind of lock ordering based on stack depth at the time of initialization.  
 
This means order could be completely different depending on whether stack is growing from higher to lower address or vice versa.  This seems kind of brittle.

In  mutex_pool.c we are initializing
```
for (int i = 0; i < MUTEX_POOL_SIZE; ++i) {
		if (malloc_mutex_init(&pool->mutexes[i], name, rank,
		    malloc_mutex_address_ordered)) {
			return true;
		}
	} 
```
Here all poll->mutexes[i] ends up having same address stored as each call malloc_mutex_init() happens to have same stack address where  poll->mutexes[i] is pushed.  Here we get impression by malloc_mutex_address_ordered argument that mutexes are somehow ordered by address, but all addresses are same, so we are actually getting same behavior as  malloc_mutex_address_exclusive.  Confusing for code reader.

Similarly
```
for (unsigned i = 0; i < ncpus; i++) {
		background_thread_info_t *info = &background_thread_info[i];
		/* Thread mutex is rank_inclusive because of thread0. */
		if (malloc_mutex_init(&info->mtx, "background_thread",
		    WITNESS_RANK_BACKGROUND_THREAD,
		    malloc_mutex_address_ordered)) {
			return true;
		}
.
```

malloc_mutex_address_ordered here seems meaningless as all mutexes for background threads have same address.

During post_fork_init in child process, it might be difficult (although not impossible) to simulate same stack address ordering when we again re initializes mutexes.  For example (theoretical), let say we have 3 mutexes, mutex_1, mutex_2 & mutex_3 with _same_ rank, and lets say they get initialized as below during boot up

```
malloc_mutex_t mutex_1;
malloc_mutex_t mutex_2;
malloc_mutex_t mutex_3;

a_func() {
   b_func(...);
   malloc_mutex_init(&mutex_2,...);
   malloc_mutex_init(&mutex_3,...);
}

b_func(<some arguments>) {
.
   c_func(...);
.
}

c_func(<some arguments>) {
.
   malloc_mutex_init(&mutex_1,...);
.
}
```
Here we know that mutex_1 in argument has lowest stack address (much deeper in stack and assuming stack is growing from high -> low) and thus in witness ordering we expect mutex_1 < mutex_2 = mutex_3, or mutex_1 must be locked first before mutex_2 or mutex_3.

In post fork child, if we are not careful, we might assume the original init order is mutex_1 , mutex_2, mutex_3 and initialize it like below, but then we end up changing the original stack address based ordering
```
wrong_post_fork_child() {
    malloc_mutex_init(&mutex_1,...);
    malloc_mutex_init(&mutex_2,...);
    malloc_mutex_init(&mutex_3,...);
} 
```
Above all 3 mutexes have same stack address, and thus not in same order as original initialization (current ordering mutex_1 = mutex_2 = mutex_3).  During post init, we might need to artificially put malloc_mutex_init(&mutex_1,...); in deeper function call in stack to get lower stack address for mutex_1, i.e. something like below: 
```
post_fork_child() {
    bogus_function_call_to_get_lower_address(...);
    malloc_mutex_init(&mutex_2,...);
    malloc_mutex_init(&mutex_3,...);
}

bogus_function_call_to_get_lower_address(...) {
    malloc_mutex_init(&mutex_1,...);
}
```
This is seems brittle and error prone (and may seem illogical).

In pull request, #1153, we tried addressing this by not changing stack address of mutex during initialization from child fork (once mutex address is obtained, it is never changed) but that seems workaround for the core issue of using stack address.
 
I think the only thing stack address guaranteed is a "predictable ordering, on a given architecture, for mutexes with same rank".  May be this is the current best way available to get predictable ordering, but should we explore this a bit further to see if there is a better alternative?

rkmisra:
This is to fix #1161

### DIFF
--- a/src/mutex.c
+++ b/src/mutex.c
@@ -174,7 +174,7 @@ malloc_mutex_init(malloc_mutex_t *mutex, const char *name,
 		mutex->lock_order = lock_order;
 		if (lock_order == malloc_mutex_address_ordered) {
 			witness_init(&mutex->witness, name, rank,
-			    mutex_addr_comp, &mutex);
+			    mutex_addr_comp, mutex);
 		} else {
 			witness_init(&mutex->witness, name, rank, NULL, NULL);
 		}


davidtgoldblatt:
Looks good! I'll merge once tests come back green. Thanks for the diff!

davidtgoldblatt:
(And of course, thanks for catching the issue in the first place).

PR #1176, opened 2018-04-16
Proposal: track the peak memory used by each thread

fwyzard:
In the context of profiling a modular application, I am interested in the peak memory used by each thread, within each module. Here "module" could be a function call, a TBB task, _etc._ that is going to run within a given thread.

Something like
  - at the begin of each module, check the memory usage and reset the peak value;
  - while the module is running, track each memory allocation and deallocation and update the peak memory usage accordingly;
  - at the end of the module, read the peak memory usage.

`jemalloc` can already track each allocation and deallocation; if we define the "memory usage" as "allocated memory - deallocated memory" (and let the client code take care of the offset) a possible solution is to add one more field to the `tsd` and track the peak "allocated - deallocated memory".

Note that an alternative approach would be store both the current and peak usage in the `tsd`, relieving the user form the need to compute the offset. However the current usage would need to be updated for all allocations and deallocations, while the peak usage can be checked only for the allocations, so it might be cheaper.

This PR is a working implementation of this idea.
The usage could be, for example:
```c++
{
  thread_local bool initialized = false;
  thread_local uint64_t * thread_allocated_p   = nullptr;
  thread_local uint64_t * thread_deallocated_p = nullptr;
  thread_local int64_t *  thread_peak_used_p   = nullptr;
  if (not initialized) {
    size_t ptr_s = sizeof(uint64_t *);
    mallctl("thread.allocatedp",   & thread_allocated_p,   & ptr_s, nullptr, 0);
    mallctl("thread.deallocatedp", & thread_deallocated_p, & ptr_s, nullptr, 0);
    mallctl("thread.peakusedp",    & thread_peak_used_p,   & ptr_s, nullptr, 0);
    initialized = true;
  }
  // reset the peak memory use before running the task and read the offset
  * thread_peak_used_p = 0;
  int64_t offset = * thread_allocated_p - * thread_deallocated_p;

  // here goes the task to be profiled
  // ...
  
  // read the peak memory use after running the task
  std::cout << "peak memory usage: " << thread_peak_used_p - offset << std::endl;
```

Before working on the documentation (both in `tsd.c` and in the man pages), test case, etc. I would like to ask for comments and feedback, whether this would be acceptable, if the implementation is reasonable, etc.

### DIFF
--- a/include/jemalloc/internal/tsd.h
+++ b/include/jemalloc/internal/tsd.h
@@ -67,6 +67,7 @@ typedef void (*test_callback_t)(int *);
     O(narenas_tdata,		uint32_t,		uint32_t)	\
     O(thread_allocated,		uint64_t,		uint64_t)	\
     O(thread_deallocated,	uint64_t,		uint64_t)	\
+    O(thread_peakused,		int64_t,		int64_t)	\
     O(prof_tdata,		prof_tdata_t *,		prof_tdata_t *)	\
     O(rtree_ctx,		rtree_ctx_t,		rtree_ctx_t)	\
     O(iarena,			arena_t *,		arena_t *)	\
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -66,6 +66,8 @@ CTL_PROTO(thread_allocated)
 CTL_PROTO(thread_allocatedp)
 CTL_PROTO(thread_deallocated)
 CTL_PROTO(thread_deallocatedp)
+CTL_PROTO(thread_peakused)
+CTL_PROTO(thread_peakusedp)
 CTL_PROTO(config_cache_oblivious)
 CTL_PROTO(config_debug)
 CTL_PROTO(config_fill)
@@ -252,6 +254,8 @@ static const ctl_named_node_t	thread_node[] = {
 	{NAME("allocatedp"),	CTL(thread_allocatedp)},
 	{NAME("deallocated"),	CTL(thread_deallocated)},
 	{NAME("deallocatedp"),	CTL(thread_deallocatedp)},
+	{NAME("peakused"),	CTL(thread_peakused)},
+	{NAME("peakusedp"),	CTL(thread_peakusedp)},
 	{NAME("tcache"),	CHILD(named, thread_tcache)},
 	{NAME("prof"),		CHILD(named, thread_prof)}
 };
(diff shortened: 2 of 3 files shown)


interwq:
Thanks for sharing the thoughts @fwyzard! The big concern I have is the added branch on the fast path, which is critical for performance in many large applications. A branch may not sound much on modern architectures, however they do make a meaningful impact for applications with a lot of allocation activities. The other thing to consider is when memory is passed between threads, where memory allocated by one thread may be freed by another.

I'd prefer this feature to be built on top of jemalloc -- since we will soon be providing the allocation hooks (as an experimental feature at first, cc: @davidtgoldblatt), and this should be straightforward to add with the hooks. Since the hooks will be user defined, the logic like switch / reset counters could be implemented in the hooks. Does that sound reasonable to you @fwyzard?

fwyzard:
@interwq thanks for your comments.

Performance is something I'm concerned about as well - in fact, if jemalloc has any relevant benchmarks I will be happy to try them, in addition to measuring the impact on our application (https://github.com/cms-sw/cmssw/).

I agree that a hard-to-predict branch can impact performance; I thought the compiler could implement it with `CMOVcc` conditional move instruction, but looking at the generated assembler that's not the case. After benchmarking the impact of the patch, I could try to force the generation of  `CMOVcc` instruction and see if that fares any better (or worse).

If there are other suggestions to improve performance with respect to my patch, I'll be happy to try them as well, and possibly maintain a patched `jemalloc` for our use case until the hooks are available.

Instead, memory moving across threads is not a concern for my use case: measuring the peak memory used by each module when processing an "event". Different modules and/or different event can run on different threads, but each module processes each event with a single thread (so far).

Implementing the logic in a hook sounds fine by me (with the usual caveat of the performance impact). I'll be happy to try them in due time.

jasone:
As @interwq says, per thread memory usage isn't a well defined concept in C, because allocations can be passed among threads.  That is why jemalloc only tracks per thread allocation/deallocation volume, which you based your patch on ([thread.allocatedp](http://jemalloc.net/jemalloc.3.html#thread.allocatedp)/[thread.deallocatedp](http://jemalloc.net/jemalloc.3.html#thread.deallocatedp)).  That said, jemalloc also supports per thread heap profiling, for which allocations are considered to be associated with their allocating threads for their entire lifetimes, thus providing a per thread memory ownership statistic than cannot go negative.  Right now there is a mechanism for triggering a profile dump if a global high-water mark is exceeded (see [opt.prof_gdump](http://jemalloc.net/jemalloc.3.html#opt.prof_gdump)), but what you would want is a way to trigger on per thread high water marks.  It looks to me like it would be possible to store atomic counters (current and high-water usage) in ```struct prof_tdata_s``` and use them to trigger dumps in the heap sampling slow path.  There would probably be some locking order challenges to deal with -- dump events would probably have to be enqueued similarly to how gdump triggering is.  However, this would provide the statistic you're looking for, in a way that wouldn't significantly impact performance.

fwyzard:
FYI, I got around benchmarking the impact of the proposed patch (rebased on top of jemalloc 5.2.1) on our application ([CMSSW](http://cms-sw.github.io/), running 16 concurrent jobs each with 4 threads and 4 TBB parallel tasks):

image

davidtgoldblatt:
I wonder -- how accurate do you need the data to be?

We currently force ourselves down a slow path every 64k worth of allocation or free activity. So we could compute the max gap between allocated and freed at those intervals more or less for free. This could give you accuracy to within 64k bytes, without adding any fast-path slowdown.

fwyzard:
My use case is in the range of the megabytes, so an accuracy of 64 kB should be sufficient.

Is there a way to force going down the slow path and updating the counters ?

PR #1178, opened 2018-04-16
Add UNUSED to avoid compiler warnings.

interwq:


interwq:
Note that in a few cases, parameters are truly unused but still kept and marked UNUSED, to maintain a consistent naming scheme with similar functions.

PR #1180, opened 2018-04-17
Add the --disable-initial-exec-tls configure option.

### LINKED ISSUE #937: static TLS errors from jemalloc 5.0.0 built on CentOS 6

davidtgoldblatt:
Right now we always make our TLS use the initial-exec model if the compiler
supports it.  This change allows configure-time disabling of this setting, which
can be helpful when dynamically loading jemalloc is the only option.

This resolves #937 (or at least, resolves it as much as we're going to in the short term).


davidtgoldblatt:
To test, I used the following program and verified that the errors went away:

#include <dlfcn.h>
#include <stdio.h>

int main() {
        void *handle = dlopen("lib/libjemalloc.so", RTLD_NOW | RTLD_LOCAL);
        if (!handle) {
                fprintf(stderr,"Couldn't open\n");
                fprintf(stderr, "%s\n", dlerror());
        }
        return 0;
}

davidtgoldblatt:
Done.

nehaljwani:
Will this change be available in the next patch / minor release? How soon is that going to happen? 😅

davidtgoldblatt:
Yep -- in fact, the upcoming 5.1 release was the impetus for getting around for this; we expect that it should be out shortly.

PR #1181, opened 2018-04-17
Avoid a resource leak down extent split failure paths.

### LINKED ISSUE #1106: extent leak path not handled correctly?
@davidtgoldblatt: I think this might be some left over from the no dump stack of diffs: https://github.com/jemalloc/jemalloc/blob/dev/src/extent.c#L1340

can you please check if things need to be fixed, or the comments are stale?

davidtgoldblatt:
Previously, we would leak the extent and memory associated with a salvageable
portion of an extent that we were trying to split in three, in the case where
the firs split attempt succeeded and the second failed.

This resolves #1106.


src/extent.c
```diff
@@ -1319,21 +1319,17 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
 		 * cant_alloc case should not occur.
 		 */
 		assert(result == extent_split_interior_error);
+		if (to_salvage != NULL) {
+			extent_deregister(tsdn, to_salvage);
```
interwq: Should we do extent_deregister_no_gdump_sub here as well?

davidtgoldblatt: I don't think so; whatever we do to the salvageable extent counted as a "real" extent operation (unlike the leak path).

interwq: Hmm however we added the extent with `extent_register_no_gdump_add(tsdn, extent)` L1282 this file?

davidtgoldblatt:
Actually, now that you mention it, this whole commit was nonsense (we of course don't want to leak the salvageable extent in a different way -- we want to salvage it!). Should be less nonsense now.

interwq:
Looks good so long as tests come back clean.

PR #1182, opened 2018-04-17
Fix abort_conf processing.

interwq:
When abort_conf is set, make sure we always error out at the end of the options
processing loop.


src/jemalloc.c
```diff
@@ -1052,9 +1050,6 @@ malloc_conf_init(void) {
 
 			CONF_HANDLE_BOOL(opt_abort, "abort")
 			CONF_HANDLE_BOOL(opt_abort_conf, "abort_conf")
-			if (opt_abort_conf && had_conf_error) {
```
interwq: Previously we may never reach this point, since CONF_HANDLE does `continue` when matches.

PR #1185, opened 2018-04-18
Remove an incorrect assertion.

### LINKED ISSUE #1183: Intermittent background thread failures
Seems to be a timing issue; it's hard to reproduce unless the test is somewhat starved for CPU time. Commands that will repro at dev (on my laptop):

```
./configure --enable-debug --with-malloc-conf=tcache:false,background_thread:true
make -j tests
seq 1000 | parallel -P128 test/unit/pack
```

Some runs will fail with: `<jemalloc>: src/background_thread.c:401: Failed assertion: "info->state != background_thread_paused"`

interwq:
Background threads are created without holding the global background_thread
lock, which mean paused state is possible (and fine).

This resolves #1183.


interwq:
Discussed offline. Merging.

Issue #1174, opened 2018-04-12
5.1 release


interwq:
This issue is for tracking communication/progress related to the 5.1 release.

interwq:
A few issues we want to resolve before the release: #937 #1133 #1106

interwq:
We have the current release [candidate](https://github.com/jemalloc/jemalloc/commit/b8f4c730eff28edee4b583ff5b6ee1fac0f26c27) running in production now, and plan to release 5.1 in a week.

CC folks who might be interested: @gnzlbg, @rkmisra, @hatvik, @rustyx,  @davidlt, @paravoid, @marxin,  @strejda, @sharkcz, @ingvarha, @cferris1000, @maksqwe, @trasz, @tamird, @xhochy, @rlibby, @zonyitoo, @wqfish, @djwatson, @jasone

marxin:
Thanks for the candidate. I can confirm it builds and survives regression tests on:
i586, x86_64, armv7l, aarch64, ppc64 and ppc64le.

sharkcz:
also s390x should be OK, I've got a successful build with all tests passed on one Fedora 27 host, but got THP related failure on another host, I'm looking into it.

sharkcz:
I can confirm that x86_64, armv7l, aarch64, ppc64, ppc64le and s390x are all OK on Fedora 27.

marxin:
@sharkcz can you please help me how you handle this on Linux:
https://github.com/jemalloc/jemalloc/issues/231

Thanks

jasone:
I have locally integrated b8f4c730eff28edee4b583ff5b6ee1fac0f26c27 into r332893 of FreeBSD-current and tested it on amd64, as well as cross-compiling for numerous other platforms.  There appear to be no outstanding problems.

gnzlbg:
I get way too many warnings when compiling jemalloc 5.1 with `-Wextra` enabled :/

interwq:
@gnzlbg: thanks for sharing the info and digging into the issue. I was able to make it work by adding `make distclean` after `./autogen.sh`. Can you please try if that solves it for you too?

gnzlbg:
@interwq so that fixes it for me and I was able to upgrade most Rust targets to use jemalloc 5.1rc1 from jemalloc 4.5 without many issues. The ones that are failing are probably still my fault, or were failing in 5.0.1 as well, but those can be fixed later.

[...]

PR #1192, opened 2018-04-27
Declare je_malloc_stats_print before use.

interwq:


interwq:
@davidtgoldblatt: this fixes the conflicting type issue reported in #1174. Let me know if you think there's a better way here.

interwq:
Actually I don't think this is the right fix.

PR #1193, opened 2018-04-29
aarch64: Add ILP32 support.

cmuellner:
Instead of setting a fix value of 48 allowed VA bits,
we distiguish between LP64 and ILP32.

Testsuite result with LP64:
Test suite summary: pass: 13/13, skip: 0/13, fail: 0/13

Testsuit result with ILP32:
Test suite summary: pass: 13/13, skip: 0/13, fail: 0/13

Reviewed-by: Philipp Tomsich <philipp.tomsich@theobroma-systems.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -412,7 +412,13 @@ AC_DEFINE_UNQUOTED([CPU_SPINWAIT], [$CPU_SPINWAIT])
 case "${host_cpu}" in
   aarch64)
     AC_MSG_CHECKING([number of significant virtual address bits])
-    LG_VADDR=48
+    if test "x${ac_cv_sizeof_void_p}" = "x4" ; then
+      #aarch64 ILP32
+      LG_VADDR=32
+    else
+      #aarch64 LP64
+      LG_VADDR=48
+    fi
     AC_MSG_RESULT([$LG_VADDR])
     ;;
   x86_64)


apinski-cavium:
Note 48 bits might be changing soon to 52.

cmuellner:
@apinski-cavium: increasing to 52 bits would needlessly decrease the number of usable bits for pointer tagging on kernels, where the VA space is actually 48 bits.
As the number of VA bits is not exposed by Linux header files and can only be fetched at runtime, I would propose a change which allows to set the number of VA bits as an argument to the configure script (or providing the VA space area via kernel header files, but that requires a kernel patch).

PS: anyone who does not know what we are talking about, here is the documentation of aarch64's memory model:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/arm64/memory.txt?h=v4.17-rc3

davidtgoldblatt:
We've been agonizing for a while about what to do about the bigger ARM address spaces; with jemalloc specifically, it forces a less efficient metadata lookup strategy. For now, this has settled on "assume 48, and wait for someone to complain".

(Still trying to debug why tests aren't running on this).

interwq:
Ran the arm CI manually and everything was fine. Merging.

Issue #1162, opened 2018-04-09
`prof_active` should be atomic


davidtgoldblatt:
Modifications are mutex-protected, but reads aren't necessarily.

PR #1195, opened 2018-05-01
making prof_active flag atomic and removing its corresponding mutex

rkmisra:
Made prof_active flag atomic to address #1162

### DIFF
--- a/include/jemalloc/internal/prof_externs.h
+++ b/include/jemalloc/internal/prof_externs.h
@@ -22,7 +22,7 @@ extern char	opt_prof_prefix[
     1];
 
 /* Accessed via prof_active_[gs]et{_unlocked,}(). */
-extern bool	prof_active;
+extern atomic_b_t	prof_active;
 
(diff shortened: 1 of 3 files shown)


gnzlbg:
@davidtgoldblatt do you have links to the bug report?

davidtgoldblatt:
I didn't file one upstream; I just threw it at a couple compiler people I know.

I dug around the llvm source a little the last time around, and it seems like it should be an easy fix by introducing an acquire_cmp primitive (analogous to the acquire_mov it currently uses to prevent reordering). The "generate a cmp here" and "generate an acquire_mov" logic is already in the same pass. Then again I'm not really a compiler person so maybe this is hard for some reason I don't know about.

gnzlbg:
There is one now: https://bugs.llvm.org/show_bug.cgi?id=38405 We'll see what happens.

davidtgoldblatt:
Missed optimization opportunity still exists -- if it gets fixed, we can fix this. For now, clearing out old PRs.

PR #1197, opened 2018-05-02
Fix a typo in INSTALL.md.

interwq:


interwq:
Merging as only doc was changed.

PR #1202, opened 2018-05-03
arm64: Correct LG_VADDR to 56 bits.

cmuellner:
The AArch64 architecture reference manual permits to use
the upper eight bits of an virtual address for address tagging
if the translation table is set up accordingly.
The Linux kernel enables address tagging [1], so it is safe to use
it on Linux-based systems.

That gives us an allowed LG_VADDR value of 56 bits.
The previous value of 48 bits is clearly a violation of the specification.

[1] https://www.kernel.org/doc/Documentation/arm64/tagged-pointers.txt

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -412,12 +412,12 @@ AC_DEFINE_UNQUOTED([CPU_SPINWAIT], [$CPU_SPINWAIT])
 case "${host_cpu}" in
   aarch64)
     AC_MSG_CHECKING([number of significant virtual address bits])
-    if test "x${ac_cv_sizeof_void_p}" = "x4" ; then
+    if test "x${LG_SIZEOF_PTR}" = "x2" ; then
       #aarch64 ILP32
       LG_VADDR=32
     else
-      #aarch64 LP64
-      LG_VADDR=48
+      #aarch64 LP64 allows to use the upper 8 bits for tagging
+      LG_VADDR=56
     fi
     AC_MSG_RESULT([$LG_VADDR])
     ;;


cmuellner:
This is basically a followup for #1193.
As @davidtgoldblatt mentioned, the 48 bits have been taken because it seemed to work.
The 48 bits seem to be derived from the fact that the typical Linux distribution provides a kernel with 48 VA bits. However, that does not allow to use upper bits for tagging unfortunately.
Instead we need to follow the ARMv8 ARM and the kernel here.

I have to admit, that I was not able to trigger a crash with the 48 bits.
So if anyone complains now, that this patch is to strict (well, just obeying the spec...)
then I can add a configure flag (--with-lg-vaddr).

make check says:
`Test suite summary: pass: 13/13, skip: 0/13, fail: 0/13`

I would be happy if that could make it into the release.

interwq:
Thanks for the patch @cmuellner! As David mentioned in #1193, with the current design, increasing vaddr bits beyond 48 will force jemalloc to use a less efficient metadata lookup -- namely changing the internal radix tree layout which is pretty important for the deallocation path.

Given that kernel doesn't seem to utilize the higher bits currently, I'm leaning to say that we should avoid paying the cost for now. Providing a config flag sound good to me. When high bits become more popular, we can think about how to optimize that case.

cmuellner:
I just had a look in the radix tree implementation and understand now what going on.

I initially thought you use pointer tagging (using the ability of the MMU to ignore the topmost 8-bits of a pointer), which allows to store information there and de-reference a pointer without masking.
But actually your radix tree implementation masks anyways (I just verified its a single `and` instruction).

So everything is fine and my patch indeed needlessly to strict.
I will close this for now and maybe open a new PR with a configure option.

interwq:
@cmuellner: FYI we are planning to release 5.1 tomorrow, in case you still want to make it to the release.

PR #1208, opened 2018-05-05
print() is a function in Python 3

cclauss:


davidtgoldblatt:
I don't see much benefit in cluttering these files for python3 compatibility when we're not using it anywhere. If we want a python3 feature at some point we can convert then.

PR #1209, opened 2018-05-05
Fix compiler for undefined macro HAVE_CPU_SPINWAIT.

cmuellner:
GCC 7 complains about the following, when HAVE_CPU_SPINWAIT
is not defined:

src/jemalloc/include/jemalloc/internal/spin.h:12:7:
warning: "HAVE_CPU_SPINWAIT" is not defined, evaluates to 0 [-Wundef]
 #  if HAVE_CPU_SPINWAIT
       ^~~~~~~~~~~~~~~~~

This patch addresses the warning, by checking if the symbol
exists and comparing it with 1.

### DIFF
--- a/include/jemalloc/internal/spin.h
+++ b/include/jemalloc/internal/spin.h
@@ -9,7 +9,7 @@ typedef struct {
 
 static inline void
 spin_cpu_spinwait() {
-#  if HAVE_CPU_SPINWAIT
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
Did you try a `make relclean` before configuring at dev? The `configure.ac` logic here is fairly straightforward, and we define `HAVE_CPU_SPINWAIT` down all paths.

cmuellner:
You are right.
My failure (same as in #1210).
Sorry for the noise.

PR #1210, opened 2018-05-05
Fix strerror_r() wrapper for cross-compilation.

cmuellner:
Commit f78d4ca3f moved the compile-time check for the strerror_r() return type
into a configure-time check. Since configure-time checks don't work for
cross-compilers, we put the old static check additionally back in as well.

### DIFF
--- a/src/malloc_io.c
+++ b/src/malloc_io.c
@@ -98,7 +98,8 @@ buferror(int err, char *buf, size_t buflen) {
 	FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM, NULL, err, 0,
 	    (LPSTR)buf, (DWORD)buflen, NULL);
 	return 0;
-#elif defined(JEMALLOC_STRERROR_R_RETURNS_CHAR_WITH_GNU_SOURCE) && defined(_GNU_SOURCE)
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
Why doesn't the configure-time check work? All we're doing is check if the snippet is compilable; it seems that if the cross-compilation succeeds, then we've detected that it works as well as we need it to, no?

cmuellner:
The configure check does not work on cross-compilation, because all compile tests are not performed during cross-compilation (because they cannot be executed anyways). autoconf manual [1] says:

```
If [...] configure enters cross compilation mode, so it doesn't run any tests that require execution.
```

I've just noticed that my compiler complains about not matching return type for strerror_r() and I therefore
analysed the situation.

[1] https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Hosts-and-Cross_002dCompilation.html

cmuellner:
After reading the configure check I understand what you mean and take back this PR.

Long story: I had an old configure script (the one from the last jemalloc release - 5.0.1) laying around in my source directory (leftover from a cross-compile experiment). And that does not do the configure check...

Sorry for the noise.

PR #1211, opened 2018-05-05
Fix include path order for out-of-tree builds.

cmuellner:
When configuring out-of-tree (source directory is not build directory),
the generated include files from the build directory should have higher
priority than those in the source dir.

This is especially helpful when cross-compiling.


davidtgoldblatt:
LGTM, thanks!

PR #1212, opened 2018-05-05
Build the docs on make install.

cmuellner:
Typically users compile with:

  ./configure
  make
  make install

This will fail for the git tree, since `make dist` is needed
to build the docs, which are required for `make install`.
The reason is, that building the library with `make` should not
require xsltproc to be installed. However, calling `make install`
requires the docs to be built. Therefore this patch introduces
a dependency for `build_doc` to `install_doc`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,15 +9,8 @@ If building from unpackaged developer sources, the simplest command sequence
 that might work is:
 
     ./autogen.sh
-    make dist
     make
     make install
-
-Note that documentation is not built by the default target because doing so
-would create a dependency on xsltproc in packaged releases, hence the
-requirement to either run 'make dist' or avoid installing docs via the various
-install_* targets documented below.
-
 
 ## Advanced configuration
 
--- a/Makefile.in
+++ b/Makefile.in
@@ -465,7 +465,7 @@ install_doc_man:
 	$(INSTALL) -m 644 $$d $(MANDIR)/man3; \
 done
 
-install_doc: install_doc_html install_doc_man
+install_doc: build_doc install_doc_html install_doc_man
 
 install: install_bin install_include install_lib install_doc
 


jasone:
The rules are intentionally set up to avoid a dependency on XML tools for official release tarballs.  Does the added dependency in this pull request trigger doc builds even if the docs are pre-built?  I'm guessing that it does trigger a rebuild, and if so, it breaks the intended behavior.

That said, as we de-emphasize official releases over time due to increased dev branch stability, keeping the set of build dependencies minimal may not remain a justifiable complication.

interwq:
It sounds to me that avoiding XML dependency on the release tarballs is still a valid argument. I'd vote for keeping it as is.

cmuellner:
This turns out to be an invalid fix for a valid problem.
The problem is, that the built docs in the release tarball can't be used for out-of-tree builds and the build process fails in this case. This can be reproduced with the following steps:

```
wget https://github.com/jemalloc/jemalloc/releases/download/5.1.0/jemalloc-5.1.0.tar.bz2
tar xf jemalloc-5.1.0.tar.bz2
cd jemalloc-5.1.0
mkdir build
cd build
../configure --prefix $(pwd)/install
make -j$(nproc)
make -j$(nproc) install
```

This will end up in:

```
/usr/bin/install -c -d /home/cm/jemalloc/jemalloc-5.1.0/build/install/share/man/man3
/usr/bin/install: cannot stat 'doc/jemalloc.html': No such file or directory
make: *** [Makefile:456: install_doc_html] Error 1
make: *** Waiting for unfinished jobs....
/usr/bin/install -c -m 644 doc/jemalloc.3 /home/cm/jemalloc/jemalloc-5.1.0/build/install/share/man/man3
/usr/bin/install: cannot stat 'doc/jemalloc.3': No such file or directory
make: *** [Makefile:463: install_doc_man] Error 1
```

The reason is, that the pre-shipped files are stored in `$(srcroot)/doc`, but the installation
copies from `$(objroot)doc`. That's no problem for in-tree builds, but breaks out-of-tree builds.

I could imagine to address this with introducing a `DOCS_HTML_SRC` variable in the Makefile
and use that as a fallback in case no file in `DOCS_HTML` exists. But it would look like a hack.

Another idea would be to copy everything from `$(srcroot)/doc` to `$(objroot)doc` during `make all`.
But I'm not sure if that would be an acceptable solution for you.

Any ideas?

interwq:
I'm not sure if we should workaround this too much, just to avoid building the doc and the dependency on xsltproc. As mentioned in [INSTALL.md](https://github.com/jemalloc/jemalloc/blob/dev/INSTALL.md), doing `make dist` will always work to build the doc in `$objroot/doc` (but of course requires xsltproc). Or if wanted we can install without doc using the  install_* targets.

I think for anything not a straight build on a packaged release, adapting the sequence for "unpackaged developer sources" in INSTALL.md is the right thing to do.

jasone:
I'm okay with us switching to building docs by default, as long as the configure script continues to warn (perhaps more loudly than it currently does) rather than failing if xsltproc is not found, and that we continue to have documented build subtargets that allow people to work around not being able to build the docs.  That said, I expect us to get complaints about builds failing, unless we configure such that we provide a fallback xsltproc that prints an informative error message with alternate build target instructions.

jasone:
@interwq, your variant sounds like a reasonable alternative.  I don't have any strong opinions one way or the other.

interwq:
Sorry for the long delay here. I'll take care of this in a couple of weeks.

interwq:
Replaced with #1430. Thanks for the input @cmuellner!

PR #1217, opened 2018-05-09
Fix an incorrect assertion.

interwq:
When configured with --with-lg-page, it's possible for the configured page size to be
greater than the system page size, in which case the page address may only be
aligned with the system page size.

### DIFF
--- a/src/pages.c
+++ b/src/pages.c
@@ -261,7 +261,7 @@ pages_decommit(void *addr, size_t size) {
 
 bool
 pages_purge_lazy(void *addr, size_t size) {
-	assert(PAGE_ADDR2BASE(addr) == addr);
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
LGTM, but, have you looked to make sure we're not doing this anywhere else?

src/pages.c
```diff
@@ -261,7 +261,7 @@ pages_decommit(void *addr, size_t size) {
 
 bool
 pages_purge_lazy(void *addr, size_t size) {
-	assert(PAGE_ADDR2BASE(addr) == addr);
+	assert(((uintptr_t)addr & ~(os_page - 1)) == (uintptr_t)addr);
```
jasone: How about this instead?
```assert(ALIGNMENT_ADDR2BASE(addr, os_page) == addr);```

PR #1220, opened 2018-05-11
Fix background thread index issues with max_background_threads.

interwq:


interwq:
After this fix, did a grep on "ncpus" to make sure we do not use it for background thread indexing.

PR #1226, opened 2018-05-24
Avoid taking extents_muzzy mutex when muzzy is disabled.

interwq:
When muzzy decay is disabled, no need to allocate from extents_muzzy.  This
saves us a couple of mutex operations down the extents_alloc path.


interwq:
We have seen some high # of lock operations on muzzy, even with muzzy decay disabled, e.g. 
`extents_muzzy                14564128`

Let's add a quick exit for that case.

PR #1227, opened 2018-05-25
fixing couple of warnings

devnexen:


davidtgoldblatt:
The build failures here look "real"; casting to (int) is not correct on (e.g.) 64-bit machines with 32-bit ints.

Closing for now, feel free to update if you don't mind fixing them.

PR #1228, opened 2018-05-29
Avoid taking large_mtx for auto arenas.

interwq:
On tcache flush path, we can avoid touching the large_mtx for auto arenas, since
it was only needed for manual arenas where arena_reset is allowed.


interwq:
# of lock operations with this diff:
`large 27`
W/o this diff:
`large 150553`

PR #1231, opened 2018-05-30
Only use dlsym if JEMALLOC_LAZY_LOCK is set.

neunhoef:
This avoids problems when using jemalloc in a completely static binary (for example with libmusl on Alpine), at least when JEMALLOC_LAZY_LOCK is not set.
The approach here works, since there is only a local `pthread_create` function if `JEMALLOC_LAZY_LOCK` is defined. If not, one can easily avoid `dlsym` by using the global symbol `pthread_create` directly.

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -812,7 +812,21 @@ pthread_create_fptr_init(void) {
 	if (pthread_create_fptr != NULL) {
 		return false;
 	}
+
+	/* If JEMALLOC_LAZY_LOCK is defined, we have overwritten
+         * pthread_create with our own function. In this case we have
+         * to use dlsym with RTLD_NEXT to find the real symbol
+         * "pthread_create" in the library.
+	 * If JEMALLOC_LAZY_LOCK is not defined, pthread_create
+         * is not overwritten and we better use the symbol directly
+         * to allow for static linking.
+	 */
+#ifdef JEMALLOC_LAZY_LOCK
 	pthread_create_fptr = dlsym(RTLD_NEXT, "pthread_create");
+#else
+	pthread_create_fptr = pthread_create;
+#endif
+
 	if (pthread_create_fptr == NULL) {
 		can_enable_background_thread = false;
 		if (config_lazy_lock || opt_background_thread) {


interwq:
@neunhoef: thanks for the input! Currently `pthread_create` is needed when 1) LAZY_LOCK as you mentioned, and 2) when [background thread](http://jemalloc.net/jemalloc.3.html#background_thread) is used (which is default off).  In the most recent [release 5.1](https://github.com/jemalloc/jemalloc/releases/tag/5.1.0) if you are not using these two features then `dlsym` should not be invoked -- this was just changed with 5.1, so if you are using 5.0 you may want to upgrade.  Does that solve your problem?

As for the case with background thread, `dlsym` is still necessary since users may wrapper pthread_create as well, and their wrapper may call malloc which would trigger reentrancy issues (e.g. we experienced such situation in HHVM).

interwq:
Closing since dlsym is avoided with 5.1 release (when possible). Let us know if anything.

neunhoef:
Unfortunately, this does not help us. Let me explain our use case: In ArangoDB we want to use jemalloc as allocator because it gives us better performance. We do not need LAZY_LOCK, but we need background threads, since as far as I understand otherwise jemalloc would never return any memory back to the operating system. In a database server we need that the process can shrink again after memory usage spikes.
On the other hand, we want to build and ship completely static binaries based on Alpine Linux and libmusl. This does not allow to use `dlsym` at all.
I understand that 5.1 (we are currently using 5.0.1) avoids calling dlsym if no background threads or LAZY_LOCK is used.
Furthermore, I do not follow your argument about users wrapping `pthread_create`. If LAZY_LOCK is off (regardless on whether there is a background thread or not), `jemalloc` itself does not compile in a function with name `pthread_create`. Other "users" defining a symbol `pthread_create` would be either other libraries or the main program using jemalloc. In any case the `pthread_create` function wrapping the pthread library function would either have to occur later in the link order than `jemalloc`. If it occurs earlier, then your `dlsym` call would miss the earlier `pthread_create`, since you use `RTLD_NEXT`, this cannot be right. If the other lib occurs in the link order later than `jemalloc`, then it is equally correct to simply use the symbol `pthread_create` directly, and the linker will resolve it correctly at link time.
This was my thinking behind the suggested patch.
If this is not right or not acceptable, I would like to suggest that we introduce a separate switch to allow to disable the usage of dlsym even with a background thread.
What do you think?

interwq:
In one of our use cases, the main program (with jemalloc statically linked) wraps pthread_create, and we need RTLD_NEXT to find the underlying pthread_create (since __real_pthread_create is not universally available), because the user defined pthread_create may call malloc which causes reentrancy issues (i.e. jemalloc calling into application code, which again calls malloc). But now that you mention it, I realize we should fall back to the default symbol if RTLD_NEXT finds nothing, which should help us avoid linking order issue like in #907.

To support a full static build, a separate config option sounds reasonable to me. I'll add that soon.

On a related note, although background_thread is recommended, it doesn't mean jemalloc won't return memory w/o it. When background thread is disabled, purge is triggered based on ticks (i.e. every N allocator calls). However it's less reliable since application threads may go idle and won't trigger the purging on unused dirty memory.

PR #1233, opened 2018-05-31
Fix MSVC build

maksqwe:


davidtgoldblatt:
Looks good, but we may delay merging for a little bit; the hooks functionality showed a surprising regression during one of our internal canaries, so we might have to revert hooks in general while we sort it out.

BTW, are these files manually edited? Can you test a workspace file without some sort of GUI action? We're constantly breaking these things, since we don't have CI on them or use them as part of our development process.

Thanks for the help on this!

PR #1240, opened 2018-06-04
Optimize ixalloc by avoiding a size lookup.

interwq:


interwq:
This avoids an unnecessary `isalloc(tsdn, ptr)` on the `xallocx` path.

src/arena.c
```diff
@@ -1606,17 +1608,23 @@ arena_ralloc_no_move(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size,
...
-		return false;
+		ret = false;
 	} else if (oldsize >= LARGE_MINCLASS && usize_max >= LARGE_MINCLASS) {
-		return large_ralloc_no_move(tsdn, extent, usize_min, usize_max,
+		ret = large_ralloc_no_move(tsdn, extent, usize_min, usize_max,
 		    zero);
+	} else {
+		ret = true;
 	}
+lable_done:
```
djwatson: spelling - label_done? Or even better, just done:

src/arena.c
```diff
@@ -1644,7 +1652,9 @@ arena_ralloc(tsdn_t *tsdn, arena_t *arena, void *ptr, size_t oldsize,
 
 	if (likely(usize <= SMALL_MAXCLASS)) {
 		/* Try to avoid moving the allocation. */
-		if (!arena_ralloc_no_move(tsdn, ptr, oldsize, usize, 0, zero)) {
+		size_t newsize;
```
djwatson: I like to name these types of things 'unused' for clarity

src/arena.c
interwq: Ah good catch. Will fix.

src/arena.c
interwq: Agreed. Will add.

interwq:
Addressed feedback @djwatson.

PR #1244, opened 2018-06-07
Add configure option --disable-libdl

### LINKED ISSUE #907: 5.0.0: Error in dlsym(RTLD_NEXT, "pthread_create")

interwq:
This makes it possible to build full static binary.

This resolves #907

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2130,8 +2130,7 @@ fi
 dnl ============================================================================
 dnl Enable background threads if possible.
 
-if test "x${have_pthread}" = "x1" -a "x${have_dlsym}" = "x1" \
-    -a "x${je_cv_os_unfair_lock}" != "xyes" ; then
+if test "x${have_pthread}" = "x1" -a "x${je_cv_os_unfair_lock}" != "xyes" ; then
   AC_DEFINE([JEMALLOC_BACKGROUND_THREAD])
 fi
 
--- a/configure.ac
+++ b/configure.ac
@@ -1588,6 +1588,21 @@ fi
 AC_DEFINE_UNQUOTED([LG_HUGEPAGE], [${je_cv_lg_hugepage}])
 
 dnl ============================================================================
+dnl Enable libdl by default.
+AC_ARG_ENABLE([libdl],
+  [AS_HELP_STRING([--disable-libdl],
+  [Do not use libdl])],
+[if test "x$enable_libdl" = "xno" ; then
+  enable_libdl="0"
+else
+  enable_libdl="1"
+fi
+],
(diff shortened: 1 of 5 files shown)


interwq:
Will rebase past #1242

configure.ac
```diff
@@ -1495,6 +1495,21 @@ AC_SUBST([jemalloc_version_bugfix])
...
+AC_ARG_ENABLE([libdl],
+  [AS_HELP_STRING([--disable-libdl],
+  [Do not use libdl])],
+[if test "x$enable_libdl" = "xno" ; then
+  enable_libdl="0"
+else
+  enable_libdl="1"
+fi
+],
+[enable_libdl="0"]
```
interwq: CI only. Will change back.

davidtgoldblatt:
Would it be possible to detect libdl at runtime (say, by declaring as weak symbols the things we want)? Every new configure flag is a personal affront to my dignity.

interwq:
Chatted with @davidtgoldblatt offline. This is to provide a way to not link against -ldl. Ready for review now.

interwq:
All tests passed. Will change the default setting and rerun again to make sure.

davidtgoldblatt:
Let's add a --disable-libdl option to CI, too?

interwq:
@davidtgoldblatt: We explicitly force lazy lock on FreeBSD, which can't work w/o libdl. I'm actually not too worried about testing the option w/ CI, the scope it affects is very limited (a single `#ifdef`pretty much, which likely won't be touched anytime soon). Does that sound reasonable?

davidtgoldblatt:
We don't have to test it on freebsd; but we should check to make sure that we don't see something unexpected start happening with our interception otherwise.

I'm just suggesting adding `--disable-libdl` to  `https://github.com/jemalloc/jemalloc/blob/dev/scripts/gen_travis.py#L47`

interwq:
Sounds good. Just added to Travis.

PR #1248, opened 2018-06-11
Avoid using atomic "acquire" with reader doing spin and waiting for writer to finish

rkmisra:
seqlock "try" from reader side could fail if writer interrupts reader by changing seq number.  In that case, reader could retry by "spinning" and waiting for writer to finish.  During this spinning, reader thread is continuously doing "acquire" on seq number load which may not be a good idea.  We could avoid that acquire during spinning if we put "acquire" fence after seq "Odd" number check and load seq number first time using relaxed memory order. 

Since atomic_thread_fence() with same memory order is stronger fence than corresponding load/store with same memory order,  It is a trade off i.e. saving large number of "acquire" fence when writer interrupts readers and reader spins waiting to acquire lock vs using stronger fence when reader is not interrupted by writer.


davidtgoldblatt:
I think we don't want this for now -- in our current use case (and most of the ones I foresee), we don't ever actually loop for data, so this is a strict loss. More generally, if we expect many reader failures, a seqlock is probably not the right data structure.

PR #1250, opened 2018-06-15
RFC: Use clang-format formatting.

davidtgoldblatt:
Most choices are configurable; see https://clang.llvm.org/docs/ClangFormatStyleOptions.html for details.

In most of the places where it breaks stuff up differently, I like the new way better. The only thing I don't like is the places where we do

```
case 'A': case 'B': case 'C':
  do_something();
```

it rewrites it into

```
case 'A':
case 'B':
case 'C':
  do_something();
```

(Which can be annoying if it's like every letter, and not just 3 of them). My thought is that we can just locally disable it there if we want.


davidtgoldblatt:
(I checked manually that `git blame -w` looks sensible, as well.)

davidtgoldblatt:
+cc @interwq @djwatson @jasone

include/jemalloc/internal/hash.h
```diff
@@ -13,305 +13,394 @@
 /* Internal implementation. */
 static inline uint32_t
 hash_rotl_32(uint32_t x, int8_t r) {
-	return ((x << r) | (x >> (32 - r)));
+  return ((x << r) | (x >> (32 - r)));
```
jasone: This entire file was imported with minimal changes from upstream in order to facilitate later merges from upstream.

include/jemalloc/internal/jemalloc_internal_decls.h
```diff
@@ -3,56 +3,56 @@
...
+#define LG_VADDR 32
+#endif
+#endif
 #else
-#  include <sys/param.h>
-#  include <sys/mman.h>
-#  if !defined(__pnacl__) && !defined(__native_client__)
-#    include <sys/syscall.h>
-#    if !defined(SYS_write) && defined(__NR_write)
-#      define SYS_write __NR_write
```
jasone: Nested cpp logic gets a lot harder to grok without whitespace....

include/jemalloc/internal/jemalloc_internal_inlines_a.h
```diff
@@ -9,164 +9,160 @@
...
-		/* Associated arena == NULL implies tcache init in progress. */
-		assert(tsd_tcachep_get(tsd)->arena == NULL ||
-		    tcache_small_bin_get(tsd_tcachep_get(tsd), 0)->avail !=
-		    NULL);
-		return true;
-	}
-
-	return false;
+  /*
+   * Thread specific auto tcache might be unavailable if: 1) during tcache
```
jasone: Comments like this one should be re-wrapped as part of the change.

src/jemalloc.c
```diff
@@ -698,534 +694,538 @@ check_entry_exit_locking(tsdn_t *tsdn) {
...
-	*opts_p = opts;
-	return false;
+  bool accept;
+  const char *opts = *opts_p;
+
+  *k_p = opts;
+
+  for (accept = false; !accept;) {
+    switch (*opts) {
+      case 'A':
```
jasone: Ow.

include/jemalloc/internal/jemalloc_internal_decls.h
davidtgoldblatt: I think this is a config setting actually, I'll double-check tomorrow

davidtgoldblatt: Note to self for tomorrow: `IndentPPDirectives: AfterHash`

davidtgoldblatt:
Agreed that the biggest benefit is automatability.

interwq:
Looks good to me! I assume the multiline comments can be re-wrapped automatically?

include/jemalloc/internal/hash.h
gnzlbg: We can use:

```c++
// clang-format off
... whole file goes here
// clang-format on
```

to easily disable clang-format for just this file without having to create a "white-list".

gnzlbg:
Which clang-format version is going to be used? Different versions produce slightly different formatting, so we probably want to use it in CI as well to catch "newcomers" maybe using a different version.

Updating the version then typically introduces minimal formatting changes.

djwatson:
We're going to make 'git blame' much harder to use after this diff.  Many other projects only run format on the new diff itself, and not the whole codebase. 

Also, this looks like a change from tabs to spaces, which looks like the majority of lines changed.

davidtgoldblatt:
Yeah; I think we'd have to do git blame -w for a while afterwards. OTOH, almost all of my git blame tracing recently has had to go through several levels of refactorings/cleanups recently to find the diff I was looking for, so I'm not sure I'd be that worried even if we couldn't ignore whitespace changes.

gnzlbg:
This is purely anecdotal, but I've worked on both, projects that did a full re-format (and then reformatted again every time clang-format was updated, that is ~twice a year), and projects that only formatted the new diffs only.

The projects that reformated the new diffs only were the worse of both worlds.

When the whole project is properly formatted, I just have tabs in emacs and vim configured to call clang-format on the whole file. That is, I dump code without spaces and hit tab, and that's pretty much it. You basically don't have to ever worry about formatting, it just happens, and if you forget it, CI catches it and its basically a one liner away.

If you only format the diffs you have to be much much careful not to accidentally format anything that you didn't change. And if you forget it, running clang-format in your whole project (which is typically a `make fmt` away) doesn't fix it, you have to use one of the many clang format python scripts that just format the diffs.

At the end of the day, having to go to one previous blame is just a minor inconvenience when you actually do a blame (and github makes this trivial). But formatting code is something that we implicitly do all the time, I'll optimize for that over the blame situation every time.

davidtgoldblatt:
Still going to happen, but in another PR.

PR #1251, opened 2018-06-16
Make use of MAP_EXCL on FreeBSD.

trasz:
The "O_FIXED | MAP_EXCL" combination works just like MAP_FIXED_NOREPLACE
recently introduced in Linux - except it's for FreeBSD and was added
some four years ago.

### DIFF
--- a/src/pages.c
+++ b/src/pages.c
@@ -588,6 +588,11 @@ pages_boot(void) {
 
 	init_thp_state();
 
+#ifdef __FreeBSD__
+	/*
+	 * FreeBSD doesn't need the check; madvise(2) is known to work.
+	 */
+#else
 	/* Detect lazy purge runtime support. */
 	if (pages_can_purge_lazy) {
 		bool committed = false;
@@ -601,6 +606,7 @@ pages_boot(void) {
 		}
 		os_pages_unmap(madv_free_page, PAGE);
 	}
+#endif
 
 	return false;
 }
--- a/src/pages.c
+++ b/src/pages.c
@@ -180,6 +180,31 @@ pages_map(void *addr, size_t size, size_t alignment, bool *commit) {
 	assert(alignment >= PAGE);
 	assert(ALIGNMENT_ADDR2BASE(addr, alignment) == addr);
 
+#if defined(__FreeBSD__) && defined(MAP_EXCL)
+	/*
+	 * FreeBSD has mechanisms both to mmap at specific address without
+	 * touching existing mappings, and to mmap with specific alignment.
+	 */
+	{
+		int prot = *commit ? PAGES_PROT_COMMIT : PAGES_PROT_DECOMMIT;
+		int flags = mmap_flags;
+
+		if (addr != NULL) {
+			flags |= MAP_FIXED | MAP_EXCL;
+		} else {
+			unsigned alignment_bits = ffs_zu(alignment);
+			assert(alignment_bits > 1);
+			flags |= MAP_ALIGNED(alignment_bits - 1);
+		}
+
+		void *ret = mmap(addr, size, prot, flags, -1, 0);
+		if (ret == MAP_FAILED) {
+			ret = NULL;
+		}
+
+		return ret;
+	}
+#endif
 	/*
 	 * Ideally, there would be a way to specify alignment to mmap() (like
 	 * NetBSD has), but in the absence of such a feature, we have to work


trasz:
Ping?

trasz:
Okay, so this results in something I didn't quite expect; it looks like this (it's the diff of truss(1) (think strace) output between python running with the vanilla jemalloc vs the same binary with patched jemalloc:

```
 mmap(0x0,262144,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON,-1,0x0) = 34391683072 (0x801e77000)
-mmap(0x801e77000,65536,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON,-1,0x0) = 34391945216 (0x801eb7000)
-munmap(0x801eb7000,65536)                       = 0 (0x0)
-mmap(0x0,528384,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON,-1,0x0) = 34391945216 (0x801eb7000)
+mmap(0x801e77000,65536,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_FIXED|MAP_ANON|MAP_EXCL,-1,0x0) ERR#12 'Cannot allocate memory'
+mmap(0x0,528384,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON|MAP_ALIGNED(12),-1,0x0) = 34391945216 (0x801eb7000)
 mmap(0x0,262144,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON,-1,0x0) = 34392473600 (0x801f38000)
-mmap(0x801f38000,131072,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON,-1,0x0) = 34392735744 (0x801f78000)
-munmap(0x801f78000,131072)                      = 0 (0x0)
-mmap(0x0,659456,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON,-1,0x0) = 34392735744 (0x801f78000)
+mmap(0x801f38000,131072,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_FIXED|MAP_ANON|MAP_EXCL,-1,0x0) ERR#12 'Cannot allocate memory'
+mmap(0x0,659456,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON|MAP_ALIGNED(12),-1,0x0) = 34392735744 (0x801f78000)
 mmap(0x0,262144,PROT_READ|PROT_WRITE,MAP_PRIVATE|MAP_ANON,-1,0x0) = 34393395200 (0x802019000)

```

... and so on.  So, the patch does improve things by reducing the number of syscalls (and the number of actual map modifications), but why does jemalloc request ranges that are obviously busy, because they were returned by the previous call to mmap?

trasz:
Ping?  The #1260 isn't a blocker for this one; it works just fine even with the bug.

davidtgoldblatt:
Sorry about the delay here -- we're tracking down a couple of slight regressions, and so don't want to add even more range to the search.

trasz:
Ping again?

src/pages.c
```diff
@@ -180,6 +181,29 @@ pages_map(void *addr, size_t size, size_t alignment, bool *commit) {
...
+	 * touching existing mappings, and to mmap with specific alignment.
+	 */
+	{
+		int prot = *commit ? PAGES_PROT_COMMIT : PAGES_PROT_DECOMMIT;
+		int flags = mmap_flags;
+
+		if (addr != NULL) {
+			flags |= MAP_FIXED | MAP_EXCL;
+		} else {
+			flags |= MAP_ALIGNED(ffs(alignment) - 1);
```
davidtgoldblatt: I think this should be `ffs_zu` (and that, therefore, we don't need to include string.h above.

src/pages.c
```diff
@@ -193,7 +193,10 @@ pages_map(void *addr, size_t size, size_t alignment, bool *commit) {
 		if (addr != NULL) {
 			flags |= MAP_FIXED | MAP_EXCL;
 		} else {
-			flags |= MAP_ALIGNED(ffs(alignment) - 1);
+			int alignment_bits = ffs(alignment);
```
davidtgoldblatt: Ditto with `ffs_zu`.

Probably even better, we should assert the alignment; we should always have tweaked it as necessary before we get to the pages module.

trasz:
Done.

trasz:
Ping?  :-)

davidtgoldblatt:
Looks great, thanks!

PR #1255, opened 2018-06-22
RFC: Tracking producer-consumer data in profiling

tyleretzel:
This isn't perfect but I wanted to get something out there for review.

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -151,8 +151,8 @@ struct prof_prod_cons_key_s {
  * Maps producer-consumer pair (i.e. [prof_prod_cons_key_t]) to bytes and object
  * counts (i.e. [prof_prod_cons_data_t]).
  */
-static ckh_t		prod_cons;
-static malloc_mutex_t 	prod_cons_mtx;
+ckh_t *prod_cons_tables;
+malloc_mutex_t *prod_cons_locks;
 
 /******************************************************************************/
 /*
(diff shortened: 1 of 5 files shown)


src/prof.c
```diff
@@ -131,6 +132,28 @@ static int		prof_dump_fd;
...
+	uint64_t prod;
+	uint64_t cons;
+};
+
+/*
+ * Maps producer-consumer pair (i.e. [prof_prod_cons_key_t]) to bytes and object
+ * counts (i.e. [prof_prod_cons_data_t]).
+ */
+static ckh_t		prod_cons;
+static malloc_mutex_t 	prod_cons_mtx;
```
tyleretzel: one small question I have about naming: is anything implied semantically by naming variables lock vs. mutex? It seems like both are used.

interwq:
Biggest question for me is how we plan to use this data -- how to report them or help figuring out potential issues. I think we should have some ideas before landing.

The other thought is, the sampling mechanism is built to capture memory usage, which means it tends to sample large objects. I'd suggest checking if the large allocations in production tend to be remote freed often -- e.g. if they have a very high chance of being prod-cons in all cases, then this feature doesn't provide much info.

src/prof.c
```diff
@@ -2317,10 +2415,42 @@ prof_boot2(tsd_t *tsd) {
 			return true;
 		}
 
+		prod_cons_tables = (ckh_t *)base_alloc(tsd_tsdn(tsd),
```
interwq: Only alloc and init `if (opt_prof_prod_cons)`.

src/prof.c
```diff
@@ -253,6 +276,53 @@ prof_malloc_sample_object(tsdn_t *tsdn, const void *ptr, size_t usize,
...
+	prof_prod_cons_key_t pair;
+	pair.prod = prod_uid;
+	pair.cons = cons_uid;
+	prof_prod_cons_data_t *data;
+
+	uint64_t i = prof_prod_cons_index_choose(pair);
+	ckh_t *table = &prod_cons_tables[i];
+	malloc_mutex_t *lock = &prod_cons_locks[i];
+
+	/* global lock for now. TODO: make this reader-writer, perhaps? */
```
interwq: nit: update the comment.

src/prof.c
```diff
@@ -2317,10 +2415,42 @@ prof_boot2(tsd_t *tsd) {
...
+		}
+
+		for (i = 0; i < PROF_PROD_CONS_LOCKS; i++) {
+			if (malloc_mutex_init(&prod_cons_locks[i],
+			    "prof_prod_cons", WITNESS_RANK_OMIT,
+			    malloc_mutex_rank_exclusive)) {
+				return true;
+			}
+
+			if (ckh_new(tsd, &prod_cons_tables[i],
```
interwq: maybe also create the table on demand?

djwatson:
> The other thought is, the sampling mechanism is built to capture memory usage, which means it tends to sample large objects. 

This is actually a big issue - remote frees probably only really matter for lots of small objects, not large ones.

djwatson:
> We can get approximate that data using this data, by biasing according to the likelihood of a given size class getting sampled.

I think this is a reasonable approach, but should be in jemalloc and not built on top externally.  Adding an extra branch in the fast path would stuck, but maybe that's what it takes.

Related, having stats for frequency of allocations, not related to allocation size, would be nice to try and spot places where we should be stack allocating instead of heap allocating.

src/prof.c
```diff
@@ -261,6 +331,20 @@ prof_free_sampled_object(tsd_t *tsd, size_t usize, prof_tctx_t *tctx) {
...
 
+	prof_tdata_t *cons_tdata = prof_tdata_get(tsd, false);
+
+	/*
+	 * cons_tdata can be null if the current thread is in an unusual state:
+	 * e.g. it's being destroyed and tsd is in state tsd_state_purgatory.
+	 * We just ignore these cases.
+	 */
+	if (cons_tdata != NULL && opt_prof_prod_cons) {
+		uint64_t prod_uid, cons_uid;
```
djwatson: we probably also want arena id and not just threadid?

davidtgoldblatt:
> Related, having stats for frequency of allocations, not related to allocation size, would be nice to try and spot places where we should be stack allocating instead of heap allocating.

In a separate PR to come, Tyler is also working on recording lifetime information for profiled allocations, which could help for this as well. I don't know that we have great ideas for ways to surface that information, though.

PR #1258, opened 2018-06-26
Fix arena locking in tcache_bin_flush_large().

interwq:
This regression was introduced in c834912 (incorrect arena used).

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -193,8 +193,8 @@ tcache_bin_flush_large(tsd_t *tsd, cache_bin_t *tbin, szind_t binind,
 	assert(binind < nhbins);
 	assert((cache_bin_sz_t)rem <= tbin->ncached);
 
-	arena_t *arena = tcache->arena;
-	assert(arena != NULL);
+	arena_t *tcache_arena = tcache->arena;
+	assert(tcache_arena != NULL);
 	unsigned nflush = tbin->ncached - rem;
 	VARIABLE_ARRAY(extent_t *, item_extent, nflush);
 	/* Look up extent once per item. */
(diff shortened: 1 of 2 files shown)


src/tcache.c
```diff
@@ -212,7 +212,7 @@ tcache_bin_flush_large(tsd_t *tsd, cache_bin_t *tbin, szind_t binind,
 			idump = false;
 		}
 
-		bool lock_large = !arena_is_auto(arena);
```
interwq: This was the bug.

davidtgoldblatt:
Is there a test case we could add that would catch this?

interwq:
re: test case. Thought about this a bit, but leaning to say this should be taken care of by a stress test instead. We have some assertions for cases like this, e.g. https://github.com/jemalloc/jemalloc/blob/c834912aa9503d470c3dae2b2b7840607f0d6e34/src/large.c#L334
however in this instance we'd need to do a producer-consumer. I'll try adding that pattern somewhere.

interwq:
Just added a new test case for remote free + tcache_flush. Verified that it catches the issue being fixed here.

test/integration/mallocx.c
```diff
@@ -71,6 +71,38 @@ TEST_BEGIN(test_overflow) {
...
+	    "Unexpected mallctl() failure");
+	size_t large_sz;
+	sz = sizeof(size_t);
+	assert_d_eq(mallctl("arenas.lextent.0.size", (void *)&large_sz, &sz,
+	    NULL, 0), 0, "Unexpected mallctl failure");
+
+	void *ptr = mallocx(large_sz, MALLOCX_ARENA(arena)
+	    | MALLOCX_TCACHE_NONE);
+	void **ret = (void **)arg;
+	*ret = ptr;
```
interwq: Had to pass back the address this way, since the thread exit code on Windows is 32bit.

interwq:
cc @davidtgoldblatt in case you want to take a look at the test case.

Issue #1149, opened 2018-03-20
Allocation lifetime profiling


davidtgoldblatt:
We could record lifetime information along with our other sampled data.

Potential uses:
- Move long-lived allocations into a dedicated hugepage arena.
- Find allocation sites where allocations are short-lived and usually small, and convert them to stack allocations where possible. (This one is trickier, since it's not easy to infer from a live heap).

PR #1259, opened 2018-06-26
RFC: Profiling for lifetime of objects

tyleretzel:
Currently just the average lifetime for each size class is aggregated, just to have some meaningful output. It doesn't seem to me like this is very good for the uses in #1149, but some sort of per-allocation site tracking could be built off of this.

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -57,6 +57,32 @@ arena_prof_tctx_reset(tsdn_t *tsdn, const void *ptr, UNUSED prof_tctx_t *tctx) {
 	large_prof_tctx_reset(tsdn, extent);
 }
 
+JEMALLOC_ALWAYS_INLINE nstime_t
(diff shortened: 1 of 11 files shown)


tyleretzel:
I guess today we basically agreed that we need data per-stacktrace for this feature to be useful. One thought I had afterwards: what about understanding lifetimes in terms of where memory from a particular allocation site is freed (i.e. another stacktrace, rather than some number of nanoseconds)? This seems like it would be better for finding things that could be stack allocated, and also seems useful for just trying to understand the behavior of a program. Of course this requires another call to the libunwind stuff on `free`, which could really slow things down... thoughts on this?

davidtgoldblatt:
I think the extra backtrace is fine; this is already a slow path anyways.

Finding stack traces where the stack trace on free is the same as the stack trace on alloc (or perhaps, an extension of it) is maybe a good indicator of heap allocations that could be moved to the stack.

davidtgoldblatt:
(Especially if we, say, impose a threshold on the max size we care about -- small, short-lived objects are good candidates).

PR #1261, opened 2018-06-27
Move tsd link and in_hook after tcache.

davidtgoldblatt:
This can lead to better cache utilization down the common paths where we don't
touch the link.


interwq:
LGTM assuming tests come back clean.

Issue #1265, opened 2018-06-28
Proposal: `mallctl` option for continuous dumping of profiled allocations


tyleretzel:
There are some opportunities in profiling that don't fit into the current model for profiling. As of now, profiling keeps track of allocated memory that hasn't been freed, and entirely forgets about allocations that have been freed (ignoring `opt.prof_accum`). This is nice because it allows profiling to be on all the time without runaway memory usage, but it's not useful if the information you want to collect talks about when/how an object is freed. Two potentially useful sorts of data in this category are:

a) data about allocation lifetimes (#1149): we could see the average lifetime for an allocation site (e.g. in nanoseconds), and we could also see the set of stack traces where allocations from a particular allocation site are freed.

b) data about threads: we could see if allocations from a particular allocation site are often allocated on one thread and freed on a different thread.

One easy way to collect this data is to just dump all sampled allocations to a log, and then parse out this information in another program. This would be something that can be turned on and off via `mallctl`.

To be concrete, each allocation would be appended to a dump file when it is freed, so the dump file would be a list of these allocations with appropriate metadata. This could be e.g. json or some binary format, but we should include at least the following information

- Size
- Timestamp for `malloc` call
- Timestamp for `free` call
- Stack trace for `malloc` call
- Stack trace for `free` call
- Thread id/name for `malloc` call

davidtgoldblatt:
Looks good. Could you also sketch out:
- Proposed mallctl names and semantics
- Some sort of log rotation strategy so this could be left on all the time (probably logs emitted in a directory, with timestamps? And a setting to set the rotation delay).
- Possible json format (e.g. we might want some way to compress stack traces, we want some way to represent thread info without replicating it per stack trace, etc.).

tyleretzel:
It seems like the most natural thing to do is to copy `prof.dump` and take a `char *` for the file name, providing some reasonable default if `NULL` (perhaps `<prefix>.<pid>.<seq>.json`?). I suggest we have:

> prof.log_start (const char *) [--enable-prof]
> Log sampled allocations to the specified file name, or if NULL is specified, to a file according to the pattern `<prefix>.<pid>.<seq>.json`, where `<prefix>` is controlled by the opt.prof_prefix option.

And I suppose this would do nothing if we already started a log. Maybe it somehow makes sense to allow logging to multiple files at once, but I don't see that being very useful.

To stop allocating, we would have:

> prof.log_stop (void) [--enable-prof]
[...]

PR #1267, opened 2018-07-03
Logging sampled allocations to a file

tyleretzel:
This adds support for the `mallctl` calls outlined in #1265, as well as a "prof_log" option to start logging automatically when the program starts.

The json format is approximately the same as that outlined in the issue: stack traces and threads are cached in arrays, and then allocations reference traces/threads by index. The output json looks like this:

```
{
  "threads": [
      {
        "thr_uid": 0,
        "thr_name": "thread name 0"
      },
      {
        "thr_uid": 1,
        "thr_name": "thread name 1"
      },
      ...
  ],
  "stack_traces": [
      [
        "0xfffffffff",
        "0xffffffffe",
        "0xffffffffd",
        "0xffffffffc"
      ],
      ...
   ],
  "allocations": [
      {
        "alloc_thread": 0,
        "free_thread": 1,
        "alloc_trace": 0,
        "free_trace": 1,
        "alloc_timestamp": 240009293498,
        "free_timestamp": 240009293499
      },
      ...
  ]
}
```

As a general summary of the design, allocations and metadata are stored in memory each time `free` is called on a sampled allocation, and then this is dumped to a json file either when the prof_log_stop `mallctl` call is made, or when the program exits.


djwatson:
I didn't notice any code issues, very clean.  

Output format looks good.  Do you have a plan for how to parse this and present it?  jeprof?  Something else?

tyleretzel:
> Output format looks good. Do you have a plan for how to parse this and present it? jeprof? Something else?

We talked about this offline, but, just for the record, we agreed that for now it's up to external programs to do something with the data. Maybe adding something to jeprof makes sense sometime in the future.

davidtgoldblatt:
I think the format could also use an "info" section; with profiling duration and sampling rate (anything else)?

tyleretzel:
Maybe the pid and jemalloc version would be useful as well. Since it doesn't cost anything significant to add these in, should I just air on the side of adding anything that might be useful?

Another thing: would the requested size be useful, in addition to the allocated size? I'm not sure if this is crucial to have, but it might be nice just for trying to understand what your program is doing. I think this would need to be put in the `extent_t` as well, since we forget about the requested size after allocation, but this seems easy enough to implement.

davidtgoldblatt:
> Maybe the pid and jemalloc version would be useful as well. Since it doesn't cost anything significant to add these in, should I just err on the side of adding anything that might be useful?

Sure, seems reasonable

> Another thing: would the requested size be useful, in addition to the allocated size? I'm not sure if this is crucial to have, but it might be nice just for trying to understand what your program is doing. I think this would need to be put in the extent_t as well, since we forget about the requested size after allocation, but this seems easy enough to implement.

I think probably, eventually, but that that may belong in a different diff.

Issue #1074, opened 2017-11-14
Avoid recomputing size on allocation and reallocation


gnzlbg:
A single `nallocx` call takes ~10-15ns in my test system.

Currently, Rust needs to,  call `nallocx` on allocation: 

```rust
let flags = align_to_flags(align);
let p = mallocx(size, flags) as *mut u8;
if !p.is_null() {
   // unnecessary nallocx call?
    real_size = nallocx(size, flags) as usize;
    return (p, real_size);
}
```

and reallocation:

```rust
let flags = align_to_flags(align);
let ptr = rallocx(ptr as *mut c_void, new_size, flags) as *mut u8;
if !p.is_null() {
   // unnecessary nallocx call?
   real_size = nallocx(size, flags) as usize;
   return (p, real_size);
}
```

Given that:

- rust always asks for aligned allocations,
- mallocx always internally computes the real size for aligned allocations (see: https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L1777)
- rallocx always internally computes the real size for aligned allocations (see: https://github.com/jemalloc/jemalloc/blob/b5d071c26697813bcceae320ba88dee2a2a73e51/src/arena.c#L1815 ) 

it seems that we could save those 10ns if jemalloc wouldn't throw this information away. 

Would it be possible to provide a pair of `mallocx`/`rallocx` methods that always return the allocation size for aligned allocations?

(I don't know whether this has a cost for unaligned allocations, or what that cost might be, but it might be worth it to check that out as well).

davidtgoldblatt:
Can you share your test system details + how you're building and testing nallocx speed? That sounds high to me.

`nallocx` has an `__attribute__((pure))` attached to it, so that the result need not be computed if not used; can Rust do something similar to propagate that information to the Rust-side compiler, so that you only pay the cost if necessary? The nallocx computations are also highly pipelinable, so I think that the "true" cost is likely lower than might be reflected in microbenchmarks.

Committing to a new allocation API is a big enough deal for us that I can't really see us doing it for this case. In some of the other threads, we've suggested fastpathing "freebie" alignment values into the malloc / nallocx-with-zero-flags pathways; have you tried this?

Does nallocx actually show up in application-level profiles?

gnzlbg:
[...]

PR #1270, opened 2018-07-10
Add experimental API to support P0901r0: "smallocx_return_t smallocx(size, flags)"

gnzlbg:
This is related to #1074 and [P0901r0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0901r0.html).

@davidtgoldblatt this is my old branch that added a new API, `smallocx` (sized mallocx), that returns the allocation size on allocation.

I'd like to have feedback on the general direction taken by the PR before I invest more time in tests, docs, putting this all behind an opt-in configure flag... The discussion on `gitter` suggested that, if this was ever implemented, it would be an experimental API, opt-in, without any backwards compatibility between `jemalloc` versions, etc. (at least initially).

I also have a couple of questions:

*  in my old branch the signature of the API is `void* smallocx(requested_size, flags, usize* allocation_size)`, but @davidtgoldblatt IIRC on gitter you suggested that it would be better to return `usize` instead. Could you write down your thoughts about the signature that this API should have?

* @davidtgoldblatt you mentioned in #1074 it that the main purpose of this API would be for better telemetry and better feedback about fragmentation. I don't think this PR addresses those concerns, but maybe you could expand on what exactly do you mean by this and maybe form your thoughts about how this should be addressed? 

---

Implementation: I tried to come up with the simplest implementation possible that allowed us to return the size class of an allocation without having to recompute it, at least in some cases. For that, I added a new `usize` option to `static_opts_s` and `dynamic_opts_s` and basically cache the result of `sz_index2size` when its invoked, and if it is not invoked, `smallocx` will call `sz_index2size` , as opposed to `mallocx`.

And that's pretty much it. There might be more situations in which we can avoid calling `sz_index2size`. I did this a long time ago, but IIRC I actually tried to propagate this back from the code that actually uses the arenas and that was not trivial, and also, the size class was still not always available (this confused me back then, but maybe I should take a look at this again now that I am slightly more familiar with the code base). 

--

cc @hunterae @ckennelly


ckennelly:
If you're looking for the coauthor of P0901, I think you're looking for @ahh, not @hunterae.

re telemetry:  This is discussed in passing in P0901.  It's more useful for us (TCMalloc) to see the size distribution demanded from `malloc(x)` than see `malloc(nallocx(x))`.  The latter will always line up perfectly with our existing size classes, so we lose signal on the internal fragmentation induced by our (potentially poor) choices.  (We could instrument `nallocx`, but userspace code can also cache the result and use it repeatedly, which then gives us a trickier problem to scale the profiles with...)

gnzlbg:
@ckennelly thanks, that makes sense. @davidtgoldblatt the current PR logs on entry the requested size, and on exit the actual usable size, but it doesn't change anything about the stats at all.

davidtgoldblatt:
> in my old branch the signature of the API is void* smallocx(requested_size, flags, usize* allocation_size), but @davidtgoldblatt IIRC on gitter you suggested that it would be better to return usize instead. Could you write down your thoughts about the signature that this API should have?

```
typedef struct {
  void *ptr;
  size_t usize;
} smallocx_return_t;

smallocx_return_t smallocx(requested_size, flags);
```

This is for the reason P0901 noted: "In practice, underlying malloc implementations provide actual definitions of ::operator new symbols which are called like any other function. Passing a reference parameter requires us to actually return the size via memory; most important ABIs support returning at least two scalar values in registers (even if they’re members of a trivially copyable struct) which can be dramatically more efficient."

> @davidtgoldblatt you mentioned in #1074 it that the main purpose of this API would be for better telemetry and better feedback about fragmentation. I don't think this PR addresses those concerns, but maybe you could expand on what exactly do you mean by this and maybe form your thoughts about how this should be addressed?

Suppose you have the distribution of requested sizes from the allocator (from your sampling). Then you can pick size classes so as to minimize space wasted due to internal fragmentation. This sort of telemetry becomes less accurate though, if you want to tie your data structure to your allocator (i.e. "from nallocx, I know my vector really has enough space for 12 items, even though I only asked for 11; I don't need to reallocate if another one is pushed on"); the user might be calling nallocx out of band to determine what its request patterns will be. nallocx-influenced calls (or those that become nallocx-influenced after the fact) shouldn't count as fragmentation (at least, not in the same way; a vector that ends up not using all the space it got morally ought to report space wastage according to the max memory it ever used, but I don't think that's necessarily feasible).

(To be clear, I'm mostly just restating insights Chris and Andrew had; the ideas here are mostly theirs).

In practice, I don't expect that this will be useful for anyone except us, when we do tuning.

> Implementation: I tried to come up with the simplest implementation possible that allowed us to return the size class of an allocation without having to recompute it, at least in some cases. For that, I added a new usize option to static_opts_s and dynamic_opts_s and basically cache the result of sz_index2size when its invoked, and if it is not invoked, smallocx will call sz_index2size , as opposed to mallocx.

This is essentially the right way to do this, I think. We'd want it configure-guarded with a scary name. One way to make this easy would be to have

```
static inline smallocx_return_t smallocx_impl(...) { ... }
#ifdef MY_CONFIG_OPT
smallocx_return_t smallocx(...) {
  return smallocx_impl(...);
}
#endif
```

gnzlbg:
So I've updated the function API, put everything behind a configure option, adapted the mallocx integration tests, and updated the travis script [0]

--

[0] Once #1272 is merged I'll update this part because the gen travis script wasn't properly updated in the last PRs and this commit reverts a couple of things..

gnzlbg:
I've fixed the bug that was causing the tests to fail and rebased and cleaned the history to make it ready for merging.

gnzlbg:
@jasone ping, this is blocked on your review :)

jasone:
I don't have time this weekend, but will try to get to this on Monday.

davidtgoldblatt:
Some of the gitter discussion: https://gitter.im/jemalloc/jemalloc?at=5b71de95937eee24231226ae

jasone:
Having been once burned by "experimental" APIs that ended up permanent symbols (see `*allocm` in FreeBSD), I'm not keen on even providing predictable symbols that can be linked via trivial special knowledge.  This pull request does just that, and I worry that we will be put under pressure to preserve `smallocx` long-term regardless of our experiences with the API's practical utility.

A year or two ago @davidtgoldblatt and I chatted about ideas for exposing experimental APIs, and he made a suggestion I would like to put forward for discussion.  The idea is to provide a header macro for `smallocx` that concatenates the current git revision id to the symbol, so that it is straightforward to link with the resulting jemalloc library and run the resulting executable, but intentionally cause library load failures  for any libjemalloc.so that did not come from precisely the same git revision.

gnzlbg:
Thank your for looking into this!

> The idea is to provide a header macro for smallocx that concatenates the current git revision id to the symbol, so that it is straightforward to link with the resulting jemalloc library and run the resulting executable, but intentionally cause library load failures for any libjemalloc.so that did not come from precisely the same git revision.

For experimentation purposes, on the Rust side at least, we wouldn't even need this. As in, we would be ok if the symbol is not exposed in any header, but the symbol name has a git hash attached to it.

I can try getting that to work in this PR, and if there is consensus about making this more public we can also expose it in a header to make it more convenient for C and C++ users to use.

---

As for the particulars about how to do this, are we already doing something similar anywhere? I suppose I just need to query the git version from the autoconf files when this feature flag is enabled, and pass the current hash as a macro define to the C compiler, so that I can concatenate it in the symbol name. One consequence of this is that `git` now becomes a required dependency to compile the library with this feature enabled, but I think this is ok since its behind a feature anyways.

jasone:
We already grab the git id as `jemalloc_version_gid` in `configure.ac`, so I think this just boils down to propagating that everywhere it's needed.  As an alternative, see what [hhvm does](https://github.com/facebook/hhvm/blob/master/hphp/hhvm/generate-buildinfo.sh#L48), but in my opinion it's probably best to just use the git id, since anyone using an experimental feature can reasonably be expected to access jemalloc via git.

I'm open to leaving experimental APIs out of the headers, but to me that seems like it might be more friction than necessary for those trying out experimental features.

gnzlbg:
@jasone so I've applied the suggested change: the symbol exported is now `smallocx_{jemalloc_version_gid}`. Let me know what you think.

src/jemalloc.c
```diff
@@ -1747,6 +1747,11 @@ struct static_opts_s {
 	 * initialization) options.
 	 */
 	bool slow;
+	/*
+	 * Return size
```
jasone: Nit: Add period and remove extra line.

src/jemalloc.c
```diff
@@ -2746,6 +2756,71 @@ int __posix_memalign(void** r, size_t a, size_t s) PREALIAS(je_posix_memalign);
...
+				dopts.tcache_ind = MALLOCX_TCACHE_GET(flags);
+			}
+		} else {
+			dopts.tcache_ind = TCACHE_IND_AUTOMATIC;
+		}
+
+		if ((flags & MALLOCX_ARENA_MASK) != 0)
+			dopts.arena_ind = MALLOCX_ARENA_GET(flags);
+	}
+
```
jasone: Nit: Remove extra blank lines.

test/integration/smallocx.c
```diff
@@ -0,0 +1,292 @@
...
+	    "Unexpected mallctl() failure");
+	size_t large_sz;
+	sz = sizeof(size_t);
+	assert_d_eq(mallctl("arenas.lextent.0.size", (void *)&large_sz, &sz,
+	    NULL, 0), 0, "Unexpected mallctl failure");
+
+	smallocx_return_t r = smallocx(large_sz, MALLOCX_ARENA(arena)
+                                 | MALLOCX_TCACHE_NONE);
+	void *ptr = r.ptr;
+  assert_zu_eq(r.size, nallocx(large_sz, MALLOCX_ARENA(arena)
```
jasone: Nit: Whitespace.

test/integration/smallocx.c
```diff
@@ -0,0 +1,292 @@
...
+	size_t sz = sizeof(unsigned);
+	assert_d_eq(mallctl("arenas.create", (void *)&arena, &sz, NULL, 0), 0,
+	    "Unexpected mallctl() failure");
+	size_t large_sz;
+	sz = sizeof(size_t);
+	assert_d_eq(mallctl("arenas.lextent.0.size", (void *)&large_sz, &sz,
+	    NULL, 0), 0, "Unexpected mallctl failure");
+
+	smallocx_return_t r = smallocx(large_sz, MALLOCX_ARENA(arena)
+                                 | MALLOCX_TCACHE_NONE);
```
jasone: Nit: Whitespace.

test/integration/smallocx.c
```diff
@@ -0,0 +1,292 @@
...
+	assert_ptr_null(smallocx(0x80000000UL, MALLOCX_ALIGN(0x80000000UL)).ptr,
+	    "Expected OOM for smallocx()");
+#endif
+}
+TEST_END
+
+/* Re-enable the "-Walloc-size-larger-than=" warning */
+JEMALLOC_DIAGNOSTIC_POP
+
+TEST_BEGIN(test_basic) {
```
jasone: This whole function has whitespace issues.  (The corresponding `mallocx` test code it was copied from appears to have different issues.)

test/integration/smallocx.c
```diff
@@ -0,0 +1,292 @@
...
+		for (sz = 1;
+		    sz < 3 * alignment && sz < (1U << 31);
+		    sz += (alignment >> (LG_SIZEOF_PTR-1)) - 1) {
+			for (i = 0; i < NITER; i++) {
+				nsz = nallocx(sz, MALLOCX_ALIGN(alignment) |
+				    MALLOCX_ZERO);
+				assert_zu_ne(nsz, 0,
+				    "nallocx() error for alignment=%zu, "
+				    "size=%zu (%#zx)", alignment, sz, sz);
+        smallocx_return_t ret = smallocx(sz, MALLOCX_ALIGN(alignment) |
```
jasone: Nit: Whitespace.  And more later in this function.

jasone:
It looks like one of my comments got dropped by the review interface:

Re: 8e92ca6, I'm fine with it being omitted from the set of commits, now that you've added name version-based symbol mangling.  I don't have a strong opinion either way.

gnzlbg:
@jasone I hope I fixed all the stylistic issues. I'll prefer to merge this as is and, after getting some experience with using this in the Rust std library, then reconsider exposing it in a more easier to use way (and adding some docs for it).

jasone:
Cool, thanks for putting in the effort to work through some of the details on exposing experimental APIs.  This is going to be a big help the next time we want to do something similar.

PR #1271, opened 2018-07-10
Add valgrind build bots to CI

gnzlbg:
This commit adds two build-bots to CI that test the release builds
of jemalloc on linux and macOS under valgrind.

The macOS build bot is allowed to fail because valgrind reports
errors about reads of uninitialized memory, and in some cases,
segfaults.

### DIFF
--- a/scripts/gen_travis.py
+++ b/scripts/gen_travis.py
@@ -113,8 +113,28 @@ for combination in unusual_combinations_to_test:
 
 # Development build
 include_rows += '''\
+    # Development build
     - os: linux
       env: CC=gcc CXX=g++ COMPILER_FLAGS="" CONFIGURE_FLAGS="--enable-debug --disable-cache-oblivious --enable-stats --enable-log --enable-prof" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
 '''
+
+# Valgrind build bots
+include_rows += '''
+    # Valgrind
+    - os: linux
+      env: CC=gcc CXX=g++ COMPILER_FLAGS="" CONFIGURE_FLAGS="" EXTRA_CFLAGS="-Werror -Wno-array-bounds" JEMALLOC_TEST_PREFIX="valgrind"
+      addons:
+        apt:
+          packages:
+            - valgrind
+'''
+
+# To enable valgrind on macosx add:
+#
+#  - os: osx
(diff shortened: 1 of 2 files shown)


gnzlbg:
Once #1272 is merged I'll rebase this to add these changes to the gen_travis script.

gnzlbg:
This should be ready to go as well.

scripts/gen_travis.py
```diff
@@ -113,8 +113,31 @@ def format_job(combination):
...
+        apt:
+          packages:
+            - valgrind
+    - os: osx
+      env: CC=gcc CXX=g++ COMPILER_FLAGS="" CONFIGURE_FLAGS="" EXTRA_CFLAGS="-Werror -Wno-array-bounds" JEMALLOC_TEST_PREFIX="valgrind"
+      install: brew install valgrind
+'''
+
+# Builds allowed to fail
+include_rows += '''
```
davidtgoldblatt: What's the point of an "allow_failures" line? Shouldn't we just not include it?

gnzlbg:
> What's the point of an "allow_failures" line? Shouldn't we just not
include it?

I often prefer to leave builds that should work, but are currently broken,
as an allow failure build, particularly when the big is reported upstream.

I don’t know your preference, CI resources, and don’t really mind either
way. (I can just comment the code out in the gen travis script but I would
prefer not to have to delete it from there and then re add it once it’s
fixed).

gnzlbg:
@davidtgoldblatt I've disabled the macosx valgrind build but left a comment in the gen_travis.py script with the instructions to enable it. I hope that is ok?

gnzlbg:
@davidtgoldblatt ping

davidtgoldblatt:
So, I mean, I'm OK with tests that check our level of support for running under Valgrind, but I don't want to endorse full Valgrind integration per se. So I'm OK merging, but I don't want to imply that like, we'll take any diff that makes Valgrind work fully; it would depend on how "intrusive" that diff turns out to be.

This looks good insofar as it goes right now, though.

gnzlbg:
@davidtgoldblatt yes, my here is not to ensure that "jemalloc supports valgrind", but that "valgrind supports jemalloc". Currently, valgrind supports jemalloc in Linux but it does not do so in OSX because valgrind does not support all syscalls that jemalloc uses, but that's a valgrind bug. Once it is fixed, I want to enable a second valgrind build bot to detect that valgrind does not break jemalloc.

PR #1272, opened 2018-07-10
Simplify .travis.yml a bit

gnzlbg:


gnzlbg:
Oh I see, I update that one then with the previous changes also and regenerate .travis.yml.

scripts/gen_travis.py
```diff
@@ -12,6 +12,7 @@
 
 before_script:
   - autoconf
+  - scripts/gen_travis.py > travis_script && diff .travis.yml travis_script
```
davidtgoldblatt: Good idea

scripts/gen_travis.py
```diff
@@ -61,47 +62,62 @@
...
+for combination in unusual_combinations_to_test:
+    include_rows += format_job(combination)
+
+# Development build
+include_rows += '''\
+    - os: linux
+      env: CC=gcc CXX=g++ COMPILER_FLAGS="" CONFIGURE_FLAGS="--enable-debug --disable-cache-oblivious --enable-stats --enable-log --enable-prof" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
+      script:
+        - make check
+        - make -j test/unit/log
```
davidtgoldblatt: Probably don't need the log-specific lines

gnzlbg: Indeed. I've removed them and the test is still compiled and run, and appropriately fails when its broken: https://travis-ci.org/jemalloc/jemalloc/jobs/403002462#L970

gnzlbg:
@davidtgoldblatt ping, I think I've addressed all comments, could you give this another look?

davidtgoldblatt:
Could you rebase as appropriate into however it would be merged? It's useful not just for historical purposes, but also reviewing; we can look at a PR as a series of smaller reasonable modifications.

gnzlbg:
@davidtgoldblatt I've rebased it into two commits, the one that changes the `gen_travis.py` to generate smaller output and the one that adds the test that the current travis.yml file is the output of the current script.

davidtgoldblatt:
Looks good, thanks; I'll merge after travis comes back clean.

Thanks for the diff!

PR #1286, opened 2018-07-19
SC: Shrink global data consumption.

davidtgoldblatt:
The amount of global data we consume can force some hot-path variables onto
separate pages, slightly increasing TLB miss rate.  Since we don't need the sc_t
data after each module has done its own initialization, we move it to the stack,
during configure-time.


davidtgoldblatt:
(Going to wait to merge until some canaries have finished)

davidtgoldblatt:
Abandoning in favor of https://github.com/jemalloc/jemalloc/pull/1288, in which we leave this data out entirely after initialization.

PR #1292, opened 2018-07-25
AIX port

2128506:
This is my initial attempt on making jemalloc work on AIX. Main issues to be addressed for AIX port:

1. Compiler selection and command line options - configure was modified to select xlc_r on AIX and sett appropriate flags.
2. Missing ffs* functions - FreeBSD implementation bundled and used when target platform doen't have them.
3. awk on AIX do not understand \ in string constants - so private_symbols.sh was changed to generate one line string constant.
4. index is a macro defined somewhere in system headers - this is the reason for changes in src/ctl.c.
5. Library initialization - on AIX pthreads initialization code calls malloc and free - at this time nothing from pthreads library should be called. I've added appropriate define and forward malloc/free calls to bootstrap_malloc/bootstrap_free until pthreads fully initialized.
6. Simple ifdef to avoid including missing sys/syscall.h
7. Correct way for drop-in malloc/free replacement on AIX is to provide user defined memory subsystem via wrapper object - such wrapper object added. I've made some additional changes to build jemalloc functions with je_ prefix because they are supposed to be called from wrapper functions provided by wrapper archive.

Currently it passes all tests except integration/extent, which fails as following:

=== test/integration/extent ===
test_extent_body:test/integration/extent.c:61: Failed assertion: (did_purge_lazy || did_purge_forced) == (true) --> false != true: Expected purge
test_extent_manual_hook (non-reentrant): fail
test_extent_body:test/integration/extent.c:61: Failed assertion: (did_purge_lazy || did_purge_forced) == (true) --> false != true: Expected purge
test_extent_manual_hook (libc-reentrant): fail
test_extent_body:test/integration/extent.c:61: Failed assertion: (did_purge_lazy || did_purge_forced) == (true) --> false != true: Expected purge
test_extent_manual_hook (arena_new-reentrant): fail
test_extent_body:test/integration/extent.c:61: Failed assertion: (did_purge_lazy || did_purge_forced) == (true) --> false != true: Expected purge
test_extent_auto_hook (non-reentrant): fail
test_extent_body:test/integration/extent.c:61: Failed assertion: (did_purge_lazy || did_purge_forced) == (true) --> false != true: Expected purge
test_extent_auto_hook (libc-reentrant): fail
test_extent_body:test/integration/extent.c:61: Failed assertion: (did_purge_lazy || did_purge_forced) == (true) --> false != true: Expected purge
test_extent_auto_hook (arena_new-reentrant): fail
--- pass: 0/6, skip: 0/6, fail: 6/6 ---

Any hints at what should be checked are very welcome.

Best regards,
Victor Kirhenshtein

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -73,6 +73,7 @@ AWK := @AWK@
 CC_MM = @CC_MM@
 LM := @LM@
 INSTALL = @INSTALL@
+AIX_WRAPPER = @AIX_WRAPPER@
 
 ifeq (macho, $(ABI))
 TEST_LIBRARY_PATH := DYLD_FALLBACK_LIBRARY_PATH="$(objroot)lib"
@@ -132,6 +133,9 @@ ifdef PIC_CFLAGS
 STATIC_LIBS += $(objroot)lib/$(LIBJEMALLOC)_pic.$(A)
 else
 STATIC_LIBS += $(objroot)lib/$(LIBJEMALLOC)_s.$(A)
+endif
+ifdef AIX_WRAPPER
+WRAPPER_LIBS = $(objroot)lib/$(LIBJEMALLOC)_wrapper.$(A)
 endif
 DSOS := $(objroot)lib/$(LIBJEMALLOC).$(SOREV)
 ifneq ($(SOREV),$(SO))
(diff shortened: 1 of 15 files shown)


davidtgoldblatt:
Nothing strikes me as the single obvious option. Maybe we mistakenly autodetect AIX as having a purge option it doesn't, leading us to error out down purging paths and not change the state?

That being said, I think the bigger impediment to this won't be the code; it'll be convincing us that we can reliably set up continuous integration on AIX and avoid breakages without a lot of effort. Any ideas on how we could e.g. easily spin up a debug environment there? Or run it on a cheap VPS host?

2128506:
We likely can provide AIX environment for CI - depending on requirements and how it should be connected. We have AIX development system running 24x7. There is also option to get free AIX access through IBM PartnerWorld - although I didn't try it yet.

You probably right about purge option. From madvise man page: "The madvise subroutine has no functionality and is supported for compatibility only". So it is detected by configure but actually don't do anything. Will try to fix that and check again.

davidtgoldblatt:
Closing out old / unmergeable PRs.

PR #1295, opened 2018-07-31
c does not have array as argument

rkmisra:
c does not have concept of array[] being passing as argument and compiler silently converts array[] argument into pointer.  Hiding the real thing (pointer) behind array[] might not be the right thing to do.  Someone might inadvertently do something  (such as sizeof() to get size of array) assuming that it is an actual array that was being passed in argument, while under the hood, it is just a pointer.

### DIFF
--- a/include/jemalloc/internal/bin.h
+++ b/include/jemalloc/internal/bin.h
@@ -79,7 +79,7 @@ struct bin_s {
 	bin_stats_t	stats;
 };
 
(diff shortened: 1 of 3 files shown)


davidtgoldblatt:
Including the static size can be a useful hint to static checks or readers (who know the expected size, and can ensure that they size their locals appropriately). I'm sympathetic to the confusion argument, but I think on average most jemalloc developers will benefit more from the expected-length hints than the syntax confusion.

Certainly, in using some of the relevant features, I've already found them helpful.

PR #1297, opened 2018-08-01
Add hook microbenchmark.

davidtgoldblatt:
On my machine, it shows around a 40ns overhead to execute a hook.

### DIFF
--- a/test/stress/hookbench.c
+++ b/test/stress/hookbench.c
@@ -1,0 +1,73 @@
+#include "test/jemalloc_test.h"
+
+static void
+noop_alloc_hook(void *extra, hook_alloc_t type, void *result,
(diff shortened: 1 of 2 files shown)


test/stress/hookbench.c
```diff
@@ -0,0 +1,67 @@
...
+static void
+malloc_free_loop(int iters) {
+	for (int i = 0; i < iters; i++) {
+		void *p = mallocx(1, 0);
+		free(p);
+	}
+}
+
+static void
+test_hooked(int iters) {
```
interwq: Add another hook and show the cost trend with more hooks?

davidtgoldblatt: Done

PR #1302, opened 2018-08-10
Avoid potential issues on extent zero-out.

interwq:
When custom extent_hooks or transparent huge pages are in use, the purging
semantics may change, which means we may not get zeroed pages on repopulating.
Fixing the issue by manually memset for such cases.

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -8,6 +8,11 @@
 #include "jemalloc/internal/sz.h"
 #include "jemalloc/internal/ticker.h"
 
+JEMALLOC_ALWAYS_INLINE bool
+arena_has_default_hooks(arena_t *arena) {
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
Nit in the commit message: I don't think this is actually a correctness issue in the "calloc not returning zero'd bytes" sense. Just that morally, we shouldn't be (e.g.) un-hugepagifying custom extent hooks, etc.

interwq:
Relaxed the commit message. Note that when we set opt.thp to "always", it may indeed happen since the madvise call may do nothing depending on the defrag policy.

djwatson:
Why do we need this enabled for thp?

interwq:
Forgot to follow up here. We rely on madvised memory to get zeroed pages on future accesses; however with THP set to always, that might not be the case, e.g. madvise partially on a hugepage may not purge the hugepage depending on the defrag setting.

interwq:
This is still needed -- @davidtgoldblatt @djwatson

PR #1309, opened 2018-08-17
Support multiple sizes for the extent_t struct

tyleretzel:
This enables bitmaps that are larger than 512 regions. Using the `slab_sizes` MALLOC_CONF option allows slabs up to 4096 regions. Normal slabs use a regularly sized `extent_t`, and larger slabs use a larger `extent_t`. Production results maybe indicate a slight (< 1.0% CPU) win on `malloc` and `free` calls, but I don't have enough experience reading those to feel comfortable concluding.

### DIFF
--- a/include/jemalloc/internal/bitmap.h
+++ b/include/jemalloc/internal/bitmap.h
@@ -8,29 +8,13 @@
 typedef unsigned long bitmap_t;
 #define LG_SIZEOF_BITMAP	LG_SIZEOF_LONG
 
-/* Maximum bitmap bit count is 2^LG_BITMAP_MAXBITS. */
-#if LG_SLAB_MAXREGS > LG_CEIL(SC_NSIZES)
-/* Maximum bitmap bit count is determined by maximum regions per slab. */
-#  define LG_BITMAP_MAXBITS	LG_SLAB_MAXREGS
-#else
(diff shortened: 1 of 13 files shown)


include/jemalloc/internal/bitmap.h
```diff
@@ -8,30 +8,14 @@
...
 #define BITMAP_GROUP_NBITS		(1U << LG_BITMAP_GROUP_NBITS)
 #define BITMAP_GROUP_NBITS_MASK		(BITMAP_GROUP_NBITS-1)
 
-/*
- * Do some analysis on how big the bitmap is before we use a tree.  For a brute
- * force linear search, if we would have to call ffs_lu() more than 2^3 times,
- * use a tree instead.
- */
-#if LG_BITMAP_MAXBITS - LG_BITMAP_GROUP_NBITS > 3
-#  define BITMAP_USE_TREE
```
djwatson: Does this mean it always uses a linear search now?

davidtgoldblatt:
This and @tyleretzel's other diff are just so we don't have any zombie code from the internship; the intent isn't for these to get merged, just that we'll have something to pull from if we want to carry on in this direction.

davidtgoldblatt:
Closing old unmergeable PRs.

PR #1311, opened 2018-08-20
run tests on macosx with valgrind

gnzlbg:


gnzlbg:
Running into newer issues, gotta figure out where the bugs are and report them / fix them first.

Issue #1307, opened 2018-08-17
pow2_ceil change breaks tests on s390x


sharkcz:
Seems the recent change in pow2_ceil (commit 4c548a61c89b0472b9952fcc4090eb00c2a88870) breaks the test-suite on s390x (Fedora 28, gcc-8.1.1-5.fc28.s390x).

I get
````
=== test/unit/bit_util ===
test_pow2_ceil_u64 (non-reentrant): pass
test_pow2_ceil_u64 (libc-reentrant): pass
test_pow2_ceil_u64 (arena_new-reentrant): pass
test_pow2_ceil_u32:test/unit/bit_util.c:42: Failed assertion: (pow2_ceil_u32(((uint32_t)1) << i)) == (((uint32_t)1) << i) --> 1 != 2: Unexpected result
test_pow2_ceil_u32 (non-reentrant): fail
test_pow2_ceil_u32:test/unit/bit_util.c:42: Failed assertion: (pow2_ceil_u32(((uint32_t)1) << i)) == (((uint32_t)1) << i) --> 1 != 2: Unexpected result
test_pow2_ceil_u32 (libc-reentrant): fail
test_pow2_ceil_u32:test/unit/bit_util.c:42: Failed assertion: (pow2_ceil_u32(((uint32_t)1) << i)) == (((uint32_t)1) << i) --> 1 != 2: Unexpected result
test_pow2_ceil_u32 (arena_new-reentrant): fail
test_pow2_ceil_zu (non-reentrant): pass
test_pow2_ceil_zu (libc-reentrant): pass
test_pow2_ceil_zu (arena_new-reentrant): pass
test_lg_ceil_floor (non-reentrant): pass
test_lg_ceil_floor (libc-reentrant): pass
test_lg_ceil_floor (arena_new-reentrant): pass
--- pass: 9/12, skip: 0/12, fail: 3/12 ---
````
with the mentioned commit, it passes without it. All other arches (aarch64, ppc64, ppc64le, x86_64) we test are OK.

**Comment by rkmisra** [comment]
Above seems to be failing for i=1 or x=2 as ((((uint32_t)1) << i) == 2)
This means we should have executed below code 

```
size_t msb_on_index = (31 ^ __builtin_clz(2 - 1));
1U << (msb_on_index + 1);
```

[...]

PR #1315, opened 2018-08-26
bit_util.h pow2_ceil_u32() #1307 fallback to older XOR based logic to…

rkmisra:
This is workaround for #1307 s390x optimizer bug https://bugzilla.redhat.com/show_bug.cgi?id=1619354 where we are falling back to older XOR version of code for s390 platform.

### DIFF
--- a/include/jemalloc/internal/bit_util.h
+++ b/include/jemalloc/internal/bit_util.h
@@ -93,7 +93,7 @@ pow2_ceil_u64(uint64_t x) {
 
 BIT_UTIL_INLINE uint32_t
 pow2_ceil_u32(uint32_t x) {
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
Nit: could you reword the commit message so that history (in e.g. github, git log, etc.) looks a little nicer (it'll get cut off as is)? A default commit for this sort of module-local tweak would be:
```
module_name: brief description of contents

Longer description of contents that explains the why, the details, etc.
```

E.g.

```
bit_util: Don't use __builtin_clz on s390x

There's an optimizer bug upstream that results in test failures; reported at
https://bugzilla.redhat.com/show_bug.cgi?id=1619354.  This works around the
failure reported at https://github.com/jemalloc/jemalloc/issues/1307.
```

rkmisra:
Updated commit message as per above recommendation.

PR #1316, opened 2018-08-31
Add sc.c to the MSVC project

rustyx:
This fixes the MSVC build error due to the missing source file sc.c.


davidtgoldblatt:
Thanks!

PR #1337, opened 2018-10-04
check return value of `malloc_read_fd`

jsteemann:
in case `malloc_read_fd` returns a negative error number, the result
would afterwards be casted to an unsigned size_t, and may have
theoretically caused an out-of-bounds memory access in the following
`strncmp` call.


jsteemann:
Ok, I made the requested changes.

jasone:
Oh, I just noticed this is currently two commits.  Can you please squash them?

jsteemann:
Fused the two commits into one.

PR #1338, opened 2018-10-08
Tcache locking refactor

djwatson:
Currently, tcache fill/flush (as well as small alloc/dalloc on the arena) may potentially drop the bin lock for slab_alloc and slab_dalloc.  This patchset refactors the logic so that the slab calls happen in the same function as the bin locking/unlocking.  The main purpose is to be able to use flat combining without having to keep track of stack state, but as a side benefit, we are able to reduce the locking, especially for slab_dalloc calls, where nothing happens after the call.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -58,8 +58,6 @@ static void arena_decay_to_limit(tsdn_t *tsdn, arena_t *arena,
     size_t npages_decay_max, bool is_background_thread);
 static bool arena_decay_dirty(tsdn_t *tsdn, arena_t *arena,
     bool is_background_thread, bool all);
-static void arena_dalloc_bin_slab(tsdn_t *tsdn, arena_t *arena, extent_t *slab,
-    bin_t *bin);
 static void arena_bin_lower_slab(tsdn_t *tsdn, arena_t *arena, extent_t *slab,
     bin_t *bin);
 
@@ -989,7 +987,7 @@ arena_decay(tsdn_t *tsdn, arena_t *arena, bool is_background_thread, bool all) {
 	arena_decay_muzzy(tsdn, arena, is_background_thread, all);
 }
 
-static void
+void
 arena_slab_dalloc(tsdn_t *tsdn, arena_t *arena, extent_t *slab) {
 	arena_nactive_sub(arena, extent_size_get(slab) >> LG_PAGE);
 
@@ -1323,8 +1321,9 @@ arena_bin_malloc_hard(tsdn_t *tsdn, arena_t *arena, bin_t *bin,
 						bin->stats.curslabs--;
 					}
 
-					arena_dalloc_bin_slab(tsdn, arena, slab,
-					    bin);
+					malloc_mutex_unlock(tsdn, &bin->lock);
+					arena_slab_dalloc(tsdn, arena, slab);
+					malloc_mutex_lock(tsdn, &bin->lock);
 				} else {
 					arena_bin_lower_slab(tsdn, arena, slab,
 					    bin);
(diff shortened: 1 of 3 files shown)


src/arena.c
```diff
@@ -1324,6 +1310,23 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
 		} else {
 			ptr = arena_bin_malloc_hard(tsdn, arena, bin, binind);
 		}
+		if (ptr == NULL) {
```
interwq: Move this to right after calling `arena_bin_malloc_hard`?

src/arena.c
```diff
@@ -1306,6 +1290,8 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
     cache_bin_t *tbin, szind_t binind, uint64_t prof_accumbytes) {
 	unsigned i, nfill;
 	bin_t *bin;
+	extent_t *fresh_slab = NULL;
+	bool dalloc = false;
```
interwq: nit: avoid the initial value here.

src/arena.c
```diff
@@ -1350,6 +1353,9 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
 		tbin->tstats.nrequests = 0;
 	}
 	malloc_mutex_unlock(tsdn, &bin->lock);
+	if (fresh_slab) {
```
interwq: Let's use `if (dalloc)` directly, and get rid of the branch @ L1326?

src/arena.c
```diff
@@ -1385,9 +1392,23 @@ arena_malloc_small(tsdn_t *tsdn, arena_t *arena, szind_t binind, bool zero) {
 	} else {
 		ret = arena_bin_malloc_hard(tsdn, arena, bin, binind);
 	}
+	if (ret == NULL) {
```
interwq: similarly, move this to right after bin_malloc_hard?

src/arena.c
```diff
@@ -1292,67 +1292,87 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
-		bin->stats.curregs += i;
-		bin->stats.nfills++;
-		tbin->tstats.nrequests = 0;
+		malloc_mutex_unlock(tsdn, &bin->lock);
+		if (done) {
+			break;
+		}
+		assert(!fresh_slab);
+		fresh_slab = arena_slab_alloc(tsdn, arena, binind, bin_info);
+                alloced = true;
```
interwq: alignment seems off.

src/arena.c
```diff
@@ -1292,67 +1292,87 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
+				ptr = arena_bin_malloc_hard(
+					tsdn, arena, bin, binind);
 			}
-			dalloc = false;
-
-			ptr = arena_bin_malloc_with_fresh_slab(tsdn, arena, bin, binind, &fresh_slab, &dalloc);
-			if (!dalloc) {
-				fresh_slab = NULL;
+			if (ptr == NULL) {
+				if (!fresh_slab && !alloced) {
```
interwq: Is `!alloced` alone sufficient here?

src/arena.c
djwatson: I don't think we can do this, since dalloc is not set after fresh_slab, we may be using slabcur again instead, yet we still need to free fresh_slab since it is unused.

interwq:
Replaced by #1763.

PR #1340, opened 2018-10-09
Remove malloc_init() off the fastpath

djwatson:
Move malloc_init() off of the fast path in imalloc(). If tsd can't allocate, then it is safe to check the tsd_state (which will be uninitialized the first time), and then only do the malloc_init() check on the slowpath.   

This appears to pass our internal testing, but I'm not sure this is enough for freebsd libc.

### DIFF
--- a/src/tsd.c
+++ b/src/tsd.c
@@ -280,11 +280,13 @@ tsd_fetch_slow(tsd_t *tsd, bool minimal) {
 		tsd_slow_update(tsd);
 	} else if (tsd_state_get(tsd) == tsd_state_uninitialized) {
 		if (!minimal) {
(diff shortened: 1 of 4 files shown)


src/tsd.c
```diff
@@ -280,6 +280,10 @@ tsd_fetch_slow(tsd_t *tsd, bool minimal) {
 		tsd_slow_update(tsd);
 	} else if (tsd_state_get(tsd) == tsd_state_uninitialized) {
 		if (!minimal) {
+			if (!tsd_booted) {
```
jasone: Given that this is the only early return in the function, IMO this would read better if the following code (lines 287-291) were instead wrapped in `if (tsd_booted) { }`.

PR #1341, opened 2018-10-09
Zero size class

djwatson:
Roll the size-0 check in to the size class lookup table (or slower compute path).  This removes a branch from the fastpath if we are using the table (and has the same number of branches if using _compute()).

### DIFF
--- a/include/jemalloc/internal/sz.h
+++ b/include/jemalloc/internal/sz.h
@@ -122,6 +122,10 @@ sz_size2index_compute(size_t size) {
 	if (unlikely(size > SC_LARGE_MAXCLASS)) {
 		return SC_NSIZES;
 	}
+
+	if (size == 0) {
+		return 0;
+	}
 #if (SC_NTINY != 0)
 	if (size <= (ZU(1) << SC_LG_TINY_MAXCLASS)) {
 		szind_t lg_tmin = SC_LG_TINY_MAXCLASS - SC_NTINY + 1;
@@ -150,14 +154,14 @@ sz_size2index_compute(size_t size) {
 JEMALLOC_ALWAYS_INLINE szind_t
 sz_size2index_lookup(size_t size) {
 	assert(size <= SC_LOOKUP_MAXCLASS);
-	szind_t ret = (sz_size2index_tab[(size-1) >> SC_LG_TINY_MIN]);
+	szind_t ret = (sz_size2index_tab[(size + (ZU(1) << SC_LG_TINY_MIN) - 1)
+					 >> SC_LG_TINY_MIN]);
 	assert(ret == sz_size2index_compute(size));
 	return ret;
 }
 
 JEMALLOC_ALWAYS_INLINE szind_t
 sz_size2index(size_t size) {
-	assert(size > 0);
 	if (likely(size <= SC_LOOKUP_MAXCLASS)) {
 		return sz_size2index_lookup(size);
 	}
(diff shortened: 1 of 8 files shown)


test/integration/aligned_alloc.c
```diff
@@ -138,10 +138,26 @@ TEST_BEGIN(test_alignment_and_size) {
 }
 TEST_END
 
+TEST_BEGIN(test_zero_alloc) {
+	void* res = aligned_alloc(8, 0);
```
jasone: `void* res` --> `void *res`

test/integration/aligned_alloc.c
```diff
@@ -138,10 +138,26 @@ TEST_BEGIN(test_alignment_and_size) {
...
+	assert(res);
+	size_t usable = malloc_usable_size(res);
+	assert(usable > 0);
+	free(res);
+
+	res = malloc(0);
```
jasone: Shouldn't this test go in `test/integration/malloc.c`?

include/jemalloc/internal/sz.h
```diff
@@ -150,14 +154,13 @@ sz_size2index_compute(size_t size) {
 JEMALLOC_ALWAYS_INLINE szind_t
 sz_size2index_lookup(size_t size) {
 	assert(size <= SC_LOOKUP_MAXCLASS);
-	szind_t ret = (sz_size2index_tab[(size-1) >> SC_LG_TINY_MIN]);
+	szind_t ret = (sz_size2index_tab[(size+7) >> SC_LG_TINY_MIN]);
```
jasone: How is it generally correct to change from `size-1` to `size+7`?  My memory of how this all works is a bit fuzzy, but on the surface this sure seems like it's assuming the word size is 8.

src/jemalloc.c
```diff
@@ -1713,8 +1713,7 @@ typedef struct static_opts_s static_opts_t;
 struct static_opts_s {
 	/* Whether or not allocation size may overflow. */
 	bool may_overflow;
-	/* Whether or not allocations of size 0 should be treated as size 1. */
```
jasone: Wait, this pull request adds implicit support for `mallocx(0, 0)`?  If so, I don't think we should go down this path.

include/jemalloc/internal/sz.h
djwatson: Ah, right, this should be size+SC_LG_TINY_MIN?

src/sz.c
```diff
@@ -37,18 +37,18 @@ sz_boot_index2size_tab(const sc_data_t *sc_data) {
...
-	size_t dst_max = (SC_LOOKUP_MAXCLASS >> SC_LG_TINY_MIN);
+	size_t dst_max = (SC_LOOKUP_MAXCLASS >> SC_LG_TINY_MIN) + 1;
 	size_t dst_ind = 0;
 	for (unsigned sc_ind = 0; sc_ind < SC_NSIZES && dst_ind < dst_max;
 	    sc_ind++) {
 		const sc_t *sc = &sc_data->sc[sc_ind];
 		size_t sz = (ZU(1) << sc->lg_base)
 		    + (ZU(sc->ndelta) << sc->lg_delta);
-		size_t max_ind = ((sz - 1) >> SC_LG_TINY_MIN);
+		size_t max_ind = ((sz + 7) >> SC_LG_TINY_MIN);
```
interwq: This should also be updated with `(ZU(1) << SC_LG_TINY_MIN) - 1`

djwatson: Indeed.  Fixed, thanks.

interwq:
LGTM!

PR #1342, opened 2018-10-10
Move bytes_until_sample to tsd from tdata

djwatson:
In the fastpath case (where we aren't going to sample the object), only bytes_until_sample is needed from tdata.  By moving it to tsd we can avoid loading tdata at all.

Additionally, if we change bytes_until_sample to be signed, we can generate better code on x86.


src/prof.c
```diff
@@ -1168,9 +1168,11 @@ prof_sample_threshold_update(prof_tdata_t *tdata) {
 	 */
 	r = prng_lg_range_u64(&tdata->prng_state, 53);
 	u = (double)r * (1.0/9007199254740992.0L);
-	tdata->bytes_until_sample = (uint64_t)(log(u) /
+	uint64_t bytes_until_sample = (uint64_t)(log(u) /
```
jasone: Did we decide to allow mixed declarations and statements in our coding conventions?  (I think we did...)  If so, please refactor the declarations for `r` and `u` to make the function consistent.

src/sc.c
```diff
@@ -244,6 +244,15 @@ size_classes(
...
 	assert(sc_data->large_maxclass == SC_LARGE_MAXCLASS);
+
+	/* 
+	 * In the allocation fastpath, we want to assume that we can
+	 * unconditionally subtract the requested allocation size from
+	 * a ssize_t, and detect passing through 0 correctly.  This
+	 * results in optimal generated code.  For this to work, the
+	 * maximum allocation size must be less than SSIZE_MAX.
+	 */
+	assert(SC_LARGE_MAXCLASS < SSIZE_MAX);
```
jasone: `SC_LARGE_MAXCLASS == PTRDIFF_MAX <= SSIZE_MAX`, right?  In practice, will this assertion ever fail on 32-bit platforms?

djwatson: Correct.  Right, this should never fail on 32-bit platforms.  It should really be a static compile-time assert, but it relies on some define math in SC_LARGE_MAXCLASS and I couldn't convince the compiler it was compile-time only.

PR #1344, opened 2018-10-11
Remove a branch from cache_bin_alloc_easy

djwatson:
Combine the branches for checking for an empty cache_bin, and
checking for the low watermark.


include/jemalloc/internal/cache_bin.h
```diff
@@ -101,12 +106,7 @@ cache_bin_alloc_easy(cache_bin_t *bin, bool *success) {
 	 * cacheline).
 	 */
 	*success = true;
-	ret = *(bin->avail - bin->ncached);
-	bin->ncached--;
-
-	if (unlikely(bin->ncached < bin->low_water)) {
-		bin->low_water = bin->ncached;
-	}
+	ret = *(bin->avail - (bin->ncached + 1));
```
interwq: Looks great!

We probably should also consider moving the load to the beginning of this function, so that we can avoid the "+1" plus make this load happen sooner. We would need to make sure avail always points to somewhere valid though.

include/jemalloc/internal/cache_bin.h
```diff
@@ -88,11 +88,16 @@ JEMALLOC_ALWAYS_INLINE void *
 cache_bin_alloc_easy(cache_bin_t *bin, bool *success) {
 	void *ret;
 
-	if (unlikely(bin->ncached == 0)) {
-		bin->low_water = -1;
-		*success = false;
-		return NULL;
+	bin->ncached--;
+	if (unlikely(bin->ncached <= bin->low_water)) {
```
interwq: nit: a short note about avoid checking ```ncached == 0``` separately?

include/jemalloc/internal/cache_bin.h
djwatson: Right I think cache_bin_t would need to be one larger to avoid a potential invalid load, will leave this for another diff.

PR #1345, opened 2018-10-15
malloc: Add a fastpath

djwatson:
This diff adds a fastpath that assumes size <= SC_LOOKUP_MAXCLASS, and that we hit tcache.  If either of these is false, we fall back to the previous codepath (renamed 'malloc_default').

Crucially, we only tail call malloc_default, and with the same kind and number of arguments, so that both clang and gcc tail-calling will kick in - therefore malloc() gets treated as a leaf function, and there are *no* caller-saved registers.   Previously malloc() contained 5 caller saved registers on x64, resulting in at least 10 extra memory-movement instructions.
    
In microbenchmarks this results in up to ~10% improvement in malloc() fastpath.  In real programs, this is a ~1% CPU and latency improvement overall.

Fastpath now looks like this:
https://gist.github.com/djwatson/f2d8d9993f5736c74c280941bb67c0c9

On top of #1341, or requires a zero-size check.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2151,15 +2151,9 @@ imalloc(static_opts_t *sopts, dynamic_opts_t *dopts) {
 		return imalloc_body(sopts, dopts, tsd);
 	}
 }
-/******************************************************************************/
-/*
- * Begin malloc(3)-compatible functions.
- */
 
-JEMALLOC_EXPORT JEMALLOC_ALLOCATOR JEMALLOC_RESTRICT_RETURN
-void JEMALLOC_NOTHROW *
-JEMALLOC_ATTR(malloc) JEMALLOC_ALLOC_SIZE(1)
-je_malloc(size_t size) {
+void *
+malloc_default(size_t size) {
 	void *ret;
 	static_opts_t sopts;
 	dynamic_opts_t dopts;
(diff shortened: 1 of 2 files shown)


interwq:
It looks there are fork related failures on Mac?

djwatson:
> It looks there are fork related failures on Mac?

fixed with an extra if(tsd) check.

src/jemalloc.c
```diff
@@ -2193,6 +2187,90 @@ je_malloc(size_t size) {
...
+JEMALLOC_ATTR(malloc) JEMALLOC_ALLOC_SIZE(1)
+je_malloc(size_t size) {
+	LOG("core.malloc.entry", "size: %zu", size);
+
+	if (tsd_get_allocates() && unlikely(!malloc_initialized())) {
+		return malloc_default(size);
+	}
+
+	tsd_t *tsd = tsd_get(false);
+	if (unlikely(!tsd || !tsd_fast(tsd) || (size > SC_LOOKUP_MAXCLASS))) {
```
interwq: This assumes SC_LOOKUP_MAXCLASS is always small size class, which likely will stay true always. Up to you if you want to make it explicit with an assert or comment.

djwatson: added a couple asserts.

gnzlbg:
Is it possible to apply any of these new optimizations to `mallocx`/`smallocx` ?

djwatson:
@gnzlbg - Yea probably.  In the short term, I have a free() + sdallocx() fastpath coming.

PR #1352, opened 2018-10-23
Restrict lwsync to powerpc64 only

chmeeedalf:
Nearly all 32-bit powerpc hardware treats lwsync as sync, and some cores
(Freescale e500) trap lwsync as an illegal instruction, which then gets
emulated in the kernel.  To avoid unnecessary traps on the e500, use
sync on all 32-bit powerpc.  This pessimizes 32-bit software running on
64-bit hardware, but those numbers should be slim.

### DIFF
--- a/include/jemalloc/internal/atomic_gcc_sync.h
+++ b/include/jemalloc/internal/atomic_gcc_sync.h
@@ -27,8 +27,10 @@ atomic_fence(atomic_memory_order_t mo) {
 	asm volatile("" ::: "memory");
 #  if defined(__i386__) || defined(__x86_64__)
 	/* This is implicit on x86. */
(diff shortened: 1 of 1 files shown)


chmeeedalf:
This change (paired with MALLOC_PRODUCTION=) resulted in a ~20% speedup in compiles on a P1022 system (devel/llvm60 build goes from ~40 hours down to ~32 hours), as it no longer has to go through the trap-and-emulate route, saving a lot of unnecessary work.

interwq:
This sounds reasonable to me. I'll let @davidtgoldblatt to double check since he may have more context here.

davidtgoldblatt:
Makes sense to me, thanks for the diff!

PR #1354, opened 2018-10-24
Make use of pthread_set_name_np(3) on FreeBSD.

trasz:


interwq:
Thanks for the patch!

PR #1356, opened 2018-10-26
Optimize large deallocation.

interwq:
We eagerly coalesce large buffers when deallocating, however the previous logic
    around this introduced extra lock overhead -- when coalescing we always lock the
    neighbors even if they are active / small, while for active extents nothing can
    be done; and the intent was to coalesce only large buffers to reduce
    fragmentation.
    
    This commit checks if the neighbor extents are potentially active before
    locking, and avoids coalescing small extents if not intended to.  This speeds up
    large_dalloc by ~20%.  It also fixes some undesired behavior: we could stop
    coalescing because a small buffer was merged, while a large neighbor was ignored
    on the other side.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -134,13 +134,16 @@ typedef enum {
 
 static lock_result_t
 extent_rtree_leaf_elm_try_lock(tsdn_t *tsdn, rtree_leaf_elm_t *elm,
-    extent_t **result) {
+    extent_t **result, bool inactive_only) {
 	extent_t *extent1 = rtree_leaf_elm_extent_read(tsdn, &extents_rtree,
 	    elm, true);
 
-	if (extent1 == NULL) {
+	/* Slab implies active extents and should be skipped. */
+	if (extent1 == NULL || (inactive_only && rtree_leaf_elm_slab_read(tsdn,
+	    &extents_rtree, elm, true))) {
 		return lock_result_no_extent;
 	}
+
 	/*
 	 * It's possible that the extent changed out from under us, and with it
 	 * the leaf->extent mapping.  We have to recheck while holding the lock.
@@ -163,7 +166,8 @@ extent_rtree_leaf_elm_try_lock(tsdn_t *tsdn, rtree_leaf_elm_t *elm,
  * address, and NULL otherwise.
  */
 static extent_t *
-extent_lock_from_addr(tsdn_t *tsdn, rtree_ctx_t *rtree_ctx, void *addr) {
+extent_lock_from_addr(tsdn_t *tsdn, rtree_ctx_t *rtree_ctx, void *addr,
+    bool inactive_only) {
 	extent_t *ret = NULL;
 	rtree_leaf_elm_t *elm = rtree_leaf_elm_lookup(tsdn, &extents_rtree,
 	    rtree_ctx, (uintptr_t)addr, false, false);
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
I think the reasoning here is a little confused; the reason to avoid trying to merge with slab allocations is that slabs are definitionally active and therefore not subject to merging, right?

Do we need a fence between the rtree leaf update and the check of the neighbor? It seems like it might introduce a store (my extent is now free) -> load (is my neighbor extent free?) race, so that we miss updates we ought to coalesce.

interwq:
> I think the reasoning here is a little confused; the reason to avoid trying to merge with slab allocations is that slabs are definitionally active and therefore not subject to merging, right?

Good point. Let me update that part.

> Do we need a fence between the rtree leaf update and the check of the neighbor? It seems like it might introduce a store (my extent is now free) -> load (is my neighbor extent free?) race, so that we miss updates we ought to coalesce.

Adding a fence sound good. I'm honestly not too worried here, since eagerly merging large extents can be seen as a hint / optimization instead of a requirement (the old logic was buggy and we were missing merging opportunities). With or w/o fence both sound fine to me. What's your preference?

davidtgoldblatt:
> Adding a fence sound good. I'm honestly not too worried here, since eagerly merging large extents can be seen as a hint / optimization instead of a requirement (the old logic was buggy and we were missing merging opportunities). With or w/o fence both sound fine to me. What's your preference?

That's worrying :(. Let's try with a fence, and we can always back off if we see it in profiles?

src/extent.c
```diff
@@ -134,13 +134,17 @@ typedef enum {
 
 static lock_result_t
 extent_rtree_leaf_elm_try_lock(tsdn_t *tsdn, rtree_leaf_elm_t *elm,
-    extent_t **result) {
-	extent_t *extent1 = rtree_leaf_elm_extent_read(tsdn, &extents_rtree,
-	    elm, true);
+    extent_t **result, bool large_only) {
+	uintptr_t bits = rtree_leaf_elm_bits_read(tsdn, &extents_rtree, elm,
+	    false);
```
interwq: changed to a non dependent read to enforce the barrier.

interwq:
Updated the comments and some logic. Will rerun the prod test to verify everything still ok.

src/extent.c
davidtgoldblatt: I don't think that's the right fence; I think we'd need a seq_cst fence after our modification but before the read of the neighbor (dependent vs. non-dependent is just for the lookups being acquire-y or not).

interwq: Ah do you mean also adding a fence on the rtree update path? In that case I would suggest doing that part in a separate change, since this diff doesn’t really affect that part, and the fence overhead on a fast-ish path will need its own benchmark run to be better measured.

davidtgoldblatt: No, we just need it on the merge path (and maybe not even then, it depends on how exactly).

davidtgoldblatt: Actually, thinking about it even more, this only matters if the two extents might be about to be coalesced, which would mean that they're in the same state, which means that they're protected by the same extents_t lock, right? So even if the race occurs, all it means is that we do a little extra work of following the extent, and so allowing it by remaining barrier-free should be OK.

src/extent.c
```diff
@@ -1567,9 +1573,10 @@ extent_coalesce(tsdn_t *tsdn, arena_t *arena, extent_hooks_t **r_extent_hooks,
 }
 
 static extent_t *
-extent_try_coalesce(tsdn_t *tsdn, arena_t *arena,
+extent_try_coalesce_impl(tsdn_t *tsdn, arena_t *arena,
     extent_hooks_t **r_extent_hooks, rtree_ctx_t *rtree_ctx, extents_t *extents,
-    extent_t *extent, bool *coalesced, bool growing_retained) {
+    extent_t *extent, bool *coalesced, bool growing_retained,
+    bool inactive_only) {
```
djwatson: maybe add a comment similar to the commit message about when to pass inactive_only

interwq:
Tried a bunch of different approaches and the current implementation was the best one. This is still only for large dealloc; for small size classes since they don't really go down this route (because of delay_coalesce), it doesn't seem to matter for them.

I'll update the comments as Dave suggested. Re: the fence comments, agree with David that even if a race occurs it should still be fine.

PR #1359, opened 2018-10-30
Optimize arena_tcache_fill_small

djwatson:
This patchset optimizes arena_tcache_fill_small(), for the cases where more than one item is requested from the same slab.  This should be fairly common.  The case of only a single item fill should be fairly neutral.

The batch fill portion is refactored to a separate function, and the linear scan of the bitmap_t's is merged with the fill itself (it's unclear to me if this might better fit in bitmap.h instead, since it needs parts of both arena and bitmap logic).  We use popcount() and bmi (bit manipulation instruction set) when available.

In a microbenchmark, this diff changes the perf overhead of arena_tcache_fill_small from a high of 12% down to 3-4%.   In real world canaries, this shows up as a 1% latency improvement.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -273,19 +273,46 @@ arena_slab_reg_alloc_batch(extent_t *slab, const bin_info_t *bin_info,
 			   unsigned cnt, void** ptrs) {
 	arena_slab_data_t *slab_data = extent_slab_data_get(slab);
 
-	assert(extent_nfree_get(slab) > 0);
+	assert(extent_nfree_get(slab) >= cnt);
 	assert(!bitmap_full(slab_data->bitmap, &bin_info->bitmap_info));
 
-	size_t regind = 0;
+#if (! defined JEMALLOC_INTERNAL_POPCOUNTL) || (defined BITMAP_USE_TREE)
 	for (unsigned i = 0; i < cnt; i++) {
-		void *ret;
-
-		regind = bitmap_sfu(slab_data->bitmap, &bin_info->bitmap_info);
-		ret = (void *)((uintptr_t)extent_addr_get(slab) +
+		size_t regind = bitmap_sfu(slab_data->bitmap,
+					   &bin_info->bitmap_info);
+		*(ptrs + i) = (void *)((uintptr_t)extent_addr_get(slab) +
 		    (uintptr_t)(bin_info->reg_size * regind));
-
-		*(ptrs + i) = ret;
 	}
+#else
+	unsigned group = 0;
+	bitmap_t g = slab_data->bitmap[group];
+	unsigned i = 0;
+	while (i < cnt) {
+		while (g == 0) {
+			g = slab_data->bitmap[++group];
+		}
+		size_t shift = group << LG_BITMAP_GROUP_NBITS;
+		size_t pop = popcount_lu(g);
+		if (pop > (cnt - i)) {
+			pop = cnt - i;
+		}
+
+		/*
+		 * Load from memory locations only once, outside the
+		 * hot loop below.
+		 */
+		uintptr_t base = (uintptr_t)extent_addr_get(slab);
+		uintptr_t regsize = (uintptr_t)bin_info->reg_size;
+		while (pop--) {
+			size_t bit = cfs_lu(&g);
+			size_t regind = shift + bit;
+			*(ptrs + i) = (void *)(base + regsize * regind);
+
+			i++;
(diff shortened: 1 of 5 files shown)


src/arena.c
```diff
@@ -1297,32 +1345,43 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
 	bin = &arena->bins[binind];
 	malloc_mutex_lock(tsdn, &bin->lock);
 	for (i = 0, nfill = (tcache_bin_info[binind].ncached_max >>
-	    tcache->lg_fill_div[binind]); i < nfill; i++) {
+	    tcache->lg_fill_div[binind]); i < nfill; i += cnt) {
 		extent_t *slab;
-		void *ptr;
+		cnt = 1;
```
interwq: Move this line to the `else` branch @L1360.

src/arena.c
```diff
@@ -1297,32 +1345,43 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
+			cnt = tofill < extent_nfree_get(slab) ?
+				tofill : extent_nfree_get(slab);
+			arena_slab_reg_alloc_batch(
+			   slab, &bin_infos[binind], cnt,
+			   tbin->avail - nfill + i);
 		} else {
-			ptr = arena_bin_malloc_hard(tsdn, arena, bin, binind);
-		}
-		if (ptr == NULL) {
+			void *ptr = arena_bin_malloc_hard(tsdn, arena, bin,
```
interwq: Another small optimization we may try: split `arena_bin_malloc_hard` into `bin_slab_refill` and calling `slab_reg_alloc`. In this case we'd only need the refill part and continue the loop with cnt = 0. OOM handling might be a bit tricky (maybe fall back to bin_malloc_hard if re-fill didn't get a new nonfull slab).

src/arena.c
```diff
@@ -1297,32 +1345,43 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
+				break;
 			}
-			break;
+			/* Insert such that low regions get used first. */
+			*(tbin->avail - nfill + i) = ptr;
 		}
-		if (config_fill && unlikely(opt_junk_alloc)) {
-			arena_alloc_junk_small(ptr, &bin_infos[binind], true);
+		for (unsigned j = 0; j < cnt; j++) {
+			if (config_fill && unlikely(opt_junk_alloc)) {
```
interwq: nit: move this branch to cover the entire loop.

src/arena.c
djwatson: I split the function in https://github.com/jemalloc/jemalloc/pull/1338, I'll rebase that one and then test again

interwq: Ah I see. If you want to land this one as is, that's totally fine too.

src/arena.c
```diff
@@ -1297,32 +1345,43 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
+			arena_slab_reg_alloc_batch(
+			   slab, &bin_infos[binind], cnt,
+			   tbin->avail - nfill + i);
 		} else {
-			ptr = arena_bin_malloc_hard(tsdn, arena, bin, binind);
-		}
-		if (ptr == NULL) {
+			cnt = 1;
+			void *ptr = arena_bin_malloc_hard(tsdn, arena, bin,
+							  binind);
```
interwq: nit: indent is off.

src/arena.c
```diff
@@ -1297,32 +1345,43 @@ arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
-			break;
+			/* Insert such that low regions get used first. */
+			*(tbin->avail - nfill + i) = ptr;
 		}
 		if (config_fill && unlikely(opt_junk_alloc)) {
-			arena_alloc_junk_small(ptr, &bin_infos[binind], true);
+			for (unsigned j = 0; j < cnt; j++) {
+				void* ptr = *(tbin->avail - nfill + i + j);
+				arena_alloc_junk_small(ptr, &bin_infos[binind],
+						       true);
```
interwq: nit: indent too.

src/arena.c
djwatson: Right, I'll land this one and test this optimization separately

PR #1360, opened 2018-10-31
implement malloc_getcpu for windows

santagada:
The only note is that this doesn't support multiple processor groups, but so does the ncpu on windows (will do another PR for it)

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_inlines_a.h
+++ b/include/jemalloc/internal/jemalloc_internal_inlines_a.h
@@ -10,7 +10,9 @@
 JEMALLOC_ALWAYS_INLINE malloc_cpuid_t
 malloc_getcpu(void) {
 	assert(have_percpu_arena);
-#if defined(JEMALLOC_HAVE_SCHED_GETCPU)
+#if defined(_WIN32)
+	return GetCurrentProcessorNumber();
+#elif defined(JEMALLOC_HAVE_SCHED_GETCPU)
 	return (malloc_cpuid_t)sched_getcpu();
 #else
 	not_reached();
--- a/include/jemalloc/internal/jemalloc_preamble.h.in
+++ b/include/jemalloc/internal/jemalloc_preamble.h.in
@@ -161,7 +161,7 @@ static const bool config_log =
     false
 #endif
     ;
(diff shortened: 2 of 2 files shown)


include/jemalloc/internal/jemalloc_internal_inlines_a.h
```diff
@@ -10,7 +10,9 @@
 JEMALLOC_ALWAYS_INLINE malloc_cpuid_t
 malloc_getcpu(void) {
 	assert(have_percpu_arena);
-#if defined(JEMALLOC_HAVE_SCHED_GETCPU)
+#if defined(_WIN32)
+	return GetCurrentProcessorNumber();
```
interwq: It looks this function has some limitation with >64 cores (https://msdn.microsoft.com/en-us/windows/desktop/ms683181).

Also can you please add some config time checks of the function in https://github.com/jemalloc/jemalloc/blob/dev/configure.ac? It looks different headers are needed on some versions.

santagada: Yes, this is what "doesn't support multiple processor groups" meant. But so does the ncpu on windows, that's why I wanted to have both with the same support, then move both to support multiple processor groups.

interwq: Yeah if you could add a follow up diff to support the processor group that would be great.

Re: the header, I only check the MSDN page and it mentioned WinBase.h and Processthreadsapi.h. I don't really see them in the code base, so may they are covered in some other ways?

santagada: When I have the time I will propose a processor group aware version of ncpus and this.

Then I could come up with a numa aware arena creation, how does linux/freebsd deal with numa aware memory allocation?

interwq: jemalloc currently is not numa aware, so even with percpu arena they may still use memory from a remote numa node.

We are trying to cut down number of arenas (by making them scale better). It might be feasible to have one arena per numa node by the next release.

interwq:
Thanks for the patch! (and sorry about the delay in merging)

PR #1363, opened 2018-11-05
Tick decay on ipallocztm / idalloctm.

### LINKED ISSUE #1347: spinning on an atomic leaks memory when using jemalloc
When keeping spawning new threads doing this spin and exits, process RSS keeps going up and finally gets OOM killed when linking with jemalloc. libc malloc doesn't have this problem.  
```
#include <atomic>
#include <thread>
#include <deque>
#include <chrono>

inline int do_spin(std::atomic<unsigned> & generation, unsigned int val)
{
    for (;;)
        if (generation.load(std::memory_order_relaxed) != val)
            return 0;
}

int main(int argc, char *argv[]) {
    std::atomic<unsigned> generation = 0;
    for(;;) {
        std::deque<std::thread> q;
        for (auto i = 1u; i < 40; ++i) {
            q.push_back(std::thread([&]() {
                do_spin(generation, generation);
            }));
        }
        using namespace std::chrono_literals;
        std::this_thread::sleep_for(100ms);
        generation ++;
        for (auto & t : q) {
            t.join();
        }
    }
}
```

interwq:
Tcache init / cleanup should trigger decay ticks as well.

This resolves #1347.

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_inlines_c.h
+++ b/include/jemalloc/internal/jemalloc_internal_inlines_c.h
@@ -80,6 +80,7 @@ ipallocztm(tsdn_t *tsdn, size_t usize, size_t alignment, bool zero,
 	if (config_stats && is_internal && likely(ret != NULL)) {
 		arena_internal_add(iaalloc(tsdn, ret), isalloc(tsdn, ret));
 	}
+	arena_decay_tick(tsdn, arena);
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


interwq:
Nope this isn't it.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -116,6 +117,7 @@ idalloctm(tsdn_t *tsdn, void *ptr, tcache_t *tcache, alloc_ctx_t *alloc_ctx,
 		assert(tcache == NULL);
 	}
 	arena_dalloc(tsdn, ptr, tcache, alloc_ctx, slow_path);
+	arena_decay_tick(tsdn, arena);
```
djwatson: arena is not a variable

interwq:
Still investigating. Will reopen soon.

PR #1365, opened 2018-11-08
Add a fastpath for free and sdallocx

djwatson:
Add unsized and sized deallocation fastpaths.  Similar to the malloc() fastpath, this removes all frame manipulation for the majority of free() calls.  The fastpath must hit the rtree cache, or be a sized dalloc.  Profiled objects go to the slowpath.

### DIFF
--- a/include/jemalloc/internal/rtree.h
+++ b/include/jemalloc/internal/rtree.h
@@ -452,6 +452,42 @@ rtree_extent_szind_read(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
 	return false;
 }
 
+/*
+ * Try to read szind_slab from the L1 cache.  Returns true on a hit,
+ * and fills in r_szind and r_slab.  Otherwise returns false.
+ *
+ * Key is allowed to be NULL in order to save an extra branch on the
+ * fastpath.  returns false in this case.
+ */
+JEMALLOC_ALWAYS_INLINE bool
+rtree_szind_slab_read_fast(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
+			    uintptr_t key, szind_t *r_szind, bool *r_slab) {
+	rtree_leaf_elm_t *elm;
+
+	size_t slot = rtree_cache_direct_map(key);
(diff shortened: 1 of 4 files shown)


include/jemalloc/internal/rtree.h
```diff
@@ -452,6 +452,42 @@ rtree_extent_szind_read(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
...
+/*
+ * Try to read szind_slab from the L1 cache.  Returns true on a hit,
+ * and fills in r_szind and r_slab.  Otherwise returns false.
+ *
+ * Key is allowed to be NULL in order to save an extra branch on the
+ * fastpath.  returns false in this case.
+ */
+JEMALLOC_ALWAYS_INLINE bool
+rtree_szind_slab_read_fast(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
+				uintptr_t key, szind_t *r_szind, bool *r_slab) {
```
interwq: nit: spacing is off.

src/jemalloc.c
```diff
@@ -2685,6 +2685,67 @@ je_free(void *ptr) {
...
+bool free_fastpath(void *ptr, size_t size, bool size_hint) {
+	tsd_t *tsd = tsd_get(false);
+	if (unlikely(!tsd || !tsd_fast(tsd))) {
+		return false;
+	}
+
+	tcache_t *tcache = tsd_tcachep_get(tsd);
+
+	alloc_ctx_t alloc_ctx;
+	if (!size_hint || config_cache_oblivious) {
```
interwq: Why is cache_oblivious involved here? I don't really follow.

djwatson: I'll add a comment

Similar to isfree() - if cache_oblivious is *off*, we can check for sampled objects by checking Page alignment.  If we are cache_oblivious, the start address is randomized, and we can't do that.

                /*                                                                                                                                                                                                                       
                 * When cache_oblivious is disabled and ptr is not page aligned,                                                                                                                                                         
                 * the allocation was not sampled -- usize can be used to                                                                                                                                                                
                 * determine szind directly.                                                                                                                                                                                             
                 */

PR #1366, opened 2018-11-08
Properly trigger decay on tcache destory.

### LINKED ISSUE #1347: spinning on an atomic leaks memory when using jemalloc

interwq:
When destroying tcache, decay may not be triggered since tsd is non-nominal.
Explicitly decay to avoid pathological cases.

This resolves #1347.

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -497,6 +497,7 @@ tcache_flush(tsd_t *tsd) {
 static void
 tcache_destroy(tsd_t *tsd, tcache_t *tcache, bool tsd_tcache) {
 	tcache_flush_cache(tsd, tcache);
+	arena_t *arena = tcache->arena;
 	tcache_arena_dissociate(tsd_tsdn(tsd), tcache);
 
 	if (tsd_tcache) {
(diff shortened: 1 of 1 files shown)


interwq:
This replaces #1363 which I couldn't seem to reopen. This time solves #1347 for real. Took me a while to realize the root of the ticking issue.

Issue #843, opened 2017-05-18
Consider using pthread mutexes on macOS


jasone:
893a0ed7c8c11962524ba6f2adeb304d038be2a9 dodged reentrant allocation due to ```pthread_mutex_lock()``` on macOS by using ```OSSpinLock```s, and since then we've also added support for ```os_unfair_lock``` as a result of the former becoming deprecated.  It appears that our reentrant support (see #501) may be sufficient to allow ```pthread_mutex_lock()``` calls.  If so, we can expand background thread support (#149) to macOS by using pthreads mutexes and removing the non-standard API support.

davidtgoldblatt:
We still potentially need some mutex acquisition down the reentrancy paths -- I don't think we can necessarily use an allocating lock() there.

We could have two mutex types; the regular one (OSSpinLock), and a condvar-supporting one (pthreads), and avoid using the condvar one in core arena code.

davidtgoldblatt:
Really all we need is a sem_timedwait alternative (which, helpfully, OS X also doesn't support). We could make something on top of kqueue (or some other poll with timeout operation).

jasone:
Hmm, this sounds hard enough that we shouldn't block the 5.0.0 release for it.

interwq:
@davidtgoldblatt: it seems our tests are happy with pthread_mutex on MacOS (#914) -- do you think reentrancy is still causing issues the tests are not covering?

davidtgoldblatt:
Have we tested this in a multithreaded scenario? If OS X does lazy locking internally, we might not notice.

It's plausible that our regular levels of a0 initialization are careful enough that the mutex initialization there works, and that later mutexes can be initialized out of a0, but past that I'm not sure. (Even though we seem to be slowly moving towards "reentrancy really just means 'hostile environment'", I really only looked at most of those paths in the context of an already-initialized a0).

Assuming a0 starts off initialized, and the locking itself doesn't allocate after the first time (i.e. only mutex initialization), I believe we should be fine from a reentrancy standpoint.

michaeleisel:
`OSSpinLock` is not just deprecated it's actually unsafe and can cause livelocks: https://blog.postmates.com/why-spinlocks-are-bad-on-ios-b69fc5221058

here's one example of a repo dealing with that: https://github.com/facebook/pop/issues/336

interwq:
The tests are still fine -- https://travis-ci.org/jemalloc/jemalloc/builds/452649535

[...]

PR #1367, opened 2018-11-08
Deprecate OSSpinLock

interwq:
OSSpinLock is deprecated and we have os_unfair_lock support for Mac.

This solves #843


interwq:
Since OSSpinLock is deprecated, also Jason mentioned that pthread_mutex may cause reentrancy problems on Mac. Let's remove OSSpinLock but still prefer os_unfair_lock on Mac.

PR #1368, opened 2018-11-09
Fix tcaches_flush.

interwq:


djwatson:
Should we have a test for this?

interwq:
> Should we have a test for this?

Yeah I'll update with a test case.

interwq:
Hmm I think I know why our tests didn't catch this in the first place -- we actually create the manual tcache if ID not found: https://github.com/jemalloc/jemalloc/blob/55e5cc1341de87ad06254d719946a5ecd05f06ab/include/jemalloc/internal/tcache_inlines.h#L220

However it's a bit tricky -- we pass in the tcache id for the allocation function; what do we do for an invalid tcache id? Returning NULL seems quite confusing. Manual says `Use the thread-specific cache (tcache) specified by the identifier tc, which must have been acquired via the tcache.create mallctl. This macro does not validate that tc specifies a valid identifier.` So it can be either UB; or maybe better crash immediately.

@djwatson you have been using the API recently, does `abort()` for invalid tcache id sound reasonable to you?

djwatson:
Yea docs should leave it defined as UB, but we should abort()

interwq:
Okay I'll change that in a separate PR.

interwq:
Reading more into this, I realized that the tcache_destroy was intentional in 3a1363b -- it kills off the entire tcache struct, and relies on tcaches_get to re-initialize the tcache when the same id is passed in later.

So the only downside was, we cannot catch an invalid random tcache id, which will actually triggers a tcache to be created. The benefit is we actually "flush" the tcache completely, including the cache stack (which is 28k, not exactly small).

Will do a follow up fix to still destroy the tcache, but only recreate if it is still valid.

PR #1369, opened 2018-11-09
Fix tcache_flush (follow up cd2931a).

interwq:
Also catch invalid tcache id.


interwq:
Followup of #1368. I don't think we can easily add a test case, given that we decided to crash on invalid tcache id.

PR #1371, opened 2018-11-12
Avoid touching all pages in extent_recycle for debug build.

interwq:
We may have a large number of pages with *zero set (since they are populated on
demand).  Only check the first page to avoid paging in all of them.


interwq:
cc: @trasz

trasz:
Works for me, thank you :-)

PR #1378, opened 2018-11-21
Sharded arena bins

interwq:
This makes it possible to have multiple set of bins in an arena, which improves
arena scalability because the bins (especially the small ones) are always the
limiting factor in production workload.

A bin shard is picked on allocation; each extent tracks the bin shard id for
deallocation.  The shard size will be determined using runtime options.


src/arena.c
```diff
@@ -1283,19 +1294,38 @@ arena_bin_malloc_hard(tsdn_t *tsdn, arena_t *arena, bin_t *bin,
...
 
+/* Choose a bin shard and return the locked bin. */
+bin_t *
+arena_bin_choose_lock(tsdn_t *tsdn, arena_t *arena, szind_t binind,
+    unsigned *binshard) {
+	bin_t *bin;
+	if (tsdn_null(tsdn) || tsd_arena_get(tsdn_tsd(tsdn)) == NULL) {
+		*binshard = 0;
+	} else {
+		*binshard = tsd_binshard_get(tsdn_tsd(tsdn)) %
```
interwq: The div instruction accounts for ~8% of tcache_fill_small. I'll see if our `div_compute` function can help here.

include/jemalloc/internal/tsd.h
```diff
@@ -93,6 +94,7 @@ typedef void (*test_callback_t)(int *);
     NULL,								\
     NULL,								\
     NULL,								\
+    ((unsigned)-1),							\
```
djwatson: can we use UINT_MAX?

src/arena.c
djwatson: Why do we need a % here at all?  Doesn't binshard_set already do the mod?

djwatson:
Needs a tes

davidtgoldblatt:
The overall strategy looks reasonable, but ditto with testing.

I wonder if all arenas need to share bin sharding decisions; you could imagine this being determined at arena creation time. (Not necessary for V1 obviously, but something to think about).

src/arena.c
davidtgoldblatt: We don't have an extent we can key off of here.

`div_compute` can't help, since it needs to know that the division is exact. However, we don't need to use `%` for range reduction; there are faster methods (e.g. https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/ ).

interwq: yeah I ended up doing a precompute per thread instead. In the long run we might want to make this more dynamic, instead of the current simple round-robin. So I figure a per thread bin-binding should be useful anyway.

include/jemalloc/internal/tsd.h
interwq: Will fix.

interwq:
re: test case. Yep will add soon.

interwq:
@davidtgoldblatt and @djwatson : you guys mind take another look? The last two commits are new: one implementing the per thread bin binding, the other adding the test case.

Note that this is still based on a slightly older base, to make the prod comparison easier (since we have the base commit running in prod right now). The merge conflict is trivial though.

The prod testing results is still great. We are getting 2-3% CPU win with this setting "bin_shards:1-128:16|129-512:8|513-2048:4".

include/jemalloc/internal/tsd.h
```diff
@@ -74,7 +75,7 @@ typedef void (*test_callback_t)(int *);
     O(iarena,			arena_t *,		arena_t *)	\
     O(arena,			arena_t *,		arena_t *)	\
     O(arenas_tdata,		arena_tdata_t *,	arena_tdata_t *)\
-    O(binshard,			unsigned,		unsigned)	\
+    O(binshards,		tsd_binshards_t,	tsd_binshards_t)\
```
davidtgoldblatt: This should live after the tcache, right? (Or at least, in the tcache bonus area). We don't touch it on any fast paths.

I mean, for the record I still think trying to micro-optimize to that extent is a penny-wise pound-foolish thing that costs us more in dev velocity than it gains us in cycles, but if that's what we're going to decide we should decide it explicitly, and stop trying to do it elsewhere, too.

test/unit/binshard.c
```diff
@@ -0,0 +1,101 @@
+#include "test/jemalloc_test.h"
+
+const char *malloc_conf = "narenas:1,bin_shards:1-128:16|129-512:4";
```
davidtgoldblatt: I think for Windows, this needs to be set in a wrapper script, the way we do with other malloc_conf tests.

Let's test a couple weirder settings, too? (e.g. too many bins for us to represent, a bin with 0, individual size classes picked out from larger ranges for a different setting, etc.).

include/jemalloc/internal/tsd.h
interwq: Oh I was mainly thinking we will want to use this to do per thread, per size binding decisions, e.g. when we have enough contentions, switch this thread to another shard for a specific size class. Agree that the few cycles it saves isn't really a big deal, but combined with the flexibility it gives, might as well have it from the beginning.

The place for this is pretty awkward, it's not exactly tiny (36B), but coupling it into tcache doesn't feel right too. I can imagine for manual arenas people will want to bypass tcache with the help of the sharded bins to avoid too much contention. Put it after tcache may touch an extra page; also likely introduce TLB miss for a semi-fast path (tcache fill).

I slightly prefer to leave it as it, but also fine switching back to a single field and reduce to the range on the fly.

test/unit/binshard.c
interwq: re: the weird settings, the current conf_init errors out on the invalid settings (in `bin_update_shard_size()`), which should be better than silently dropping the option. Combined with abort_conf enabled in dbg mode, it will crash in these cases. I did verify it locally though. Do you prefer to add "abort_conf:false" explicitly for the test case? It will still output the error msg which could be misleading as well.

interwq: re: the wrapper -- yep will change it to there.

interwq: Added test case for overlapping ranges and a individual value from a previous larger range.

include/jemalloc/internal/tsd.h
davidtgoldblatt: I mean, I agree with what you're saying, but that's not the decision procedure we've done in the past; if we think cleanliness is more important than struct layout micro-optimization, we should move (e.g.) tsd_link and in_hook here too, and not bother with the large vs. small distinction. That was something you insisted on, so if you've changed your mind on how important it is we ought to clean up the code, too.

test/unit/binshard.c
davidtgoldblatt: TBH I've never been clear on how we feel about contact-violating malloc_conf options. One the one hand, we try to do something reasonable in the case of nonsensical configurations, which implies that maybe we want to be resilient to them. But on the other, we don't really test or validate the parsing code the way we would if it had to deal with genuinely hostile inputs.

include/jemalloc/internal/tsd.h
interwq: This one is quite on the boardline. On one hand it's required on every tcache fill / flush, so I didn't want it to be at the tail (also combined with the possibility of touching another page); on the other hand it's not really fast path and takes half a cacheline here. I agree that the tcache bonus area is a good fit; the reason pushed me away from that was the recent struggling the rocksDB guys had "wanting a single manual arena; wanting manual tcache to improve scalability; sadly manual tcache increases RSS (w/ >1k threads)". So I made this independent from tcache.

But I think we should try a bit harder here; how about I do a follow up patch to move this into tcache; and then for the allocations w/o tcache we make on-the-fly sharding decisions?

PR #1382, opened 2018-11-28
mutex: fix trylock spin wait contention

djwatson:
If there are 3 or more threads spin-waiting on the same mutex,
there will be excessive exclusive cacheline contention because
pthread_trylock() immediately tries to CAS in a new value, instead
of first checking if the lock is locked.

This diff adds a 'locked' hint flag, and we will only spin wait
without trylock()ing while set.  I don't know of any other portable
way to get the same behavior as pthread_mutex_lock().

This is pretty easy to test via ttest, e.g.

./ttest1 500 3 10000 1 100

Throughput is nearly 3x as fast.

This blames to the mutex profiling changes, however, we almost never
have 3 or more threads contending in properly configured production
workloads, but still worth fixing.

### DIFF
--- a/include/jemalloc/internal/mutex.h
+++ b/include/jemalloc/internal/mutex.h
@@ -43,6 +43,11 @@ struct malloc_mutex_s {
 #else
 			pthread_mutex_t		lock;
 #endif
+			/* 
+			 * Hint flag to avoid exclusive cache line contention
+			 * during spin waiting
+			 */
+			atomic_b_t		locked;
 		};
 		/*
 		 * We only touch witness when configured w/ debug.  However we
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/mutex.h
```diff
@@ -43,6 +43,11 @@ struct malloc_mutex_s {
 #else
 			pthread_mutex_t		lock;
 #endif
+			/* 
+			 * Hint flag to avoid exclusive cache line contention
+			 * during spin waiting
+			 */
+			bool			locked;
```
davidtgoldblatt: Should be `atomic_bool`

davidtgoldblatt: Or rather, `atomic_b_t`.

djwatson: done

interwq:
It seems the arm CI is down. Let me check on that.

PR #1383, opened 2018-11-30
Mention different mmap(2) behaviour with retain:true.

trasz:


davidtgoldblatt:
Sorry for the long delay on this, it completely fell off my radar somehow.

PR #1386, opened 2018-12-04
Only read arena index from extent on the tcache flush path.

interwq:
Add exten_arena_ind_get() to avoid loading the actual arena ptr in case we just
need to check arena matching.


interwq:
When working on the sharded bins, I noticed that avoiding some extent_*_get functions can actually make a measurable difference (at function level of course) in prod workload. This patch removes a load we don't need on the tcache flush path.

Haven't done any prod measurements. Will try one later to confirm it's useful.

davidtgoldblatt:
LGTM assuming tests come back clean.

Any other places where this pops up? IIRC the washing through the arenas array was one of the common sources of circular dependencies with the arena code.

PR #1387, opened 2018-12-05
Move the bin shard assignment from TSD to tcache.

interwq:
The bin shard binding is not accessed on fast path, plus it's not a tiny
struct. So we want it out of TSD hot region. Place it in the tcache bonus area
where the less hot data gather.


interwq:
Will do some benchmark later to make sure no regression.

src/arena.c
```diff
@@ -1343,16 +1343,26 @@ arena_bin_malloc_hard(tsdn_t *tsdn, arena_t *arena, bin_t *bin,
...
+	malloc_mutex_lock(tsdn, &bin->lock);
+
+	return bin;
+}
+
+/* Used on non tcache path only.  */
+static bin_t *
+arena_bin_choose_lock_no_tcache(tsdn_t *tsdn, arena_t *arena, szind_t binind,
+    unsigned *binshard) {
+	bin_t *bin = &arena->bins[binind].bin_shards[0];
```
djwatson: we always choose 0?

interwq:
Not going this direction anymore, considering the bin sharding approach we are moving towards.

PR #1388, opened 2018-12-06
Add rate counters to stats

zinoale:
Work in progress. Let's talk.

### DIFF
--- a/include/jemalloc/internal/emitter.h
+++ b/include/jemalloc/internal/emitter.h
@@ -45,7 +45,9 @@ struct emitter_col_s {
 		int int_val;
 		unsigned unsigned_val;
 		uint32_t uint32_val;
+		uint32_t uint32_t_val;
 		uint64_t uint64_val;
+		uint64_t uint64_t_val;
 		size_t size_val;
 		ssize_t ssize_val;
 		const char *str_val;
--- a/include/jemalloc/internal/mutex_prof.h
+++ b/include/jemalloc/internal/mutex_prof.h
@@ -35,22 +35,31 @@ typedef enum {
 	mutex_prof_num_arena_mutexes
 } mutex_prof_arena_ind_t;
 
+/*
+ * The forth parameter is a boolean value that is true for derived rate counters
+ * and false for real ones.
+ */
 #define MUTEX_PROF_UINT64_COUNTERS					\
-    OP(num_ops, uint64_t, "n_lock_ops")					\
-    OP(num_wait, uint64_t, "n_waiting")					\
-    OP(num_spin_acq, uint64_t, "n_spin_acq")				\
-    OP(num_owner_switch, uint64_t, "n_owner_switch")			\
-    OP(total_wait_time, uint64_t, "total_wait_ns")			\
-    OP(max_wait_time, uint64_t, "max_wait_ns")
+    OP(num_ops, uint64_t, "n_lock_ops", false, num_ops)					\
+    OP(num_ops_ps, uint64_t, "(#/sec)", true, num_ops)				\
+    OP(num_wait, uint64_t, "n_waiting", false, num_wait)				\
+    OP(num_wait_ps, uint64_t, "(#/sec)", true, num_wait)				\
+    OP(num_spin_acq, uint64_t, "n_spin_acq", false, num_spin_acq)			\
+    OP(num_spin_acq_ps, uint64_t, "(#/sec)", true, num_spin_acq)			\
+    OP(num_owner_switch, uint64_t, "n_owner_switch", false, num_owner_switch)		\
+    OP(num_owner_switch_ps, uint64_t, "(#/sec)", true, num_owner_switch)	\
+    OP(total_wait_time, uint64_t, "total_wait_ns", false, total_wait_time)		\
+    OP(total_wait_time_ps, uint64_t, "(#/sec)", true, total_wait_time)		\
+    OP(max_wait_time, uint64_t, "max_wait_ns", false, max_wait_time)
 
 #define MUTEX_PROF_UINT32_COUNTERS					\
-    OP(max_num_thds, uint32_t, "max_n_thds")
+    OP(max_num_thds, uint32_t, "max_n_thds", false, max_num_thds)
 
 #define MUTEX_PROF_COUNTERS						\
 		MUTEX_PROF_UINT64_COUNTERS				\
 		MUTEX_PROF_UINT32_COUNTERS
 
-#define OP(counter, type, human) mutex_counter_##counter,
+#define OP(counter, type, human, derived, base_counter) mutex_counter_##counter,
 
 #define COUNTER_ENUM(counter_list, t)					\
 		typedef enum {						\
--- a/src/stats.c
+++ b/src/stats.c
@@ -52,6 +52,20 @@ char opt_stats_print_opts[stats_print_tot_num_options+1] = "";
 
 /******************************************************************************/
 
+static uint64_t
+rate_per_second(uint64_t value, uint64_t uptime_ns) {
+	uint64_t billion = 1000000000;
+	if (uptime_ns == 0 || value == 0) {
+		return 0;
+	}
+	if (uptime_ns < billion) {
+		return value;
+	} else {
+		uint64_t uptime_s = uptime_ns / billion;
+		return value / uptime_s;
+	}
+}
+
 /* Calculate x.yyy and output a string (takes a fixed sized char array). */
 static bool
 get_rate_str(uint64_t dividend, uint64_t divisor, char str[6]) {
@@ -104,12 +118,12 @@ mutex_stats_init_cols(emitter_row_t *row, const char *table_name,
 
 #define WIDTH_uint32_t 12
 #define WIDTH_uint64_t 16
-#define OP(counter, counter_type, human)				\
+#define OP(counter, counter_type, human, derived, base_counter)	\
 	col = &col_##counter_type[k_##counter_type];			\
 	++k_##counter_type;						\
 	emitter_col_init(col, row);					\
 	col->justify = emitter_justify_right;				\
-	col->width = WIDTH_##counter_type;				\
+	col->width = derived ? 8 : WIDTH_##counter_type;		\
 	col->type = emitter_type_title;					\
 	col->str_val = human;
 	MUTEX_PROF_COUNTERS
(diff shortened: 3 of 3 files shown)


include/jemalloc/internal/mutex_prof.h
```diff
@@ -36,21 +36,26 @@ typedef enum {
...
 
 #define MUTEX_PROF_UINT64_COUNTERS					\
-    OP(num_ops, uint64_t, "n_lock_ops")					\
-    OP(num_wait, uint64_t, "n_waiting")					\
-    OP(num_spin_acq, uint64_t, "n_spin_acq")				\
-    OP(num_owner_switch, uint64_t, "n_owner_switch")			\
-    OP(total_wait_time, uint64_t, "total_wait_ns")			\
-    OP(max_wait_time, uint64_t, "max_wait_ns")
+    OP(num_ops, uint64_t, "n_lock_ops", false, num_ops)					\
+    OP(num_ops_ps, uint64_t, "n_lock_ops_ps", true, num_ops)				\
```
interwq: We don't really need the per second values to be real counters. They are derived so we only need to compute on demand, i.e. when output. It can be done similar to `util` value https://github.com/jemalloc/jemalloc/blob/dev/src/stats.c#L364, which is also derived and only output to the human format only.

src/stats.c
```diff
@@ -52,6 +52,10 @@ char opt_stats_print_opts[stats_print_tot_num_options+1] = "";
 
 /******************************************************************************/
 
+uint64_t rate_per_second(uint64_t value, uint64_t uptime_ns) {
+    return uptime_ns ? (uint64_t)((double)value * 1e9 / (double)uptime_ns) : 0;
```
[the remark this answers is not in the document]
davidtgoldblatt: I think this should be OK in stats printing, which is only called via at_exit (where it's just another user function) or via an explicit malloc_stats_print call. No?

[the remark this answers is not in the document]
interwq: That might be the case. However I have always wanted an option to dump malloc_stats once in a while similar to prof_interval. Plus we can get away without using floating point.

include/jemalloc/internal/mutex_prof.h
zinoale: They are not real. The 4th argument is "synthetic" that is true for rates and false for originals. The synthetic counters are never ctl-ed but computed on demand based on where the "real" counter is. And the name of the "real" counter is provided with the 5th argument of the macro.

Having the synthetics inserted inside the sequence of OP() makes us able to have synthetics next to the real countners and not to keep them separate.

src/stats.c
```diff
@@ -52,6 +52,19 @@ char opt_stats_print_opts[stats_print_tot_num_options+1] = "";
...
+		return 0;
+	}
+	uint64_t uptime_s = uptime_ns / 1000000000;
+	if (!uptime_s) {
+		/* Hope there is no overflow if the app ran for less than a second */
+		return value * 1000000000 / uptime_ns;
+	} else {
+		return value / uptime_s;
+	}
+}
```
interwq: Since there are counters for rare-ish events; e.g. the slow path waiting for some locks may only happen a few times a second, I'd prefer we calculate the per milli-second rate, and convert to `x.yyy` if the rate is <=10. The function should help there.

This can be done in a new commit.

include/jemalloc/internal/mutex_prof.h
interwq: I see. To keep the terms consistent, let's use `derived` for the synthetic ones. Also we would need some comments to mention that the derived ones are used for rates currently.

src/stats.c
```diff
@@ -204,19 +234,38 @@ mutex_stats_emit(emitter_t *emitter, emitter_row_t *row,
...
-#define OP(counter, type, human)					\
-	col = &col_##type[k_##type];						\
-	++k_##type;							\
-	emitter_json_kv(emitter, #counter, EMITTER_TYPE_##type,		\
-	    (const void *)&col->bool_val);
+#define OP(counter, type, human, synthetic, base_counter)		\
+	if (!synthetic) {                    \
+		col = &col_##type[k_##type];                        \
+		++k_##type;                            \
+		emitter_json_kv(emitter, #counter, EMITTER_TYPE_##type,        \
```
interwq: We don't want the derived counters to be included for JSON format (e.g. similarly the `util` counter is not included in JSON output` -- they are parsed by applications and can be easily derived). Let's simply ignore them for JSON.

src/stats.c
```diff
@@ -298,7 +339,7 @@ stats_arena_bins_print(emitter_t *emitter, bool mutex, unsigned i) {
 		uint64_t nmalloc, ndalloc, nrequests, nfills, nflushes;
 		uint64_t nreslabs;
 
-		CTL_M2_M4_GET("stats.arenas.0.bins.0.nslabs", i, j, &nslabs,
+        CTL_M2_M4_GET("stats.arenas.0.bins.0.nslabs", i, j, &nslabs,
```
interwq: nit: alignment is off.

src/stats.c
zinoale: They are not included into JSON as the if() statement at line 238 disables it. I can rename "synthetic" to be "derived" if it makes the code more readable.

zinoale:
A few lines from the original report:

                            allocated         nmalloc         ndalloc       nrequests
small:                          11648              20               2               5
large:                          28672               2               1               2
total:                          40320              22               3               7

The corresponding lines for the updated report:

                            allocated         nmalloc      nmalloc_ps         ndalloc      ndalloc_ps       nrequests    nrequests_ps
small:                          11648              20           48861               2            4886               5           12215
large:                          28672               2            4886               1            2443               2            4886
total:                          40320              22            4886               3            2443               7            4886

zinoale:
[old.txt](https://github.com/jemalloc/jemalloc/files/2681796/old.txt)
[new.txt](https://github.com/jemalloc/jemalloc/files/2681797/new.txt)

include/jemalloc/internal/emitter.h
```diff
@@ -45,7 +45,9 @@ struct emitter_col_s {
 		int int_val;
 		unsigned unsigned_val;
 		uint32_t uint32_val;
+		uint32_t uint32_t_val;
 		uint64_t uint64_val;
+		uint64_t uint64_t_val;
```
interwq: Are these two lines used anywhere?

include/jemalloc/internal/mutex_prof.h
```diff
@@ -35,22 +35,31 @@ typedef enum {
...
+    OP(num_wait, uint64_t, "n_waiting", false, num_wait)				\
+    OP(num_wait_ps, uint64_t, "n_waiting_ps", true, num_wait)				\
+    OP(num_spin_acq, uint64_t, "n_spin_acq", false, num_spin_acq)			\
+    OP(num_spin_acq_ps, uint64_t, "n_spin_acq_ps", true, num_spin_acq)			\
+    OP(num_owner_switch, uint64_t, "n_owner_switch", false, num_owner_switch)		\
+    OP(num_owner_switch_ps, uint64_t, "n_owner_switch_ps", true, num_owner_switch)	\
+    OP(total_wait_time, uint64_t, "total_wait_ns", false, total_wait_time)		\
+    OP(total_wait_time_ps, uint64_t, "total_wait_ns_ps", true, total_wait_time)		\
```
interwq: Since the name can get a bit lengthy, let's change all the ps string to simply `(#/sec)`? That way it's probably clear enough, w/o using too much space.

src/stats.c
```diff
@@ -52,6 +52,23 @@ char opt_stats_print_opts[stats_print_tot_num_options+1] = "";
...
+rate_per_second(uint64_t value, uint64_t uptime_ns) {
+	static uint64_t billion = 1000000000;
+	if (uptime_ns == 0 || value == 0) {
+		return 0;
+	}
+	if (uptime_ns < billion) {
+		if (opt_abort) {
+			assert(value * billion > value);
+		}
+		return value * billion / uptime_ns;
```
interwq: Well I changed my mind here; let's simply consider the subsecond case as 1 second, i.e. 
  if (uptime_ns < billion) {
       return value; 
  }

src/stats.c
```diff
@@ -103,8 +120,8 @@ mutex_stats_init_cols(emitter_row_t *row, const char *table_name,
 	}
 
 #define WIDTH_uint32_t 12
-#define WIDTH_uint64_t 16
-#define OP(counter, counter_type, human)				\
+#define WIDTH_uint64_t 18
```
interwq: I assume no need to increase this after we use the `#/sec` for title?

src/stats.c
```diff
@@ -230,44 +283,36 @@ stats_arena_bins_print(emitter_t *emitter, bool mutex, unsigned i) {
...
-#undef COL
+	COL_HDR(row, justify_spacer, right, 1, title)
+	COL_HDR(row, util, right, 6, title)
+	COL_HDR(row, nfills, right, 13, uint64)
+	COL_HDR(row, nfills_ps, right, 13, uint64)
+	COL_HDR(row, nflushes, right, 13, uint64)
+	COL_HDR(row, nflushes_ps, right, 13, uint64)
+	COL_HDR(row, nslabs, right, 13, uint64)
+	COL_HDR(row, nreslabs, right, 13, uint64)
+	COL_HDR(row, nreslabs_ps, right, 13, uint64)
```
interwq: Please also try tightening the spacing after using the `#/sec` for header. The stats looks pretty messy when line wraps around.

include/jemalloc/internal/emitter.h
zinoale: Yes, unfortunately they are used inside macros like
    dst->counter_type##_val = rate_per_second(base->counter_type##_val, uptime);
where for the case the provided counter_type is uint64_t not uint64.

If I use the char* union member and cast it to the desired type I get compilation warning or errors at least on MacOS.

I can just rename the union member not to add a new one but this will lead to more places in the code being fixed. Please let me know if you want me to do so.

interwq: I see. Please leave it as is then.

zinoale:
I replaced titles for rates to become shorter strings and shrank spaces accordingly.
The new version of the "new" file is attached.
[new.txt](https://github.com/jemalloc/jemalloc/files/2688710/new.txt)

interwq:
Thanks for working on this @zinoale!

PR #1389, opened 2018-12-08
Fix incorrect stats mreging with sharded bins.

interwq:
With sharded bins, we may not flush all items from the same arena in one run.
Adjust the stats merging logic accordingly.

The regression was introduced in #1378.


interwq:
Thanks to David's quick response! Merging w/o waiting for the tests since this is a bit urgent for a production patch.

interwq:
Some note: we didn’t catch this because the tests don’t really try hard on the remote deallocation / producer-consumer model. It didn’t cause any damage in production, but triggered some alarms in our test system. I’ll add some tests to explicitly stress the remote free path.

PR #1390, opened 2018-12-10
refactoring, more readable code: lock outside of arena_bin_choose

liutongxuan:
arena_bin_choose_lock function lock the bin inside of it, that's unclear in caller's code.


interwq:
Thanks for submitting the PR. However the bin choose and lock functions are combined this way intentionally. First, they should always be used together; second which bin to choose may be dependent on the locking status — we have also been experimenting a trylock based bin binding strategy, where we won’t know if the bin is preferred until after the lock operation.

Feel free to ask if anything.

liutongxuan:
thanks for the explanation. When read this part of code, hard to find malloc_mutex_lock is inside of arena_bin_choose_lock (this make the code logic separated, weird), using lock/unlock to be pairs probably more clear.

PR #1394, opened 2018-12-14
Add --{enable,disable}-{static,shared} to configure script

Ericson2314:
My distro (NixOS) offers a custom toolchain where it's not possible to make static libs, so it's insufficient to just delete the libs I don't want. I actually need to avoid building them in the first place.

CC @vaibhavsagar @matthewbauer


davidtgoldblatt:
This seems reasonable to me.

Speak up now if anyone has objections, or I'll merge in a couple days.

davidtgoldblatt:
Thanks for the diff!

vaibhavsagar:
Thanks!

PR #1405, opened 2019-01-11
Force purge on thread death only when w/o bg thds.

interwq:
We can rely on bg thds instead of force purging everything on thread exit.


interwq:
@jasone : may need your insights on the cirrus-ci failure -- it looks to be failing during fetching packages. The log doesn't say much https://cirrus-ci.com/task/5381651179765760. I think it feels this way randomly. Any idea?

jasone:
I just manually kicked off a re-run (have to be logged in to the cirrus-ci.com site to do so), since the failure does look spurious.

PR #1409, opened 2019-01-14
Avoid creating bg thds for huge arena lone.

interwq:
For low arena count settings, the huge threshold feature may trigger an unwanted
bg thd creation.  Given that the huge arena does eager purging by default,
bypass bg thd creation when initializing the huge arena.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -2276,6 +2276,17 @@ arena_i_decay_ms_ctl_impl(tsd_t *tsd, const size_t *mib, size_t miblen,
 			ret = EINVAL;
 			goto label_return;
 		}
+		if (arena_is_huge(arena_ind) && *(ssize_t *)newp > 0) {
+			/*
+			 * By default the huge arena purges eagerly.  If it is
+			 * set to non-zero decay time afterwards, background
+			 * thread might be needed.
+			 */
+			if (background_thread_create(tsd, arena_ind)) {
+				ret = EFAULT;
+				goto label_return;
+			}
+		}
 		if (dirty ? arena_dirty_decay_ms_set(tsd_tsdn(tsd), arena,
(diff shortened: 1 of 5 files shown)


interwq:
I just realized that, with `huge_threshold` enabled, the huge arena might trigger an unnecessary bg thd creation. This matters for the small application with low arena count, e.g. with `narenas:1`. Let's avoid the extra bg thd in that case, since the huge arena purges immediately.

An edge case is, if the decay time of the huge arena is set to non-0, we'd need to try and create bg thd in that case. Added support for this as well, since I don't think we want to explicitly disallow that (setting decay time of the huge arena).

src/ctl.c
```diff
@@ -2276,6 +2276,17 @@ arena_i_decay_ms_ctl_impl(tsd_t *tsd, const size_t *mib, size_t miblen,
 			ret = EINVAL;
 			goto label_return;
 		}
+		if (arena_is_huge(arena_ind) && *(ssize_t *)newp > 0) {
+			/*
+			 * By default the huge arena purges eagerly.  If it is
```
davidtgoldblatt: I'm sort of inclined to say that the user shouldn't be allowed to change purging settings for the huge arena; WDYT?

interwq: I was thinking about that too. However if there's some valid case where huge allocations are freed and reused often (or rather bad application code), maybe we should have some mechanism in place, which might be easier to fix the usage?

interwq: Oh I agree with you. I'm thinking this as a hack (i.e. not documented). I also think the huge arena thing should not be user visible. Right now we don't even provide a way to figure out which arena is for huge (other than doing a huge allocation and check).

I was thinking this case for, when someone comes in and says they have a weird and hard-to-change workload, do we have a fix for them  -- I also agree that we should not encourage such solutions; but given that we have the huge arena enabled by default, I feel we should be more prepared.

PR #1412, opened 2019-01-16
Set huge_threshold to 8M by default.

interwq:
This feature uses an dedicated arena to handle huge requests, which
significantly improves VM fragmentation.  In production workload we tested it
often reduces VM size by >30%.


davidtgoldblatt:
The Windows failures look real.

interwq:
The test is fixed. Ready for review now.

interwq:
As discussed offline, will rename this feature soon.

PR #1417, opened 2019-01-25
Update copyright dates.

interwq:


davidtgoldblatt:
On some open source projects, we just say "Copyright blah blah blah, 2002-Present". Can we do that here? I mean, I know the cost is just 1 commit a year, but it feels so silly.

PR #1420, opened 2019-01-28
Tweak the spacing for the total_wait_time per second.

interwq:


interwq:
The per second total_wait_time has nano second as unit. Make it more lengthy as production data does go pretty high.

As a side note, I did find the rate counters very helpful. I tend to look for them instead of the raw total.

PR #1432, opened 2019-02-14
Fix enabling of both background thread and profiling on macOS

BenHuddleston:


interwq:
Background thread is not supported on MacOS because of it only supports pthread at the moment (and we haven't moved to pthread based API on Mac yet). See https://github.com/jemalloc/jemalloc/issues/843

PR #1437, opened 2019-02-17
Restructure jeprof's documentation and generate manpage/HTML

paravoid:
jeprof currently has most of its documentation in the middle of the code, in a string in the usage_string sub, plus some of it in comments at the top.
    
This converts all that to standard POD and uses Pod::Usage to generate the usage output, as well as pod2man and pod2html to generate a manpage and HTML documentation for it. All of three of the above are shipped by core Perl and do not introduce an additional build dependency.

The --help output is not 1:1, and differs on horizontal and vertical whitespace. Also, neither the manpage nor the HTML match exactly DocBook's output, which is a bit of an annoying inconsistency. I played around with pod2docbook which, apart from being an additional external dependency, it does not seem to generate a DocBook XML that can be used to generate a manpage.


paravoid:
Ping? Would be neat if this (or a variation of it) was part of 5.2 :)

davidtgoldblatt:
+cc @interwq? I think you are the least-uninformed about the documentation generation stuff.

interwq:
@paravoid: thanks a lot for the patch! Unfortunately I don't think we want to ship this in 5.2 yet, given that we have had quite some users reporting doc generation related issues (mostly dependency config), and we recently already added a new workaround and switched to build docs by default (even for release tarballs, see #1212 and #1430). Given that this involves new dependencies and a fair amount of code movement, I plan to come back to this after 5.2.

paravoid:
@interwq thanks, I understand. Would it perhaps make sense to just merge the first of these commits (the conversion to POD) and not hook it up to the build system for the manpage/HTML in the meantime? That way us downstream distributors could call pod2man from the packaging Makefiles etc. but wouldn't have to carry an invasive patch like that :)

paravoid:
It's been a while and 5.2 was released in the meantime -- apologies for pestering you! Any feedback on this? @interwq?

interwq:
Thanks for reminding @paravoid! My concern / question is still about the new dependency on pod2html and pod2man -- any insights regarding its availability on other platforms (BSDs, MacOS and Windows)? I'd be reluctant to accept it as is, given that relying on `xsltproc` already caused confusions to quite some folks. You can check the detection / error handling around xsltproc for details.

paravoid:
I haven't used a BSD/MacOS/Windows system in many years, but pod2man and pod2html are both part of Perl (as in, the main Perl distribution, not a CPAN module or something like that), so I'd expect them to be present there.

interwq:
Sorry again for the delay here. Although the refactored doc certainly looks better, I have a bit mixed feeling regarding the additional pod2* dependencies. @jasone any input?

FYI that we might need to do a 5.2.1 because of #1545 which fixed a long standing VM leak on Windows.

jasone:
I prefer relying on the `--help` output over adding a build dependency.  We will always have rough edges around `jeprof` due to its external origin, and although I don't have strong opinions in this case, in general `jeprof` is best left loosely integrated.

paravoid:
FWIW the only build dependency would be Perl itself. But, if having Perl as a build-dependency is a problem, then we can do as I suggested in a comment above: merge the first two commits (= move the `--help` output in a dedicated machine-parseable section in the Perl script) and let downstream distributors optionally generate a manpage if required.

davidtgoldblatt:
Something like this still appeals to me, but I think it probably won't happen in the near term -- closing out for now.

Eventually, I think we'll end up with a jeprof rewrite, at which point revisiting makes sense.

Issue #1438, opened 2019-02-17
libgcc profiling initialization & fakechroot


paravoid:
OK, this is not really a jemalloc bug, but I'm wondering if you have any insight or workarounds to provide for this.

In Debian, I enabled (libgcc) `--enable-prof` for amd64/i386 builds. One of the bugs I encountered while enabling this was a deadlock in the [fakechroot](http://packages.debian.org/sid/fakechroot) package. fakechoot is a piece of software which does a bunch of nasty LD_PRELOAD hacks to give a chroot environment to unprivileged users in which they can do root-like operations like bootstrap Debian.

I tracked down the issue and it seems to be sourced in the fact that with profiling build-enabled (`--enable-prof`), jemalloc calls `_Unwind_Backtrace` to initialize the unwind code. This happens unconditionally, even if profiling is not runtime-enabled.

This in turn seems to call `dl_iterate_phdr`, which is one of the libc calls that fakechroot captures; fakechroot then calls `dlsym`/`dlerror`, which allocates memory, which then ends up in jemalloc territory again, and (naturally) locks up. I have a full backtrace over at the Debian BTS, at [#918742](https://bugs.debian.org/918742).

Obviously when one intercepts and replaces libc functions like `dl_iterate_phdr`... all bets are off. Still, I was wondering if there was any way that e.g. `_Unwind_Backtrace` could be initialized more lazily, to minimize the initialization code path avoid such a thing? Or any other advice you may have :)

(Note that this isn't triggered when using libunwind due to a different initialization path. @jasone had previously recommended using unwind in #338, but I wanted to avoid the additional dependency for now. This is something I can and may have to revisit, though.)

interwq:
Will simply guarding the `_Unwind_Backtrace` init call with opt.prof work for you? https://github.com/jemalloc/jemalloc/pull/1441

paravoid:
It would, but I assumed there was a reason this wasn't the case as it felt quite deliberate. Would be happy if my assumption proved to be wrong!

interwq:
Thanks for reporting this @paravoid!

PR #1441, opened 2019-02-20
Guard libgcc unwind init with opt_prof.

interwq:
Only triggers libgcc unwind init when prof is enabled.  This helps workaround
some bootstrapping issues.


davidtgoldblatt:
Can you add an explanation as to why in the commit summary (or as a comment if needed)?

interwq:
@davidtgoldblatt : just updated the commit msg. This should workaround #1438.

Issue #1444, opened 2019-02-22
Windows: unresolved external symbol tls_callback


daverigby:
When building jemalloc v5.1.0 on Windows Server 2016 with MSVC2017 (and the following configure options), I see a linker error related to `tls_callback`:

```
sh -c "./autogen.sh CC=cl --prefix=C:/Jenkins/workspace/kv_engine-windows-master/build-deps/jemalloc/install"
```

Error:

```
cl -nologo -LD  -Felib/jemalloc.dll src/jemalloc.obj src/arena.obj src/background_thread.obj src/base.obj src/bin.obj src/bitmap.obj src/ckh.obj src/ctl.obj src/div.obj src/extent.obj src/extent_dss.obj src/extent_mmap.obj src/hash.obj src/hook.obj src/large.obj src/log.obj src/malloc_io.obj src/mutex.obj src/mutex_pool.obj src/nstime.obj src/pages.obj src/prng.obj src/prof.obj src/rtree.obj src/stats.obj src/sc.obj src/sz.obj src/tcache.obj src/test_hooks.obj src/ticker.obj src/tsd.obj src/witness.obj   -link -DEBUG
   Creating library lib/jemalloc.lib and object lib/jemalloc.exp
tsd.obj : error LNK2001: unresolved external symbol tls_callback
lib/jemalloc.dll : fatal error LNK1120: 1 unresolved externals
```

Compiler version: Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27026.1 for x64
> where cl
C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64\cl.exe

I _believe_ the problem is the linker directive given to MSVC to forcibly include the `tls_callback` symbol (to fix #318) - this isn't subject to the `JEMALLOC_N()` symbol prefix, but the actual symbol declaration _is_ - as such we end up with a linker directive referring to a symbol which isn't present.

The below patch fixes the problem in my environment - the linker string is tokenised (and stringified) so the correct replacement from `private_namespace.h` can be applied.

```
diff --git a/src/tsd.c b/src/tsd.c
index f317d486..2eceed90 100644
--- a/src/tsd.c
+++ b/src/tsd.c
@@ -472,7 +472,7 @@ _tls_callback(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved) {
[...]

PR #1445, opened 2019-02-22
Stringify tls_callback linker directive

daverigby:
Proposed fix for #1444 - ensure that `tls_callback` in the `#pragma comment(linker)`directive gets the same prefix added as it does i the C declaration.


interwq:
Thanks for the patch!

nimaforoughi:
#pragma comment (linker, "/INCLUDE:" STRINGIFY(__tls_used))

PR #1448, opened 2019-03-08
Remove best fit

djwatson:
This option saves a few CPU cycles, but potentially adds a lot of fragmentation - so much so that there are workarounds like max_active.  Instead, let's just drop it entirely.  It only made a difference in one service I tested (.3% cpu regression), while many services saw a memory win (also small, less than 1% mem P99)

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -441,30 +441,6 @@ extents_fit_alignment(extents_t *extents, size_t min_size, size_t max_size,
 	return NULL;
 }
 
-/* Do any-best-fit extent selection, i.e. select any extent that best fits. */
-static extent_t *
-extents_best_fit_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents,
-    size_t size) {
-	pszind_t pind = sz_psz2ind(extent_size_quantize_ceil(size));
-	pszind_t i = (pszind_t)bitmap_ffu(extents->bitmap, &extents_bitmap_info,
-	    (size_t)pind);
-	if (i < SC_NPSIZES + 1) {
-		/*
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
Love it. That being said, maybe let's wait until after we cut the release to merge, in case we want to get any last-minute fixes in cleanly?

interwq:
Yeah let’s wait until we cut 5.2. I’m still curious about the effects of this on long running binaries.

djwatson:
kept the max_active_fit check in first_fit path instead of removing.

src/extent.c
```diff
@@ -483,7 +459,16 @@ extents_first_fit_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents,
 		assert(!extent_heap_empty(&extents->heaps[i]));
 		extent_t *extent = extent_heap_first(&extents->heaps[i]);
 		assert(extent_size_get(extent) >= size);
-		if (ret == NULL || extent_snad_comp(extent, ret) < 0) {
+                bool size_ok = true;
+		/*
+		 * In order to reduce fragmentation, avoid reusing and splitting
+		 * large extents for much smaller sizes.
+		 */
+		if ((sz_pind2sz(i) >> opt_lg_extent_max_active_fit) > size) {
```
interwq: The intention was to only apply this to dirty_extents, maybe add a new arg to `extents_fit_locked` and specify if max_active_fit should be checked?

djwatson: Sure, update with restriction to delay_coalesce (which is only set for dirty).

interwq:
Should we land this and then #1455 next?

Issue #1429, opened 2019-02-04
5.2 release


interwq:
This issue is for tracking communication/progress related to the 5.2 release.

Current release candidate: 8e9a613

We have a few notable changes that warrant this release -- 
- the `max_background_threads` fix
- a series of fast path optimizations
- the `oversize_threshold` feature
- allow `void *p = ::operator new(0);  ::operator delete(p, 0);`

interwq:
The current release [candidate](https://github.com/jemalloc/jemalloc/commit/8e9a613122251d4c519059f8e1e11f27f6572b4c) has been up in production for a while. We will give it a few more weeks before cutting the release. I'll prepare a draft of the changelog next week.

@jasone: could you help with testing the RC on FreeBSD-current? As far as I can think of, the only visible "behavior" change is that we will use a dedicated arena for >=8M allocations (the `oversize_threshold` feature which is to avoid mixing "huge" and smaller allocations).

jasone:
I'll work on FreeBSD integration/testing during the next week.

ingvarha:
8e9a613 does not run full test suite on s390x with gcc-9.0.1, see #1434.

Ingvar

ldcsaa:
Would you tell us when will 5.2.0 be released ?

antirez:
Are there critical issues (that can lead to instability of programs using Jemalloc) that 5.2 will fix and are present in 5.1.0? Thanks.

davidtgoldblatt:
> Would you tell us when will 5.2.0 be released?

Probably order weeks, unless some undiscovered bug appears.

> Are there critical issues (that can lead to instability of programs using Jemalloc) that 5.2 will fix and are present in 5.1.0?

5.2 has better behavior w.r.t. fragmentation of virtual memory; in some pathological cases 5.1 can have a very high ratio of virtual address space consumed to pages actually in memory.

interwq:
In addition to what David mentioned, 5.2 will have a few bug fixes for the  ‘background_thread’ feature (which is not enabled by default)

jasone:
I think I may have run into a serious portability problem for the mips platform on FreeBSD.  Shortly after the 5.1.0 release, 982c10de3566f38628770e57c62d1a6cdc5a09f9 added 8-bit atomics, i.e. in `include/jemalloc/internal/atomic.h`:

```c
JEMALLOC_GENERATE_INT_ATOMICS(uint8_t, u8, 0)
```

[...]

PR #1449, opened 2019-03-08
Workaround when  8-bit atomics are unavailable

interwq:
Detect if 8-bit atomics are available.  In some rare cases (older compiler, e.g. gcc 4.2 w/ MIPS), 8-bit atomics might be unavailable.  Detect such cases and fallback to 32bit instead.


interwq:
@jasone : can you please try build this with MIPS and see if it works? This is to workaround the issue you encountered in #1429

Also apologize for the delay. Had a few other things going on.

jasone:
Will do; I won't have much time tomorrow, but should be able to get this done by the end of the weekend.

PR #1455, opened 2019-03-11
Page cache

djwatson:
Add a pagecache in front of the slabs.  The vast majority of extent tree manipulations comes from the slabs and not large requests. Currently we are depending on delay_coalesce to prevent lots of merge/split operations, instead let's make this very explicit with a pagecache.
    
It supports watermarks and fill/flush similar to how tcache works.
    
Current prod tests show this as pretty memory-neutral, and a small CPU win, mostly from splitting what would be a single extent mutex in to page sized mutexes.

(builds on top of #1448 )


davidtgoldblatt:
No longer mergeable -- we also have the SEC implemented, which is fast as-is

Issue #1454, opened 2019-03-11
Memory leak introduced between 5.01 and 5.1


mpghf:
I've been experiencing a new memory leak that was introduced between 5.01 and 5.1 releases.  I've narrowed it down to a particular commit:  it starts on 3e64dae802b9f7cd4f860b0d29126cd727d5166b and affects all subsequent versions, including head revision.

The system is a fairly standard Gentoo Linux system:

4.9.72-gentoo-64Bit # 1 SMP PREEMPT Thu Aug 2 00:25:30 -00 2018 x86_64 Intel(R) Xeon(R) CPU E5-2695 v2 @ 2.40GHz GenuineIntel GNU/Linux

The compiler is a standard gcc 5.4:

gcc (Gentoo 5.4.0-r4 p1.7, pie-0.6.5) 5.4.0

Prior to the commit above  (or using the stock Linux malloc), memory usage stays constant (~300 MB) indefinitely.  From that commit onward, memory usage grows quickly until the system runs out of memory (3.6 GB).

Running jemalloc with leak profiling enabled (as detailed on the wiki) shows no sign of the missing memory.

As for what I'm doing to provoke this:  

I've narrowed it down to a small piece of code, but the same code in a separate program doesn't provoke the same error, so I assume that it's the exact context that causes the problem.  The program in which this code is found is a medium size, proprietary application, with a number of threads (5 or so) running, although the other threads are doing less and allocating less memory.

The sample in questions looks like this (_**note that this code does not provoke the problem by itself**_):
```
#include <iostream>
#include <thread>
#include <list>
#include <chrono>

using namespace std;


class TestClass
{
public:
	TestClass(unsigned uLength)
	{
		p = new unsigned char[uLength];
		//The empty space is filled with useful data
		for(unsigned u = 0; u < std::min(uLength, 256u); u++) {
			p[u] = std::rand() | 0xff;
		}
	}
	~TestClass()
	{
		delete[] p;
	}
private:
	unsigned char *p = nullptr;
};

shared_ptr<TestClass> ExtractCaptures()
{
	shared_ptr<TestClass> p = make_shared<TestClass>(49340);
	return p;
}

int main()
{
	do {
		//It doesn't seem to matter whether this is a list or vector
		list<shared_ptr<TestClass>> captures;
		for(unsigned u = 0; u < 256; u++) {
			shared_ptr<TestClass> p = ExtractCaptures();
			//Removing the push_back below makes the problem disappear
[...]

PR #1458, opened 2019-03-13
Eagerly purge oversized extents.

interwq:
This change improves memory usage slightly, at virtually no CPU cost.

Also resolves the issue in #1454


interwq:
Verified again that production workload benefits from this change. Landing now with the intention to include this in the coming 5.2 release. Will give it a couple more weeks to test widely in production, before cutting 5.2.

This also resolves #1454

PR #1460, opened 2019-03-15
Add format annotation to the format generator function

zoulasc:
Convert the format generator function to an annotated format function,
so that the generated formats can be checked by the compiler.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -851,6 +851,18 @@ if test "x${je_cv_format_printf}" = "xyes" ; then
   AC_DEFINE([JEMALLOC_HAVE_ATTR_FORMAT_PRINTF], [ ])
 fi
 
+dnl Check for format_arg(...) attribute support.
+JE_CFLAGS_SAVE()
+JE_CFLAGS_ADD([-Werror])
+JE_CFLAGS_ADD([-herror_on_warning])
+JE_COMPILABLE([format(printf, ...) attribute], [#include <stdlib.h>],
+              [const char * __attribute__((__format_arg__(1))) foo(const char *format);],
+              [je_cv_format_arg])
+JE_CFLAGS_RESTORE()
+if test "x${je_cv_format_arg}" = "xyes" ; then
+  AC_DEFINE([JEMALLOC_HAVE_ATTR_FORMAT_ARG], [ ])
+fi
+
 dnl Support optional additions to rpath.
 AC_ARG_WITH([rpath],
   [AS_HELP_STRING([--with-rpath=<rpath>], [Colon-separated rpath (ELF systems only)])],
--- a/include/jemalloc/internal/emitter.h
+++ b/include/jemalloc/internal/emitter.h
@@ -86,10 +86,11 @@ emitter_printf(emitter_t *emitter, const char *format, ...) {
 	va_end(ap);
 }
 
-static inline void
+static inline const char * __attribute__((__format_arg__(3)))
 emitter_gen_fmt(char *out_fmt, size_t out_size, const char *fmt_specifier,
     emitter_justify_t justify, int width) {
 	size_t written;
+	fmt_specifier++;
 	if (justify == emitter_justify_none) {
 		written = malloc_snprintf(out_fmt, out_size,
 		    "%%%s", fmt_specifier);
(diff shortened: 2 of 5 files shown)


davidtgoldblatt:
This mostly looks good, thanks for the patch. A couple nits:
- This should be written in terms of `JEMALLOC_FORMAT_PRINTF` (not all the compilers we support have that attribute syntax).
- Any reason for the return type change of `emitter_gen_fmt`? I find this version slightly less easy to read.

davidtgoldblatt:
(Also, @interwq is trying to cut a release, so this merge will probably be delayed for a week or two).

zoulasc:
1. Will fix.
2. Can't set the format to an intermediate variable because of compiler brain-damage (it loses track that it is a format).

zoulasc:
Perhaps it is better to delay the release and because I have more patches (some of them minor bugs). For example a missing unlock https://github.com/jemalloc/jemalloc/blob/dev/src/extent.c#L799, etc.

davidtgoldblatt:
Makes sense to me unless I'm missing something w.r.t. the locking.

+cc @interwq -- the unmatched locking looks troubling (side question: why didn't top-level mutex assertions catch this?)

zoulasc:
We caught it because our unit tests reduce process memory limits enough to catch causes that do not regularly happen. For example, this one was caught here: https://nxr.netbsd.org/xref/src/tests/lib/libc/stdio/t_printf.c#154

interwq:
@zoulasc : thanks for catching the issue! We have been delaying the 5.2 release for a while and I'm trying to make it happen in the next few days -- the context is that some compiler folks are waiting the release, in order to land a patch which utilizes the sized deallocation API; and it sounds like gcc 9 will be cut very soon and we really want the patch to make it.

If you don't mind, I'll make a PR on your behalf to fix the unmatched locking bug. Please let us know if there's more patches you'd like to include in 5.2.

zoulasc:
I have a whole bunch of cleanups;
1. to use c99 initializers
2. to clean up nested variable warnings
3. fix missing prototypes

And other minor things like, I think this is change is needed:
                        CONF_HANDLE_SIZE_T(opt_lg_extent_max_active_fit,
                            "lg_extent_max_active_fit", 0,
-                           (sizeof(size_t) << 3), yes, yes, false)
+                           (sizeof(size_t) << 3), no, yes, false)
Because the compiler complains. As a matter of fact the compiler has found many more minor issues when all the warnings are enabled. 

But I think none should delay the release, so finish that and I will send patches later.

interwq:
@zoulasc -- thanks again for all the info! Yeah at this time with all the production testing already done, we prefer only landing correctness related changes.

I'll make the unlock fix, also the CONF_HANDLE_SIZE_T change since both are very safe.

interwq:
PR for the two things in #1472. 

@davidtgoldblatt re: the missing unlock bug not discovered by the mutex assertions. It is a tricky one, the path is only triggered if there were concurrent coalesce operations involving the extent we are trying to lock. Our current CI probably will never catch it; even our production dbg builds don't seem to encounter it (maybe partially because we enable bg thds so decay / coalescing happens way less often).

In the short term, I looked through extent.c and arena.c for similar unmatched patterns (and fortunately didn't find any). For the long term, let's discuss in our team meeting next week.

zoulasc:
Hi folks, it's been more than two weeks and this has not been merged. Gentle nudge :-) 

Here's another bug: The tls_model attribute needs to be present on all external declarations not just the definitions of the thread variables, in order to be visible during compilation. Otherwise the model is not enforced properly. You can reproduce this by nm -go *.o | grep __tsd_get_addr in the current build, then add the JEMALLOC_TLS_MODEL attribute to all the __thread variables in the headers where it is missing (tsd_malloc_thread_cleanup.h, tsd_tls.h -- why do we have an dup one?), recompile and try the nm command again. This was causing a locking issue on NetBSD/arm...

interwq:
Thanks for the patch @zoulasc!

Re: the TLS_MODEL, are you guys using `enable_initial_exec_tls`? AFAICT that's the only option we have there. I'll do a PR to fix the missing attribute.  Regarding the multiple headers / implementation, see https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/tsd.h#L241

interwq:
Never mind. I just realized that `enable_initial_exec_tls` is the default.

interwq:
re: the TLS_MODEL attribute, fix in #1482.

Issue #566, opened 2017-01-18
expose hints that will allow applications to perform runtime active defragmentation


oranagra:
Hi, I recently added a mechanism for runtime active defragmentation in redis, for which i added some code to jemalloc so that the application can decide which pointers are worth moving, and which aren't.
this hint can be merged into jemalloc so that other applications can choose to use it, and with your assistance it can maybe be improved so that defragmentation is more efficient.

first i'll describe what i already did:
after checking that the allocation is in a small bin (not large or huge), we must make sure it's not in the current run (used for new allocations), and that it doesn't reside in a fully utilized run. then we need to decide which non-full runs to evict and which non-full runs to fill.
ideally, the least amount of "work" would be to move the allocations from runs with low %run_utilization into runs with high %run_utilization, or in other words, re-allocate pointers that reside in runs with %run_utilization that is smaller than the average (i.e. the %bin_utilization).
additionally when re-allocating, we must make sure to bypass the thread cache, otherwise we'll "end up eating our own shit".

if want, you can take a quick look at the raw code i added: https://github.com/antirez/redis/pull/3720/files
(it also contains a showcase with malloc_stats output)
this was currently implemented on top of jemalloc 4.0.3, but i can PR that into the latest.

improvements:
1) when calculating the % bin utilization, it may be good to excludes full runs. so that we know if a run is less utilized than average, only among the non full runs, and not the entire bin utilization. for that all we need is to know the count of non-full runs in the bin.

2) when we re-allocate a pointer, it always moves into the non full run with the lowest address. there's a chance that this allocation will move out of a run which is going to be the currun soon, and we'll end up moving another allocation to the place we just evicted. 
```
consider these to be runs in a certain bin (sorted by address space)
[xxx-] [xxxx] [xa--] [bx--] [xxxx] [xxx-]
'x' = used, '-' = unused, 'a'/'b' used allocations that we'll want to move.
```
in this case, assuming we'll first try to consider moving 'a' before we consider moving 'b'. it would be more efficient if we didn't move the 'a' and then move 'b' into it's place, and instead move the just 'b' into the unused space in the first run.

in order to do that i think we need to sort all the non-full runs of each bin by their address, and divide that list into two groups: ones that reside in address that is lower than the relative point in the address space than a split point according to the average bin utilization (% utilization that excludes full runs), and ones above it.
then we need to move only pointers that reside in run above that address.

[...]

PR #1463, opened 2019-03-18
Add memory utilization analytics to mallctl

yinan1048576:
The analytics tool is put under experimental.utilization namespace in
mallctl.  Input is one pointer or an array of pointers and the output
is a list of memory utilization statistics.

### DIFF
--- a/include/jemalloc/internal/extent_externs.h
+++ b/include/jemalloc/internal/extent_externs.h
@@ -74,4 +74,10 @@ bool extent_merge_wrapper(tsdn_t *tsdn, arena_t *arena,
 
 bool extent_boot(void);
 
+void extent_util_stats_get(tsdn_t *tsdn, const void *ptr,
+    size_t *nfree, size_t *nregs, size_t *size);
+void extent_util_stats_verbose_get(tsdn_t *tsdn, const void *ptr,
+    size_t *nfree, size_t *nregs, size_t *size,
+    size_t *bin_nfree, size_t *bin_nregs, void **slabcur_addr);
+
 #endif /* JEMALLOC_INTERNAL_EXTENT_EXTERNS_H */
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -216,6 +216,8 @@ CTL_PROTO(stats_mapped)
 CTL_PROTO(stats_retained)
 CTL_PROTO(experimental_hooks_install)
 CTL_PROTO(experimental_hooks_remove)
+CTL_PROTO(experimental_utilization_query)
+CTL_PROTO(experimental_utilization_batch_query)
 
 #define MUTEX_STATS_CTL_PROTO_GEN(n)					\
 CTL_PROTO(stats_##n##_num_ops)						\
@@ -574,11 +576,17 @@ static const ctl_named_node_t stats_node[] = {
 
 static const ctl_named_node_t hooks_node[] = {
 	{NAME("install"),	CTL(experimental_hooks_install)},
-	{NAME("remove"),	CTL(experimental_hooks_remove)},
+	{NAME("remove"),	CTL(experimental_hooks_remove)}
+};
+
+static const ctl_named_node_t utilization_node[] = {
+	{NAME("query"),		CTL(experimental_utilization_query)},
+	{NAME("batch_query"),	CTL(experimental_utilization_batch_query)}
 };
 
 static const ctl_named_node_t experimental_node[] = {
-	{NAME("hooks"),		CHILD(named, hooks)}
+	{NAME("hooks"),		CHILD(named, hooks)},
+	{NAME("utilization"),	CHILD(named, utilization)}
 };
 
 static const ctl_named_node_t	root_node[] = {
@@ -2714,7 +2722,7 @@ static int
 prof_log_start_ctl(tsd_t *tsd, const size_t *mib, size_t miblen, void *oldp,
     size_t *oldlenp, void *newp, size_t newlen) {
 	int ret;
-	
+
 	const char *filename = NULL;
 
 	if (!config_prof) {
@@ -2726,7 +2734,7 @@ prof_log_start_ctl(tsd_t *tsd, const size_t *mib, size_t miblen, void *oldp,
 
 	if (prof_log_start(tsd_tsdn(tsd), filename)) {
 		ret = EFAULT;
-		goto label_return; 
+		goto label_return;
 	}
 
 	ret = 0;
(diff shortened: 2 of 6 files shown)


src/ctl.c
```diff
@@ -3083,3 +3085,44 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
 label_return:
 	return ret;
 }
+
+static int
+experimental_frag_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
+    void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	assert(oldp != NULL && newp != NULL);
```
interwq: We would need to handle the cases (return EINV).

src/ctl.c
```diff
@@ -3083,3 +3085,44 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+
+	/* trim *oldlenp and newlen so that:
```
interwq: Check the coding style here: https://github.com/jemalloc/jemalloc/wiki/Coding-Style

src/ctl.c
```diff
@@ -3083,3 +3085,44 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	 * (a) the larger is reduced to fit the smaller
+	 * (b) both are trimmed to a multiple of the unit size (if not yet)
+	 * rely on compiler to upgrade multiplication / division to shifts
+	 */
+	if (*oldlenp / (2 * sizeof(uint16_t)) < newlen / sizeof(void *)) {
```
interwq: As mentioned in person, let's use 32-bit.

src/ctl.c
```diff
@@ -3083,3 +3085,44 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	} else {
+		*oldlenp = newlen / sizeof(void *) * (2 * sizeof(uint16_t));
+		newlen -= newlen % sizeof(void *);
+	}
+
+	/* each pointer will output two short entries for the extent it
+	 * resides in, where the first one is the number of free regions and
+	 * the second one is the total number of regions
+	 */
+	void * const newend = newp + newlen;
```
interwq: It looks this is causing issues in Windows CI: https://ci.appveyor.com/project/jasone/jemalloc/builds/23165922/job/q69f6p9pi9ijulq1

src/ctl.c
```diff
@@ -3083,3 +3085,44 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	for (; newp < newend; newp += sizeof(void *)) {
+		const void *ptr = *(const void **)newp;
+		const extent_t *ext = iealloc(tsd_tsdn(tsd), ptr);
+		if (extent_slab_get(ext)) {
```
interwq: nit: `assert(ext != NULL);`

src/ctl.c
```diff
@@ -3083,3 +3085,44 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+		newlen -= newlen % sizeof(void *);
+	}
+
+	/* each pointer will output two short entries for the extent it
+	 * resides in, where the first one is the number of free regions and
+	 * the second one is the total number of regions
+	 */
+	void * const newend = newp + newlen;
+	for (; newp < newend; newp += sizeof(void *)) {
+		const void *ptr = *(const void **)newp;
```
interwq: The core logic (the lines in the loop) should live in the `extent.c`.

davidtgoldblatt:
I don't think it's the most important thing in the world while this is experimental, but I also have doubts that this is the right API. I think better would be to return some token that maps to an extent, and then have ways to access extent information given a token (I suspect that in the future we won't just care about fragmentation information, but also size, sampled status, etc. etc.). We should chat more generally tomorrow.

yinan1048576:
Revised the change based on our offline discussions.

- Return three instead of two pieces of information for each input pointer: number of free regions, the total number of regions, and the total size of the corresponding extent. Application can have whatever criteria they want afterwards for determining need for defragmentation. Did not go with returning the handles of the corresponding extents because we don't want to always guarantee that the underlying extents will stay before the caller tries to query the other information using the extent handles.
- The type of extent total size is size_t, so I bumped up the size of all three fields to size_t so that it's convenient for the application to fetch them.
- I explicitly terminate and raise error whenever seeing improper input sizes, rather than silently manipulating them as I did previously.
- I'm not exactly sure when the extent_t output from iealloc can be NULL. In theory the input pointers can be anything, and I assume it's possible that the iealloc output might output NULL. In such cases rather than doing assertion, I zero the output, continue the computation, but report an error to the caller in the end. The effect is that if the caller made a mistake when supplying the input pointers, which I assume was most likely due to carelessness, they know exactly if and where there are the mistake(s) afterwards, and all the other correct pointers got their output.
- The issue reported from the Windows compiler was a legitimate one that gcc should have catched. I made the correction and hopefully the Windows compiler is happy this time...

src/ctl.c
```diff
@@ -3083,3 +3085,57 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	 * (a) the number of free regions,
+	 * (b) the total number of regions, and
+	 * (c) the total size of the extent in terms of bytes.
+	 */
+	void * const newend = (const void **)newp + newlen;
+	for (; newp < newend; newp = (const void **)newp + 1) {
+		const extent_t *ext = iealloc(
+		    tsd_tsdn(tsd), *(const void **)newp);
+		if (unlikely(ext == NULL)) {
+			memset(oldp, 0, sizeof(size_t) * 3);
```
interwq: Not really need to zero it out. As long as we return an error code, the user is not supposed to look into the values anyway.

src/ctl.c
```diff
@@ -3083,3 +3085,57 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+
+	/*
+	 * Each pointer will output three entries for the extent it resides in:
+	 * (a) the number of free regions,
+	 * (b) the total number of regions, and
+	 * (c) the total size of the extent in terms of bytes.
+	 */
+	void * const newend = (const void **)newp + newlen;
+	for (; newp < newend; newp = (const void **)newp + 1) {
+		const extent_t *ext = iealloc(
```
interwq: Let’s move the lookup into the extent info get function as well, doesn’t look like we need the extent_t here other than as the input to that function.

src/ctl.c
```diff
@@ -3083,3 +3085,57 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+		    tsd_tsdn(tsd), *(const void **)newp);
+		if (unlikely(ext == NULL)) {
+			memset(oldp, 0, sizeof(size_t) * 3);
+			ret = EFAULT;
+		} else {
+			extent_frag_get(ext, (size_t *)oldp,
+			    (size_t *)oldp + 1, (size_t *)oldp + 2);
+		}
+		oldp = (size_t *)oldp + 3;
```
interwq: Let’s make a size_t *ptr and adjust that. Better keep the oldp intact.

include/jemalloc/internal/extent_inlines.h
```diff
@@ -151,6 +151,19 @@ extent_bsize_get(const extent_t *extent) {
 	return extent->e_bsize;
 }
 
+static inline void
+extent_frag_get(const extent_t *extent,
```
interwq: Nit: not sure about the function name

yinan1048576:
Made a couple of modifications according to feedback and offline discussions.

- Still decide to zero out the output fields for invalid input pointers. I've put the reasoning as comment. One more comment (for the comment): valid pointers (i.e. those returned from previous malloc calls) will surely give non-NULL extent, but invalid pointers may or may not always give NULL extent depending on the underlying implementation, and some "maliciously" invalid pointers may also directly lead to segfault. In my reasoning, for simplicity, I'm assuming: (a) the caller has good intention and provide us pointers they truly want to get fragmentation info about, and (b) NULL extent is a sufficient and necessary indication of invalid pointers.
- Encapsulate the extent looking-up and extent reading into a single function and put it in extent.c, with declaration put in extent_extern.h. Again, not sure whether the function name is ideal - I'm calling it extent_frag_stats, which sounds slightly better than extent_frag to me.
- Define additional local variables to make the code more intuitive and readable.

include/jemalloc/internal/extent_externs.h
```diff
@@ -74,4 +74,8 @@ bool extent_merge_wrapper(tsdn_t *tsdn, arena_t *arena,
 
 bool extent_boot(void);
 
+int
+extent_frag_stats(tsdn_t *tsdn, const void *ptr,
```
interwq: nit: put "int" on same line, use up to 80 chars per line.

re: naming -- maybe, `extent_util_stats_get`? We already use the term utilization in the stats world.

src/ctl.c
```diff
@@ -3083,3 +3085,76 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	/*
+	 * Make sure newlen is an exact multiple of sizeof(const void *),
+	 * and if so, reduce it to the multiple i.e. count of input pointers.
+	 * Rely on compiler to convert multiplication / division to shifts.
+	 */
+	if (newlen != newlen / sizeof(const void *) * sizeof(const void *)) {
+		ret = EINVAL;
+		goto label_return;
+	}
+	newlen /= sizeof(const void *);
```
interwq: nit: Similarly, keep the original input args, feel free to define a new one with explicit naming.

src/ctl.c
```diff
@@ -3083,3 +3085,76 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
 label_return:
 	return ret;
 }
+
```
interwq: Suggestion: add some example of the return format / layout, something like. 

    /*
     * input[0]:  pointer_to_query  |  output[0]: extent_n_free_items
     *                              |  output[1]: ....
      ...

Also a simple description regarding what the 4 args do.

src/extent.c
```diff
@@ -2279,3 +2279,23 @@ extent_boot(void) {
...
 	return false;
 }
+
+int
+extent_frag_stats(tsdn_t *tsdn, const void *ptr,
+    size_t *nfree, size_t *nregs, size_t *size) {
+	const extent_t *extent = iealloc(tsdn, ptr);
+	if (unlikely(extent == NULL)) {
+		*nfree = *nregs = *size = 0U;
+		return EFAULT;
```
interwq: nit: this can simply return true; the EFAULT etc is for user-facing return codes. Internal functions don't really need to use them.

src/extent.c
```diff
@@ -2279,3 +2279,23 @@ extent_boot(void) {
...
+	} else {
+		if (extent_slab_get(extent)) {
+			*nfree = extent_nfree_get(extent);
+			*nregs = bin_infos[extent_szind_get(extent)].nregs;
+		} else {
+			*nfree = 0U;
+			*nregs = 1U;
+		}
+		*size = extent_size_get(extent);
+		return 0;
```
interwq: nit: add `assert(*size >= nfree * nregs);`

src/ctl.c
```diff
@@ -578,7 +579,8 @@ static const ctl_named_node_t hooks_node[] = {
 };
 
 static const ctl_named_node_t experimental_node[] = {
-	{NAME("hooks"),		CHILD(named, hooks)}
+	{NAME("hooks"),		CHILD(named, hooks)},
+	{NAME("frag"),		CTL(experimental_frag)},
```
interwq: Naming is hard --`frag` feels a bit strange.  Let's always use fragmentation?

yinan1048576:
- Improve formatting / naming / readability.
- Do not return `EFAULT` when extent is `NULL`. It doesn't affect the caller's ability to detect it at all because the output fields are zeroed out.
- Improve comments.

yinan1048576:
Added two assertions:

- `assert(*nfree <= *nregs);`
- `assert(*nfree * extent_usize_get(extent) <= *size);`

As discussed offline, the two `<=` should always be `<`, but we allow `==` since it might not be a big issue even if it somehow appears.

src/ctl.c
```diff
@@ -3083,3 +3085,97 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+ * one for the metadata and one for the actual content.  Other types may use
+ * even more allocations.  When inquiring about fragmentation statistics, the
+ * caller often wants to examine into all such allocations, especially internal
+ * one(s), rather than just the topmost one.  The issue comes when some
+ * implementations do certain optimizations to reduce/aggregate some internal
+ * allocations, e.g. putting short strings directly into the metadata, and such
+ * decisions are not known to the caller.  Therefore, we permit pointers to
+ * memory usages that may not be returned by previous malloc calls, and we
+ * provide the caller a convenient way to identify such cases.
+ */
```
interwq: nit: let's also add a simple example for the actual defrag work, e.g. disable tcache; free / alloc; re-enable tcache.

src/ctl.c
```diff
@@ -3083,3 +3085,97 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+static int
+experimental_fragmentation_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
```
interwq: how about `experimental_fragmentation_query`, or maybe utilization instead of fragmentation?

src/ctl.c
```diff
@@ -3083,3 +3085,97 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	 * and that the multiple equals the count of input pointers.
+	 */
+	if (*oldlenp != len * sizeof(size_t) * 3) {
+		ret = EINVAL;
+		goto label_return;
+	}
+
+	const void **ptrs = (const void **)newp;
+	const void ** const ptrs_end = ptrs + len;
+	size_t *frag_stats = (size_t *)oldp;
```
interwq: nit: let's name this `util_stats`, to match the extent_util call below.

test/unit/mallctl.c
```diff
@@ -853,6 +853,66 @@ TEST_BEGIN(test_hooks_exhaustion) {
...
+	TEST_FRAG_EINVAL(out, &out_sz, in, in_sz,
+	    "*oldlenp and newlen do not match");
+	in_sz += sizeof(const void *);
+#undef TEST_FRAG_EINVAL
+
+	/* Examine output for valid call */
+	out_sz_ref = out_sz /= 2;
+	in_sz /= 2;
+	assert_d_eq(mallctl(
+			"experimental.fragmentation", out, &out_sz, in, in_sz),
```
interwq: nit: spacing is off. Let's always do 4 spaces for second line, regardless of the nested calls.

yinan1048576:

yinan1048576:
Also changed macro naming in tests.

src/ctl.c
```diff
@@ -3083,3 +3085,114 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	if (oldp == NULL || oldlenp == NULL || newp == NULL || newlen == 0) {
+		ret = EINVAL;
+		goto label_return;
+	}
+
+	/*
+	 * Make sure newlen is an exact multiple of sizeof(const void *).
+	 * Rely on compiler to convert multiplication / division to shifts.
+	 */
+	const size_t len = newlen / sizeof(const void *);
```
interwq: Let's move this to before the first branch, and then we can merge the two branches below into the first one -- note the body of the branch is the same; so cleaner to group them.

src/ctl.c
```diff
@@ -3083,3 +3085,114 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	/*
+	 * Make sure that *oldlenp is an exact multiple of sizeof(size_t) * 3,
+	 * and that the multiple equals the count of input pointers.
+	 */
+	if (*oldlenp != len * sizeof(size_t) * 3) {
+		ret = EINVAL;
+		goto label_return;
+	}
+
+	const void **ptrs = (const void **)newp;
```
interwq: nit: I slightly prefer having an integer index for the loop here, i.e.

    typedef struct {
        size_t nfree;
        ...
    } extent_util_stats_t;
    extent_util_stats_t *output_stats = (extent_util_stats_t *)oldp;
    void **input_ptrs = (void **)newp;
    for (unsigned i = 0; i < len; i++) {
        extent_util_stats_get(tsd, &input_ptrs[i], &output_stats[i].nfree, ...);
    }

yinan1048576:
Revised the change -

- Split the `mallctl` API into two: `experimental.utilization.query` for single pointer query, and `experimental.utilization.batch_query` for batch querying.
- For single pointer query, return an additional field at the beginning: the address of `slabcur`, which is the place of a potential reallocation. It's up to the application how to use it to determine fragmentation. In the implementation from Redis, fragmentation is set to be true whenever this address of potential reallocation is not equal to the extent the input pointer resides in, which might be an overly strict criteria. If really desired, the application can achieve the same effect by examining whether the deviation of the input pointer from our returned `slabcur_addr` is below a very low threshold (i.e. the size of a single extent), but it's also possible to define less strict criteria by loosening the deviation threshold.
- Do not return `slabcur_addr` for batch querying, because it'd be quickly changing throughout the defragmentation process and thus returning it for all pointers at query time may not make that much sense for determining defragmentation. The `nfree` field can also change throughout the defragmentation process, but it is relatively less volatile since it's always local to the extent the pointer resides in, and it has a nicer property: in the middle of the defragmentation process, at the time a particular pointer is defragmented, if `nfree` decreased from what it was at querying time, then it means the defragmentation process till now favored reallocating to this extent so it wasn't a problem from the beginning; if `nfree` increased, then that means the defragmentation process till now favored reallocating away from this extent so it has already been a problem earlier. In both cases the defragmentation determination at the querying time wouldn't be affected if we had returned the `nfree` at this later time.
- Explicitly define `struct`s holding the output fields so as to improve readability. The `struct`s are not exposed out and are only for internal use in `src/ctl.c`. I didn't make use of the `struct`s in the unit tests because I'd like to mimic what application would do.
- Improve the comment and code readability.

src/extent.c
```diff
@@ -2279,3 +2279,42 @@ extent_boot(void) {
...
+	assert(*nfree * extent_usize_get(extent) <= *size);
+
+	if (slabcur_addr != NULL) {
+		const arena_t *arena = extent_arena_get(extent);
+		assert(arena != NULL);
+		const unsigned binshard = extent_binshard_get(extent);
+		bin_t *bin = &arena->bins[szind].bin_shards[binshard];
+
+		malloc_mutex_lock(tsdn, &bin->lock);
+		const extent_t *cur = bin->slabcur;
```
interwq: nit: no need for the `cur` var.

src/extent.c
```diff
@@ -2279,3 +2279,42 @@ extent_boot(void) {
 
 	return false;
 }
+
+void
+extent_util_stats_get(tsdn_t *tsdn, const void *ptr, size_t *nfree,
+    size_t *nregs, size_t *size, void **slabcur_addr) {
+	*nfree = *nregs = *size = 0;
```
interwq: nit: We try avoiding double initialization, mainly for the purpose of explicitly setting the values, instead of relying on the default value to cover multiple cases (more error prone). Let's remove this line and the branch below, and see the comments below for changes.

src/extent.c
```diff
@@ -2279,3 +2279,42 @@ extent_boot(void) {
...
+	if (slabcur_addr != NULL) {
+		*slabcur_addr = NULL;
+	}
+
+	const extent_t *extent = iealloc(tsdn, ptr);
+	if (unlikely(extent == NULL)) {
+		return;
```
interwq: ... add `*nfree = *nregs = *size = 0;`, then `goto label_return;`

src/extent.c
```diff
@@ -2279,3 +2279,42 @@ extent_boot(void) {
...
+	}
+
+	*size = extent_size_get(extent);
+	if (!extent_slab_get(extent)) {
+		*nregs = 1;
+		return;
```
interwq: ... add `*nfree = 0;` and then `goto label_return`.

src/extent.c
```diff
@@ -2279,3 +2279,42 @@ extent_boot(void) {
...
+		const unsigned binshard = extent_binshard_get(extent);
+		bin_t *bin = &arena->bins[szind].bin_shards[binshard];
+
+		malloc_mutex_lock(tsdn, &bin->lock);
+		const extent_t *cur = bin->slabcur;
+		*slabcur_addr = extent_addr_get(cur);
+		assert(*slabcur_addr != NULL);
+		malloc_mutex_unlock(tsdn, &bin->lock);
+	}
+}
```
interwq: ... and in the end, have a `label_return` (after the normal return) which does
```
        if (slabcur_addr != NULL) {
		*slabcur_addr = NULL;
	}
        return;
```

With this refactor, the values are set exactly once for each case, and it may enable the compiler to warn us if some values are missing initialization.

yinan1048576:
Revised the following:

- Added two additional output fields `bin_nfree` and `bin_nregs` for single pointer query. This is to ensure that this new API is fully compatible with redis usage case.
- Improved code quality & readability, comment and unit tests.

FYI in the latest redis usage (see https://github.com/antirez/redis/blob/167705519b5f21b2aa3954527be15dc653131221/deps/jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h#L219 for the implementation and https://github.com/antirez/redis/blob/22e9321c3ee238f498980cf18781e10caaa01589/src/defrag.c#L57 for the caller), the fragmentation is triggered at the following condition:

```
je_get_defrag_hint(ptr, &bin_util, &run_util)
    && run_util <= bin_util
    && run_util < 1<<16
```

which is equivalent to:

```
ptr >= SLABCUR_READ(out) && ptr < SLABCUR_READ(out) + SIZE_READ(out)
    && NFREE_READ(out) * BIN_NREGS_READ(out) >= NREGS_READ(out) * BIN_NFREE_READ(out)
    && NFREE_READ(out) > 0
```

assuming `out` points to the `mallctl` output.

yinan1048576:
Tests fail when `--disable-stats` is set, which is expected. Fixing it...

yinan1048576:
Fixed the failure. It also makes me realize that the redis implementation assumed stats are turned on, and might crash otherwise since there can be a "divide by zero".

src/ctl.c
```diff
@@ -3083,3 +3091,224 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+ * Note that "(e)" and "(f)" are only available when stats are enabled;
+ * otherwise both are set zero.
+ *
+ * This API is mainly intended for small class allocations, where extents are
+ * used as slab.
+ *
+ * In case of large class allocations, "(a)" will be NULL, and "(e)" and "(f)"
+ * will be zero.  The other three fields will be properly set though the values
+ * are trivial: "(b)" will be 0, "(c)" will be 1, and "(d)" will be the original
+ * allocation size rounded up to the nearest size class.
```
interwq: nit: the term here is `usable size`.

src/ctl.c
```diff
@@ -3083,3 +3091,224 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+ *
+ * In the rare case where there's no associated extent found for the input
+ * pointer, the function zeros out all output fields and return.  Please refer
+ * to the comment for experimental_utilization_batch_query_ctl to understand the
+ * motivation from C++.
+ */
+static int
+experimental_utilization_query_ctl(tsd_t *tsd, const size_t *mib,
+    size_t miblen, void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret = 0;
```
interwq: nit: similarly, try avoiding double initialization. Here we only need to define `ret`. And on the success path, do `ret = 0` before label_return.

src/ctl.c
```diff
@@ -3083,3 +3091,224 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+	void *ptr = NULL;
```
interwq: nit: no need to initialize.

src/ctl.c
```diff
@@ -3083,3 +3091,224 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+
+	assert(sizeof(extent_util_stats_verbose_t)
+	    == sizeof(void *) + sizeof(size_t) * 5);
+
+	WRITE(ptr, void *);
```
interwq: nit: let's move this to after the input check below; and also check for newp in the branch.

src/ctl.c
```diff
@@ -3083,3 +3091,224 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+ * type size_t for each input pointer about the extent it resides in:
+ *
+ * (a) number of free regions in the extent,
+ * (b) number of regions in the extent, and
+ * (c) size of the extent in terms of bytes.
+ *
+ * This API is mainly intended for small class allocations, where extents are
+ * used as slab.  In case of large class allocations, the outputs are trivial:
+ * "(a)" will be 0, "(b)" will be 1, and "(c)" will be the original allocation
+ * size rounded up to the nearest size class.
```
interwq: nit: term usable size too.

src/ctl.c
```diff
@@ -3083,3 +3091,224 @@ experimental_hooks_remove_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+ * implementations do certain optimizations to reduce/aggregate some internal
+ * allocations, e.g. putting short strings directly into the metadata, and such
+ * decisions are not known to the caller.  Therefore, we permit pointers to
+ * memory usages that may not be returned by previous malloc calls, and we
+ * provide the caller a convenient way to identify such cases.
+ */
+static int
+experimental_utilization_batch_query_ctl(tsd_t *tsd, const size_t *mib,
+    size_t miblen, void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret = 0;
```
interwq: similarly, not double initialize ret.

yinan1048576:

test/unit/mallctl.c
```diff
@@ -883,5 +1075,7 @@ main(void) {
 	    test_arenas_lookup,
 	    test_stats_arenas,
 	    test_hooks,
-	    test_hooks_exhaustion);
+	    test_hooks_exhaustion,
+	    test_utilization_query,
```
interwq: The util query tests deserve their own tests I feel. Let's move them to a `util_query.c` in a follow up diff.

zvi-code:
I plan to issue PR [link TBD] to [Valkey](https://github.com/valkey-io/valkey), that replaces the existing patch [valkey](https://github.com/valkey-io/valkey/blob/71dd85dc5a89a4e07d837efa14937990b01e317f/deps/jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h#L341) with this API.  One peace of information that current algorithm uses and is not included in the API is the number of fullslabs in the bin. The motivation for having the number of fullslabs is to improve the convergence of defrag(the fullslabs are shifting the average and may cause stagnation in defrag process). 
What is the best alternative way I can fetch this information with one of the existing API? 

Probably this API will do the work:
> stats.arenas.<i>.bins.<j>.curslabs (size_t) r- [--enable-stats]
> stats.arenas.<i>.bins.<j>.nonfull_slabs (size_t) r- [--enable-stats]

zuiderkwast:
It's interesting that there are no links between this PR and #566. Until today I was not aware of this PR.

PR #1465, opened 2019-03-23
Add a feature to do redzone checks for small sampled allocations

davidtgoldblatt:


djwatson:
lgtm other than the contbuild failure

davidtgoldblatt:
Test failure was a silly config thing. Holding off on merging until the release, though.

include/jemalloc/internal/safety_check.h
```diff
@@ -0,0 +1,26 @@
+#ifndef JEMALLOC_INTERNAL_SAFETY_CHECK_H
+#define JEMALLOC_INTERNAL_SAFETY_CHECK_H
+
+void safety_check_fail(const char *format, ...);
+/* Can set to NULL for a default. */
+void safety_check_set_abort(void (*abort_fn)());
+
+JEMALLOC_ALWAYS_INLINE void
+safety_check_set_redzone(void *ptr, size_t usize, size_t bumped_usize) {
```
interwq: I'm worried that this might increase the memory overhead from profiling. IIRC, 3 of the 4 pages may not be touched at all. Maybe let's only populate and check the first page?

interwq: I missed that the red zone is only 32B. Never mind.

interwq:
LGTM! @davidtgoldblatt can you try rebasing? Did you benchmark, or should I try that first?

davidtgoldblatt:
I tried microbench both ways (both with lg_prof_sample at the default and at 0), and any difference was in the noise. I didn't try large-scale canaries, though.

PR #1469, opened 2019-03-28
 Allow low values of oversize_threshold to disable the feature.

interwq:
We should allow a way to easily disable the feature (e.g. not reserving the
arena id at all).


interwq:
We used to set the oversize_threshold to 0 for a few months, before enabling it by default. This change allows using 0 to disable the feature through the runtime option.

PR #1472, opened 2019-03-29
By @zoulasc: fix a missing unlock bug and a warning

interwq:
Both issues were discovered by @zoulasc, in #1460

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1252,7 +1252,7 @@ malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS]) {
 			    false)
 			CONF_HANDLE_SIZE_T(opt_lg_extent_max_active_fit,
 			    "lg_extent_max_active_fit", 0,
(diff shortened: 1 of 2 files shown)


src/jemalloc.c
```diff
@@ -1252,7 +1252,7 @@ malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS]) {
 			    false)
 			CONF_HANDLE_SIZE_T(opt_lg_extent_max_active_fit,
 			    "lg_extent_max_active_fit", 0,
-			    (sizeof(size_t) << 3), yes, yes, false)
+			    (sizeof(size_t) << 3), no, yes, false)
```
davidtgoldblatt: I hate these raw "yes"/"no" things a long ways after the macro definition. I wish it were check_min and check_max or something.

interwq: I find myself constantly looking to figure out which is which too. A quick refactoring task for our new team member maybe ;)

PR #1473, opened 2019-04-01
Update Changelog for 5.2.0.

interwq:


interwq:
Note that experimental / undocumented features are not included in the ChangeLog.

ChangeLog
```diff
@@ -4,6 +4,110 @@ brevity.  Much more detail can be found in the git revision history:
...
+  setting the number of background threads.  In addition, peak / spike memory
+  usage is improved with certain allocation patterns.  As usual, the release and
+  prior dev versions have gone through large-scale production testing.
+
+  New features:
+  - Implement oversize_threshold, which uses a dedicated arena for allocations
+    crossing the specified threshold to reduce fragmentation.  (@interwq)
+  - Add extents usage information to stats.  (@tyleretzel)
+  - Log time information for sampled allocations.  (@tyleretzel)
+  - Support 0 size in sdallocx.  (@djwatson)
```
davidtgoldblatt: Is this a documented guarantee (i.e. an ABI strengthening)? I thought we were doing it to support our own operator delete overloads, but not otherwise promising anything.

ChangeLog
```diff
@@ -4,6 +4,110 @@ brevity.  Much more detail can be found in the git revision history:
...
+  - Lower the default number of background threads to 4 (when the feature
+    is enabled).  (@interwq)
+  - Optimize the trylock spin wait.  (@djwatson)
+  - Use arena index for arena-matching checks.  (@interwq)
+  - Avoid forced decay on thread termination when using background threads.
+    (@interwq)
+  - Disable muzzy decay by default.  (@djwatson, @interwq)
+  - Only initialize libgcc unwinder when profiling is enabled.  (@paravoid,
+    @interwq)
+  - Eagerly purge extents crossing the oversize_threshold when enabled.
```
davidtgoldblatt: Might as well mention this alongside oversize_threshold, there was never a point-numbered version that didn't include both features.

ChangeLog
interwq: The manual does say `The minimum valid input size is the original requested size of the allocation`, to me it sounds a requirement to support 0 (as it can be the original size of malloc). No strong opinion though. Do you think we should not mention it here?

interwq:
cc: @jasone, @djwatson in case you guys want to take a look too.

ChangeLog
davidtgoldblatt: That sounds convincing to me.

jasone: I was alarmed when I read this particular item as well.  The `*allocx` API has always been intended as providing precisely the semantics needed by a well written application.  IMO the memory allocator shouldn't be universally burdened with sanity-checking bogus inputs.  If our sized deallocation support for C++ needs this guarantee, fine, but I would really rather not add this to the `sdallocx` spec.

PR #1474, opened 2019-04-01
Tweak the wording about oversize_threshold.

interwq:


interwq:
Makes sense. Just changed.

Issue #1477, opened 2019-04-03
[  188s] test_prof_log_many_traces:test/unit/prof_log.c:133: Failed assertion: (jet_prof_log_bt_count()) == (8) --> 16 != 8: Wrong number of backtraces given sample workload since 5.2.0


marxin:
I see the failure on multiple targets (aarch64, ppc64le) and also on an older x86_64 system.
Is there a way how to print all the backtraces collected in ```prof.c```. That will help to identify why is the number ```16``` and not ```8``` expected. Thanks.

marxin:
[jemalloc-gdb-jet_prof_backtrace-break-point.txt](https://github.com/jemalloc/jemalloc/files/3037988/jemalloc-gdb-jet_prof_backtrace-break-point.txt)

There's a log file from gdb, where I print ```bt``` in all invocations of ```jet_prof_backtrace``` function. Hope it will help.

gnzlbg:
We are seeing a similar issue when using the dev branch in our CI: 

```
test_prof_log_many_traces:../jemalloc/test/unit/prof_log.c:133: Failed assertion: (jet_prof_log_bt_count()) == (8) --> 9 != 8: Wrong number of backtraces given sample workload
```

We are compiling jemalloc like this on `x86_64-unknown-linux-gnu`: https://travis-ci.com/gnzlbg/jemallocator/jobs/187725813#L5489

interwq:
Thanks for reporting this @marxin @gnzlbg. This is a new test added in 5.2; so probably just an issue on the test itself.

It looks to me that there are more than 8 different traces, namely these are all the direct called from `test_prof_log_many_traces ()`

```
#5  0x0000000000404586 in f2 () at test/unit/prof_log.c:105
#5  0x000000000040451f in f2 () at test/unit/prof_log.c:105

#7  0x000000000040458e in f2 () at test/unit/prof_log.c:106
#7  0x0000000000404527 in f2 () at test/unit/prof_log.c:106

#7  0x000000000040450e in f1 () at test/unit/prof_log.c:101
#7  0x0000000000404575 in f1 () at test/unit/prof_log.c:101

#5  0x00000000004044f1 in f1 () at test/unit/prof_log.c:99
#5  0x0000000000404558 in f1 () at test/unit/prof_log.c:99

#6  f1 () at test/unit/prof_log.c:100

#5  0x000000000040459f in f3 () at test/unit/prof_log.c:94
[...]

PR #1478, opened 2019-04-03
Fix test/unit/prof_log

interwq:
The unwinder may produce traces more than expected.  Instead verify the lower
bound only.

Corresponding to #1477


interwq:
@djwatson : could certainly use your expertise on the unwinder. Details in #1477

djwatson:
In my experience these types of tests don't work very well across compilers/platforms.  I'd suggest maybe just disabling this one except in CI, since you can control the CI platform.

interwq:
Details in #1477. The unexpected traces are from loop unrolling.

PR #1480, opened 2019-04-09
Separate tests for extent utilization API

yinan1048576:
As title.


interwq:
nit: @yinan1048576 you can simply only have the title in the commit msg, i.e. no need for the "as title" in the msg.

PR #1483, opened 2019-04-16
Fix crash on NULL iealloc return  

QwertyZW:
Our only extra configure option is `--enable-prof`

This crash occurs once about over thousands of hours of runtime. This is the patch that we're going with to band-aid it in-house

iealloc returns NULL which is then passed down to atomic_store_p as 0x40 (via the &extent->..)

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -47,7 +47,11 @@ arena_prof_tctx_get(tsdn_t *tsdn, const void *ptr, alloc_ctx_t *alloc_ctx) {
 		}
 	} else {
 		if (unlikely(!alloc_ctx->slab)) {
-			return large_prof_tctx_get(tsdn, iealloc(tsdn, ptr));
+			extent_t *extent = iealloc(tsdn, ptr);
+			if (unlikely(extent == NULL)) {
+				return (prof_tctx_t *)(uintptr_t)1U;
+			}
+			return large_prof_tctx_get(tsdn, extent);
 		}
 	}
 	return (prof_tctx_t *)(uintptr_t)1U;
@@ -67,7 +71,11 @@ arena_prof_tctx_set(tsdn_t *tsdn, const void *ptr, size_t usize,
 		}
 	} else {
 		if (unlikely(!alloc_ctx->slab)) {
-			large_prof_tctx_set(tsdn, iealloc(tsdn, ptr), tctx);
+			extent_t *extent = iealloc(tsdn, ptr);
+			if (unlikely(extent == NULL)) {
+				return;
+			}
+			large_prof_tctx_set(tsdn, extent, tctx);
 		}
 	}
 }


davidtgoldblatt:
The most likely cause of this behavior is application-induced corruption; have you run your binary under ASAN/valgrind/jemalloc debug mode? In any case, if this is a jemalloc bug, this isn't the right fix for it; iealloc should never return null for a valid pointer.

Also +cc @interwq; CI not triggering again.

interwq:
Like David mentioned, heap corruption is likely the root cause here. Closing but feel free to ask if anything.

Garfielt:
Hi, @QwertyZW ，we got this crash too, on our online service with several and random hours then it happen，do you have any further progress to share，thanks?

yinan1048576:
Can anyone share a stack trace?

Garfielt:
@yinan1048576 
```
Program terminated with signal 11, Segmentation fault.
#0  atomic_load_p (mo=atomic_memory_order_acquire, a=0x48) at include/jemalloc/internal/atomic.h:62
62      JEMALLOC_GENERATE_ATOMICS(void *, p, LG_SIZEOF_PTR)
Missing separate debuginfos, use: debuginfo-install glibc-2.17-292.el7.x86_64 libgcc-4.8.5-39.el7.x86_64 sqlite-3.7.17-8.el7.x86_64
(gdb) bt
#0  atomic_load_p (mo=atomic_memory_order_acquire, a=0x48) at include/jemalloc/internal/atomic.h:62
#1  extent_prof_tctx_get (extent=0x0) at include/jemalloc/internal/extent_inlines.h:185
#2  je_large_prof_tctx_get (tsdn=tsdn@entry=0x7fa4f90e0eb0, extent=0x0) at src/large.c:374
#3  0x000000000043c95f in arena_prof_tctx_get (alloc_ctx=<synthetic pointer>, ptr=0x7fa4f90fbc20, tsdn=0x7fa4f90e0eb0) at include/jemalloc/internal/arena_inlines_b.h:50
#4  prof_tctx_get (alloc_ctx=<synthetic pointer>, ptr=0x7fa4f90fbc20, tsdn=0x7fa4f90e0eb0) at include/jemalloc/internal/prof_inlines_b.h:44
#5  prof_free (alloc_ctx=<synthetic pointer>, usize=<optimized out>, ptr=0x7fa4f90fbc20, tsd=0x7fa4f90e0eb0) at include/jemalloc/internal/prof_inlines_b.h:239
#6  ifree (slow_path=false, tcache=0x7fa4f90e10a0, ptr=0x7fa4f90fbc20, tsd=0x7fa4f90e0eb0) at src/jemalloc.c:2497
#7  je_free_default (ptr=0x7fa4f90fbc20) at src/jemalloc.c:2710
```

Garfielt:
Hi @yinan1048576 , I saw you commited some code about profiling, are these changes related to our question?

yinan1048576:
@Garfielt the commits are not related. The same crash should most likely still be there, though the stack trace will be different. The stack trace you provided might be a result of double free on the application side. I suggest that you run some address sanitizer to check.

Garfielt:
@yinan1048576 OK, thanks for your advice!

PR #1486, opened 2019-04-18
Add nonfull_slabs to bin_stats_t.

doronrk:
When config_stats is enabled track the size of bin->slabs_nonfull in
the new nonfull_slabs counter in bin_stats_t. This metric should be
useful for establishing an upper ceiling on the savings possible by
meshing.


interwq:
One more nit: let's add a small test case similar to https://github.com/jemalloc/jemalloc/blob/dev/test/unit/stats.c#L310

PR #1487, opened 2019-04-18
Improve size class header

yinan1048576:
Mainly fixing typos.  The only non-trivial change is in the
computation for SC_NPSIZES, though the result wouldn't be any
different when SC_NGROUP = 4 as is always the case at the moment.


yinan1048576:
Only change `2 * SC_NGROUP` to `SC_LG_NGROUP * SC_NGROUP` and leave everything else unchanged.

(I removed the `#if SC_LG_TINY_MIN == 0` check because it's defined to be `3` just two lines above it, but I get the point that it should still be there for the sake of the rigor of the code, so keep it.)

PR #1488, opened 2019-04-19
configure.ac: Add an option to disable doc

ffontaine:
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


interwq:
Thanks for the patch!

PR #1489, opened 2019-04-19
meshing: bootstrapping and some stats/ctl

doronrk:
These 3 commits add features that are primarily all behind the new `opt_mesh` option which defaults to false. Overhead for programs with `opt_mesh:false` should hopefully be little to none: 1 NULL ptr in arena_t, as well as some branches based off of the value of `opt_mesh` on removal/insertion into slabs->nonfull and arena_dalloc_bin_locked_impl in src/arena.c

The end result of this stack of commits is that programs with `opt_mesh` will have stats available on the number of nonfull slabs with each possible 8 bit mesh shape. This should be useful in assessing the possible benefits of a mesh approach based on an exhaustive search of slabs with nregs <= 8.

I'm not a fan of the pervasive `(1 << 8)` everywhere that needs to logically represent the number of possible 8 bit mesh shapes, so I would definitely welcome feedback on how to improve that.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -110,6 +110,7 @@ C_SRCS := $(srcroot)src/jemalloc.c \
 	$(srcroot)src/large.c \
 	$(srcroot)src/log.c \
 	$(srcroot)src/malloc_io.c \
+	$(srcroot)src/mesh.c \
 	$(srcroot)src/mutex.c \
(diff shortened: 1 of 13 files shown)


doronrk:
need to fix bugs revealed by CI.

PR #1490, opened 2019-04-22
Fix typo derived from tcmalloc's pprof

prehistoric-penguin:
The same pr is submitted into gperftools:

https://github.com/gperftools/gperftools/pull/1105


interwq:
Thanks for the patch!

PR #1493, opened 2019-04-22
Mesh5

doronrk:
These 3 commits add features that are primarily all behind the new opt_mesh option which defaults to false. Overhead for programs with opt_mesh:false should hopefully be little to none: 1 NULL ptr in arena_t, as well as some branches based off of the value of opt_mesh on removal/insertion into slabs->nonfull and arena_dalloc_bin_locked_impl in src/arena.c

The end result of this stack of commits is that programs with opt_mesh will have stats available on the number of nonfull slabs with each possible 8 bit mesh shape. This should be useful in assessing the possible benefits of a mesh approach based on an exhaustive search of slabs with nregs <= 8.

I'm not a fan of the pervasive (1 << 8) everywhere that needs to logically represent the number of possible 8 bit mesh shapes, so I would definitely welcome feedback on how to improve that.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -110,6 +110,7 @@ C_SRCS := $(srcroot)src/jemalloc.c \
 	$(srcroot)src/large.c \
 	$(srcroot)src/log.c \
 	$(srcroot)src/malloc_io.c \
+	$(srcroot)src/mesh.c \
 	$(srcroot)src/mutex.c \
 	$(srcroot)src/mutex_pool.c \
 	$(srcroot)src/nstime.c \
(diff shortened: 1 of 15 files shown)


emeryberger:
Hi @doronrk - would love to know the status of this!

interwq:
@emeryberger -- unfortunately Doron left the team a few months ago. Currently we don't have people working on the project; however we are still looking for engineers to take over.

We did some profiling in production regarding the potential mesh-able pages. The vast majority of our binaries have ~5-10% of fragmentation overall. With the production stats, we were hoping for ~1% or up to 2% of overall RSS savings (which is a big deal given the scale). However the added overhead on CPU / latency is pretty hard to estimate.

We'll update if any news arise.

interwq:
Closing for now.

PR #1498, opened 2019-05-01
Add confirm_conf option

yinan1048576:
If the confirm_conf option is set, when the program starts, each of
the four malloc_conf strings will be printed, and each option will
be printed when being set.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -43,6 +43,8 @@ bool	opt_abort_conf =
     false
 #endif
     ;
+/* Intentionally default off, even with debug builds. */
+bool	opt_confirm_conf = false;
 const char	*opt_junk =
 #if (defined(JEMALLOC_DEBUG) && defined(JEMALLOC_FILL))
     "true"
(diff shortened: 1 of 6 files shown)


src/jemalloc.c
```diff
@@ -1119,11 +1143,30 @@ malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS]) {
 				    sizeof(o)-1;			\
 				strncpy(o, v, cpylen);			\
 				o[cpylen] = '\0';			\
+				CONTINUE;				\
+			}
+
+#define CONTINUE continue
```
davidtgoldblatt: Why do we need this? This seems weird.

yinan1048576:
Modified the `#define CONTINUE` for the `search_confirm_conf_only` case.

davidtgoldblatt:
Sorry, what I mean is, why do we need to do this `define CONTINUE` thing at all? I mean in two senses:
- Why does the continue logic change at all?
- Why does it need to be done with a macro instead of just writing `continue` or not (as appropriate).

yinan1048576:
The reason why I'm setting up things like this is:

- In the first call to `malloc_conf_init`, I'm only doing two things: (a) getting the `confirm_conf` option and (b) getting the four `opts`s (and having them cached so that there's no need to make system calls to get them again the next time.
- In the second call to `malloc_conf_init`, I'm doing exactly the same as what would be done before this change only except that I am additionally (a) printing out the four `opts`s and (b) printing out each option when being set, if the `opt_confirm_conf` is set to be true (as determined in the first call).
- I'm calling the exact same `CONF_HANDLE_BOOL` macro in both the first call (for handling `opt_confirm_conf`) and the second call (for handling other boolean options).
- Not all options are handled via macros; some are directly handled via code e.g. `metadata_thp`. The only common part in all the handling code is the final `continue`.

Therefore, I'm deciding to "overwrite" the `continue`, so that each option when being set can be printed out, if `opt_confirm_conf` is set. Moreover, the "overwrite" should be different for the first call to `malloc_conf_init`: it shouldn't print based on the `opt_confirm_conf == true` test.

So, following this line of thinking, there can be my current solution - defining a `CONTINUE` macro and define it differently for the two calls, or, there can be another solution to define it only once like the following:

```
#define CONTINUE                                                            \
                        if (!search_confirm_conf_only && opt_confirm_conf) {\
                                malloc_printf(                              \
                                    "<jemalloc>: conf set: %.*s:%.*s\n",    \
                                    (int)klen, k, (int)vlen, v);            \
                                continue;                                   \
                        } else {                                            \
                                continue;                                   \
                        }
```

This second solution may be making it slightly more complicated to understand, given that this `CONTINUE` would be put under a `if (search_confirm_conf_only) { ... continue; }` clause where there's already a `search_confirm_conf_only` and a `continue`.

Any suggestions for some other alternatives?

yinan1048576:
Will improve the code logic a bit more and add some more comments...

yinan1048576:
Improved the code:

- Only define one `CONTINUE`.
- Also defined a `MALLOC_CONF_ERROR` macro and added logic so that the options won't be printed if an error is encoutered.
- Added a wrapper function so there won't be large space taken by static variables within functions (i.e. the `buf` variable).
- Added some minor comments to explain the logic but tried to rely mainly on the code to explain itself.

yinan1048576:
Revised the naming of the new macros so that they all follow `CONF_*` format.

src/jemalloc.c
```diff
@@ -37,6 +37,13 @@ bool	opt_abort =
 #endif
     ;
 bool	opt_abort_conf =
+#ifdef JEMALLOC_DEBUG
```
interwq: We don't really want to output conf by default in debug mode; we have many testing frameworks running in dbg mode, with an expected test output format.

src/jemalloc.c
```diff
@@ -930,92 +937,124 @@ malloc_slow_flag_init(void) {
...
+malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
+    bool search_confirm_conf_only, const char *opts_cache[4],
+    char buf[PATH_MAX + 1], const char *empty_opts) {
+	static const char *opts_explain[4] = {
+		"string specified via --with-malloc-conf",
+		"string pointed to by the global variable malloc_conf",
+		"\"name\" of the file referenced by the symbolic link named "
+		    "/etc/malloc.conf",
+		"value of the environment variable MALLOC_CONF"
+	};
```
interwq: 👍 matching the language in the manual certainly makes sense.

src/jemalloc.c
```diff
@@ -930,92 +937,124 @@ malloc_slow_flag_init(void) {
...
 	unsigned i;
-	char buf[PATH_MAX + 1];
 	const char *opts, *k, *v;
 	size_t klen, vlen;
 
 	for (i = 0; i < 4; i++) {
```
interwq: Nit: now that we have the magic number 4 populated in multiple places, let's define a macro somewhere.

src/jemalloc.c
```diff
@@ -930,92 +937,124 @@ malloc_slow_flag_init(void) {
...
-			} else {
-				/* No configuration specified. */
-				buf[0] = '\0';
-				opts = buf;
-			}
-			break;
-		case 2: {
-			ssize_t linklen = 0;
+		if (opts_cache[i] == NULL) {
+			switch (i) {
```
interwq: Move the big switch in its own function, given that the logic around this gets a bit tricky.

src/jemalloc.c
```diff
@@ -1382,6 +1425,18 @@ malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS]) {
 	atomic_store_b(&log_init_done, true, ATOMIC_RELEASE);
 }
 
+static void
+malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS]) {
+	const char *opts_cache[4] = {NULL, NULL, NULL, NULL};
+	char buf[PATH_MAX + 1];
+	const char empty_opts = '\0';
```
interwq: Let's get rid of this hack; and pass a explicit flag "use_cached_opts" to the helper -- first call will be false which reads the opts and fill the cache.

src/jemalloc.c
```diff
@@ -930,92 +937,124 @@ malloc_slow_flag_init(void) {
...
+#define CONF_CONTINUE							\
+			if (!search_confirm_conf_only			\
+			    && opt_confirm_conf && !error) {		\
+				malloc_printf(				\
+				    "<jemalloc>: Set conf value: "	\
+				    "%.*s:%.*s\n", (int)klen, k,	\
+				    (int)vlen, v);			\
+				continue;				\
+			} else {					\
+				continue;				\
```
interwq: Let's only have 1 `continue` statement (move it outside of the branch). Also `{}` the whole thing.

src/jemalloc.c
```diff
@@ -1119,7 +1155,14 @@ malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS]) {
 				    sizeof(o)-1;			\
 				strncpy(o, v, cpylen);			\
 				o[cpylen] = '\0';			\
-				continue;				\
+				CONF_CONTINUE;				\
+			}
+
+			bool error = false;
```
interwq: name the var better, maybe `cur_opt_valid`?

yinan1048576:
Revised according to feedback.  Only one thing left: what do you mean by:

> Let's get rid of this hack; and pass a explicit flag "use_cached_opts" to the helper -- first call will be false which reads the opts and fill the cache.

What I used for checking whether I should read the raw opts or not is to test whether `opts_cache[i] == NULL`, and even with a flag passed in, `opts_cache` will be set to all `NULL`s initially anyway, so they can be used as flags themselves, and for free. :)

yinan1048576:
Added documentation in `doc/jemalloc.xml.in`.

interwq:
re: explicit flags VS checking null value; discussed offline. Yinan will update with the flag indicating whether to access the actual option or not.

yinan1048576:
Revised according to feedback -
- Used explicit flag rather than null-checking to differentiate the first and second call. The flag is effectively identical to the `search_confirm_conf_only` flag, so consolidated the two into a single `initial_call` flag;
- Removed `empty_opts` (to improve locality) and used `NULL` to indicate an empty opts.

src/jemalloc.c
```diff
@@ -929,93 +930,130 @@ malloc_slow_flag_init(void) {
...
-			}
-			break;
-		case 2: {
-			ssize_t linklen = 0;
+/* Number of sources for initializing malloc_conf */
+#define MALLOC_CONF_NSOURCES 4
+
+static const char *
+obtain_malloc_conf(unsigned which_source, char buf[PATH_MAX + 1]) {
+	assert(which_source < MALLOC_CONF_NSOURCES);
```
interwq: nit: let's add some more explicit checking

    if (config_debug) {
        static unsigned read_source = 0;
        /* Each source should only be read once, to minimize # of syscalls on init. */
        assert(read_source++ == which_source);
    }

yinan1048576:
Revised according to feedback. Some example output (in debug mode):
```
$ ./test
$ MALLOC_CONF="confirm_conf:true,asdf:fdsa" ./test
<jemalloc>: malloc_conf (string specified via --with-malloc-conf): ""
<jemalloc>: malloc_conf (string pointed to by the global variable malloc_conf): ""
<jemalloc>: malloc_conf ("name" of the file referenced by the symbolic link named /etc/malloc.conf): ""
<jemalloc>: malloc_conf (value of the environment variable MALLOC_CONF): "confirm_conf:true,asdf:fdsa"
<jemalloc>: Set conf value: confirm_conf:true
<jemalloc>: Invalid conf pair: asdf:fdsa
Aborted (core dumped)
$ MALLOC_CONF="confirm_conf:true" ./test
<jemalloc>: malloc_conf (string specified via --with-malloc-conf): ""
<jemalloc>: malloc_conf (string pointed to by the global variable malloc_conf): ""
<jemalloc>: malloc_conf ("name" of the file referenced by the symbolic link named /etc/malloc.conf): ""
<jemalloc>: malloc_conf (value of the environment variable MALLOC_CONF): "confirm_conf:true"
<jemalloc>: Set conf value: confirm_conf:true
$ MALLOC_CONF="confirm_conf:true,abort_conf:false,asdf:fdsa" ./test
<jemalloc>: malloc_conf (string specified via --with-malloc-conf): ""
<jemalloc>: malloc_conf (string pointed to by the global variable malloc_conf): ""
<jemalloc>: malloc_conf ("name" of the file referenced by the symbolic link named /etc/malloc.conf): ""
<jemalloc>: malloc_conf (value of the environment variable MALLOC_CONF): "confirm_conf:true,abort_conf:false,asdf:fdsa"
<jemalloc>: Set conf value: confirm_conf:true
<jemalloc>: Set conf value: abort_conf:false
<jemalloc>: Invalid conf pair: asdf:fdsa
```

interwq:
One last nit: let's add an index to the malloc_conf output, e.g. `<jemalloc>: malloc_conf #1 (string specified via --with-malloc-conf): ""`

yinan1048576:
Revised the print-outs. Now they look like:
```
<jemalloc>: malloc_conf #1 (string specified via --with-malloc-conf): ""
<jemalloc>: malloc_conf #2 (string pointed to by the global variable malloc_conf): ""
<jemalloc>: malloc_conf #3 ("name" of the file referenced by the symbolic link named /etc/malloc.conf): ""
<jemalloc>: malloc_conf #4 (value of the environment variable MALLOC_CONF): "confirm_conf:true"
```

PR #1501, opened 2019-05-06
Track nfills and nflushes for arenas.i.small / large

interwq:
Small is added purely for convenience.  Large flushes wasn't tracked before and
can be useful in analysis.  Large fill simply reports nmalloc, since there is no
batch fill for large currently.


interwq:
Reviewed by Yinan. Merging.

PR #1506, opened 2019-05-15
Fix assert in free fastpath

yinan1048576:
rtree_szind_slab_read_fast() may have not initialized
alloc_ctx.szind, unless after confirming the return is true.


yinan1048576:
cc @djwatson

PR #1508, opened 2019-05-20
Add experimental.arenas.i.pactivep.

interwq:
The new experimental mallctl exposes the arena pactive counter to applications,
which allows fast read w/o going through the mallctl / epoch steps.  This is
particularly useful when frequent balancing is required, e.g. when having
multiple manual arenas, and threads are multiplexed to them based on usage.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -224,6 +224,8 @@ CTL_PROTO(experimental_hooks_install)
 CTL_PROTO(experimental_hooks_remove)
 CTL_PROTO(experimental_utilization_query)
 CTL_PROTO(experimental_utilization_batch_query)
+CTL_PROTO(experimental_arenas_i_pactivep)
+INDEX_PROTO(experimental_arenas_i)
 
 #define MUTEX_STATS_CTL_PROTO_GEN(n)					\
 CTL_PROTO(stats_##n##_num_ops)						\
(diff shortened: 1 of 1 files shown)


interwq:
@yinan1048576 for review

Note that the test case is intentionally omitted, as the new API only support certain platforms (because of atomics). Manually tested and it has the expected results and behavior.

src/ctl.c
```diff
@@ -3348,3 +3370,56 @@ experimental_utilization_batch_query_ctl(tsd_t *tsd, const size_t *mib,
...
+	return ret;
+}
+
+static int
+experimental_arenas_i_pactivep_ctl(tsd_t *tsd, const size_t *mib,
+    size_t miblen, void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	if (!config_stats) {
+		return ENOENT;
+	}
+	if (*oldlenp != sizeof(size_t *)) {
```
yinan1048576: It might be better to have `oldlenp == NULL || *oldlenp != sizeof(size_t *)`, just in case `oldlenp` happens to be `NULL`?

`oldp == NULL` can also be tested here to be rigorous, but the `READ(pactivep, size_t *)` below tests it anyway, so it doesn't make much difference, only except that it may save a few operations to test it here if the input is invalid.

interwq: Good point. Will update.

PR #1510, opened 2019-05-22
Avoid blocking on background thread lock for stats.

### LINKED ISSUE #1502: mallctl("epoch") is very slow in some scenario
Hi, I'm trying to enable the background_thread in Redis to solve a problem i reported in #1398.
while testing for possible regression i found a problem which i'm not yet sure what it is related to.

I'm attaching a small tester that reproduces it, but the general scenario is this:
Application allocates some massive amount of allocations (about 100M allocations of 100 bytes, 100M allocations of 24 bytes, and another 8MB of big chunk).
It then proceeds to release that memory in a background thread, while the main thread starts working on a new data set (but in some cases the main thread may also go idle).
Since the main thread may go idle, i'm enabling the `background_thread`.
Another thing that the main thread is doing from time to time is collect some stats from jemalloc, for which it needs to set the `epoch` using `mallctl.

In some cases, i observed that triggering the `epoch` calculations takes over a second!!!
This doesn't seem to occur when the background_thread is disabled, and it also doesn't seem to occur if the old allocations are released at the same order in which they were allocated (this is why the attached tester uses Redis's dict.c, i.e. to scramble the order in which deallocations are make)

Please let me know if there's a way to mitigate this? or if this is maybe a bug / edge case that can be fixed in jemalloc.

Thanks.

[test_epoch_calc_time.tgz.txt](https://github.com/jemalloc/jemalloc/files/3171307/test_epoch_calc_time.tgz.txt)
 (sorry for the dirty coding)

interwq:
Background threads may run for a long time, especially when the # of dirty pages
is high.  Avoid blocking stats calls because of this (which may cause latency
spikes).

Resolves #1502.

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -799,7 +799,13 @@ background_thread_stats_read(tsdn_t *tsdn, background_thread_stats_t *stats) {
 	nstime_init(&stats->run_interval, 0);
 	for (unsigned i = 0; i < max_background_threads; i++) {
 		background_thread_info_t *info = &background_thread_info[i];
-		malloc_mutex_lock(tsdn, &info->mtx);
+		if (malloc_mutex_trylock(tsdn, &info->mtx)) {
+			/*
+			 * Each background thread run may take a long time;
+			 * avoid waiting on the stats if the thread is active.
+			 */
+			continue;
+		}
 		if (info->state != background_thread_stopped) {
 			num_runs += info->tot_n_runs;
 			nstime_add(&stats->run_interval, &info->tot_sleep_time);


src/background_thread.c
```diff
@@ -799,7 +799,13 @@ background_thread_stats_read(tsdn_t *tsdn, background_thread_stats_t *stats) {
 	nstime_init(&stats->run_interval, 0);
 	for (unsigned i = 0; i < max_background_threads; i++) {
 		background_thread_info_t *info = &background_thread_info[i];
-		malloc_mutex_lock(tsdn, &info->mtx);
+		if (malloc_mutex_trylock(tsdn, &info->mtx)) {
+			/*
+			 * Each background thread run may take a long time;
+			 * avoid waiting on the stats if the thread is active.
+			 */
+			continue;
```
yinan1048576: Do you want to have a counter or something so that the caller can know how many background threads succeeded? E.g. instead of returning a `bool`, return an `unsigned` counter so that the caller can know more details if desired? May not be immediately useful though given the current way this function is called.

interwq: Yep it could be useful if we decide to report detailed per thread stats -- right now only the aggregated numbers are shown.

PR #1515, opened 2019-05-24
Add missing safety_check.c to MSBuild projects

frederik-h:
The file is included in the list of source files in Makefile.in,
but it is missing from the project files. This causes the
build to fail due to unresolved symbols.


interwq:
Thanks for the patch!

PR #1516, opened 2019-05-28
RFC: fast mutex

djwatson:
There are likely large wins to be had by explicitly controlling the adaptive mutex rate.  We can also experiment with MCS or flat-combining style locks (see also folly::DistributedMutex that we've been rolling out internally). 

This diff reduced P99 tail latencies by ~2% last time I tested it. 

One downside is that the background thread logic also requires a condition variable with timeout support, so that also has to be re-implemented in terms of the new mutex's waiter list.


krytarowski:
atomics are not generally portable and can be missing on certain architectures. Better to stay with regular mutexes or make this conditional.

davidtgoldblatt:
Closing out PRs that are no longer mergeable.

PR #1520, opened 2019-06-04
Fix logic in printing

yinan1048576:
`cbopaque` can now be overriden without overriding `write_cb` in
the first place.  (Otherwise there would be no need to have the
`cbopaque` parameter in `malloc_message`.)


yinan1048576:
(Also included a fix of a typo in the doc.)

PR #1524, opened 2019-06-06
Buffer prof_accumbytes using bytes_until_sample

yinan1048576:
`prof_accumbytes` in `tcache` and `bytes_until_sample` in `tsd`
move in sync with each other, so trying to buffer the more slowly
cycling one i.e. `prof_accumbytes` using the faster cycling one i.e.
`bytes_until_sample`, in order to save some cycles in fast path.

### DIFF
--- a/include/jemalloc/internal/tcache_inlines.h
+++ b/include/jemalloc/internal/tcache_inlines.h
@@ -93,9 +93,6 @@ tcache_alloc_small(tsd_t *tsd, arena_t *arena, tcache_t *tcache,
 	if (config_stats) {
 		bin->tstats.nrequests++;
 	}
-	if (config_prof) {
-		tcache->prof_accumbytes += usize;
-	}
 	tcache_event(tsd, tcache);
 	return ret;
 }
@@ -150,9 +147,6 @@ tcache_alloc_large(tsd_t *tsd, arena_t *arena, tcache_t *tcache, size_t size,
 
 		if (config_stats) {
 			bin->tstats.nrequests++;
-		}
-		if (config_prof) {
-			tcache->prof_accumbytes += usize;
 		}
 	}
 
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2342,6 +2342,10 @@ je_malloc(size_t size) {
 		tsd_bytes_until_sample_set(tsd, bytes_until_sample);
 
 		if (unlikely(bytes_until_sample < 0)) {
+			tcache->prof_accumbytes +=
+			    (uint64_t)-bytes_until_sample +
+			    ((uint64_t)1U << lg_prof_sample);
+
 			/*
 			 * Avoid a prof_active check on the fastpath.
 			 * If prof_active is false, set bytes_until_sample to
@@ -2363,9 +2367,6 @@ je_malloc(size_t size) {
 		if (config_stats) {
 			*tsd_thread_allocatedp_get(tsd) += usize;
 			bin->tstats.nrequests++;
-		}
-		if (config_prof) {
-			tcache->prof_accumbytes += usize;
 		}
 
 		LOG("core.malloc.exit", "result: %p", ret);


yinan1048576:
This change may have some critical effect on the profiling dumps, and I'm putting it up in a somewhat premature state to seek some early feedback. The change depends on the following simplification assumptions:

- `lg_prof_interval`, whenever set to be some value other than the default `-1`, is typically significantly larger than `lg_prof_sample` in practice, since a dumping rate higher than the sampling rate would make the dumps less interesting. Therefore I'm buffering `prof_accumbytes` using `bytes_until_sample` rather than the other way around. In the rare case where `lg_prof_interval` is smaller than `lg_prof_sample` (and is not `-1`), the dumps are generated at the sampling frequency, which is also the limit for the dumps to be interesting.
- I'm increasing `prof_accumbytes` by `(uint64_t)1U << lg_prof_sample` Whenever `bytes_until_sample` runs below 0. This is not exactly precise, since the sampling interval is stochastically distributed with `(uint64_t)1U << lg_prof_sample` as its mean. I think using the mean to approximate may not be a super critical issue since (a) the dumping intervals were not exact anyway, and (b) on average it shouldn't deviate much especially since a dump is generated typically after many sampling cycles whose deviations cancel each other.
- I noticed that `bytes_until_sample` doesn't seem to take into consideration allocation sizes beyond `SC_LOOKUP_MAXCLASS`, since it would have branched out to `malloc_default` [early](https://github.com/jemalloc/jemalloc/blob/40a3435b8dc225ad61329aca89d9c8d0dfbc03ab/src/jemalloc.c#L2320)? Not sure if I'm missing anything, or if this was by design. In any case, it's not affected by this change.

interwq:
> I noticed that bytes_until_sample doesn't seem to take into consideration allocation sizes beyond SC_LOOKUP_MAXCLASS, since it would have branched out to malloc_default early? Not sure if I'm missing anything, or if this was by design. In any case, it's not affected by this change.

malloc_default handles `bytes_until_sample` as well, specifically down this path: 
`imalloc_body`
`prof_alloc_prep`
and eventually
https://github.com/jemalloc/jemalloc/blob/21cfe59ff7b10a61dabe26cd3dbfb7a255e1f5e8/include/jemalloc/internal/prof_inlines_b.h#L83

So this change needs to update that part accordingly as well.

yinan1048576:
Great point! Things are getting more interesting here. :) I browsed through some of the earlier PRs from Dave on this subject. A few things to make sure before I continue on:

(1) It seems to me that the best place to adjust `tcache->prof_accumbytes` would be in `prof_sample_accum_update()`. If I put it in `prof_sample_check()`, we might end up double counting if `tdata` has not been initialized before the `prof_sample_accum_update()` call?

(2) It also seems to me that adjusting `tcache->prof_accumbytes` in `prof_sample_accum_update()` can eliminate the need to adjust it in `malloc()` fast path, or otherwise there'd be double counting since `prof_sample_accum_update()` would eventually be called in the slow path?

(3) Having looked at `prof_sample_accum_update()` more closely, I realize that it might be a better idea not to increase `tcache->prof_accumbytes` by `(uint64_t)1U << lg_prof_sample` i.e. the average, but by the exact amount as determined by `prof_sample_threshold_update()`. My previous approach was changing the `malloc()` fast path, which lives quite far away from the `prof_sample_threshold_update()` call. Now it's much closer, only despite the fact that the `prof_sample_threshold_update()` call obtains the next countdown rather than the previous one which we need. Maybe I can allocate another variable in the `tsd` for storing the previous countdown?

interwq:
re: when to adjust tcache->prof_accumbytes, like you mentioned, to avoid double accounting, we should only accumulate when we reset `bytes_until_sample` in this case, otherwise we always risk incrementing twice. Although the idump isn't totally strict in terms of the interval, that still won't be right.

re: adjusting the actual bytes instead of an average. Agreed. Let's design this change towards #1583 directly? I think it's fine adding two constants like we discussed, to be combined with `allocated` for the trigger, i.e. by checking if `allocated - last_threshold > current_gap`.

interwq:
Discussed with Yinan offline; we should be able to make the fast path only modifying `thread.allocated` and a single branch to compare against a precomputed threshold for triggering medium / slow paths.

yinan1048576:
Covered by #1622 and #1646. Closing.

PR #1525, opened 2019-06-11
Add buffered writer

yinan1048576:
The buffered writer adopts a signature identical to `write_cb`,
so that it can be plugged into anywhere `write_cb` appears.
Currently it's only used in `malloc_stats_print` when both the
`write_cb` parameter and `je_malloc_message` are not set - without
buffering (as is the case before this change) it would invoke
`malloc_write_fd` for every single line in the stats output.

### DIFF
--- a/src/malloc_io.c
+++ b/src/malloc_io.c
@@ -664,6 +664,36 @@ malloc_printf(const char *format, ...) {
 	va_end(ap);
 }
 
+void
+buf_writer_flush(buf_writer_arg_t *arg) {
+	assert(arg->buf_end <= arg->buf_size);
+	arg->buf[arg->buf_end] = '\0';
+	if (arg->write_cb == NULL) {
+		arg->write_cb = je_malloc_message != NULL ?
+		    je_malloc_message : wrtmessage;
+	}
+	arg->write_cb(arg->cbopaque, arg->buf);
+	arg->buf_end = 0;
+}
+
+void
+buffered_write_cb(void *buf_writer_arg, const char *s) {
+	buf_writer_arg_t *arg = (buf_writer_arg_t *)buf_writer_arg;
+	size_t i, slen, n, s_remain, buf_remain;
+	assert(arg->buf_end <= arg->buf_size);
+	for (i = 0, slen = strlen(s); i < slen; i += n) {
+		if (arg->buf_end == arg->buf_size) {
+			buf_writer_flush(arg);
+		}
+		s_remain = slen - i;
+		buf_remain = arg->buf_size - arg->buf_end;
+		n = s_remain < buf_remain ? s_remain : buf_remain;
+		memcpy(arg->buf + arg->buf_end, s + i, n);
+		arg->buf_end += n;
+	}
+	assert(i == slen);
+}
+
 /*
  * Restore normal assertion macros, in order to make it possible to compile all
  * C files as a single concatenation.
--- a/test/unit/buf_writer.c
+++ b/test/unit/buf_writer.c
@@ -1,0 +1,64 @@
+#include "test/jemalloc_test.h"
+
+#define TEST_BUF_SIZE 16
+#define UNIT_MAX (TEST_BUF_SIZE * 3)
+
+static size_t test_write_len;
+static char test_buf[TEST_BUF_SIZE];
+static uint64_t arg_store;
+
+static void test_write_cb(void *cbopaque, const char *s) {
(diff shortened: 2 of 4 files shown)


yinan1048576:
It is also feasible to enable buffered writing by default for `malloc_stats_print`. I imagine there might be cases where the call back is a "`grep`"-like function, in which case buffering might not be what the caller wants, but most of the times the call back should just be some writing function and buffering can be very helpful.

yinan1048576:
Separated out the change to `malloc_stats_print` as a subsequent change so as to improve modularity. Also improved & added more comments.

yinan1048576:
Rebase, and put unit tests into the same PR.

src/malloc_io.c
```diff
@@ -664,6 +664,33 @@ malloc_printf(const char *format, ...) {
 	va_end(ap);
 }
 
+void
+buf_writer_flush(buf_writer_arg_t *arg) {
+	arg->buf[arg->buf_end] = '\0';
```
interwq: nit: assert end < size.

src/malloc_io.c
```diff
@@ -664,6 +664,33 @@ malloc_printf(const char *format, ...) {
...
+
+void
+buffered_write_cb(void *buf_writer_arg, const char *s) {
+	buf_writer_arg_t *arg = (buf_writer_arg_t *)buf_writer_arg;
+	size_t i, n, slen;
+	for (i = 0, slen = strlen(s); i < slen; i += n) {
+		if (arg->buf_end == arg->buf_size) {
+			buf_writer_flush(arg);
+		}
+		n = slen - i < arg->buf_size - arg->buf_end ?
```
interwq: nit: add a `buf_remain = arg->buf_size - arg->buf_end` to help readability.

src/malloc_io.c
```diff
@@ -664,6 +664,33 @@ malloc_printf(const char *format, ...) {
...
+	arg->write_cb(arg->cbopaque, arg->buf);
+	arg->buf_end = 0;
+}
+
+void
+buffered_write_cb(void *buf_writer_arg, const char *s) {
+	buf_writer_arg_t *arg = (buf_writer_arg_t *)buf_writer_arg;
+	size_t i, n, slen;
+	for (i = 0, slen = strlen(s); i < slen; i += n) {
+		if (arg->buf_end == arg->buf_size) {
```
interwq: Move this to the end of the loop?

test/unit/buf_writer.c
```diff
@@ -0,0 +1,65 @@
...
+}
+
+TEST_BEGIN(test_buf_write) {
+	char s[UNIT_MAX + 1];
+	size_t unit, n_unit, remain, i;
+	size_t arg = 1;
+	buf_writer_arg_t test_buf_arg =
+	    {test_write_cb, &arg, test_buf, TEST_BUF_SIZE - 1, 0};
+	memset(s, 'a', UNIT_MAX);
+	for (unit = UNIT_MAX; /* include 0 for edge case */; --unit) {
```
interwq: nit: make `unit` signed and check `unit >= 0`?

test/unit/buf_writer.c
```diff
@@ -0,0 +1,65 @@
...
+		/* unit keeps decreasing, so strlen(s) is always unit. */
+		s[unit] = '\0';
+		for (n_unit = 1; n_unit <= 3; ++n_unit) {
+			test_write_len = 0;
+			remain = 0;
+			for (i = 1; i <= n_unit; ++i) {
+				/* A toy pseudo-random number generator */
```
interwq: Any particular reason to not use the prng functions?

src/malloc_io.c
yinan1048576: The logic might be slightly different. Having the test and the flush upfront makes flushing lazy: if the input exactly fits the entire buffer, the flush won't take place until the next call of `buffered_write_cb()` or `buf_writer_flush()`. (My unit tests are particularly testing for this behavior.) The advantage is that the final `buf_writer_flush()` call would never unnecessarily call the real `write_cb()` one more time with an empty buffer. (Of course I can guard against an empty buffer in the `buf_writer_flush()` function, but that feels a bit redundant.)

yinan1048576:
Revised according to feedback. Added more asserts and better variable namings.

src/malloc_io.c
interwq: Ah I see. Didn't think of the final `buf_writer_flush()` call. Makes sense.

PR #1526, opened 2019-06-12
Buffer prof log stop

yinan1048576:
Make use of the new buffered writer for the output of
`prof_log_stop`.

### DIFF
--- a/src/prof_log.c
+++ b/src/prof_log.c
@@ -160,6 +160,7 @@ prof_log_bt_index(tsd_t *tsd, prof_bt_t *bt) {
 		return node->index;
 	}
 }
+
 static size_t
 prof_log_thr_index(tsd_t *tsd, uint64_t thr_uid, const char *name) {
 	assert(prof_logging_state == prof_logging_state_started);
@@ -575,7 +576,7 @@ prof_log_emit_metadata(emitter_t *emitter) {
 	emitter_json_object_end(emitter);
 }
 
-
+#define PROF_LOG_STOP_BUFSIZE PROF_DUMP_BUFSIZE
 bool
 prof_log_stop(tsdn_t *tsdn) {
 	if (!opt_prof || !prof_booted) {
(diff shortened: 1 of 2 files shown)


yinan1048576:
The only reason why I changed the return to `close(fd) || arg.ret == -1` is that I didn't find any other place `arg.ret` is used; adding it to the return can at least make it somewhat useful. However if we want to be more rigorous, the return value should be checked each time the call back is called, i.e. similar to what we're doing for the prof dump, but that would involve a much more extensive revision. On the other hand, the caller (`prof_log_stop_final`) doesn't interpret the return value anyway. So I think it's also fine to completely get rid of the `ret` field, in which case `struct prof_emitter_cb_arg_s` is also no longer needed.

yinan1048576:
Made the buffer larger but dependent on `JEMALLOC_PROF`. Added and improved comments.

yinan1048576:
Rebase.

src/prof.c
```diff
@@ -2691,6 +2691,20 @@ prof_log_emit_metadata(emitter_t *emitter) {
 	emitter_json_object_end(emitter);
 }
 
+/*
+ * The buffer doesn't need a lock, since only one thread is able to use it at a
+ * time due to the protection by prof_logging_state.
+ */
+#define PROF_LOG_STOP_BUFSIZE PROF_DUMP_BUFSIZE
+static char prof_log_stop_buf[
```
interwq: Since prof_log isn't used by default, even with prof:true, I'd say allocating the buffer in prof_log_start(), e.g. `if (prof_log_stop_buf == NULL) { ...`, so we only pay the price when in use.

yinan1048576:
Great point. In the revision I'm now allocating the buffer even later, in `prof_log_stop()`, so that there's no extra memory consumption throughout the logging process.

yinan1048576:
Removing irrelevant comments.

src/prof_log.c
```diff
@@ -623,11 +624,19 @@ prof_log_stop(tsdn_t *tsdn) {
...
 	}
 
-	/* Emit to json. */
 	struct prof_emitter_cb_arg_s arg;
 	arg.fd = fd;
-	emitter_init(&emitter, emitter_output_json, &prof_emitter_write_cb,
-	    (void *)(&arg));
+
+	char *prof_log_stop_buf = (char *)iallocztm(tsdn,
+	    PROF_LOG_STOP_BUFSIZE * sizeof(char),
```
interwq: No need to do the sizeof(char), it's always 1.

Also, do we have similar pattern for the current prof dump buffer? If so, let's also do a similar optimization there.

yinan1048576:

> Also, do we have similar pattern for the current prof dump buffer? If so, let's also do a similar optimization there. (In a separate PR)

Yes, there's a similar pattern there. Will look into it.

PR #1527, opened 2019-06-12
Migrate prof dump

yinan1048576:
Replace the original `prof_dump_write` by the new buffered writer.

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -221,7 +221,6 @@ static char		prof_dump_buf[
     1
 #endif
 ];
-static size_t		prof_dump_buf_end;
 static int		prof_dump_fd;
 
 /* Do not dump any profiles until bootstrapping is complete. */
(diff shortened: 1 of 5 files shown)


yinan1048576:
Two remarks:

- The new buffered writer is supposed to be encapsulated in the sense that the outside world should not directly alter any internal data i.e. fields in a `buf_writer_arg_t` variable. However I found `prof_dump_maps` may have to be an exception: it's using the buffer to "pump" contents directly from an input file to an output file, which seems hard to be generalized, so I'm letting it directly touch internal fields of the `buf_writer_arg_t` variable. This is the only place I know of doing buffered file piping, but if it turns out that it also appears in other places, I'll think of ways to generalize it out to e.g. another API in the buffered writer.

- I removed the `propagate_err` parameter in `prof_dump_close`, but it seems also removable in many (or probably all?) other places, since the static `buf_writer_arg_t` variable already carries it and it never changes throughout the entire `prof_dump_file` call. I figured that this could be a change of a relatively larger scale, so I'm trying not to pollute this pull request for now.

yinan1048576:
Rebase.

src/prof.c
```diff
@@ -1257,70 +1265,37 @@ prof_dump_flush(bool propagate_err) {
...
 	}
-	prof_dump_buf_end = 0;
-
-	return ret;
 }
 
-static bool
-prof_dump_close(bool propagate_err) {
-	bool ret;
+static buf_writer_arg_t prof_dump_buf_arg =
```
interwq: Is this covered by any lock?

interwq:
Let's continue this @yinan1048576 ?

yinan1048576:
Yes, my plan is to resume after #1589 is in.

yinan1048576:
Becomes part of #1795 and #1794.

PR #1534, opened 2019-07-02
WIP -- test only

interwq:


Ignition:
I can confirm that this fix works. Windows is no longer leaking due to no extent split function.

Ignition:
We still have the situation where reusable extents in dirty and muzzy can't be recycled because they are not splitable. The fix no longer leaks them but they become unusable. 

`arena_slab_alloc` tries to reuse the dirty and muzzy extents. But the selected extents in `extent_recycle` require splitting which can't be done.

In `extent_recycle` if the extracted extent is too large can it be used without splitting?
Or does the `extent_recycle_extract` need changing to prefer exact match?

interwq:
Replaced by #1545.

PR #1539, opened 2019-07-03
Fix redzone setting and checking

yinan1048576:


davidtgoldblatt:
Welp, this is embarassing. Good catch.

PR #1541, opened 2019-07-04
WIP -- Modify selection to avoid fragmentation

Ignition:
In situations where a perfectly adequate extent exists, of the exact right size, prefer its selection over any other extent. This avoids unnecessary fragmentation on all platforms and on windows also avoids the un-reusable extent problem (because windows doesn't split extents).

Based on and extends #1534

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -451,33 +451,44 @@ extents_first_fit_locked(tsdn_t *tsdn, arena_t *arena, extents_t *extents,
 	extent_t *ret = NULL;
 
 	pszind_t pind = sz_psz2ind(extent_size_quantize_ceil(size));
-	for (pszind_t i = (pszind_t)bitmap_ffu(extents->bitmap,
-	    &extents_bitmap_info, (size_t)pind);
-	    i < SC_NPSIZES + 1;
-	    i = (pszind_t)bitmap_ffu(extents->bitmap, &extents_bitmap_info,
-	    (size_t)i+1)) {
+	pszind_t i = (pszind_t)bitmap_ffu(extents->bitmap, &extents_bitmap_info,
+	(size_t)pind);
+	/*
+	 * In order to reduce fragmentation, if avoidable then prefer to not split
+	 */
+	if (pind == i) {
 		assert(!extent_heap_empty(&extents->heaps[i]));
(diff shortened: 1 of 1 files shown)


interwq:
@Ignition : thanks for the patch. However we wouldn't really want to abandon first-fit, which is critical to avoiding fragmentation in the long run. We used to have best-fit (which is what this patch does), however after a bunch of evaluation we decided to remove it -- it provides some immediate benefit (avoids splitting, also less scanning), however the better layout from first-fit compensates more in the long run.

We have some ideas regarding how to improve the situation on Windows; the current issue is caused by the different semantic between VirtualAlloc / VirtualFree and mmap / munmap. We'll discuss a bit more and update some time next week.

Ignition:
I will point out this is not identical to best-fit as this is a hybrid (only using best-fit if an exact match is found), possibly worth evaluating. But I get what you are saying and trust your evaluation, I look forward to your fix.

davidtgoldblatt:
Just as an update: we think that setting retain=true on Windows should be a good fix to this, so long as we don't merge extents allocated as different VirtualAlloc calls. I think Qi is experimenting with this this week.

Ignition:
`retain=true` is not a fix. Primary reason for selecting jemalloc over other scallable allocators was for the ability to release memory back to the OS after a bursty memory consumption. IMO `retain=true` only makes sense in a dedicated HPC context where you are the only running process that matters. Many other usages exists like desktop applications and multi tenant solutions where it is unlikely they will all hit peak consumption at the same time so you can fit more tenants onto one box.

davidtgoldblatt:
Retaining is about address space reservation rather than committed memory reservation; it's mostly orthogonal to how much and when we purge.

Ignition:
Ok, good clarification, thank you.

interwq:
@Ignition : #1545 should allow proper merge & split on Windows and fix the issue. We'd appreciate much if you could help test it; it would be great if you can also get some stats, e.g. by calling `malloc_stats_print(NULL, NULL, NULL);` at the end of your program.

interwq:
cc: @frederik-h in case you are interested in testing the fix in #1545  for Windows as well.

PR #1545, opened 2019-07-12
Implement retain on Windows.

interwq:
The VirtualAlloc and VirtualFree APIs are different because MEM_DECOMMIT cannot
be used across multiple VirtualAlloc regions.  To properly support decommit, only
allow merge / split within the same region -- this is done by not merging extents
with different serial numbers (w/ opt.retain).


interwq:
@Ignition, @frederik-h, @santagada, @Cytoo, @j0t -- FYI that we plan to do a 5.2.1 release in a couple of weeks, for the Windows VM leak fix in this patch. The issues you guys experienced before on Windows are likely all related to the issue here; any chance you could help test / benchmark the changes here?

Ignition:
The way I built for windows retain was not default true. Even with `--with-malloc-conf=retain:true`, `malloc_stats_print(NULL, NULL, NULL);` still reports `opt.retain: false`. I can you my conan recipe to double check nothing odd is going on.

Ignition:
My conanfile.py https://gist.github.com/Ignition/5e2174ee6dd97a2c6e93db3978549dbf

Ignition:
Ok, never mind. It was due to my conan build cache having an old build. I now get `opt.retain: true`, initial testing perfectly fine.

src/extent.c
```diff
@@ -2151,9 +2148,23 @@ extent_split_wrapper(tsdn_t *tsdn, arena_t *arena,
...
 
 static bool
-extent_merge_default_impl(void *addr_a, void *addr_b) {
+extent_merge_default_impl(extent_t *a, void *addr_a, extent_t *b,
+    void *addr_b) {
 	if (!maps_coalesce) {
-		return true;
+		/*
+		 * When coalesce is not always allowed (Windows), only merge
+		 * extents with identical sn (i.e. from the same VirtualAlloc
```
davidtgoldblatt: Can you include some back-of-the envelope math to convince me / the reader that "same sn == same VirtualAlloc call"? This is the first time that we're making that fact a correctness issue instead of just a QoI one, and that makes me somewhat nervous.

davidtgoldblatt:
You know, it occurs to me that we can use two rtree bits to indicate "same VirtualAlloc" instead of using the extent sn. Have the bits be 0 if the extent is in the interior of a mapping, 1 if it's at the beginning of a mapping, and 2 if it's at the end of a mapping. That would let us avoid trying to play games with serial numbers entirely.

davidtgoldblatt:
Actually, we don't even need to take it from the rtree; it can just live in the extent state itself. We can take two bits from the serial number (or squirrel them away into the low bits of a pointer, etc.), and then we get a firm correctness guarantee regardless of serial number overflow.

interwq:
> Actually, we don't even need to take it from the rtree; it can just live in the extent state itself. We can take two bits from the serial number (or squirrel them away into the low bits of a pointer, etc.), and then we get a firm correctness guarantee regardless of serial number overflow.

Yeah realized we can further simplify by only tracking the `is_head` state, and disallow merging if the second extent (higher addr) is a head extent. Will update soon.

j0t:
Just have tested my app with a library built from latest `interwq:retain` in CYGWIN. At first sight, all works fine with default `JE_MALLOC_CONF` settings. The statistics shows a great amount of retained bytes and mapped quantity is not growing anymore.
~~~
Allocated: 1321568, active: 8912896, metadata: 4691552 (n_thp 0), resident: 48103424, mapped: 58261504, retained: 238485504
~~~
However, some jemalloc tests didn't pass. Don't know yet is it a specific CYGWIN issue or something else. Will try to found out more later.

src/extent.c
```diff
@@ -2045,13 +2039,16 @@ extent_purge_forced_wrapper(tsdn_t *tsdn, arena_t *arena,
...
 static bool
 extent_split_default(extent_hooks_t *extent_hooks, void *addr, size_t size,
     size_t size_a, size_t size_b, bool committed, unsigned arena_ind) {
-	return !maps_coalesce;
+	if (maps_coalesce) {
+		return false;
+	}
+
+	/* See comments in extent_merge_default_impl(). */
+	return opt_retain ? false : true;
```
davidtgoldblatt: Why does extent splitting return an error only if retain is off? I see why we might want to treat this differently, but erroring out on splitting seems like a weird way to do so.

Also, I think this whole thing is most clearly written as: `return !(maps_coalesce || opt_retain)`.

interwq: Will refactor this. I wanted to separate the Windows logic and add some comments. Let me make that happen.

I think we want to consider the returning true as "not allowed to split" instead of a real error; without retain, purging does MEM_RELEASE which only accepts the whole region. In that case we don't allow split / merge. The user custom hooks might return true as a way to say not-allowed as well, e.g. in hhvm for arenas owning reserved huge pages (but with regular pages as back up), they may check if an extent is part of the huge pages and disallow purging.

interwq:
@Ignition, @j0t : many thanks for help testing!

The high retained bytes might be expected, normally it should be a couple times of the peak RSS. Also the total VM size won't decrease because only decommit will happen instead of unmap. It would be helpful if you can share the full stats too.

interwq:
@j0t : which tests are failing for you? You can test against the dev branch to verify if it's related to retain here.

j0t:
> @Ignition, @j0t : many thanks for help testing!

You are welcome! It's great that this issue has been resolved finally.

> It would be helpful if you can share the full stats too.

Here are two stats, `dev` as baseline and `retain`:
dev
[dev.log](https://github.com/jemalloc/jemalloc/files/3423569/dev.log)
retain
[retain.log](https://github.com/jemalloc/jemalloc/files/3423571/retain.log)

j0t:
> @j0t : which tests are failing for you? You can test against the dev branch to verify if it's related to retain here.

No, the same failures are originated from `dev`. They are specific to CYGWIN and related to `prof_backtrace()` in `prof.c` and thread cleanup machinery in `tsd_win.h`. BTW, I did managed to fix them, but it could require a separate PR, I believe...

src/extent.c
```diff
@@ -1323,7 +1317,7 @@ extent_grow_retained(tsdn_t *tsdn, arena_t *arena,
 
 	extent_init(extent, arena, ptr, alloc_size, false, SC_NSIZES,
 	    arena_extent_sn_next(arena), extent_state_active, zeroed,
-	    committed, true);
+	    committed, true, true);
```
davidtgoldblatt: Let's make this an enum instead of a new bool parameter? I already have trouble reading all these bool params.

src/extent.c
```diff
@@ -2162,24 +2164,58 @@ extent_merge_default_impl(void *addr_a, void *addr_b) {
 	return false;
 }
 
-#ifdef JEMALLOC_MAPS_COALESCE
+static bool
```
davidtgoldblatt: Let's add a function-level comment? I.e.

"Returns true if the given extents can't be merged because of their head bit settings."

src/extent.c
```diff
@@ -2162,24 +2164,58 @@ extent_merge_default_impl(void *addr_a, void *addr_b) {
...
+	}
+	assert(extent_sn_comp(a, b) == 0);
+
+	return false;
+}
+
 static bool
 extent_merge_default(extent_hooks_t *extent_hooks, void *addr_a, size_t size_a,
     void *addr_b, size_t size_b, bool committed, unsigned arena_ind) {
+	if (!maps_coalesce) {
```
davidtgoldblatt: I think this logic should live in extent_merge_default_impl, no? It feels weird that we're splitting up the coalescing logic this way.

interwq: I have been back and forth on this a couple of times; the reason I wanted `extent_head_no_merge` to be outside of `extent_merge_default_impl` is to make sure in `extent_merge_impl`, we check the extents mergeability before calling the merge hook, so that if user has custom merge hook, they won't get two unmergeable extents (since the head state is tracked internally). I do agree the split logic this way is strange.

How about, we move the `extent_head_no_merge` into `extent_merge_default_impl`; but in the custom merge hook branch here https://github.com/jemalloc/jemalloc/pull/1545/files#diff-7b26b977303fe92c093a2245b0eaf255R2239, we check `extent_head_no_merge` before calling the user hook?

davidtgoldblatt: I sort of thought we wouldn't track head state at all for extent-hook-managed extents; it seems like keeping track of that is the user's job.

I don't think that there's a good way of separating this out though, especially in the presence of potentially-mutable extent hooks. So I think you're right that the current way is better.

We should probably add some mechanism for this to the extent hook wishlist, though.

interwq: Okay I won't change this part then. Update coming soon; will add it to the wishlist as well.

src/extent.c
interwq: Added.

src/extent.c
interwq: Done.

src/extent.c
interwq: Just updated @davidtgoldblatt

interwq:
@j0t : thanks again for sharing the details! I checked the stats and they looked good; also added a new counter in #1553 to detect problematic behavior just in case.

For the profiling related fix, please feel free to submit when ready; we have a 5.2.1 release coming in a couple of weeks, if it's a small-ish change we can still include it.

j0t:
> @j0t : thanks again for sharing the details! I checked the stats and they looked good; also added a new counter in #1553 to detect the problematic behavior just in case.

@interwq, oh, thanks, that's nice! Will check this counter later.
> 
> For the profiling related fix, please feel free to submit when ready; we have a 5.2.1 release coming in a couple of weeks, if it's a small-ish change we can still include it.

Currently I'm struggling with some build/makefile issues related to CYGWIN specific, carefully trying 
to keep everything in buildable state. (In particular CYGWIN has different dll naming conventions and requires import libraries to be present for proper linking).

PR #1547, opened 2019-07-14
Reorder the configs for AppVeyor.

interwq:
Enable-debug and 64-bit runs tend to be more relevant. Run them first.


davidtgoldblatt:
This is such a good idea that I'm embarrassed that we don't already do it.

PR #1552, opened 2019-07-15
extent_dalloc instead of leak when register fails.

interwq:
extent_register may only fail if the underlying extent and region got stolen /
coalesced before we lock.  Avoid doing extent_leak (which purges the region)
since we don't really own the region.


interwq:
@davidtgoldblatt -- definitely needs your sanity check here.

PR #1553, opened 2019-07-15
Track the leaked VM space via the dropped_vm counter.

interwq:
The counter is 0 unless metadata allocation failed (indicates OOM), and is
mainly for sanity checking.

Tested with local code path emulation and verified that the counter reported properly.

Based on #1552.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -210,6 +210,7 @@ CTL_PROTO(stats_arenas_i_internal)
 CTL_PROTO(stats_arenas_i_metadata_thp)
 CTL_PROTO(stats_arenas_i_tcache_bytes)
 CTL_PROTO(stats_arenas_i_resident)
+CTL_PROTO(stats_arenas_i_abandoned_vm)
 INDEX_PROTO(stats_arenas_i)
 CTL_PROTO(stats_allocated)
 CTL_PROTO(stats_active)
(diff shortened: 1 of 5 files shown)


interwq:
Sampled output under the per arena section -- 

    active:                         36864
    mapped:                       2146304
    retained:                     2048000
    base:                          249232
    internal:                           0
    metadata_thp:                       0
    tcache_bytes:                       0
    resident:                      299008
    dropped_vm:                         0
    extent_avail:                       0

src/ctl.c
```diff
@@ -543,6 +544,7 @@ static const ctl_named_node_t stats_arenas_i_node[] = {
 	{NAME("metadata_thp"),	CTL(stats_arenas_i_metadata_thp)},
 	{NAME("tcache_bytes"),	CTL(stats_arenas_i_tcache_bytes)},
 	{NAME("resident"),	CTL(stats_arenas_i_resident)},
+	{NAME("dropped_vm"),	CTL(stats_arenas_i_dropped_vm)},
```
interwq: @davidtgoldblatt -- as we discussed a while ago, adding the counter to track the extents_leak (which also got renamed to extents_drop_vm in this diff), but under a different name. Does this sound okay / any suggestions?

davidtgoldblatt: Sounds good ("abandoned" or "orphaned" would work too if you like one of those better). Saying "leak" in user-visible stats is a recipe for confusion imo.

interwq: I think "abandoned" sounds better. Making the switch now.

PR #1554, opened 2019-07-16
Fix posix_memalign with input size 0.

interwq:
Return a valid pointer instead of failed assertion.

Size 0 fails at https://github.com/jemalloc/jemalloc/blob/ac34afb4037d7e9e87efde2b8e913d87aae131da/include/jemalloc/internal/jemalloc_internal_inlines_c.h#L72 before this patch.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1817,6 +1817,11 @@ struct static_opts_s {
 	bool may_overflow;
 
 	/*
+	 * Whether or not allocations (with alignment) of size 0 should be
+	 * treated as size 1.
+	 */
+	bool bump_empty_aligned_alloc;
+	/*
 	 * Whether to assert that allocations are not of size 0 (after any
 	 * bumping).
 	 */
@@ -1857,6 +1862,7 @@ struct static_opts_s {
 JEMALLOC_ALWAYS_INLINE void
 static_opts_init(static_opts_t *static_opts) {
 	static_opts->may_overflow = false;
+	static_opts->bump_empty_aligned_alloc = false;
 	static_opts->assert_nonempty_alloc = false;
 	static_opts->null_out_result_on_error = false;
 	static_opts->set_errno_on_error = false;
@@ -2044,11 +2050,6 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd) {
 		goto label_oom;
 	}
 
-	/* Validate the user input. */
-	if (sopts->assert_nonempty_alloc) {
-		assert (size != 0);
-	}
-
 	if (unlikely(dopts->alignment < sopts->min_alignment
 	    || (dopts->alignment & (dopts->alignment - 1)) != 0)) {
 		goto label_invalid_alignment;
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
I think probably the right way to do this would be to make it a static_options_t field, and put it in imalloc_body.

It was a hard fight to unify all that logic, even small deviations from the "all logic lives in imalloc_body" scares me.

This is actually the way things used to be (there was a bump_empty_alloc static option), but IIRC Dave killed it in the last round of fast-path optimizations.

I think probably we should do the same optimization as we did there (subtract 1 from size unconditionally, handle the zero case in the sa2u slow path), but if you want to save the effort and just make it a static_option I think that's fine.

interwq:
@davidtgoldblatt : ah that's right. I believe Dave killed the static option because the non-aligned size 0 got changed to be handled by the index 0 in the lookup table. Quick summary about the 0 size handling -- regular non-align malloc()s, size 0 is taken care of by Dave's optimization on the lookup table; mallocx()s (also aligned_alloc) simply disallow size 0; while posix_memalign is left behind here:

Like you mentioned, the two options: 1) add a static option, maybe `bump_empty_aligned_allc`, and only check it down the sa2u path. However it's a bit awkward since the posix_memalign will be the only API doing that (and unlikely to change soon). 2) handle the 0 case in sa2u. However I'd rather not enforce a branch for all other cases (e.g. we know some workload that does mallocx with alignment often).

Unless you think the this special case can live top level like the patch right now, I'm leaning towards (1) with static `bump_empty_aligned_alloc`. Does that sound good?

davidtgoldblatt:
Yeah, I think 1 is fine. 

Another thing: we probably also want to set this in aligned_alloc.

interwq:
Re: aligned_alloc I’m less sure about supporting 0 there. The manual says ‘size should be a multiple of alignment.’ I’ll allow 0 there as well, unless you interpret it differently @davidtgoldblatt

interwq:
Also verified that the regression was introduced in https://github.com/jemalloc/jemalloc/commit/ac34afb4037d7e9e87efde2b8e913d87aae131da, and we did allow 0 for aligned_alloc as well. Update the PR to fix both cases.

src/jemalloc.c
```diff
@@ -2068,13 +2069,22 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd) {
...
+			}
+		}
 		usize = sz_sa2u(size, dopts->alignment);
 		dopts->usize = usize;
 		if (unlikely(usize == 0
 		    || usize > SC_LARGE_MAXCLASS)) {
 			goto label_oom;
 		}
 	}
+	/* Validate the user input. */
```
davidtgoldblatt: I don't think this needs to move? assert_nonempty_alloc should just be false for posix_memalign and aligned_alloc.

davidtgoldblatt:
One nit, otherwise LGTM.

src/jemalloc.c
interwq: Moved this as in the comment for assert_nonempty_alloc we say “Whether to assert that allocations are not of size 0 (after any bumping).” So wanted to make sure all the bumping are done before the assertion. Does that make sense or a bit too much?

davidtgoldblatt: Oh I see. Yeah that makes sense.

PR #1559, opened 2019-07-18
Unit tests for buffered writer

yinan1048576:


yinan1048576:
Confirmed that the unit test would fail if I introduce a bug to the buffered writer implementation like the one just fixed by https://github.com/jemalloc/jemalloc/pull/1521 - we would have avoided the bug if we had unit test! :)

yinan1048576:
Relocating the unit tests to https://github.com/jemalloc/jemalloc/pull/1525.

PR #1560, opened 2019-07-22
Buffer stats printing

yinan1048576:
The buffered writer is only used in `malloc_stats_print` when both
the `write_cb` parameter and `je_malloc_message` are not set;
without buffering (as is the case before this change)
`malloc_stats_print` would invoke `malloc_write_fd` for every single
`printf` (including printing just the line break in many places).

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -3704,6 +3704,7 @@ je_mallctlbymib(const size_t *mib, size_t miblen, void *oldp, size_t *oldlenp,
 	return ret;
 }
 
+#define STATS_PRINT_BUFSIZE 65536
 JEMALLOC_EXPORT void JEMALLOC_NOTHROW
 je_malloc_stats_print(void (*write_cb)(void *, const char *), void *cbopaque,
     const char *opts) {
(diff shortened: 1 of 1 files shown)


yinan1048576:
This is separated out from https://github.com/jemalloc/jemalloc/pull/1525. Also changed the allocation for the buffer to be dynamic, so that there's no additional memory consumed when not needed.

yinan1048576:
Rebase.

yinan1048576:
Rebase and minor fix.

src/jemalloc.c
```diff
@@ -3713,10 +3714,22 @@ je_malloc_stats_print(void (*write_cb)(void *, const char *), void *cbopaque,
 
 	tsdn = tsdn_fetch();
 	check_entry_exit_locking(tsdn);
-	stats_print(write_cb, cbopaque, opts);
+	if (write_cb != NULL || je_malloc_message != NULL) {
+		stats_print(write_cb, cbopaque, opts);
+	} else {
```
interwq: With this change, how do we test the buffered path? IIRC, the tests all customize the two things above?

yinan1048576:
Right, it's hard to test the buffered path. That was a reason why I added the unit tests for the buffered writer, so that it becomes less crucial to have "integration tests" each time we use the buffered writer.

If we really want to test, I can have an `#ifdef JEMALLOC_JET` guard but it would still be hard to test both the buffered and unbuffered paths. Any good suggestion?

On the other hand, it might not be a terrible idea to always buffer the stats print output. In theory the output may differ if the call back function and argument provided by the application are "special". See:

https://github.com/jemalloc/jemalloc/blob/22746d3c9fddd5486e9ec5c0c6b2e25230db9a8e/include/jemalloc/internal/malloc_io.h#L107

for some more explanations I wrote. But in general this shouldn't be a big issue.

interwq:
I think always use buffered print is fine; there's no guarantee that the output has to be "fine-grained".

Maybe guard it only by `config_debug`? In debug build we may want immediate output; e.g. say if stats_output got stuck / blocked on a lock.

yinan1048576:
Revised according to feedback. (Used `JEMALLOC_DEBUG` rather than `config_debug` to save one indentation level. :) )

yinan1048576:
Stick to `config_debug`.

yinan1048576:
Increase buffer size to 65536. It's the same as the buffer size for prof dumping.

PR #1564, opened 2019-07-24
Invoke arena_dalloc_promoted() properly w/o tcache.

interwq:
When tcache was disabled, the dalloc promoted case was missing.

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -267,6 +267,22 @@ arena_dalloc_no_tcache(tsdn_t *tsdn, void *ptr) {
 }
 
 JEMALLOC_ALWAYS_INLINE void
+arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
+    bool slow_path) {
+	if (szind < nhbins) {
+		if (config_prof && unlikely(szind < SC_NBINS)) {
+			arena_dalloc_promoted(tsdn, ptr, tcache, slow_path);
+		} else {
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -310,7 +319,7 @@ arena_dalloc(tsdn_t *tsdn, void *ptr, tcache_t *tcache,
 	}
 }
 
-static inline void
+JEMALLOC_ALWAYS_INLINE void
```
davidtgoldblatt: Why are we making these slow paths always inline?

interwq: ah that wasn't intentional. will fix.

davidtgoldblatt:
Can you explain the bug a little? It seems like the original code was correct; an allocation was promoted if and only if szind is small but we're in the large dalloc pathway; I don't get how the tcache is involved.

Are we missing a test?

interwq:
For sampled allocations, we ought to call `arena_dalloc_promoted`, regardless if tcache is on / off. However it was missing on the no tcache path, i.e. https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/arena_inlines_b.h#L255-L256 Only that part is fixed with this PR.

The issue was caught with tcache:false combined with prof:true, by the safety_check tests, e.g. here https://github.com/jemalloc/jemalloc/blob/dev/test/unit/safety_check.c#L27

I was running `run_test.sh` which had this combination (apparently the CI didn't have this).

davidtgoldblatt:
Makes sense, thanks.

PR #1565, opened 2019-07-25
Make small size section of tcache_bin_info static

yinan1048576:
Optimization of free fast path - there's one less redirection.
The total memory consumption is kept unchanged.

### DIFF
--- a/include/jemalloc/internal/tcache_inlines.h
+++ b/include/jemalloc/internal/tcache_inlines.h
@@ -8,6 +8,18 @@
 #include "jemalloc/internal/ticker.h"
 #include "jemalloc/internal/util.h"
 
+JEMALLOC_ALWAYS_INLINE cache_bin_info_t *
+tcache_small_bin_info_get(szind_t i) {
+	assert(i < SC_NBINS);
+	return &tcache_small_bin_info[i];
+}
+
+JEMALLOC_ALWAYS_INLINE cache_bin_info_t *
+tcache_large_bin_info_get(szind_t i) {
+	assert(i >= SC_NBINS && i < nhbins && tcache_large_bin_info != NULL);
+	return &tcache_large_bin_info[i - SC_NBINS];
+}
+
 static inline bool
 tcache_enabled_get(tsd_t *tsd) {
 	return tsd_tcache_enabled_get(tsd);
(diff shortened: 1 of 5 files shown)


src/tcache.c
```diff
@@ -747,30 +752,31 @@ tcache_boot(tsdn_t *tsdn) {
 	nhbins = sz_size2index(tcache_maxclass) + 1;
 
 	/* Initialize tcache_bin_info. */
-	tcache_bin_info = (cache_bin_info_t *)base_alloc(tsdn, b0get(), nhbins
-	    * sizeof(cache_bin_info_t), CACHELINE);
-	if (tcache_bin_info == NULL) {
+	assert(nhbins >= SC_NBINS);
+	tcache_large_bin_info_data = (cache_bin_info_t *)base_alloc(tsdn,
```
interwq: nhbins can be equal to NBINS; should not alloc in that case.

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -8,6 +8,18 @@
 #include "jemalloc/internal/ticker.h"
 #include "jemalloc/internal/util.h"
 
+JEMALLOC_ALWAYS_INLINE cache_bin_info_t *
+tcache_small_bin_info(szind_t i) {
```
interwq: nit: let's name these ..._bin_info_get()

yinan1048576:
Revised according to feedback. Also added stronger asserts.

yinan1048576:
(And also added cacheline alignment for small bin info.)

src/tcache.c
```diff
@@ -747,30 +753,34 @@ tcache_boot(tsdn_t *tsdn) {
...
-	if (tcache_bin_info == NULL) {
-		return true;
+	assert(nhbins >= SC_NBINS);
+	if (nhbins > SC_NBINS) {
+		tcache_large_bin_info = (cache_bin_info_t *)base_alloc(tsdn,
+		    b0get(), (nhbins - SC_NBINS) * sizeof(cache_bin_info_t),
+		    CACHELINE);
+		if (tcache_large_bin_info == NULL) {
+			return true;
+		}
```
interwq: nit: let's move this section to below the first loop, to make it clear that this is allocating for the large sizes.

Also there you could simply say, `if (nhbins == SC_NBINS) return;`

yinan1048576:

yinan1048576:
Looking back at this PR, it seems now, after your cache bin optimization, `tcache_bin_info` is not accessed on malloc / free fast path anymore, so optimizing for it becomes no longer relevant?

interwq:
That's true. Also it became trickier to apply this; feel free to drop it.

PR #1566, opened 2019-07-25
Lower nthreads in test/unit/retained on 32-bit to avoid OOM.

interwq:


test/unit/retained.c
```diff
@@ -107,6 +107,9 @@ TEST_BEGIN(test_retained) {
 	atomic_store_u(&epoch, 0, ATOMIC_RELAXED);
 
 	unsigned nthreads = ncpus * 2;
+	if (sizeof(void *) < 8 && nthreads > 16) {
```
yinan1048576: nit: use `LG_SIZEOF_PTR < 3`?

interwq: Done.

PR #1569, opened 2019-07-25
Fix redzone test

yinan1048576:


yinan1048576:
I haven't looked into why tcache affects the safety check. I'm just trying to quickly fix `./run_tests.sh` failures.

interwq:
We do need to figure out why these are failing. BTW did you try the most recent dev? #1564 was to fix issues like this.

yinan1048576:
Good point! The test passed on the latest dev. Closing.

Issue #1414, opened 2019-01-23
failed tests under CYGWIN


thefallentree:
Hi there,

I'm trying to get CYGWIN compilation going (again) with release 5.1.0 , here what I found working and not working.

Working:
1. no <sys/syscall.h> under CYGWIN, so that need fixing.
2. remove override of DUMPSYM in configure.ac (search for cygwin) , I can successful compile.

Not working:
1. TSD unit test has issue. 
```
test_tsd_sub_thread:test/unit/tsd.c:83: Failed assertion: (data_cleanup_count) >= (3) --> 0x1 < 0x3: Cleanup function should have executed multiple times.
```
2. integration test get stackdump
```
=== test/integration/cpp/basic ===
test/test.sh: line 34:  6732 Segmentation fault      (core dumped) $JEMALLOC_TEST_PREFIX ${t}.exe /home/aunyx/src/jemalloc-5.1.0/ /home/aunyx/src/jemalloc-5.1.0/
Test harness error: test/integration/cpp/basic w/ MALLOC_CONF=""
Use prefix to debug, e.g. JEMALLOC_TEST_PREFIX="gdb --args" sh test/test.sh test/integration/cpp/basic
make: *** [Makefile:496: check_integration] Error 1

aunyx@wks-home ~/src/jemalloc-5.1.0
$ JEMALLOC_TEST_PREFIX="gdb --args" sh test/test.sh test/integration/cpp/basic
=== test/integration/cpp/basic ===
GNU gdb (GDB) (Cygwin 8.0.1-1) 8.0.1
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-pc-cygwin".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from test/integration/cpp/basic.exe...done.
(gdb) run
[...]

PR #1570, opened 2019-07-26
Cygwin port

j0t:
Here is an attempt to make jemalloc work properly in CYGWIN. The issued being resolvedː
### Build
* different executable naming conventions (see [Building and Using DLLs](https://cygwin.com/cygwin-ug-net/dll.html))
* import libraries are required for proper linking
* added some `defined(__CYGWIN__)` in order to fix compile errors
* CYGWIN is some kind of 'neutral waters' territory where one can use either POSIX  or WinAPI functions, but I've choosed to be near to WinAPI. This ended up with a mix of macros: `-D_GNU_SOURCE -D_WIN32 -D_WIN32_WINNT=0x0600`.
### CYGWIN limitations and other notes
* background thread is not supported (no `RTLD_NEXT`)
* limited pthread callback level (2 vs 4), so I've decided to use `tsd_win.h`.
* mixing pthreads with WIN32 is not wise: the construct `pthread_cond_wait(&info->cond, &info->mtx.lock);` will fail since `info->mtx` under WIN32 is a `CRITICAL_SECTION` rather cygwin structure.
* MS-originated `tls_callback` machinery is not supported - this it the rationale for moving from `TlsGetValue` to `FlsGetValue` (and friends), which supports a relevant thread cleanup callbacks.
* also, there was something wrong with `BT_FRAME / ` macro in `prof_backtrace`, so I've just switched to native `RtlCaptureStackBackTrace`, which works fine.

In spite of all this, all the tests (for `x86_64 Cygwin`) are passed.
Don't know yet how it works in MINGW or VS, because I don't have ones.


j0t:
Oops, it seems I've broke everything :( (Funny, in my GNU/Linux box everything was fine)

interwq:
@j0t -- FYI it looks the AppVeyor failures are real.

j0t:
Yep, @interwq, thanks, I've seen these failures. Will try to prepare my MSYS environment and take a closer look to this.

j0t:
There is probably something wrong with my MSYS/MINGW setup. A first, I was trying to build the `dev` branch (as a baseline) with `MSYSTEM=MINGW64, CPU=x86_64`, but the compiler asks for `sys/syscall.h`, which is [obviously] not present. 
```
# mingw32-make
gcc -std=gnu11 -Wall -Wextra -Wsign-compare -Wundef -Wno-format-zero-length -pipe -g3 -O3 -funroll-loops -c -D_REENTRANT -Iinclude -I../include -DJEMALLOC_NO_PRIVATE_NAMESPACE -o src/jemalloc.sym.o ../src/jemalloc.c
In file included from include/jemalloc/internal/jemalloc_preamble.h:5,
                 from ../src/jemalloc.c:2:
../include/jemalloc/internal/jemalloc_internal_decls.h:22:14: fatal error: sys/syscall.h: No such file or directory
   22 | #    include <sys/syscall.h>
      |              ^~~~~~~~~~~~~~~
compilation terminated.
mingw32-make: *** [Makefile:389: src/jemalloc.sym.o] Error 1
```
Indeed, file `jemalloc_internal_decls.h` is trying to include `<sys/syscall.h>` when `_WIN32` is not defined:
```
#ifdef _WIN32
#  include <windows.h>
...
#else
...
#  if !defined(__pnacl__) && !defined(__native_client__)
#    include <sys/syscall.h>
'''
#endif
```
However, it seems that an succeeded AppVeyor build `MSYSTEM=MINGW64, CPU=x86_64` doesn't define `_WIN32` macro and at the same time has no such error. 
Could someone help to figure out what is going on?

j0t:
Well, I've fixed all **build** issues for AppVeyor case. What remains is the `test/unit/hook` failure which is specific for `MSVC` variant only. I suppose the safest thing I should do now is to leave the original `tsd_win.h` as it is and introduce a new `tsd_cyg.h` which uses `Fls*` functions. Then, if `_WIN32` is defined, then include `tsd_win.h` only when `_MSC_VER` is defined, otherwise use `tsd_cyg.h`.

interwq:
@j0t : could you explain more why FLS is used instead of TLS? Does `tls_callback` not work in Cygwin? Also, what happens when the applications is using fibers?

j0t:
> @j0t : could you explain more why FLS is used instead of TLS? 

Here is some discussion about releasing TLS [SO: How to release the object in a TLS-slot at thread exit on Windows?](https://stackoverflow.com/questions/18650956/how-to-release-the-object-in-a-tls-slot-at-thread-exit-on-windows). They suggests using `Fls*` instead of `Tls*`.
>Does `tls_callback` not work in Cygwin? 

Yes, exactly, `tls_callback` doesn't. It is just a reference in a special section `.CRT$...` which is very specific to MSVC CRTL. [Thread Local Storage, part 3: Compiler and linker support for implicit TLS](http://www.nynaeve.net/?p=183)
> Also, what happens when the applications is using fibers?

Do you mean, "when the applications is using fibers **explicitly**"? Have no idea at all, since my application doesn't call any fiber-specific API (except the recently introduced stuff in `tds_cyg.h`).
But if I understood it correctly, there is nothing wrong with FLS for threads [SO: FLS vs TLS, can I use Fiber Local Storage in place of TLS?](https://stackoverflow.com/questions/19613016/fls-vs-tls-can-i-use-fiber-local-storage-in-place-of-tls). 

DISCLAIMER: By no means I'm not considering yourself as an expert in WinAPI :)

interwq:
Thanks for all the details @j0t, much appreciated. Also, please do not consider me (or honestly, any of the current full-time jemalloc devs) as WinAPI experts; we really aren't :D -- the Windows improvements couldn't be done w/o the good samaritans in this community.

I see the issue with tls_callback on non-MSVC now. However, when fibers are explicitly used by the application, is it true that the per thread cache becomes per fiber?

Also, given that the changes in this PR are fairly large scoped, I'm leaning towards merging this after the 5.2.1 release. The Windows VM leak issue has been there ever since 5.0, so I'm hoping to have a clean cut just for that fix. Does that sound reasonable to you?

j0t:
> Thanks for all the details @j0t, much appreciated. Also, please do not consider me (or honestly, any of the current full-time jemalloc devs) as WinAPI experts; we really aren't :D -- the Windows improvements couldn't be done w/o the good samaritans in this community.

No prob @interwq :). It was my way, when I've discovered VM leak in #1040, I just wanted all jemalloc tests to be passed :)

> I see the issue with tls_callback on non-MSVC now. However, when fibers are explicitly used by the application, is it true that the per thread cache becomes per fiber?

I believe, yes. At the same time, it would be nice if someone could test the proposed solution with a greater attention. As we see, in Windows we have a whole zoo of various solutions: MSVC, Cygwin+GCC and MSYS+MINGW, each with own specific differences and limitations, often partially obscured and mixed together. My current setup is Cygwin+GCC and as I said before, I didn't managed to prepare a proper MSYS build environment.

> Also, given that the changes in this PR are fairly large scoped, I'm leaning towards merging this after the 5.2.1 release. The Windows VM leak issue has been there ever since 5.0, so I'm hoping to have a clean cut just for that fix. Does that sound reasonable to you?

Agreed. Indeed, it might be too quick to merge the proposed solution right now.
However, I hoping it could be useful in the future since I've seen some people having a similar troubles with TSD in Cygwin #1414.

Anyway, your recent `retain` feature for Windows is a great step toward a better compatibility, so many thanks for that.

interwq:
@j0t: we recently added buffered prof dump. Could you please rebase and update?

j0t:
@interwq, all done! CI & GIT are happy now :)

j0t:
Ups, just noticed that there are conflicts again! Will try to resolve it ;)

interwq:
@j0t : sorry for the delay on this; I certainly intend to merge this (hopefully soon).

interwq:
@j0t : I had some more discussion with @davidtgoldblatt on this. I realized that using FLS instead of TLS has implications on the thread level API we provide, e.g. [thread.arena](http://jemalloc.net/jemalloc.3.html#thread.arena) which binds a thread to a specific arena, and [thread.tcache.enabled](http://jemalloc.net/jemalloc.3.html#thread.tcache.enabled) which controls whether to use thread local cache for the calling thread.

If we switch to FLS for cygwin and the application does use fibers explicitly, it means the thread API guarantees are broken, which can be correctness issues. Of course without the fix in this PR, missing tls callbacks / leaking tls is also a correctness issue, so we do want to resolve both.

A few of additional questions:
- by any chance, are there other tls_callback alternatives in Cygwin? e.g. how do they implement destructors for C++ thread_locals.
- if FLS is still the only way feasible, can we try still implementing TLS using FLS, such as, still use TLS as the true storage, but use FLS callbacks to cleanup?

j0t:
> @j0t : I had some more discussion with @davidtgoldblatt on this. I realized that using FLS instead of TLS has implications on the thread level API we provide, e.g. [thread.arena](http://jemalloc.net/jemalloc.3.html#thread.arena) which binds a thread to a specific arena, and [thread.tcache.enabled](http://jemalloc.net/jemalloc.3.html#thread.tcache.enabled) which controls whether to use thread local cache for the calling thread.
> 
> If we switch to FLS for cygwin and the application does use fibers explicitly, it means the thread API guarantees are broken, which can be correctness issues. Of course without the fix in this PR, missing tls callbacks / leaking tls is also a correctness issue, so we do want to resolve both.

Absolutely agreed.

> A few of additional questions:
> 
> * by any chance, are there other tls_callback alternatives in Cygwin? e.g. how do they implement destructors for C++ thread_locals.

I've done some research concerned to thread storage support provided by the C/C++ language in Cygwin. Destructors of `thread_local` C++ objects are registered as callbacks by `__cxa_thread_atexit` from `libstdc++` into `__thread`-type list. The access to this list is supported by compiler, naturally, there is some layering, but it ends with calls to `TlsAlloc/Tls{Get,Set}Value` through `pthread_{get,set}specific`. 

> * if FLS is still the only way feasible, can we try still implementing TLS using FLS, such as, still use TLS as the true storage, but use FLS callbacks to cleanup?

Without being an expert, I cannot give 100% guarantee. But its true that FLS stuff is used behind the scenes in recent runtime:

[SO: How to release the object in a TLS-slot at thread exit on Windows?](https://stackoverflow.com/questions/18650956/how-to-release-the-object-in-a-tls-slot-at-thread-exit-on-windows)

> Recent versions of the runtime libraries use FLS on all operating systems where they are available (FLS was introduced in Windows Vista). There are 128 FLS slots available to the process (FLS_MAXIMUM_AVAILABLE). 

BTW, I've seen there was an attempt of rewriting `tsd_win` by using `__declspec(thread)` in #1652. Looks interesting, but is seems like as sharp turn like mine as well :)
Anyway, it might be reasonable (in the bright future) to get rid of all these explicit calls to platform-specific API and delegate all the stuff directly to compiler?

davidtgoldblatt:
Closing out older/unmergeable PRs. It's not clear that we ever got to a firm conclusion on some of the correctness questions.

PR #1571, opened 2019-07-27
Workaround g++ unused variable warnings on compare_exchange

yinan1048576:
g++ complained about set but unused variable (though it's in fact
used).  This is a workaround.


interwq:
@yinan1048576 -- let's update the commit msg as well (which will show up in the commit history).

PR #1573, opened 2019-07-29
 Limit to exact fit on Windows with retain off

interwq:
W/o retain, split and merge are disallowed on Windows.  Avoid doing first-fit
which needs splitting almost always.  Instead, try exact fit only and bail out
early.


interwq:
Tested with a CI run using `default_retain="0"` on Windows and everything was fine.

PR #1576, opened 2019-07-30
Quick fix for prof log printing

yinan1048576:
The emitter APIs used were incorrect, a side effect of which was
extra lines being printed.

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -2744,12 +2744,12 @@ prof_log_stop(tsdn_t *tsdn) {
 	emitter_init(&emitter, emitter_output_json, &prof_emitter_write_cb,
 	    (void *)(&arg));
 
-	emitter_json_object_begin(&emitter);
+	emitter_begin(&emitter);
 	prof_log_emit_metadata(&emitter);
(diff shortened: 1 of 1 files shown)


src/prof.c
```diff
@@ -2744,12 +2744,12 @@ prof_log_stop(tsdn_t *tsdn) {
 	emitter_init(&emitter, emitter_output_json, &prof_emitter_write_cb,
```
interwq: It looks this does intentionally output to json only?

interwq: and the two begin functions differ in if `emitter_json_key_prefix` is called; is that required here?

yinan1048576: `emitter_begin` and `emitter_end` are designed to be called no matter what the output format is.

`emitter_json_object_begin` and `emitter_json_object_end` are designed for printing individual json objects. Despite that the entire json is also an object, from printing point of view there are some subtle differences: `emitter_json_object_begin` assumes that there can be content before it, and `emitter_json_object_end` assumes that there can be content after it, both of which affect the handling of new lines.

PR #1577, opened 2019-08-01
Add compact json option for emitter

yinan1048576:
JSON format is largely meant for machine-machine communication, so
adding the option to the emitter.  According to local testing, the
savings in terms of bytes outputted is around 50% for stats printing
and around 25% for prof log printing.

### DIFF
--- a/include/jemalloc/internal/emitter.h
+++ b/include/jemalloc/internal/emitter.h
@@ -6,6 +6,7 @@
 typedef enum emitter_output_e emitter_output_t;
 enum emitter_output_e {
 	emitter_output_json,
+	emitter_output_json_compact,
 	emitter_output_table
 };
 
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/emitter.h
```diff
@@ -6,9 +6,13 @@
 typedef enum emitter_output_e emitter_output_t;
 enum emitter_output_e {
 	emitter_output_json,
+	emitter_output_json_compact,
 	emitter_output_table
 };
 
+#define IS_JSON(format) ((format) == emitter_output_json || \
```
davidtgoldblatt: Or just, a function returning bool?

yinan1048576:
Rebase, and revise according to feedback.

> How will the user ask for this? It seems like we want another stats printing option.

I was trying to add that, but later realized that the design there might need some further thinking and discussion, so I decided to only expand the `emitter` functionality in this PR and leave the change on stats print options to a future PR. They're also logically separate: `emitter` is a standalone utility that can be called anywhere, in stats print, prof log print, or any other place.

include/jemalloc/internal/emitter.h
```diff
@@ -75,6 +76,12 @@ struct emitter_s {
 	bool emitted_key;
 };
 
+static inline bool
+is_output_json(emitter_t *emitter) {
```
davidtgoldblatt: Slight preference for `emitter_is_json` or `emitter_outputs_json` or something, since this is nominally public. But not a hill I want to die on if you disagree.

yinan1048576: Yes, it makes sense to get the "namespace" right. Slightly prefer `emitter_outputs_json`.

PR #1580, opened 2019-08-02
Update Changelog for 5.2.1.

interwq:


interwq:
cc: @davidtgoldblatt and @jasone in case you have questions.

PR #1585, opened 2019-08-08
Fix tcaches mutex pre- / post-fork handling.

interwq:
`tcaches_mtx` isn't really related to config_prof or opt_tcache. It's all about explicit tcaches in fact.

Unconditionally does pre / post fork mutex handling.

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -778,21 +778,15 @@ tcache_boot(tsdn_t *tsdn) {
 
 void
 tcache_prefork(tsdn_t *tsdn) {
-	if (!config_prof && opt_tcache) {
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
This is an odd check that I have a Chesterton's Fence worry about it. Any idea why it was added?

interwq:
I had the same feeling. The mutex was added in 1b6e43507ed330314fffe0872f48a95a9fe502fe, and the checks were added that way already. The opt_tcache part I can see why it was included. However the config prof part is indeed strange.

I went to check the prefork checks for other modules and didn’t see any similar pattern. After a while I managed to convince myself that it was a simple mistake :)

PR #1587, opened 2019-08-08
Avoid zero the tcache avail array during tcache init.

interwq:
The tcache avail stack is always accessed / guarded with ncached (i.e. filled
with valid pointer), which means there is no need to zero it out initially.
This may save some pages for threads not accessing many sizes classes.


interwq:
This isn't what I was thinking. Closing.

PR #1589, opened 2019-08-08
Better exception handling for prof dumping

yinan1048576:
The previous logic for handling I/O exceptions in prof dumping is to
use a `propagate_err` flag to indicate whether exceptions should be
handled locally or rather propagated to the caller via the function
return value.  This PR tries to fix the code in the following
aspects:

(1) The code is not consistent in: when the `propagate_err` flag is
    off, whether the exception should still be returned after it's
    handled.  Some places still raise up the exception, while some
    other places just silence it.  The code should at least stick to
    one of the two designs.

(2) Between the two designs, the probably better one should be to
    always let the caller know of the exception.  Even if the
    exception is a transient and recoverable I/O exception, the
    caller may not want to continue on even when they know that the
    exception is already handled.

(3) The naming of the `propagate_err` flag is itself ambiguous and
    against this design since "not propagating" may be interpreted
    as not to let the caller know at all.

### DIFF
--- a/src/prof_data.c
+++ b/src/prof_data.c
@@ -420,11 +420,11 @@ prof_bt_count(void) {
 #endif
 
 static int
-prof_dump_open_impl(bool propagate_err, const char *filename) {
+prof_dump_open_impl(bool propagate_err_only, const char *filename) {
 	int fd;
 
 	fd = creat(filename, 0644);
-	if (fd == -1 && !propagate_err) {
+	if (fd == -1 && !propagate_err_only) {
 		malloc_printf("<jemalloc>: creat(\"%s\"), 0644) failed\n",
 		    filename);
 		if (opt_abort) {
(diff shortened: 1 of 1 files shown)


yinan1048576:
Rebase and revise according to feedback: agree that `handle_err` seems an equally bad name. :)

I come up with the name `propagate_err_only`. The caller-callee contract is: the callee should always propagate error; when `propagate_err_only` is false, the callee additionally do warning + aborting, but other than that the behavior should be identical to what would have happened if `propagate_err_only` is on. The "always propagate error" part of the contract was obviously violated before this change; the "identical behavior" part was also violated but more subtly. For example, before this change, when `propagate_err_only` is off, `prof_dump_write()` will still try to write till the end of the buffer is reached even when an I/O error has been encountered, and if the error is a transient error, the file content will end up being different (and messed up); the more proper way is to stop early and immediately notify the caller.

yinan1048576:
Closing - it'll be part of #1794.

PR #1591, opened 2019-08-10
Fix large bin index accessed through cache bin descriptor.

interwq:


davidtgoldblatt:
Or let's just move the definition of i into the loop?

Let's add a test? Existing ones would have caught this if "screwy tcache settings" were on of the malloc conf options we tried.

interwq:
Just updated.

re: testing -- actually I have been experimenting using pointers to represent the cache bin, i.e. keeping a cur_item ptr etc instead of the numeric ncached, in #1592. There the sanity checks helped to catch this. The benchmark results so far are very promising; I'd say it's more than likely to happen. Maybe let's not add more test configs yet?

davidtgoldblatt:
Makes sense to me

PR #1592, opened 2019-08-10
Redesign the cache bin metadata for fast path.

interwq:
Implement the pointer-based metadata for tcache bins --
- 3 pointers are maintained to represent each bin;
- 2 of the pointers are compressed on 64-bit;
- is_full / is_empty done through pointer comparison;

Comparing to the previous counter based design --
- fast-path speed up ~15% in benchmarks
- direct pointer comparison and de-reference
- no need to access tcache_bin_info in common case

Fast path micro-benchmark results improved from 15ns to 13ns. Production A/B testing shows fast path malloc cost reduction 5-7%, free cost reduction 15-17% (likely from not accessing tcache_bin_info).

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -14,7 +14,16 @@ bool	opt_tcache = true;
 ssize_t	opt_lg_tcache_max = LG_TCACHE_MAXCLASS_DEFAULT;
 
 cache_bin_info_t	*tcache_bin_info;
-static unsigned		stack_nelms; /* Total stack elms per tcache. */
+/*
+ * For the total bin stack region (per tcache), reserve 2 more slots so that 1)
+ * the empty position can be safely read on the fast path before checking
+ * "is_empty"; and 2) the low_water == -1 case can go beyond the empty position
+ * by 1 step safely (i.e. no overflow).
+ */
+static const unsigned total_stack_padding = sizeof(void *) * 2;
+
+/* Total stack size required (per tcache).  Include the padding above. */
+static uint32_t total_stack_bytes;
 
 unsigned		nhbins;
 size_t			tcache_maxclass;
@@ -47,14 +56,16 @@ tcache_event_hard(tsd_t *tsd, tcache_t *tcache) {
 	} else {
 		tbin = tcache_large_bin_get(tcache, binind);
 	}
-	if (tbin->low_water > 0) {
+
+	cache_bin_sz_t low_water = cache_bin_low_water_get(tbin, binind);
+	cache_bin_sz_t ncached = cache_bin_ncached_get(tbin, binind);
+	if (low_water > 0) {
 		/*
 		 * Flush (ceiling) 3/4 of the objects below the low water mark.
 		 */
 		if (binind < SC_NBINS) {
 			tcache_bin_flush_small(tsd, tcache, tbin, binind,
-			    tbin->ncached - tbin->low_water + (tbin->low_water
-			    >> 2));
+			    ncached - low_water + (low_water >> 2));
 			/*
 			 * Reduce fill count by 2X.  Limit lg_fill_div such that
 			 * the fill count is always at least 1.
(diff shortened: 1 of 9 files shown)


interwq:
Based on #1594. Only the first commit is this PR.

Although not shown in the commit history, initially I kept both the new pointer based implementation and the existing ncached etc, and checked that they always match before removing the old counters.

src/tcache.c
```diff
@@ -502,8 +585,21 @@ tcache_create_explicit(tsd_t *tsd) {
...
+		idalloctm(tsd_tsdn(tsd), tcache, NULL, NULL, true, true);
+		size = sz_sa2u(size + cache_bin_max_offset, CACHELINE);
+		tcache = ipallocztm(tsd_tsdn(tsd), size, CACHELINE, true, NULL,
+		    true, arena_get(TSDN_NULL, 0, true));
+		if (tcache == NULL) {
+			return NULL;
+		}
+		avail_array = (void *)((uintptr_t)tcache +
+		    (uintptr_t)stack_offset);
+	}
```
yinan1048576: Add `assert(!tcache_bin_lowbits_overflowable(...))`?

src/tcache.c
```diff
@@ -416,44 +482,61 @@ tcache_init(tsd_t *tsd, tcache_t *tcache, void *avail_stack) {
...
+	if (tcache_bin_lowbits_overflowable(avail_array, total_stack_bytes)) {
+		idalloctm(tsd_tsdn(tsd), avail_array, NULL, NULL, true, true);
+		/* Add padding so the overflowable range can be skipped. */
+		size = sz_sa2u(size + cache_bin_max_offset, CACHELINE);
+		avail_array = ipallocztm(tsd_tsdn(tsd), size, CACHELINE, true,
+		    NULL, true, arena_get(TSDN_NULL, 0, true));
+		if (avail_array == NULL) {
+			return true;
+		}
+	}
```
yinan1048576: Add `assert(!tcache_bin_lowbits_overflowable(...))`?

interwq: Here the total space + total size may still appear overflowable; but in that case there will be enough padding for us to skip a bin's space that may overflow the low bits. The assertion here https://github.com/jemalloc/jemalloc/pull/1592/files#diff-777e740de5d78972a2742086c8dc7f67R514 should verify that case.

interwq:
re: testing -- I don't think there are any existing tests specifically for cache bin or tcache. Let me add a new one.

interwq:
Added `test/unit/cache_bin`.

src/tcache.c
```diff
@@ -406,6 +429,49 @@ tsd_tcache_enabled_data_init(tsd_t *tsd) {
...
+	 * will access the slots toward higher addresses (for the benefit of
+	 * adjacent prefetch).
+	 */
+	void *full_position = (void *)*stack_cur;
+	/*
+	 * Check with max_offset intentionally.  It is more conservative than
+	 * necessary (bin_stack_size + 1 ptr), where the extra 1 is for
+	 * low_water (and is easy to miss).
+	 */
+	if (tcache_bin_lowbits_overflowable(full_position,
```
davidtgoldblatt: Why do we need to check for overflow?

It seems like we can do an equality check on the low-water mark (and change the meaning of low-water slightly) instead. The current strategy wastes tcache space, right?

interwq: The pointer compression (for low_water and full_position) needs the guarantee that the space do not cross the 32-bit / 4G boundary.

When we allocate the total space for the stacks (https://github.com/jemalloc/jemalloc/pull/1592/files#diff-777e740de5d78972a2742086c8dc7f67R530), we check if the low bits are overflowable and if so add some padding. And then when initializing the individual stacks, add the padding for the specific stack -- so yep that tcache stack space is wasted in that case. It rarely happens though.

re: checking with low-water vs max_offset -- I intentionally made this conservative with max_offset. It's true that stack_size + sizeof(ptr) (i.e. the limit of low_water) is sufficient; however in case we change the limit in the future, it's easy to miss the change here (plus we don't really save much from it).

davidtgoldblatt: > The pointer compression (for low_water and full_position) needs the guarantee that the space do not cross the 32-bit / 4G boundary.

What I mean is, if we make it an equality check, we don't need that guarantee (right?). I.e. we can make the algorithm resilient to overflow, so we don't need to allocate extra padding.

davidtgoldblatt: Actually, come to think of it, any reason why we can't align the cache bin allocation so that overflow is impossible to begin with?

interwq: > What I mean is, if we make it an equality check, we don't need that guarantee (right?). I.e. we can make the algorithm resilient to overflow, so we don't need to allocate extra padding.

You mean we always check with `== low_water`? That should be possible; but does feel pretty risky though.

> Actually, come to think of it, any reason why we can't align the cache bin allocation so that overflow is impossible to begin with?

Good point. What do you think about, when allocating the total stack space, align with the next power of 2 ceiling of total_stack_size?

interwq: Just updated: added alignment to avoid low bits overflow.

PR #1595, opened 2019-08-12
Pull thread_(de)allocated out of config_stats

yinan1048576:
Make `thread_allocated` and `thread_deallocated` modified independently of `config_stats`.  A key purpose longer term is to make `thread_allocated` as a universal allocation size accumulator, so that it is the single counter malloc() fast path touches, and its accumulation is bulk transferred to other counters in slow path.


src/jemalloc.c
```diff
@@ -2339,11 +2325,11 @@ je_malloc(size_t size) {
 	}
 
 	szind_t ind = sz_size2index_lookup(size);
-	size_t usize;
-	if (config_stats || config_prof) {
-		usize = sz_index2size(ind);
-	}
-	/* Fast path relies on size being a bin. I.e. SC_LOOKUP_MAXCLASS < SC_SMALL_MAXCLASS */
+	size_t usize = sz_index2size(ind);
```
interwq: nit: Let's add some comments, about usize being computed unconditionally.

yinan1048576:
Good catch! `grep`ed `config_stats` and confirmed that there's no more place in the file.

yinan1048576:
More comments.

yinan1048576:
Rebase plus minor modifications. The main difference is: in my previous revision I was trying to extract all the things out of the `config_stats` flag, but I realized that the correct approach should be only focusing on extracting the `thread_(de)allocated` out of `config_stats` flag. One visible difference is that I'm now no longer incrementing `bin->tstats.nrequests` when `config_stats` is off, which is the correct thing to do. The only extra thing I'm extracting out from `config_stats` is the setting of `dopts->usize` in `imalloc_body()`: I think this is the right thing to do, and not setting it when `config_stats` is off (as we did earlier) may result in subtle bugs.

interwq:
Let's try a full `run_tests`? There I believe we do have many cases with stats off.

yinan1048576:
All tests passed. :)

src/jemalloc.c
```diff
@@ -2063,12 +2060,9 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd) {
 		if (unlikely(ind >= SC_NSIZES)) {
 			goto label_oom;
 		}
-		if (config_stats || (config_prof && opt_prof) || sopts->usize) {
-			usize = sz_index2size(ind);
-			dopts->usize = usize;
-			assert(usize > 0 && usize
-			    <= SC_LARGE_MAXCLASS);
-		}
+		usize = sz_index2size(ind);
```
interwq: Can we get rid of the 0 init of usize @ L2040 above this section? That way compiler should be able to warn us if its used w/o setting.

src/jemalloc.c
```diff
@@ -2339,11 +2325,12 @@ je_malloc(size_t size) {
 	}
 
 	szind_t ind = sz_size2index_lookup(size);
-	size_t usize;
-	if (config_stats || config_prof) {
-		usize = sz_index2size(ind);
-	}
-	/* Fast path relies on size being a bin. I.e. SC_LOOKUP_MAXCLASS < SC_SMALL_MAXCLASS */
+	/* usize is always needed to increment thread_allocated. */
```
interwq: nit: Let's beef up the comments here regarding why we always increment thread.allocated.

PR #1596, opened 2019-08-14
Store stack_size in tcache_bin_info.

interwq:
With the cache bin metadata switched to pointers, ncached_max is usually
accessed and timed by sizeof(ptr). Store the results in tcache_bin_info for
direct access, and add a helper function for the ncached_max value.

On top of #1592


interwq:
Ready for review -- this is the commit https://github.com/jemalloc/jemalloc/pull/1596/commits/7f9a5f8a729a5439c44d81ba0d13b2f7392f6474

interwq:
@yinan1048576 : yep, actually `stack_nelms` is removed here https://github.com/jemalloc/jemalloc/pull/1596/files#diff-777e740de5d78972a2742086c8dc7f67L17

Issue #1511, opened 2019-05-22
Add per background thread mutex stats


interwq:
While debugging #1502, I realized that the per thread lock is not included in malloc_stats. It may not worth reporting the lock stats of all threads (e.g. 4 threads means 28 more counters), but report the max should be helpful.

PR #1597, opened 2019-08-14
Add max_per_bg stats to reflect the maximum of counters for background thread mutexes.

zhxchen17:
Added a new stats row to aggregate the maximum value of each counters for background
threads. https://github.com/jemalloc/jemalloc/issues/1511


interwq:
Refactored in #1619.

PR #1599, opened 2019-08-16
Add hint in the bogus version string.

interwq:


configure.ac
```diff
@@ -584,6 +584,7 @@ if test ! -e "${objroot}VERSION" ; then
     AC_MSG_RESULT(
       [Missing VERSION file, and unable to generate it; creating bogus VERSION])
     echo "0.0.0-0-g0000000000000000000000000000000000000000" > "${objroot}VERSION"
+    echo "0.0.0-0-g0000no0version0000try0git0fetch0tags0000" > "${objroot}VERSION"
```
davidtgoldblatt: Might be easier to read with underscores instead of 0s as the space delimiter? Or `.` characters, if we want it to match `[a-z0-9.-]*`

interwq: Yeah I wanted to make it a “valid” hash, since the last string is supposed to be the commit hash from “git describe”, in case some parser expects it. Or you think we should break it there intentionally?

davidtgoldblatt: I would assume a parser would expect hex there, rather than anything alphanumeric. Although, yeah, maybe just flat-out breaking it wouldn't be the worst thing in the world either.

interwq: Agreed. Switched to underscores instead.

PR #1603, opened 2019-08-21
Track low_water == -1 case explicitly.

interwq:
The -1 value of low_water indicates if the cache has been depleted and
refilled.  Track the status explicitly in the tcache struct.

This allows the fast path to check if (cur_ptr > low_water), instead of >=,
which avoids reaching slow path when the last item is allocated.


interwq:
After this change, the cache_bin_sz_t doesn’t have to be signed anymore. Will try to change that in a future PR.

PR #1604, opened 2019-08-21
Bug fix for prof_active switch

yinan1048576:
The bug is subtle but critical: if application performs the following
three actions in sequence: (a) turn `prof_active` off, (b) make at
least one allocation that triggers the malloc slow path via the
`if (unlikely(bytes_until_sample < 0))` path, and (c) turn
`prof_active` back on, then the application would never get another
sample (until a very very long time later).

The fix is to properly reset `bytes_until_sample` rather than
throwing it all the way to `SSIZE_MAX`.

A side minor change is to call `prof_active_get_unlocked()` rather
than directly grabbing the `prof_active` variable - it is the very
reason why we defined the `prof_active_get_unlocked()` function.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2356,13 +2356,15 @@ je_malloc(size_t size) {
 			/*
 			 * Avoid a prof_active check on the fastpath.
 			 * If prof_active is false, set bytes_until_sample to
-			 * a large value.  If prof_active is set to true,
+			 * sampling interval.  If prof_active is set to true,
 			 * bytes_until_sample will be reset.
 			 */
-			if (!prof_active) {
-				tsd_bytes_until_sample_set(tsd, SSIZE_MAX);
+			if (!prof_active_get_unlocked()) {
+				tsd_bytes_until_sample_set(tsd,
+				    ((uint64_t)1U << lg_prof_sample));
+			} else {
+				return malloc_default(size);
 			}
-			return malloc_default(size);
 		}
 	}
 


yinan1048576:
While working on fixing this bug, I was thinking: an important reason why this bug was not identified earlier was the lack of test coverage, in two aspects:
- There's no test that performed the three actions (in my description) in sequence - I think it is hard to exhaust everything and there's no easy solution here.
- The malloc fast path is not well covered by tests - it's actually an interesting dilemma: in debug build malloc always routes to slow path here: https://github.com/jemalloc/jemalloc/blob/0043e68d4c54a305d84ead95cae27a730540451b/src/jemalloc.c#L2332 so basically the whole body of the malloc fast path function is undebuggable. I think we should have a way to solve it - it feels like a simpler problem to solve than the first one.

src/jemalloc.c
```diff
@@ -2359,8 +2359,11 @@ je_malloc(size_t size) {
 			 * a large value.  If prof_active is set to true,
 			 * bytes_until_sample will be reset.
 			 */
```
interwq: We need to update the comments. Any chance we can still keep the optimization (set to size_max) and find a way to reset it when prof active is changed?

yinan1048576: That means we'll need to go through all the threads and update their `bytes_until_sample` when `prof_active` is flipped from `false` to `true`?

interwq: I’d prefer we don’t go that far. An alternative is to add a check on some slow path; anywhere you think would fit?

yinan1048576: How about we don't route to the slow path but simply continue on in the fast path (after increasing `bytes_until_sample` back)? After all there's no need to go to the slow path. The cost would be an extra `prof_active` check and an extra `bytes_until_sample` update.

yinan1048576: I feel adding a check on the slow path is doable, and a good choice can be the GC ticker path. (Other paths to slow path all seem to have the possibility of not being triggered after a very long time.) If we rely on the ticker, then the wrong `bytes_until_sample` can be fixed within a known number of malloc calls, though within these calls the total bytes allocated can still be huge.

interwq: Yeah I think the complication of the slow path update isn’t worth it. 

We should still try optimizing a bit for the prof inactive case though, maybe set it to the sample interval? The current diff will frequently go down the reset path.

interwq: Also are we double counting when going down malloc_default? It looks we already adjusted bytes until sample.

interwq: I think the logic was added recent-ish in one of the fast path optimizations. @yinan1048576 let’s create an issue to track it (I bet it’s released
already).

yinan1048576:
I'm now resetting `bytes_until_sample = 1 << lg_prof_sample`, so that the rate of falling to the slow path is the same as when `prof_active` was on. Further, I'm not really letting it fall to the slow path, but instead I'm letting the execution flow continue on the fast path, so that it's more a really fast "slow path". I believe this is correct - even if we were resetting `bytes_until_sample = SSIZE_MAX`, we should have continued on the fast path.

One other minor remark is that I'm directly resetting via `tsd_bytes_until_sample_set(tsd, ((uint64_t)1U << lg_prof_sample))` rather than doing the more correct thing:
```
    prof_tdata_t *tdata = tsd_prof_tdata_get(tsd);
    if (tdata != NULL) {
        prof_sample_threshold_update(tdata);
    }
```
The only difference is that the reset is not randomized, but we probably don't need that anyway when `prof_active` is off. Also the `prof_sample_threshold_update()` call would take much longer.

src/jemalloc.c
yinan1048576: > We should still try optimizing a bit for the prof inactive case though, maybe set it to the sample interval? The current diff will frequently go down the reset path.

Great point! Doing that.

> Also are we double counting when going down malloc_default? It looks we already adjusted bytes until sample.

Yes we're double counting on the slow path. But I guess the impact would be minimal, since in the majority of the cases the execution won't go into the slow path. I guess there would also be some performance concerns, since we would probably need to do something extra on the fast path to indicate to the slow path that `bytes_until_sample` has already been set?

> let’s create an issue to track it (I bet it’s released already).

Yes it's released. Created #1605.

PR #1613, opened 2019-08-27
Workaround the stringop-overflow check false positives.

### LINKED ISSUE #1601: run_test fails due to -Werror (stringop-overflow)
The run_test has log.c with line 53, etc.

	log_var_t log_l1 = LOG_VAR_INIT("l1");
        ...
	strcpy(log_var_names, "l1");

I think the intent is to verify that jemalloc detects a buffer overwrite in this situation.  However, the code fails to compile with some newer GCC as the option 'stringop-overflow' detects the overwrite at compile time and the log.c fails to build aborting the test.

interwq:
Resolves #1601

### DIFF
--- a/test/src/test.c
+++ b/test/src/test.c
@@ -232,3 +232,16 @@ p_test_fail(const char *prefix, const char *message) {
 	malloc_cprintf(NULL, NULL, "%s%s\n", prefix, message);
 	test_status = test_status_fail;
 }
+
+void
+strncpy_cond(void *dst, const char *src, bool cond) {
+	if (cond) {
+		/*
+		 * Avoid strcpy and explicitly set length to 0 because the
+		 * `stringop-overflow` check may warn even if the specific test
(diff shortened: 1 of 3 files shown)


test/src/test.c
```diff
@@ -232,3 +232,16 @@ p_test_fail(const char *prefix, const char *message) {
 	malloc_cprintf(NULL, NULL, "%s%s\n", prefix, message);
 	test_status = test_status_fail;
 }
+
+void
+strncpy_cond(void *dst, const char *src, bool cond) {
+	if (cond) {
```
yinan1048576: I can't see why this `if` branch is necessary.

interwq: @yinan1048576 : This condition and the size n below are both workarounds for the false positives of the `stringop-overflow` check. The actual code will never be reached when cond == false. Since this is test code, I went extra conservative and included both.

PR #1614, opened 2019-08-27
Guard prof_active reset by opt_prof

yinan1048576:
Set `prof_active` to read-only when `opt_prof` is turned off.


yinan1048576:
Added unit tests to confirm the intended behavior.

yinan1048576:
Fixed test failure.

PR #1616, opened 2019-08-29
Use thread_allocated to count prof sampling

yinan1048576:


yinan1048576:
The goal is to get rid of `bytes_until_sample`, and instead use `thread_allocated` and a new `thread_allocated_sample` (replacing `bytes_until_sample`) to count for prof sampling. A few remarks:

- Previously `bytes_until_sample` may double count allocation bytes, either in profiling slow path, or in `tcache` slow path. The former luckily doesn't matter, since `bytes_until_sample` gets below zero and will be reset later anyway, but the latter *does* matter. The reason underlying such complication is that `bytes_until_sample` is changed in the middle of the execution flow, whereas the slow path can be routed to either before or after it takes place: neither changing nor keeping `bytes_until_sample` would be right in the middle of the fast path, unless there's an additional indicator specifying to the slow path whether `bytes_until_sample` has been changed or not in the fast path, which would also add a few cycles to the fast path.
- The way to solve such a complication is to either change `bytes_until_sample` before any slow path routing takes place (i.e. the root node), or at the end of every execution flow (i.e. all leaf nodes), including at the end of all slow path executions and at the end of the fast path execution (after all possible slow path routings). The root node approach is much simpler but it seems hard to implement, since `bytes_until_sample` needs at least a valid `tsd`, which is already after a few slow path routings. On the other hand, `thread_allocated` already faithfully adopts the leaf node approach (and thus never runs into double counting). So my change not only replaces `bytes_until_sample` by `thread_allocated`, but also sticks to its leaf node approach, so that the previous double counting issue of `bytes_until_sample` is simultaneously fixed.
- In the leaf node approach, `thread_allocated` is treated as read-only throughout the execution of both fast path and slow path, and would only be manually rolled over when an overflow issue is gonna happen for sure. The new `thread_allocated_sample` acts more similar to `bytes_until_sample`, in that it is what `prof_sample_threshold_update()` would reset, *not* `thread_allocated`.
- An overflow issue is one of the following two cases. First, `thread_allocated_sample` overflows, in which case `thread_allocated` would excessively trigger sampling. Second, `thread_allocated_sample` hasn't overflowed, but if the fast path only checks `thread_allocated + usize > thread_allocated_sample`, the `thread_allocated + usize` part can overflow, resulting in over suppression of sampling. The solution is to always limit `thread_allocated_sample` to be at most `UINT64_MAX - SC_LOOKUP_MAXCLASS`, and if `thread_allocated` is guaranteed to never exceed `thread_allocated_sample`, `thread_allocated + usize` never overflows in fast path, so that the sampling check in the fast path can be quick.
- A tricky part in the implementation is that `prof_sample_threshold_update()` not only needs to reset `thread_allocated_sample` to be a sampling interval above `thread_allocated`, but also further delay `thread_allocated_sample` by `usize`. The reason is that `thread_allocated` is read-only at this stage and is incremented when the allocation finishes. The further delay is to make sure that `thread_allocated_sample` is a sampling interval above `thread_allocated` at the end. The only downside is that `thread_allocated` won't be advanced if the allocation fails, in which case the sampling interval would be slightly larger than what it should be, but this should only happen rarely.
- An even trickier (taking >90% time from me) but eventually pretty nice part in the implementation is: `prof_sample_check()` no longer needs the boolean `update` flag. Previously the only places it needs the `update` flag (i.e. a `false` flag for suppressing update) are `rallocx` and `xallocx`, where the initial `prof_alloc_prep()` call has `update` as `false` so that the sample would not be mistakenly `prof_alloc_rollback()`ed later in the execution. Now `thread_allocated` is read-only and would never be updated until the end, when it would be "automatically" updated, so `prof_sample_check()` no longer needs to manage the updates.

yinan1048576:
Discussed offline. The other design should have little overlap with what I have now. Closing.

PR #1617, opened 2019-08-30
Make cache_bin_sz_t unsigned.

interwq:
The bin size type was made signed only because the low_water could go -1, which
was already removed.

### DIFF
--- a/include/jemalloc/internal/cache_bin.h
+++ b/include/jemalloc/internal/cache_bin.h
@@ -13,12 +13,8 @@
  * of the tcache at all.
  */
 
-/*
- * The count of the number of cached allocations in a bin.  We make this signed
- * so that negative numbers can encode "invalid" states (e.g. a low water mark
- * of -1 for a cache that has been depleted).
- */
-typedef int32_t cache_bin_sz_t;
+/* The size in bytes of each cache bin stack. */
+typedef uint16_t cache_bin_sz_t;
 
 typedef struct cache_bin_stats_s cache_bin_stats_t;
 struct cache_bin_stats_s {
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/cache_bin.h
```diff
@@ -18,7 +18,7 @@
  * so that negative numbers can encode "invalid" states (e.g. a low water mark
  * of -1 for a cache that has been depleted).
  */
-typedef int32_t cache_bin_sz_t;
+typedef uint16_t cache_bin_sz_t;
```
yinan1048576: Also change the comment?

include/jemalloc/internal/cache_bin.h
```diff
@@ -163,7 +163,7 @@ cache_bin_low_water_get(cache_bin_t *bin, szind_t ind) {
 	cache_bin_sz_t ncached_max = cache_bin_ncached_max_get(ind);
 	cache_bin_sz_t low_water = ncached_max -
 	    (bin->low_water_position - bin->full_position) / sizeof(void *);
-	assert(low_water >= 0 && low_water <= ncached_max);
+	assert(low_water <= ncached_max);
```
yinan1048576: Fix the comment for this function?

include/jemalloc/internal/cache_bin.h
interwq: Good catch! Fixed.

include/jemalloc/internal/cache_bin.h
interwq: Done.

PR #1620, opened 2019-09-04
Fix libc reentrancy level

yinan1048576:
`_Unwind_Backtrace` is most likely not calling `malloc()`/`free()`, but from correctness aspect we should have a guard rather than relying only on trust. From behavior aspect `reentrancy_level` is also important: for example, profiling is turned off when `reentrancy_level > 0`.

### DIFF
--- a/include/jemalloc/internal/prof_externs.h
+++ b/include/jemalloc/internal/prof_externs.h
@@ -56,7 +56,7 @@ void prof_malloc_sample_object(tsdn_t *tsdn, const void *ptr, size_t usize,
 void prof_free_sampled_object(tsd_t *tsd, const void *ptr, size_t usize,
     prof_tctx_t *tctx);
 void bt_init(prof_bt_t *bt, void **vec);
-void prof_backtrace(prof_bt_t *bt);
+void prof_backtrace(tsd_t *tsd, prof_bt_t *bt);
 prof_tctx_t *prof_lookup(tsd_t *tsd, prof_bt_t *bt);
 #ifdef JEMALLOC_JET
 size_t prof_tdata_count(void);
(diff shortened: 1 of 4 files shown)


src/prof.c
```diff
@@ -256,7 +256,10 @@ prof_backtrace(prof_bt_t *bt) {
 
 	cassert(config_prof);
 
+	tsd_t *tsd = tsd_fetch();
```
interwq: Let’s pass tsd down to here?

Also we should add the pre and post reentrancy for all backtrace impls (add shared begin / finish functions) not just the libc one.

yinan1048576:
Revised according to feedback. (Added a wrapper function, rather than a begin function and an end function.)

src/prof.c
```diff
@@ -423,12 +423,20 @@ prof_backtrace(prof_bt_t *bt) {
 }
 #else
 void
-prof_backtrace(prof_bt_t *bt) {
```
interwq: Let’s make the impl functions static.

include/jemalloc/internal/prof_externs.h
```diff
@@ -56,7 +56,7 @@ void prof_malloc_sample_object(tsdn_t *tsdn, const void *ptr, size_t usize,
 void prof_free_sampled_object(tsd_t *tsd, const void *ptr, size_t usize,
     prof_tctx_t *tctx);
 void bt_init(prof_bt_t *bt, void **vec);
-void prof_backtrace(prof_bt_t *bt);
+void prof_backtrace(prof_bt_t *bt, tsd_t *tsd);
```
interwq: nit: we usually put tsd first in the arg.

yinan1048576:
Revise according to feedback.

Issue #1494, opened 2019-04-23
Make `prof.prefix` writable? 


interwq:
People mentioned they want to change `prof.prefix` after fork for connivence (e.g. to make higher level tooling working properly around plain `prof.dump`). It's possible to always specify the prefix at the tooling level; however I don't see any downside of updating the prefix directly.

PR #1623, opened 2019-09-08
Add `prof.dump_prefix` to override filename prefixes for dumps.

zhxchen17:
As #1494 mentioned, in some cases it might be helpful to provide another option to override the read-only `opt.prof_prefix`. This patch adds a write-only option `prof.dump_prefix` to address the needs to this.
Implementation notes: Most places where `opt.prof_prefix` is read to determine the dumped filename is changed to a single getter function `opt_prof_dump_prefix()`. There is one extra guard added to protect the access to internal `prof_dump_prefix` buffer.


src/prof.c
```diff
@@ -106,6 +108,9 @@ static uint64_t		prof_dump_mseq;
 static uint64_t		prof_dump_useq;
 
 malloc_mutex_t	prof_dump_mtx;
+static char	prof_dump_prefix[PROF_DUMP_PREFIX_LEN] = {0};
+static bool	prof_dump_prefix_override = false;
+static malloc_mutex_t	prof_dump_prefix_mtx;
```
interwq: I would think using `prof_dump_mtx` as the guard should be fine? cc: @yinan1048576 for sanity check.

yinan1048576: Hmm. If we want to reuse an existing mutex, I'd think `prof_dump_seq_mtx` is functionally closer. Both the prefix and the sequence numbers are for the file names. (We can probably change the mutex's name to be something better.)

zhxchen17: I actually think `prof_dump_seq_mtx` is more similar, but I didn't think of using it, because I just ignored I can actually use a coarse-grained lock for multiple shared values.  Maybe we can rename it to be something like `prof_dump_filename_mtx` first, then use that in this feature? @yinan1048576

yinan1048576: `prof_dump_filename_mtx` sounds good to me. @interwq thoughts?

I remember `prof_dump_seq_mtx` might have already been used for multiple values, e.g. those different kinds of sequence numbers, so adding one more doesn't seem like crazy.

I think another benefit of consolidating `prof_dump_seq_mtx` with `prof_dump_prefix_mtx` is that you don't need to always try to acquire both of them when getting the file name. Feels cleaner and safer.

interwq: SGTM!

src/prof.c
```diff
@@ -571,6 +606,14 @@ prof_accum_init(tsdn_t *tsdn, prof_accum_t *prof_accum) {
 	return false;
 }
 
+void
+prof_dump_prefix_set(tsdn_t *tsdn, const char *prefix) {
+	malloc_mutex_lock(tsdn, &prof_dump_filename_mtx);
+	prof_dump_prefix_override = true;
+	strncpy(prof_dump_prefix, prefix, PROF_DUMP_PREFIX_LEN - 1);
```
interwq: After strncpy, we should enforce a null terminator at the end, in case the original prefix doesn't terminate.

src/prof.c
```diff
@@ -514,40 +518,71 @@ prof_getpid(void) {
...
 		    "%s.%d.%"FMTu64".%c.heap",
-		    opt_prof_prefix, prof_getpid(), prof_dump_seq, v);
+		    prof_prefix, prof_getpid(), prof_dump_seq, v);
 	}
 	prof_dump_seq++;
 }
 
+void
+prof_get_default_json(tsdn_t *tsdn, char *filename, uint64_t ind) {
+	size_t buf_size = PATH_MAX + 1;
```
interwq: nit: use PROF_DUMP_PREFIX_LEN?

include/jemalloc/internal/prof_externs.h
```diff
@@ -72,10 +72,12 @@ void prof_cnt_all(uint64_t *curobjs, uint64_t *curbytes, uint64_t *accumobjs,
     uint64_t *accumbytes);
 #endif
 int prof_getpid(void);
+void prof_get_default_json(tsdn_t *tsdn, char *filename, uint64_t ind);
```
interwq: nit: `prof_get_default_filename`

src/prof.c
```diff
@@ -106,6 +108,8 @@ static uint64_t		prof_dump_mseq;
 static uint64_t		prof_dump_useq;
 
 malloc_mutex_t	prof_dump_mtx;
+static char	prof_dump_prefix[PROF_DUMP_PREFIX_LEN] = {0};
```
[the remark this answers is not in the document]
zhxchen17: @interwq When I'm trying to use `base_alloc` in the critical section of `prof_dump_filename_mtx`, the mutex ranks are sort of inverted (base is much lower than prof_dump_filename). What could I do to tackle this?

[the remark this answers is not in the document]
interwq: Ah that’s right. We can lock — check if it’s empty — , and if so unlock then alloc with iallocztm, and lock again then try setting it. If some other threads beat us and created it already, unlock and then free the buffer with idalloctm.

Just realized this can only happen in through mallctl, right? This means everything is under the ctl mtx, so we don’t need to worry about allocating the buffer and then deallocate. Try base alloc the buffer (and assert we hold ctl mtx), and assume we don’t need to deallocate should be fine.

Issue #957, opened 2017-07-19
scripts/gen_run_tests.py assumes make is GNU make


emaste:
On FreeBSD make is BSD make, and GNU make is installed as gmake

We can probably assume that if gmake exists in the path it should be preferred, without having to resrt to OS-specific cases.

PR #1627, opened 2019-09-11
check GNU make exists in path

GiridharPrasath:
Signed-off-by: Giridhar Prasath R <cristianoprasath@gmail.com>

Refers: #957


interwq:
Thanks for the patch!

PR #1634, opened 2019-09-22
Split out some extent module bits

davidtgoldblatt:
A whole lot of stuff lives in the extent module now. This starts to split it up.

The split is:
- `extent.h`: The `extent_t` struct itself.
- `eset.h`, `eset.c`: Short for "extent set"; a renaming of `extents_t` that's easier to verbally distinguish from `extent_t`. This is just the container-y portions of the current module. It selects the extent to use for a given allocation request (and so does first-fit, maintains LRU ordering, etc.), but doesn't actually modify the OS-visible state of any of its contents.
- `extent_externs.h`, `extent_types.h`, `extent_structs.h`, `extent.c`: What's left. Mostly this is responsible for the actual OS state changes (stuff like calling extent hooks, purging, contention management, etc.).

I think that there's a few more well-scoped chunks we could pull out:
- An emap ("extent map") module, which handles the translation from addresses to `extent_t`s and manages the associated state (and so keeps the rtree, and answers questions like "are these extents mergeable", without actually doing the merging).
- The extent_avail cache.
- The extent hook infrastructure.
- What's left would be just the code that makes OS-state decisions (doing stuff like purging, actually invoking extent hooks, etc.).


include/jemalloc/internal/sz.h
```diff
@@ -315,4 +315,51 @@ sz_sa2u(size_t size, size_t alignment) {
 	return usize;
 }
 
+static inline size_t
+sz_psz_quantize_floor(size_t size) {
```
interwq: nit: maybe not inline sz_psz_quantize_floor and _ceil, i.e. move to sz.c?

src/extent.c
```diff
@@ -249,86 +249,8 @@ extent_hooks_assure_initialized(arena_t *arena,
...
-	 * atomic operations for updating eset->npages.
-	 */
-	size_t cur_extents_npages =
-	    atomic_load_zu(&eset->npages, ATOMIC_RELAXED);
-	assert(cur_extents_npages >= npages);
-	atomic_store_zu(&eset->npages,
-	    cur_extents_npages - (size >> LG_PAGE), ATOMIC_RELAXED);
-}
+ph_gen(, extent_heap_, extent_heap_t, extent_t, ph_link,
+    extent_snad_comp)
```
interwq: nit: no need for the line break

davidtgoldblatt:
Per comments.

PR #1639, opened 2019-10-03
Suppress tdata creation in reentrancy

yinan1048576:
This change suppresses tdata initialization and prof sample threshold
update in interrupting malloc calls.  Interrupting calls have no need
for tdata. Delaying tdata creation aligns better with our lazy tdata
creation principle, and it also helps us gain control back from
interrupting calls more quickly and reduces any risk of delegating
tdata creation to an interrupting call.

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -127,9 +127,14 @@ prof_tctx_should_destroy(tsdn_t *tsdn, prof_tctx_t *tctx) {
 
 void
 prof_alloc_rollback(tsd_t *tsd, prof_tctx_t *tctx, bool updated) {
-	prof_tdata_t *tdata;
-
 	cassert(config_prof);
+
+	if (tsd_reentrancy_level_get(tsd) > 0) {
+		assert((uintptr_t)tctx == (uintptr_t)1U);
+		return;
+	}
+
+	prof_tdata_t *tdata;
 
 	if (updated) {
 		/*
@@ -810,6 +815,8 @@ prof_active_set(tsdn_t *tsdn, bool active) {
 
 const char *
 prof_thread_name_get(tsd_t *tsd) {
+	assert(tsd_reentrancy_level_get(tsd) == 0);
+
 	prof_tdata_t *tdata;
 
 	tdata = prof_tdata_get(tsd, true);
@@ -821,6 +828,8 @@ prof_thread_name_get(tsd_t *tsd) {
 
 int
 prof_thread_name_set(tsd_t *tsd, const char *thread_name) {
+	assert(tsd_reentrancy_level_get(tsd) == 0);
+
 	prof_tdata_t *tdata;
 	unsigned i;
 	char *s;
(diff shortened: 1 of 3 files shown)


src/prof.c
```diff
@@ -131,7 +131,7 @@ prof_alloc_rollback(tsd_t *tsd, prof_tctx_t *tctx, bool updated) {
 
 	cassert(config_prof);
 
-	if (updated) {
+	if (updated && tsd_reentrancy_level_get(tsd) == 0) {
```
interwq: Can we reach here during reentrancy?

yinan1048576: I tried adding `assert(tsd_reentrancy_level_get(tsd) == 0);` and nothing broke.

I think logically `prof_alloc_rollback()` shouldn't be called by interrupting malloc, but our implementation doesn't forbid it.

First, my feeling is that the logic here for obtaining `tdata` and resetting the sampling threshold can be put under the `(uintptr_t)tctx > (uintptr_t)1U` branch below it. Thoughts?

If we keep the current logic unchanged, then by design the `tdata` creation and threshold reset part needs to be robust to the `(uintptr_t)tctx == (uintptr_t)1U` case, i.e. the no sample case, which is what's gonna happen in interrupting malloc calls. In `jemalloc.c`, `prof_alloc_rollback()` is in fact called whenever the internal allocation fails, regardless of whether there has been a sample or not; we'd probably never run into such an allocation failure case in the tests, but it might be possible in reality.

yinan1048576: Actually, I don't understand why we need to reset the sampling threshold here. If a sample was recorded, then the threshold should have already been reset; if a sample was not recorded, then we should continue counting rather than doing a reset which effectively serves to discard the accumulation so far. If there have been any reentrancy events, then we should count the allocations there rather than discarding them. So it seems to me that not having a reset here makes more sense in any case.

interwq: I think the resetting is incorrect as well; let's tackle that later in another PR.

If we simply return from this function when reentrancy > 0, will it break anything? (like missing the destroy below?)

yinan1048576:
Changed the logic in `prof_alloc_rollback(): if `reentrancy_level > 0`, assert that no sample was created and immediately return.

Also added a few other checks to guard against `tdata` creation in reentrancy.

src/prof.c
```diff
@@ -778,7 +783,7 @@ prof_tdata_cleanup(tsd_t *tsd) {
 		return;
 	}
 
-	tdata = tsd_prof_tdata_get(tsd);
```
interwq: Any reason to change this? This is the shutdown path; using the tsd getter might be right.

yinan1048576: `prof_tdata_get(tsd, false)` is identical to `tsd_prof_tdata_get(tsd)`. I'm trying to make the code consistent, and also make it look more explicit that we're not creating `tdata` here.

interwq: I think the point here is slightly different. On the shutdown path, we should stick to the low level getter to retrieve the underlying data. Right now they may be identical; however it won’t be crazy for us to say, add a tsd_nominal guard in the prof getter. On the cleanup path here, it would be safer to assume things are not working as normal.

interwq: See a similar example for tcache https://github.com/jemalloc/jemalloc/blob/dev/src/tcache.c#L650
where the higher level tcache_get won’t work on the shutdown path, and the tsd getter is used directly.

yinan1048576:

yinan1048576:
Confirmed that `./run_tests.sh` passed.

PR #1642, opened 2019-10-08
Optimize cache_bin_alloc_easy for malloc fast path

yinan1048576:
`tcache_bin_info` is not accessed on malloc fast path but the
compiler reserves a register for it, as well as an additional
register for `tcache_bin_info[ind].stack_size`.  The new optimization
gets rid of the need for the two registers.

### DIFF
--- a/include/jemalloc/internal/cache_bin.h
+++ b/include/jemalloc/internal/cache_bin.h
@@ -182,8 +182,11 @@ cache_bin_array_descriptor_init(cache_bin_array_descriptor_t *descriptor,
 	descriptor->bins_large = bins_large;
 }
 
+#define INVALID_SZIND ((szind_t)(unsigned)-1)
+
 JEMALLOC_ALWAYS_INLINE void *
-cache_bin_alloc_easy(cache_bin_t *bin, bool *success, szind_t ind) {
+cache_bin_alloc_easy_impl(cache_bin_t *bin, bool *success, szind_t ind,
+    const bool adjust_low_water) {
 	/*
 	 * This may read from the empty position; however the loaded value won't
 	 * be used.  It's safe because the stack has one more slot reserved.
@@ -191,20 +194,30 @@ cache_bin_alloc_easy(cache_bin_t *bin, bool *success, szind_t ind) {
 	void *ret = *(bin->cur_ptr.ptr++);
 	/*
 	 * Check for both bin->ncached == 0 and ncached < low_water in a single
-	 * branch.  This also avoids accessing tcache_bin_info (which is on a
-	 * separate cacheline / page) in the common case.
+	 * branch.  When adjust_low_water is true, this also avoids accessing
+	 * tcache_bin_info (which is on a separate cacheline / page) in the
+	 * common case.
 	 */
 	if (unlikely(bin->cur_ptr.lowbits > bin->low_water_position)) {
-		uint32_t empty_position = bin->full_position +
-		    tcache_bin_info[ind].stack_size;
-		if (unlikely(bin->cur_ptr.lowbits > empty_position)) {
-			/* Over-allocated; revert. */
+		if (adjust_low_water) {
+			assert(ind != INVALID_SZIND);
+			uint32_t empty_position = bin->full_position +
+			    tcache_bin_info[ind].stack_size;
+			if (unlikely(bin->cur_ptr.lowbits > empty_position)) {
+				/* Over-allocated; revert. */
+				bin->cur_ptr.ptr--;
+				assert(bin->cur_ptr.lowbits == empty_position);
+				*success = false;
+				return NULL;
+			}
+			bin->low_water_position = bin->cur_ptr.lowbits;
+		} else {
+			assert(ind == INVALID_SZIND);
 			bin->cur_ptr.ptr--;
-			assert(bin->cur_ptr.lowbits == empty_position);
+			assert(bin->cur_ptr.lowbits == bin->low_water_position);
 			*success = false;
 			return NULL;
 		}
-		bin->low_water_position = bin->cur_ptr.lowbits;
 	}
 
 	/*
@@ -218,6 +231,19 @@ cache_bin_alloc_easy(cache_bin_t *bin, bool *success, szind_t ind) {
 
 	return ret;
 }
+
+JEMALLOC_ALWAYS_INLINE void *
+cache_bin_alloc_easy_reduced(cache_bin_t *bin, bool *success) {
+	/* The szind parameter won't be used. */
+	return cache_bin_alloc_easy_impl(bin, success, INVALID_SZIND, false);
+}
+
+JEMALLOC_ALWAYS_INLINE void *
+cache_bin_alloc_easy(cache_bin_t *bin, bool *success, szind_t ind) {
+	return cache_bin_alloc_easy_impl(bin, success, ind, true);
+}
+
+#undef INVALID_SZIND
 
 JEMALLOC_ALWAYS_INLINE bool
 cache_bin_dalloc_easy(cache_bin_t *bin, void *ptr) {
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2372,7 +2372,7 @@ je_malloc(size_t size) {
 
 	cache_bin_t *bin = tcache_small_bin_get(tcache, ind);
 	bool tcache_success;
-	void *ret = cache_bin_alloc_easy(bin, &tcache_success, ind);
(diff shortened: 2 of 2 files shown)


interwq:
@yinan1048576 : can you also share the generated asm before and after?

include/jemalloc/internal/cache_bin.h
```diff
@@ -182,6 +182,36 @@ cache_bin_array_descriptor_init(cache_bin_array_descriptor_t *descriptor,
 	descriptor->bins_large = bins_large;
 }
 
+JEMALLOC_ALWAYS_INLINE void *
+cache_bin_alloc_opt(cache_bin_t *bin, bool *success) {
```
interwq: Let's do some refactoring to reduce duplicated code:
- add a `cache_bin_alloc_easy_impl(bin, success, const bool handle_low_water)`;
- and the prev alloc_easy will have the last arg as true, the new one will have false;
- then within the unlikely branch, we can decide if low_water related stuff should be considered.

The naming is hard. Not a fan of opt since it also means option in our code. Maybe call the new one `cache_bin_alloc_oversimplified`?

yinan1048576:
It turns out that the assembly for `malloc()` (I only checked GCC) is significantly reduced - here's the assembly where `[X]` denotes the instructions being removed (the remaining instructions are identical to the assembly after this change except for minor differences in register assignments, address computations ,`nop` insertions and instruction scheduling):
```
14410:   mov    0x29cb69(%rip),%rdx    # 2b0f80 <.got>
14417:   cmpb   $0x0,%fs:(%rdx)
1441b:   jne    144c2 <malloc+0xb2>
14421:   cmp    $0x1000,%rdi
14428:   ja     144c2 <malloc+0xb2>
1442e:   decl   %fs:0x1f8(%rdx)
14435:   js     144c2 <malloc+0xb2>
[X] 1443b:   push   %rbp
1443c:   lea    0x7(%rdi),%rax
14440:   lea    0x4c1af9(%rip),%rcx    # 4d5f40 <jemalloc_je_sz_size2index_tab>
[X] 14447:   push   %rbx
14448:   shr    $0x3,%rax
1444c:   lea    0x4c0d6d(%rip),%rsi    # 4d51c0 <jemalloc_je_sz_index2size_tab>
14453:   movzbl (%rcx,%rax,1),%ecx
14457:   mov    (%rsi,%rcx,8),%r8
1445b:   sub    %r8,%fs:0x20(%rdx)
14460:   js     144c0 <malloc+0xb0>
14462:   mov    %fs:0x0,%r11
14469:
1446b:   lea    (%rcx,%rcx,2),%rbp
1446f:   lea    (%r11,%rbp,8),%rsi
14473:   add    %rdx,%rsi
14476:   mov    0x200(%rsi),%r10
1447d:   lea    0x8(%r10),%r9
14481:   mov    %r9,0x200(%rsi)
14488:   mov    (%r10),%rax
1448b:   cmp    0x210(%rsi),%r9d
14492:   ja     144d0 <malloc+0xc0>
14494:   add    %r8,%fs:0x10(%rdx)
[X] 14499:   lea    (%rcx,%rcx,2),%rdi
[X] 1449d:   lea    (%r11,%rdi,8),%rcx
144a1:   incq   0x208(%rcx,%rdx,1)
144a8:
144a9:   add    %r8,%fs:0x1f0(%rdx)
144b0:
[X] 144b1:   pop    %rbx
[X] 144b2:   pop    %rbp
144b3:   retq
144b4:   mov    %r10,0x200(%rsi)
[X] 144bb:   nopl   0x0(%rax,%rax,1)
[X] 144c0:   pop    %rbx
[X] 144c1:   pop    %rbp
144c2:   jmpq   11e50 <jemalloc_je_malloc_default>
144c7:   nopw   0x0(%rax,%rax,1)
[X] 144ce:
[X] 144d0:   lea    0x4c1c79(%rip),%rbp    # 4d6150 <jemalloc_je_tcache_bin_info>
[X] 144d7:   mov    0x0(%rbp),%rbp
[X] 144db:   mov    0x0(%rbp,%rcx,4),%ebx
[X] 144df:   add    0x214(%rsi),%ebx
[X] 144e5:   cmp    %ebx,%r9d
[X] 144e8:   ja     144b4 <malloc+0xa4>
[X] 144ea:   mov    %r9d,0x210(%rsi)
[X] 144f1:   jmp    14494 <malloc+0x84>
[X] 144f3:   nopw   %cs:0x0(%rax,%rax,1)
[X] 144fa:
[X] 144fd:   nopl   (%rax)
```
So there are 21 lines less in total, of which 6 are in the fast path:
- Two pairs of `push` and `pop` for the two registers I described previously, and
- Two redundant `lea` instructions.

(I haven't refactored the code yet and what's described above was from my initial version.)

interwq:
Sounds good. The refactoring should yield the same code, hopefully.

Are the two lea instructions preparing for the tcache bin info stuff for the low water logic?

yinan1048576:
Not really. Their purpose is to obtain the address offset for locating `bin`. You can see that they're exactly duplicating lines `1446b` and `1446f`. I don't know why the compiler failed to directly make use of the already computed `%rsi`, but by simplifying the logic, this change probably have enabled the compiler to "see more clearly" through the code in some way and act more optimally.

yinan1048576:
(And I haven't compared with Clang - the redundant computations may or may not be there.)

yinan1048576:
Revised according to feedback, and confirmed that GCC can still generate the same assembly after the refactoring.

Naming is hard. :) I felt `handle_low_water` might not be accurate, since the previous `cache_bin_alloc_easy()` doesn't quite handle the entire case of `cur_ptr` surpassing `low_water` - it only handles it if the bin is still not empty. I don't have a better suggestion though.

I also felt that `oversimplified` may not be ideal - it sounds like we're overdoing it to the extent that it becomes wrong. I don't have a better suggestion either.

I'm temporarily putting `fast` in both places, but let's discuss more. :)

include/jemalloc/internal/cache_bin.h
```diff
@@ -183,7 +183,8 @@ cache_bin_array_descriptor_init(cache_bin_array_descriptor_t *descriptor,
 }
 
 JEMALLOC_ALWAYS_INLINE void *
-cache_bin_alloc_easy(cache_bin_t *bin, bool *success, szind_t ind) {
+cache_bin_alloc_easy_impl(cache_bin_t *bin, bool *success, szind_t ind,
+    bool fast) {
```
interwq: Let make the last param const, to make it explicit. How about call it “adjust_low_water”? Because when it’s false, we never modify low water.

include/jemalloc/internal/cache_bin.h
```diff
@@ -219,6 +227,17 @@ cache_bin_alloc_easy(cache_bin_t *bin, bool *success, szind_t ind) {
 	return ret;
 }
 
+JEMALLOC_ALWAYS_INLINE void *
+cache_bin_alloc_easy_fast(cache_bin_t *bin, bool *success) {
+	/* The szind parameter won't be used. */
+	return cache_bin_alloc_easy_impl(bin, success, 0, true);
```
interwq: Instead of 0, better use an invalid value, like max szind + 1.

src/jemalloc.c
```diff
@@ -2372,7 +2372,7 @@ je_malloc(size_t size) {
 
 	cache_bin_t *bin = tcache_small_bin_get(tcache, ind);
 	bool tcache_success;
-	void *ret = cache_bin_alloc_easy(bin, &tcache_success, ind);
+	void *ret = cache_bin_alloc_easy_fast(bin, &tcache_success);
```
interwq: _easy_reduced() ? No strong opinion.

include/jemalloc/internal/cache_bin.h
```diff
@@ -192,19 +195,28 @@ cache_bin_alloc_easy(cache_bin_t *bin, bool *success, szind_t ind) {
 	/*
 	 * Check for both bin->ncached == 0 and ncached < low_water in a single
 	 * branch.  This also avoids accessing tcache_bin_info (which is on a
-	 * separate cacheline / page) in the common case.
+	 * separate cacheline / page) in the common case (when fast is false).
```
interwq: Needs update here

yinan1048576:
Revised according to feedback. There's no handy macro for the maximum valid `szind`, so I defined a macro myself.

yinan1048576:
Updated comments.

yinan1048576:
Confirmed that the newest revision generates the same assembly.

PR #1645, opened 2019-10-13
use SYS_openat when available

RingsC:
some architectures like AArch64 may not have the open syscall, but have
openat syscall. so check and use SYS_openat if SYS_openat available if
SYS_open is not supported at init_thp_state.


interwq:
Thanks for the patch!

PR #1646, opened 2019-10-14
Pull prof_accumbytes into thread event handler

yinan1048576:


src/thread_event.c
```diff
@@ -18,6 +18,30 @@ static void thread_##event##_event_handler(tsd_t *tsd);
 ITERATE_OVER_ALL_EVENTS
 #undef E
 
+static void
+thread_prof_sample_event_handler(tsd_t *tsd) {
+	assert(config_prof && opt_prof);
+	assert(prof_sample_event_wait_get(tsd) == 0U);
+	tcache_t *tcache = tsd_tcachep_get(tsd);
+	if (tcache != NULL) {
```
yinan1048576: Not sure if `tcache` is ever possible to be `NULL`?

include/jemalloc/internal/thread_event.h
```diff
@@ -44,7 +44,8 @@ void thread_event_boot();
     C(thread_allocated_next_event_fast)					\
     C(thread_allocated_last_event)					\
     C(thread_allocated_next_event)					\
-    ITERATE_OVER_ALL_EVENTS
+    ITERATE_OVER_ALL_EVENTS						\
+    C(prof_sample_last_event)
```
interwq: IIRC, the last_event isn't required for other events (only to facilitate idump?). If that's the case, let's not add it to the list and manually invoke the getter / setter (I think call the tsd ones directly is fine).

yinan1048576: The only thing is that the getter and setter for it would be the same as those for the other counters, but different from those for other `tsd` fields in that the setter needs to invoke the pointer getter so as to work when `tsd_nominal(tsd)` is `false`.

interwq: Okay let’s keep it as is for now.

yinan1048576:
Revised according to feedback. Redesigned the counting mechanism for interval triggered profiling dumping, so that there's no `tcache` or `arena` level counters, but only a single global counter, and dumping is invoked directly within the thread event handler. Previously, the `arena` level counting made dumping [sporadic](https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/prof_externs.h#L36), which was not good; the `tcache` level counting was necessary to avoid the need to lock the `arena` level counter on the fast path, which now becomes irrelevant since the `thread_allocated` counter is the only counter touched on fast path that buffers allocation bytes to other counters only on slow path.

A few other remarks:

- In the previous logic, `prof_idump()` didn't proactively initialize `tdata`. This was fine before, because `prof_idump()` was called after `prof_alloc_prep()`, where the `tdata` was initialized. Now `prof_idump()` is triggered directly in the thread event handler, by which time `prof_alloc_prep()` hasn't been called yet. Therefore, I'm proactively initializing `tdata` in `prof_idump()` when dumping is triggered, and this should be fine since it will be very rare to run into scenarios where dumping is triggered whereas `tdata` hasn't been initialized (which BTW is the case in a unit test).

- I'm not yet very clear about the logic around `prof_accum_cancel()` - not sure whether it should also invoke a `thread_event_rollback()`?

include/jemalloc/internal/prof_inlines_a.h
```diff
@@ -3,11 +3,36 @@
...
+	 * prof_active in the fast path, so there are no guarantees regarding
+	 * how long it will take for all threads to notice state changes.
+	 */
+	return prof_active;
+}
+
 static inline bool
-prof_accum_add(tsdn_t *tsdn, prof_accum_t *prof_accum,
-    uint64_t accumbytes) {
+prof_accum_add(tsdn_t *tsdn, uint64_t accumbytes) {
```
interwq: Let's call this function `prof_idump_accum`, since that's the only purpose?

Also let's change both add and cancel to accept tsd instead of tsdn?

include/jemalloc/internal/tcache_types.h
```diff
@@ -47,8 +47,8 @@ typedef struct tcaches_s tcaches_t;
 #define TCACHE_GC_INCR							\
     ((TCACHE_GC_SWEEP / SC_NBINS) + ((TCACHE_GC_SWEEP / SC_NBINS == 0) ? 0 : 1))
 
-/* Used in TSD static initializer only. Real init in tcache_data_init(). */
-#define TCACHE_ZERO_INITIALIZER {0}
```
interwq: Any reason to remove the 0? I think as long as it doesn't hurt, we should keep it.

interwq:
The simplification to idump shows the benefits of having the thread event handler; I'm very happy with the way we can care about idump only within the event handler, w/o worrying about it anywhere else.

include/jemalloc/internal/prof_inlines_a.h
yinan1048576: I'm ending up only changing the two functions' names: `prof_accum_add()` -> `prof_idump_accum()`, and `prof_accum_cancel()` -> `prof_idump_rollback()`. (The name `prof_idump_cancel()` sounds like it attempts to cancel the dump.)

I decided to give up after changing some of the `tsdn`s into `tsd`: now `prof_idump_accum()` is only called in the event handler which is neat and easy to change, and the change for `prof_idump_rollback()` only involves a few `rallocx()` related call sites. but then I found I should also change `prof_idump()` since it's also only called in the event handler now, but after all the call sites are changed, I further found that I should probably also change `prof_gdump()` since it's essentially identical to `prof_idump()` and should also only be executed when a `tsd` is available, and then I figured that a lot of other places in the profiling codebase might need changes as well. So instead of bloating this PR, I'll stick to the `tsdn` way for now and will consider making a larger scale change in the future.

I think the discrepancy is on the level of design principle: previously, `tsdn` is used whenever the underlying implementation doesn't require a `tsd`; what we want now is that `tsd` should be used whenever it doesn't make logical sense to call the function when a `tsd` is not available. This can be a discussion on the scale of the entire jemalloc codebase.

interwq: Sounds reasonable. Let's not touch tsdn in this PR then.

include/jemalloc/internal/tcache_types.h
yinan1048576: The `0` was previously for initializing the `prof_accumbytes` field in `tcache`, or more precisely, `{0}` forms a partial initialization list for the entire `tcache` structure. Now after `prof_accumbytes` is removed from the `tcache`, the code won't build if I leave the `{0}` initializer as is, since now the first field is the GC ticker. So I removed `0` and just supplied an empty `{}` initializer and the code is able to build again.

However, interestingly, I just found in a CI run that MSVC doesn't like my empty initializer. So I'm now changing the initializer to `{{0, 0}}` and see if MSVC likes it... (The `{0, 0}` in the middle is for initializing the two internal fields of the GC ticker.)

interwq: empty initializers may be disallowed.  `{{0}}` may work as well.

include/jemalloc/internal/prof_externs.h
```diff
@@ -27,19 +27,14 @@ extern char	opt_prof_prefix[
...
+extern prof_accum_t prof_accum;
+
 /* Accessed via prof_active_[gs]et{_unlocked,}(). */
 extern bool	prof_active;
 
 /* Accessed via prof_gdump_[gs]et{_unlocked,}(). */
 extern bool	prof_gdump_val;
 
-/*
- * Profile dump interval, measured in bytes allocated.  Each arena triggers a
```
interwq: nit: let keep the first sentence which is still relevant.

include/jemalloc/internal/prof_inlines_a.h
```diff
@@ -3,11 +3,36 @@
...
 static inline bool
-prof_accum_add(tsdn_t *tsdn, prof_accum_t *prof_accum,
-    uint64_t accumbytes) {
+prof_idump_accum(tsdn_t *tsdn, uint64_t accumbytes) {
 	cassert(config_prof);
 
+	if (prof_interval == 0 || !prof_active_get_unlocked()) {
+		return false;
+	}
+
```
interwq: nit: move the stuff below to a static function, maybe prof_idump_accum_impl?

include/jemalloc/internal/prof_inlines_a.h
```diff
@@ -18,33 +43,32 @@ prof_accum_add(tsdn_t *tsdn, prof_accum_t *prof_accum,
...
+	prof_accum.accumbytes = a1;
+	malloc_mutex_unlock(tsdn, &prof_accum.mtx);
 #endif
 	return overflow;
 }
 
 static inline void
-prof_accum_cancel(tsdn_t *tsdn, prof_accum_t *prof_accum,
-    size_t usize) {
+prof_idump_rollback(tsdn_t *tsdn, size_t usize) {
```
interwq: Similarly, add an early bail with `prof_interval == 0 || !prof_active_get_unlocked()`, and make the body non inline.

src/prof.c
```diff
@@ -45,6 +45,8 @@ bool		opt_prof_leak = false;
 bool		opt_prof_accum = false;
 char		opt_prof_prefix[PROF_DUMP_FILENAME_LEN];
 
+prof_accum_t	prof_accum;
```
interwq: nit: let's name this `idump_accum`. prof_accum is also the name of the struct I believe.

include/jemalloc/internal/prof_inlines_a.h
yinan1048576: You mean a non-static and non-inlined function in `prof.c`? How about putting the entire function there as well? For example, functions like `prof_idump()` are also there and have similar early termination conditions.

include/jemalloc/internal/prof_inlines_a.h
yinan1048576: Same question as above.

include/jemalloc/internal/prof_inlines_a.h
interwq: Yeah I meant a non-static function in prof.c. None of this happens often, so this shouldn't matter. I slightly prefer keeping the above branch inlined, but the body blow non inlined.

src/prof.c
yinan1048576: Renamed it to `prof_idump_accum` to stick to the `prof_` "namespace". It clashes with the function name `prof_idump_accum()` so changing the function name to `prof_idump_increase()`.

yinan1048576:

yinan1048576:
Trying `{{0}}` and see if MSVC is happy.

src/prof.c
interwq: Hmm increase feels confusing, as in increasing interval which we don’t do.

I’d still vote for keeping the function name with accum, and name the variable maybe prof_idump_accumulated?

src/prof.c
```diff
@@ -45,6 +45,8 @@ bool		opt_prof_leak = false;
 bool		opt_prof_accum = false;
 char		opt_prof_prefix[PROF_DUMP_FILENAME_LEN];
 
+prof_accum_t	prof_idump_accumulated;
```
interwq: I think we can make this static now?

yinan1048576:
Revise according to feedback.

yinan1048576:
Rebase.

yinan1048576:
Rebase.

yinan1048576:
Rebase.

PR #1648, opened 2019-10-16
[RFC] Disable cache index randomization by default.

interwq:
As modern architectures adopt smarter cache mapping algorithms, e.g. hashing
combined with set associative, the likelihood of cache thrashing from page
aligned large allocations is reduced.  Turning off the randomization by default
to avoid the overhead (one extra page per large allocation).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1395,7 +1395,7 @@ else
   enable_cache_oblivious="1"
 fi
 ],
-[enable_cache_oblivious="1"]
+[enable_cache_oblivious="0"]
 )
 if test "x$enable_cache_oblivious" = "x1" ; then
   AC_DEFINE([JEMALLOC_CACHE_OBLIVIOUS], [ ])


davidtgoldblatt:
I like this. I think we can even make it a malloc_conf option; the only place where it would change anything is in the large allocation szind computation, which is already a slow path (and dwarfed by the cost of actually doing the allocation).

> Turning off the randomization by default to avoid the overhead (one extra page per large allocation).

The savings can be even larger, right? My guess is that large allocations are disproportionately likely to be powers of two, and so the extra page can bump the size class up, so that the effective cost might even be 25%.

interwq:
> I think we can even make it a malloc_conf option;

Good point. I think making it a runtime option with default off (no randomization) is quite reasonable. Will update the PR.

jasone:
Can you please document as part of this change which (current) modern architectures don't need cache index randomization?  I'm betting a lot of us don't have those modern marvels of engineering yet. 😉

davidtgoldblatt:
I think this probably makes sense (both because it's the configuration we have by far the most experience with, and because the median external user probably values memory over CPU), but we can reopen to revive; this isn't pending review.

PR #1649, opened 2019-10-17
Fix amd64 MSVC warning

yinan1048576:
Get rid of `warning C4267: 'initializing': conversion from 'size_t' to 'cache_bin_sz_t', possible loss of data` when compiled in an environment like `MSYSTEM=MINGW64, CPU=x86_64, MSVC=amd64`.

(Not sure if we'd better have some additional asserts as guards.)


yinan1048576:
Not sure if it's a good idea to turn on some flag like `-Werror` for all CIs, but it can also over trigger on cases we don't really care.

yinan1048576:
Fixed more warnings.

PR #1652, opened 2019-10-23
Rewrite of tsd_win to use __declspec(thread) instead of the generic wrapper

santagada:


santagada:
FYI I'm not too interested in fixing gcc errors on windows. Can we just have a very slow version on it and ask people to not use it?

davidtgoldblatt:
@jasone, any relevant context here on __declspec(thread) vs the getter / setter? I think we don't support hijacking malloc on Windows so it should be fine, but I want to be sure.

Copying from the gitter thread, I think this needs CI in two ways:
- Existing builds need to still work by falling back to the functionality.
- We should test the new tsd module implementation via a MSVC build

jasone:
There's a near total lack of documentation on intent and benefit (i.e. minimal commit messages), and the failing CI tests leave me guessing.  At a high level, I don't know of any problems with changing the implementation, except that the replacement needs to work for a relevant set of build systems/environments.  This pull request needs to explain how it meets these requirements, and it also needs to preserve the CI system.

davidtgoldblatt:
The background here is that `__declspec(thread)` is more like `__thread`, and `TlsGetValue` is more like `pthread_getspecific`. In some of the testing @santagada  has been doing the TLS accesses can be a significant portion of malloc time on Windows.

jasone:
Sounds good to me, as long as it works. :wink:  I don't remember off the top of my head whether we ended up with the current implementation because it was the only workable option at the time, or if we just didn't try hard enough.

santagada:
Exactly, I can document more, but the gist of it was tsd_get was dominating the results to the point that an allocation heavy part of one of our system was spending 6 seconds in allocation out of 30... that led me to look into it. We have our own tls implementation but its a bit too simplistic for a general memory allocator, but __declspec(thread) is supported in all os after vista so it seems the correct tool for this.

Any comments on calling tsd_cleanup directly on the DLL_THREAD_DETACH case?
```
	case DLL_THREAD_DETACH:
		_malloc_thread_cleanup();
		tsd_cleanup(tsd_get(false));
```

I couldn't understand how tsd_cleanup is called besides being called by pthread when the thread dies... and couldn't understand what is _malloc_thread_cleanup at all.

davidtgoldblatt:
Correctness/CI issues look real -- closing for now.

PR #1657, opened 2019-10-25
Pull tcache GC events into thread event handler

yinan1048576:
Currently only the tcache GC events along the allocation path are pulled into the thread event handler. Once the deallocation path also gets a thread event handler, the GC events there can be consolidated (with `thread.deallocated`).

The interval in terms of bytes are temporarily set to be 32768: given that `TCACHE_GC_INCR` is right now 228, 32768 bytes would be an equivalent interval if we assume the average allocation size is around 140-150 bytes.

### DIFF
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -33,6 +33,7 @@ void thread_event_boot();
  *  E(event,		(condition))
  */
 #define ITERATE_OVER_ALL_EVENTS						\
+    E(tcache_gc,	(TCACHE_GC_INCR_BYTES > 0))			\
     E(prof_sample,	(config_prof && opt_prof))
 
 #define E(event, condition)						\
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -27,6 +27,7 @@ void thread_event_trigger(tsd_t *tsd, bool delay_event);
 void thread_event_rollback(tsd_t *tsd, size_t diff);
 void thread_event_update(tsd_t *tsd);
 void thread_event_boot();
+void tsd_thread_event_init(tsd_t *tsd);
 
 /*
  * List of all events, in the following format:
--- a/src/thread_event.c
+++ b/src/thread_event.c
@@ -11,12 +11,31 @@
  */
 static bool thread_event_active = false;
 
+/* TSD event init function signatures. */
+#define E(event, condition)						\
+static void tsd_thread_##event##_event_init(tsd_t *tsd);
+
+ITERATE_OVER_ALL_EVENTS
+#undef E
+
 /* Event handler function signatures. */
 #define E(event, condition)						\
 static void thread_##event##_event_handler(tsd_t *tsd);
 
 ITERATE_OVER_ALL_EVENTS
 #undef E
+
+static void
+tsd_thread_tcache_gc_event_init(tsd_t *tsd) {
+	assert(TCACHE_GC_INCR_BYTES > 0);
+	thread_tcache_gc_event_update(tsd, TCACHE_GC_INCR_BYTES);
+}
+
+static void
+tsd_thread_prof_sample_event_init(tsd_t *tsd) {
+	assert(config_prof && opt_prof);
+	/* Do not set sample interval until the first allocation. */
+}
 
 static void
 thread_tcache_gc_event_handler(tsd_t *tsd) {
@@ -265,6 +284,16 @@ void thread_event_boot() {
 #define E(event, condition)						\
 	if (condition) {						\
 		thread_event_active = true;				\
+	}
+
+	ITERATE_OVER_ALL_EVENTS
+#undef E
+}
+
+void tsd_thread_event_init(tsd_t *tsd) {
+#define E(event, condition)						\
+	if (condition) {						\
+		tsd_thread_##event##_event_init(tsd);			\
 	}
 
 	ITERATE_OVER_ALL_EVENTS
--- a/src/thread_event.c
+++ b/src/thread_event.c
@@ -19,6 +19,17 @@ ITERATE_OVER_ALL_EVENTS
 #undef E
 
 static void
+thread_tcache_gc_event_handler(tsd_t *tsd) {
+	assert(TCACHE_GC_INCR_BYTES > 0);
+	assert(tcache_gc_event_wait_get(tsd) == 0U);
+	thread_tcache_gc_event_update(tsd, TCACHE_GC_INCR_BYTES);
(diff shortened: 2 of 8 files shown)


src/thread_event.c
```diff
@@ -18,6 +18,16 @@ static void thread_##event##_event_handler(tsd_t *tsd);
 ITERATE_OVER_ALL_EVENTS
 #undef E
 
+static void
+thread_tcache_gc_event_handler(tsd_t *tsd) {
+	assert(opt_tcache);
+	assert(tcache_gc_event_wait_get(tsd) == 0U);
+	if (TCACHE_GC_INCR_BYTES > 0U && tcache_available(tsd)) {
+		thread_tcache_gc_event_update(tsd, TCACHE_GC_INCR_BYTES);
```
interwq: Leave the update outside of the branch? tcache can be switched on/off per thread; if we don't reset it, we might never handle the event again?

src/thread_event.c
```diff
@@ -18,6 +18,16 @@ static void thread_##event##_event_handler(tsd_t *tsd);
...
+		tcache_event_hard(tsd, tsd_tcachep_get(tsd));
```
interwq: let's use `tcache_get` here. The tcachep getter should not be used by other modules; the branch is guarded with tcache_available already.

yinan1048576:
Revised according to feedback. Also improved the unit tests so that they can automatically adapt to new thread events.

include/jemalloc/internal/thread_event.h
```diff
@@ -33,6 +33,7 @@ void thread_event_boot();
  *  E(event,		(condition))
  */
 #define ITERATE_OVER_ALL_EVENTS						\
+    E(tcache_gc,	(opt_tcache))					\
```
interwq: maybe `TCACHE_GC_INCR_BYTES > 0 && opt_tcache`?

src/thread_event.c
```diff
@@ -18,6 +18,20 @@ static void thread_##event##_event_handler(tsd_t *tsd);
 ITERATE_OVER_ALL_EVENTS
 #undef E
 
+static void
+thread_tcache_gc_event_handler(tsd_t *tsd) {
+	assert(opt_tcache);
+	assert(tcache_gc_event_wait_get(tsd) == 0U);
+	if (TCACHE_GC_INCR_BYTES == 0U) {
```
interwq: Then we can assert TCACHE_GC_INCR_BYTES > 0 here.

interwq:
The Travis failure might be real -- https://travis-ci.org/jemalloc/jemalloc/jobs/602966009

yinan1048576:
Revised according to feedback. The test failure will be addressed in a separate PR, and will then rebase this one.

yinan1048576:
Let me first see if the fix is working...

src/extent.c
```diff
@@ -2124,7 +2124,11 @@ extent_util_stats_verbose_get(tsdn_t *tsdn, const void *ptr,
 	} else {
 		*bin_nfree = *bin_nregs = 0;
 	}
-	*slabcur_addr = extent_addr_get(bin->slabcur);
-	assert(*slabcur_addr != NULL);
+	if (bin->slabcur != NULL) {
+		*slabcur_addr = extent_addr_get(bin->slabcur);
+		assert(*slabcur_addr != NULL);
+	} else {
+		*slabcur_addr = extent_heap_first(&bin->slabs_nonfull);
```
interwq: This doesn’t look right. The heap first still returns an extent.

yinan1048576:
Corrected.

yinan1048576:
The test failure is fixed. Putting up a separate PR #1659 since it's logically unrelated to this PR.

yinan1048576:
Rebase.

yinan1048576:
Avoid doing tcache GC at TSD init.

src/tsd.c
```diff
@@ -233,6 +233,10 @@ tsd_data_init(tsd_t *tsd) {
 	*tsd_offset_statep_get(tsd) = config_debug ? 0 :
 	    (uint64_t)(uintptr_t)tsd;
 
+	if (TCACHE_GC_INCR_BYTES > 0 && opt_tcache) {
+		thread_tcache_gc_event_update(tsd, TCACHE_GC_INCR_BYTES);
+	}
+
```
interwq: We should have similar init for other events as well. Maybe add a tsd_thread_event_init() for them.

yinan1048576:
Rebase, and revise according to feedback. For now the init function for prof sampling event is kept as empty, because the current logic won't initialize `tdata` and the sampling interval until the first allocation arrives. This might be beneficial in some cases? Should we change that?

interwq:
Re: init prof sample. Not creating tdata on init sounds right. Some threads might be really short lived (not having any sample at all). However we should still set the init sample interval?

interwq:
How are we initializing sample interval the first time currently? We shouldn’t always sample the first allocation?

interwq:
I think I found it, this is the booted case in prof_sample_accum_update? If that’s the case, yes I think not doing anything in the prof sample init is fine.

yinan1048576:
Right, it's the `!booted` case in `prof_sample_accum_update()`. The `tdata` and the sample interval are set at the first allocation (not the first sampled allocation); afterwards, the thread will wait until the sample interval is reached to obtain the first sampled allocation. And this logic has been kept the same (in both the previous `bytes_until_sample` based counting and the current `thread_allocated` based counting).

In other words, unless a thread dies before making any allocation request, doing the `tdata` and sample interval initialization in `tsd_thread_prof_sample_event_init()` will do no harm, and it will simplify the `prof_sample_accum_update()` logic.

interwq:
I see. Okay let's keep it that way for now. I'll try some refactoring around that later.

nit: squash the commit `Avoid doing tcache GC at TSD init` into the previous one?

yinan1048576:
Squash first two commits. Also changed the TC interval from 32768 to 65536.

include/jemalloc/internal/thread_event.h
```diff
@@ -0,0 +1,142 @@
...
+void thread_event_update(tsd_t *tsd);
+void thread_event_boot();
+void tsd_thread_event_init(tsd_t *tsd);
+
+/*
+ * List of all events, in the following format:
+ *  E(event,		(condition))
+ */
+#define ITERATE_OVER_ALL_EVENTS						\
+    E(tcache_gc,	(TCACHE_GC_INCR_BYTES > 0 && opt_tcache))	\
```
interwq: As discussed with Yinan offline, this guard needs to be updated. tcache can still be turned on per thread even w/o opt_tcache.

yinan1048576:

yinan1048576:
Rebase.

PR #1659, opened 2019-10-28
Guard slabcur fetching in extent_util

yinan1048576:
`slabcur` can be `NULL`, and in such cases, the extent utilization querying logic should fetch the first extent from the non-full extents, because it will become `slabcur` next. If this fails, then the logic returns `NULL`.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -2124,7 +2124,12 @@ extent_util_stats_verbose_get(tsdn_t *tsdn, const void *ptr,
 	} else {
 		*bin_nfree = *bin_nregs = 0;
 	}
-	*slabcur_addr = extent_addr_get(bin->slabcur);
-	assert(*slabcur_addr != NULL);
+	extent_t *slab;
+	if (bin->slabcur != NULL) {
+		slab = bin->slabcur;
+	} else {
+		slab = extent_heap_first(&bin->slabs_nonfull);
+	}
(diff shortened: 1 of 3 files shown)


src/extent.c
```diff
@@ -2124,7 +2124,14 @@ extent_util_stats_verbose_get(tsdn_t *tsdn, const void *ptr,
...
-	assert(*slabcur_addr != NULL);
+	if (bin->slabcur != NULL) {
+		*slabcur_addr = extent_addr_get(bin->slabcur);
+		assert(*slabcur_addr != NULL);
+	} else {
+		extent_t *first_nonfull_extent =
+		    extent_heap_first(&bin->slabs_nonfull);
+		*slabcur_addr = first_nonfull_extent == NULL ? NULL :
+		    extent_addr_get(first_nonfull_extent);
+	}
```
interwq: maybe refactor a bit --

    if (bin->slabcur != NULL) {
		slab = extent_addr_get(bin->slabcur);
    } else {
		slab = extent_heap_first(&bin->slabs_nonfull);
    }

    *slabcur_addr = slab != NULL ? extent_addr_get(first_nonfull_extent) : NULL;
    
Did we document this API yet? If so we need to update the manual as well.

yinan1048576: It's an experimental API. The only documentation was a block of comment in `ctl.c`. Updated it.

yinan1048576:

interwq:
LGTM assuming tests pass.

PR #1661, opened 2019-10-29
Add -Wimplicit-fallthrough checks if supported

nickdesaulniers:
Clang since r369414 (clang-10) can now check -Wimplicit-fallthrough for
C code, and use the GNU C style attribute to denote fallthrough.

Move the test from header only to autoconf. The previous test used
brittle version detection which did not work for newer clang that
supported this feature.

The attribute has to be its own statement, hence the added `;`. It also
can only precede case statements, so the final cases should be
explicitly terminated with break statements.

Fixes commit 3d29d11ac2c1 ("Clean compilation -Wextra")
Link: https://github.com/llvm/llvm-project/commit/1e0affb6e564b7361b0aadb38805f26deff4ecfc
Signed-off-by: Nick Desaulniers <ndesaulniers@google.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -866,6 +866,26 @@ if test "x${je_cv_format_arg}" = "xyes" ; then
   AC_DEFINE([JEMALLOC_HAVE_ATTR_FORMAT_ARG], [ ])
 fi
 
+dnl Check for fallthrough attribute support.
+JE_CFLAGS_SAVE()
+JE_CFLAGS_ADD([-Wimplicit-fallthrough])
+JE_COMPILABLE([fallthrough attribute],
+              [#if !__has_attribute(fallthrough)
+               #error "foo"
+               #endif],
+              [int x = 0;
+               switch (x) {
+               case 0: __attribute__((__fallthrough__));
+               case 1: return 1;
+               }],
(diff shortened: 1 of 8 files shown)


nickdesaulniers:
hmm...looks like a clang build is failing:
```
include/jemalloc/internal/hash.h:107:32: error: declaration does not declare
      anything [-Werror,-Wmissing-declarations]
                case 3: k1 ^= tail[2] << 16; JEMALLOC_FALLTHROUGH;
                                             ^~~~~~~~~~~~~~~~~~~~
```

nickdesaulniers:
likely an old clang build. pushing a change.

interwq:
Thanks for the patch. Travis is still not happy though. Can you please take another look @nickdesaulniers?

nickdesaulniers:
Looks like my autoconf detection of fallthrough attribute is always returning true: https://godbolt.org/z/Xdx5QP

TODO: debug

nickdesaulniers:
autoconf...*shakes fist* @interwq quick, merge it before it breaks in non obvious ways! :rocket:

interwq:
Thanks for the patch @nickdesaulniers!

PR #1666, opened 2019-11-05
Check for large size first in the uncommon case of malloc.

interwq:
Larger sizes are not that uncommon comparing to !tsd_fast.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2344,11 +2344,9 @@ je_malloc(size_t size) {
 	}
 
 	tsd_t *tsd = tsd_get(false);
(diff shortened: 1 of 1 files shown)


yinan1048576:
This makes sense. Curious whether there are some measurements for the difference?

interwq:
Tried tweaking the stress benchmark to alloc 5K (which fails the size check here), and in `perf report` it showed the cost of `malloc` (which only does the fastpath check and calls `malloc_default`) reduced from ~2.5% to 1%. Verified that one less branch in that case.

And as expected, no change for the fastpath case.

PR #1667, opened 2019-11-05
Fix a low_water assertion on in cache_bin_alloc_easy_impl.

interwq:
The regression was introduced in https://github.com/jemalloc/jemalloc/commit/05681e387a3202567ff95528dbc460e92e031a3c.

However, as perviously mentioned by @yinan1048576, the assertions are not included in opt builds, while the fast path is not walked in the dbg mode. We'll need to introduce an opt + assert mode later.


interwq:
Discussed offline with @yinan1048576. I obviously misunderstood the low_water pointer comparison. The assertion is correct as is.

As a side note, we'll need to define some way to compare the pointer with clear naming, e.g. `cur_ptr > low_water` currently means gone past low water, instead of the above low water (i.e. `n > low_water_value` in the previous design.

PR #1669, opened 2019-11-05
Remove mallctl test for zero_realloc

yinan1048576:
The test fails whenever `zero_realloc` is set to be anything other
than strict.  In unit tests we'd better not assume that the default
is always turned on.


yinan1048576:
Rather than completely removing the new unit test, added `zero_realloc` to the standard test set for `opt`s, to do some basic checkings.

PR #1670, opened 2019-11-05
Refactor arena_bin_malloc_hard().

interwq:
Clean up some redundant operations (looks to be leftover from previous refactoring).


interwq:
re: insert beforehand. I don't think there's much advantage to that. We'll have to do the check after `arena_bin_nonfull_slab_get()`, since we release the bin lock in the middle. The refactor here is to get rid of redundant logic; the logic after arena_bin_nonfull_slab_get() is more general and covers the case handled by the early branch properly.

PR #1676, opened 2019-11-08
Prefer getaffinity() to detect number of CPUs.

interwq:
The `getaffinity` approach should be preferred than `sysconf(_SC_NPROCESSORS_ONLN)`, because with affinity binding the number of usable cores could be fewer.

Also prefer `sched_getaffinity` when available.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -742,18 +742,28 @@ malloc_ncpus(void) {
 	SYSTEM_INFO si;
 	GetSystemInfo(&si);
 	result = si.dwNumberOfProcessors;
-#elif defined(JEMALLOC_GLIBC_MALLOC_HOOK) && defined(CPU_COUNT)
+#elif defined(CPU_COUNT)
 	/*
 	 * glibc >= 2.6 has the CPU_COUNT macro.
 	 *
 	 * glibc's sysconf() uses isspace().  glibc allocates for the first time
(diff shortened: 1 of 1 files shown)


src/jemalloc.c
```diff
@@ -742,18 +742,28 @@ malloc_ncpus(void) {
...
+#  else
 		cpu_set_t set;
-
+#  endif
+#  if defined(JEMALLOC_HAVE_SCHED_SETAFFINITY)
+		sched_getaffinity(0, sizeof(set), &set);
+#  else
 		pthread_getaffinity_np(pthread_self(), sizeof(set), &set);
+#  endif
 		result = CPU_COUNT(&set);
```
azat, line 767: But if `sched_getaffinity` will be preferred over `_SC_NPROCESSORS_ONLN` then `percpu_arena_choose()` can return index beyond cached `ncpus`, seems that all available CPUs should be preferred not only enabled via affinity, since the affinity can be changed on fly and then you are in trouble. Am I missing something?

azat: Here is a reproducer - https://gist.github.com/azat/9770f751c3e7d2da5422cc1ca26339c4

Tested on:
- 5.2.1 + this PR
- dev branch (259c5e3e)

azat: Here is an RFC - #1939

PR #1680, opened 2019-11-12
Optimize away the tsd_fast() check on fastpath.

interwq:
Fold the tsd_state branch onto the event threshold check. After this change, we have only 3 branches on the fast-path: 1) size threshold, 2) event threshold and 3) tcache is empty.

### DIFF
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -27,6 +27,7 @@ void thread_event_trigger(tsd_t *tsd, bool delay_event);
 void thread_event_rollback(tsd_t *tsd, size_t diff);
 void thread_event_update(tsd_t *tsd);
 void thread_event_boot();
+void thread_event_recompute_fast_threshold(tsd_t *tsd);
 void tsd_thread_event_init(tsd_t *tsd);
 
 /*
@@ -43,9 +44,7 @@ void tsd_thread_event_init(tsd_t *tsd);
 /* List of all thread event counters. */
 #define ITERATE_OVER_ALL_COUNTERS					\
     C(thread_allocated)							\
-    C(thread_allocated_next_event_fast)					\
     C(thread_allocated_last_event)					\
-    C(thread_allocated_next_event)					\
     ITERATE_OVER_ALL_EVENTS						\
     C(prof_sample_last_event)
 
@@ -80,6 +79,60 @@ ITERATE_OVER_ALL_COUNTERS
  * event.
  */
 #undef E
+
+/*
+ * Two malloc fastpath getters -- use the unsafe getters since tsd may be
+ * non-nominal, in which case the fast_threshold will be set to 0.  This allows
+ * checking for events and tsd non-nominal in a single branch.
+ *
+ * Note that these can only be used on the fastpath.
+ */
+JEMALLOC_ALWAYS_INLINE uint64_t
+thread_allocated_malloc_fastpath(tsd_t *tsd) {
+	return *tsd_thread_allocatedp_get_unsafe(tsd);
+}
+
+JEMALLOC_ALWAYS_INLINE uint64_t
+thread_allocated_next_event_malloc_fastpath(tsd_t *tsd) {
+	uint64_t v = *tsd_thread_allocated_next_event_fastp_get_unsafe(tsd);
+	assert(v <= THREAD_ALLOCATED_NEXT_EVENT_FAST_MAX);
+	return v;
+}
+
+/* Below 3 for next_event_fast. */
+JEMALLOC_ALWAYS_INLINE uint64_t
+thread_allocated_next_event_fast_get(tsd_t *tsd) {
+	uint64_t v = tsd_thread_allocated_next_event_fast_get(tsd);
+	assert(v <= THREAD_ALLOCATED_NEXT_EVENT_FAST_MAX);
+	return v;
+}
+
+JEMALLOC_ALWAYS_INLINE void
+thread_allocated_next_event_fast_set(tsd_t *tsd, uint64_t v) {
+	assert(v <= THREAD_ALLOCATED_NEXT_EVENT_FAST_MAX);
+	*tsd_thread_allocated_next_event_fastp_get(tsd) = v;
+}
+
+JEMALLOC_ALWAYS_INLINE void
+thread_allocated_next_event_fast_set_non_nominal(tsd_t *tsd) {
+	/*
+	 * Set the fast threshold to zero when tsd is non-nominal.  Use the
+	 * unsafe getter as this may get called during tsd init and clean up.
+	 */
+	*tsd_thread_allocated_next_event_fastp_get_unsafe(tsd) = 0;
+}
+
+/* For next_event.  Setter also updates the fast threshold. */
+JEMALLOC_ALWAYS_INLINE uint64_t
+thread_allocated_next_event_get(tsd_t *tsd) {
+	return tsd_thread_allocated_next_event_get(tsd);
+}
+
+JEMALLOC_ALWAYS_INLINE void
+thread_allocated_next_event_set(tsd_t *tsd, uint64_t v) {
+	*tsd_thread_allocated_next_eventp_get(tsd) = v;
+	thread_event_recompute_fast_threshold(tsd);
+}
 
 /*
  * The function checks in debug mode whether the thread event counters are in
--- a/include/jemalloc/internal/tsd.h
+++ b/include/jemalloc/internal/tsd.h
@@ -110,7 +110,7 @@ typedef void (*test_callback_t)(int *);
     /* reentrancy_level */	0,					\
     /* narenas_tdata */		0,					\
     /* thread_allocated */	0,					\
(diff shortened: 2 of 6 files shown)


interwq:
@davidtgoldblatt -- need your input here.  The `tsd_fast()` check is folded onto the thread_event check in this diff. This means the remote tsd state update is deferred to, only take effect after the thread falls off the fast path. Potentially we update the `event_threshold` remotely as well? Opinions?

davidtgoldblatt:
Yeah, I think remotely modifying event_threshold is the right thing to do here. I don't think we have any firm guarantees about how "quickly" other threads notice a switch to a global slow path in the first place.

interwq:
@davidtgoldblatt and @yinan1048576 need both of your review here.

interwq:
@davidtgoldblatt I forgot the case where we don't have ATOMIC_U64... WDYT about plain uint64_t? The pattern is remote threads can only reset it to 0.

include/jemalloc/internal/tsd.h
```diff
@@ -355,8 +355,8 @@ tsd_assert_fast(tsd_t *tsd) {
 	 * counters; it's not in general possible to ensure that they won't
 	 * change asynchronously from underneath us.
 	 */
-	assert(!malloc_slow && tsd_tcache_enabled_get(tsd) &&
-	    tsd_reentrancy_level_get(tsd) == 0);
+	assert(!malloc_initialized() || (tsd_tcache_enabled_get(tsd) &&
```
yinan1048576: Why do we have the additional `!malloc_initialized()` condition here?

test/unit/thread_event.c
```diff
@@ -9,6 +9,7 @@ TEST_BEGIN(test_next_event_fast_roll_back) {
 	    THREAD_ALLOCATED_NEXT_EVENT_FAST_MAX);
 	thread_allocated_next_event_fast_set(tsd,
 	    THREAD_ALLOCATED_NEXT_EVENT_FAST_MAX);
+	thread_event_recompute_fast_threshold(tsd);
```
yinan1048576: Should we avoid the `thread_allocated_next_event_fast_set()` call on the previous line?

src/tsd.c
```diff
@@ -117,6 +117,9 @@ tsd_force_recompute(tsdn_t *tsdn) {
 		    <= tsd_state_nominal_max);
 		tsd_atomic_store(&remote_tsd->state, tsd_state_nominal_recompute,
 		    ATOMIC_RELAXED);
+		atomic_store_u64(
+		    tsd_thread_allocated_next_event_fastp_get_unsafe(remote_tsd),
+		    0, ATOMIC_RELAXED);
```
yinan1048576: Is this doing the same thing as `thread_allocated_next_event_fast_set_non_nominal(remote_tsd)`?

src/jemalloc.c
```diff
@@ -2363,13 +2363,25 @@ je_malloc(size_t size) {
...
+	 * non-nominal in a single branch.
+	 */
+	uint64_t *allocatedp = tsd_thread_allocatedp_get_unsafe(tsd);
+	/*
+	 * Not using relaxed atomic_load, because compilers right now can't tell
+	 * that this can be done as a memory comparison, and forces a load into
+	 * a register that hurts fast-path performance.
+	 */
+	uint64_t *fast_thresholdp =
+	    (uint64_t *)tsd_thread_allocated_next_event_fastp_get_unsafe(tsd);
```
yinan1048576: So by avoiding using the atomic load, the assembly code can be identical to before? What's the risk of not using the atomic load?

test/unit/thread_event.c
interwq: Good catch. Will update.

src/tsd.c
interwq: Yep will fix.

src/jemalloc.c
interwq: Right. The asm output around this doesn’t change. If we do use atomic load, the fast threshold will be loaded into a register, instead of a comparison with memory directly.

With the current implementation I don’t think there’s any real risk. We are doing the same trick with tsd_state as well.

include/jemalloc/internal/tsd.h
interwq: Let me see if I can get rid of it. I wanted to call this during tsd init, which might break the previous assertions.

src/jemalloc.c
yinan1048576: Got it. For other places where we have less concern on the performance, should we guard the atomic loads and stores in an `#if JEMALLOC_ATOMIC_U64` and always provide an `#else` part?

include/jemalloc/internal/tsd.h
interwq: Got rid of it.

src/jemalloc.c
interwq: With @davidtgoldblatt 's help, the conclusion reached is, with a couple of memory fences the pattern here can be made safe w/o the `threshold_fast` being atomic. Stay tuned for David's comments (will be included in the diffs somewhere).

davidtgoldblatt:
Thread safety comment:

```
There are two threads to consider in the synchronization here:
- The owner of the tsd being updated by a slow path change
- The remote thread, doing that slow path change.

As a design constraint, we want to ensure that a slow-path transition cannot be
ignored for arbitrarily long, and that if the remote thread causes a slow-path
transition and then communicates with the owner thread that it has occurred,
then the owner will go down the slow path on the next allocator operation (so
that we don't want to just wait until the owner hits its slow path reset
condition on its own).

Here's our strategy to do that:

The remote thread will update the slow-path stores to TSD variables, issue a
SEQ_CST fence, and then update the TSD next_event_fast counter. The owner thread
will update next_event_fast, issue an SEQ_CST fence, and then check its TSD to
see if it's on the slow path.

This is fairly straightforward when 64-bit atomics are supported. Assume that
the remote fence is sandwiched between two owner fences in the reset pathway.
The case where there is no preceding or trailing owner fence (i.e. because the
owner thread is near the beginning or end of its life) can be analyzed
similarly. The owner store to next_event_fast preceding the earlier owner fence
will be earlier in coherence order than the remote store to it, so that the
owner thread will go down the slow path once the store becomes visible to it,
which is no later than the time of the second fence.

The case where we don't support 64-bit atomics is trickier, since word tearing
is possible. We'll repeat the same analysis, and look at the two owner fences
sandwiching the remote fence. The next_event_fast stores done alongside the
earlier owner fence cannot overwrite any of the remote stores (since they
precede the earlier owner fence in sb, which precedes the remote fence in sc,
which precedes the remote stores in sb). After the second owner fence there
will be a re-check of the slow-path variables anyways, so the "owner will
notice that it's on the slow path eventually" guarantee is satisfied. To make
sure that the out-of-band-messaging constraint is as well, note that either the
message passing is sequenced before the second owner fence (in which case the
remote stores happen before the second set of owner stores, so malloc sees a
value of zero for next_event_fast and goes down the slow path), or it is not (in
which case the owner sees the tsd slow-path writes on its previous update). This
leaves open the possibility that the remote thread will (at some arbitrary point
in the future) zero out one half of the owner thread's next_event_fast, but
that's always safe (it just sends it down the slow path earlier).
```

src/thread_event.c
```diff
@@ -128,15 +129,78 @@ thread_event_assert_invariants_debug(tsd_t *tsd) {
...
+ * stores, so malloc sees a value of zero for next_event_fast and goes down the
+ * slow path), or it is not (in which case the owner sees the tsd slow-path
+ * writes on its previous update). This leaves open the possibility that the
+ * remote thread will (at some arbitrary point in the future) zero out one half
+ * of the owner thread's next_event_fast, but that's always safe (it just sends
+ * it down the slow path earlier).
+ */
+void
+thread_event_recompute_fast_threshold(tsd_t *tsd) {
+	if (tsd_state_get(tsd) != tsd_state_nominal) {
```
interwq: This is only needed so that the `_get` below doesn't fail the tsd state assertion on purgatory state (on tsd cleanup, we also update the fast threshold to be consistent).

Technically we can also change all thread_event getters to use the unsafe ones; however other places all pass the assertions so that doesn't feel right.

interwq:
Ready for review now.

interwq:
@davidtgoldblatt and @yinan1048576 need you guys to take another look here.

include/jemalloc/internal/thread_event.h
```diff
@@ -81,6 +80,58 @@ ITERATE_OVER_ALL_COUNTERS
  */
 #undef E
 
+/*
+ * Two malloc fastpath getters -- use the unsafe getters since tsd may not be
+ * nominal, in which case the fast_threshold will be set to 0.  This allows
+ * checking for events and tsd non-nominal in a single branch.
+ */
+JEMALLOC_ALWAYS_INLINE uint64_t
+thread_allocated_get_malloc_fastpath(tsd_t *tsd) {
```
yinan1048576: nit: also get rid of the `get` to be consistent with `thread_allocated_next_event_malloc_fastpath()`?

interwq: Good point. Just updated.

PR #1685, opened 2019-11-15
Fix bug in prof_realloc

yinan1048576:
Temporarily disallow `prof_realloc()` to pass in the old pointer.

The old pointer has primarily been used for logging purpose.
Logging needs to access the `extent` the old pointer sits in, which
is unfortunately lost at the time `prof_free_sampled_object()` is
called on the `prof_realloc()` code path, and it's not trivial to
refactor the code to properly reserve the `extent` till the time
`prof_free_sampled_object()` is called.

Disallowing `prof_realloc()` to pass in the old pointer will result
in not recording the associated deallocation, which is better than
recording a wrong deallocation.  Given that the majority of the
traffic will go through `prof_malloc()` and `prof_free()` instead of
`prof_realloc()`, this should be the "minimally incorrect" fix in a
short time.

### DIFF
--- a/include/jemalloc/internal/prof_inlines_b.h
+++ b/include/jemalloc/internal/prof_inlines_b.h
@@ -203,7 +203,7 @@ prof_realloc(tsd_t *tsd, const void *ptr, size_t usize, prof_tctx_t *tctx,
 	 * counters.
 	 */
 	if (unlikely(old_sampled)) {
-		prof_free_sampled_object(tsd, ptr, old_usize, old_tctx);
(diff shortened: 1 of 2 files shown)


interwq:
Let's wait until you have a full fix? Assuming there's no urgency around it.

yinan1048576:
Discussed offline. Rebase, and add comments regarding the assumptions on the input to `prof_free_sampled_object()`.

yinan1048576:
This fix is relatively more correct (following #1684), but it's not the real fix. Closing.

PR #1688, opened 2019-11-20
[RFC only] try folding the size check branch on malloc fastpath

interwq:
Based on #1680. This commit https://github.com/jemalloc/jemalloc/pull/1688/commits/33b9e74cb4d947f98ad98d832481c812046a480d tries to avoid the size > MAX_LOOKUP_SIZE branch, by abusing the fast threshold.

### DIFF
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -27,6 +27,7 @@ void thread_event_trigger(tsd_t *tsd, bool delay_event);
 void thread_event_rollback(tsd_t *tsd, size_t diff);
 void thread_event_update(tsd_t *tsd);
 void thread_event_boot();
+void thread_event_recompute_fast_threshold(tsd_t *tsd);
 void tsd_thread_event_init(tsd_t *tsd);
 
 /*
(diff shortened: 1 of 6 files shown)


interwq:
@yinan1048576 @davidtgoldblatt this one might be going too far; so wanted you opinions first. I guess Dave would love it :)

This makes sure that the fast_threshold is always at most MAX_LOOKUP_SIZE ahead of thread.allocated, so that the threshold branch covers the `if (size > MAX_LOOKUP_SIZE)` branch. One tricky part is, since we don't know the `usize` beforehand, we check `size+thread.allocated` and make the fast-threshold ahead more conservative (to be the previous size class before MAX_LOOKUP_SIZE).

Verified that the assembly of fast path has branches cut down from 3 to 2. One more computation on size+allocated though, in addition to `usize+allocated` which is needed to advance `thread.allocated`.

interwq:
oops this breaks the overflow detection of the threshold, which was the initial purpose of the fast threshold.

yinan1048576:
Yeah `size` can be anything so it's hard to handle. Not sure if there's any way around.

interwq:
@yinan1048576: the correctness shouldn't be affected; we should still be able to handle overflow as designed. I'm mainly concerned about the added complexity, and if the benefit is enough to justify.

davidtgoldblatt:
What sorts of benefit are we talking about? The complexity is enough that I'd be given pause if it were mostly a theoretical win, but not if it weren't.

interwq:
@davidtgoldblatt : agreed. The benefits on paper are:

- one less branch. The first thing in malloc is not branching, but loading tsd now.
- loading usize is not on the critical path anymore (it's only used to increment thread.allocated at the end, which goes to the store buffer).
- i.e. the first branch (compare against fast threshold) doesn't depend on usize anymore.

I'll try some production benchmark to see if it's measurable change.

interwq:
Wait... I think Yinan was correct. The overflow handling was hard. The correctness around fast_threshold depends on the size being in range. I don't see a good way to dealing with this right now. Closing.

PR #1690, opened 2019-11-21
Fix incorrect usage of cassert.

interwq:
I thought cassert was for static assert. Apparently it's intended for config assertion instead.

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -402,7 +402,7 @@ tsd_tcache_enabled_data_init(tsd_t *tsd) {
 
 static bool
 tcache_bin_init(cache_bin_t *bin, szind_t ind, uintptr_t *stack_cur) {
-	cassert(sizeof(bin->cur_ptr) == sizeof(void *));
+	assert(sizeof(bin->cur_ptr) == sizeof(void *));
 	/*
 	 * The full_position points to the lowest available space.  Allocations
 	 * will access the slots toward higher addresses (for the benefit of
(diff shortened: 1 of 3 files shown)


yinan1048576:
I actually thought about this before and I have a few comments:

- `cassert()` is essentially implemented as `assert()` without being restricted to debug build. (See https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/assert.h). So if judging only from its implementation, `cassert()` can be used to assert anything and it's always on regardless of the build mode. I wonder if we may actually desire such a strong assert in some places in our codebase.
- The implementation of `cassert()` also makes it great for asserting config variables, in which case the logic can either trigger compile time error or be optimized out.
- I think the implementation of `cassert()` also makes it great for more general static assert; the only thing is its name doesn't sound like a good fit.

So what I have in mind is:
- Define a strong assert (with some good name) that can be used to check conditions in run time regardless of the build mode. I suspect that we have some good usage cases for it.
- Define a static assert (with some good name) that simply calls the strong assert. It should be used for doing any static assertions other than asserting config.
- Define the `cassert()` that simply calls the static assert. It should be particularly used for asserting config variables. (I think this is optional; it's more just for having a nice name when we're only asserting a config variable.)
- Keep the current `assert()` but change its name to reflect that it should only be used when we'd only like to assert in debug mode.

Thoughts on the design of this hierarchy?

interwq:
Discussed offline with Yinan. A couple of more things we need to consider:
- fast path assertions currently are never executed; debug build will set malloc_slow and always bypass fast fast-path.
- assertions on slow path in opt build. Down a few paths often triggered by heap corruption, assertions with hints should avoid confusion better than just segfaults.

That being said, let's solve those separately in a later PR @yinan1048576?

PR #1691, opened 2019-11-23
Correct stats for realloc() / rallocx()

yinan1048576:
`thread.allocated` and `thread.deallocated` should not be increased
when `realloc()` / `rallocx()` does not change the allocation.  This
also aligns with what `xallocx()` has been doing.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -3116,7 +3116,11 @@ do_rallocx(void *ptr, size_t size, int flags, bool is_realloc) {
 	}
 	assert(alignment == 0 || ((uintptr_t)p & (alignment - 1)) == ZU(0));
 
-	*tsd_thread_deallocatedp_get(tsd) += old_usize;
+	if (p == ptr && usize == old_usize) {
+		thread_event_rollback(tsd, usize);
+	} else {
+		*tsd_thread_deallocatedp_get(tsd) += old_usize;
+	}
 
 	UTRACE(ptr, size, p);
 	check_entry_exit_locking(tsd_tsdn(tsd));
--- a/test/unit/zero_realloc_free.c
+++ b/test/unit/zero_realloc_free.c
@@ -2,9 +2,6 @@
 
 static uint64_t
 deallocated() {
-	if (!config_stats) {
-		return 0;
-	}
 	uint64_t deallocated;
 	size_t sz = sizeof(deallocated);
 	assert_d_eq(mallctl("thread.deallocated", (void *)&deallocated, &sz,
@@ -19,10 +16,8 @@ TEST_BEGIN(test_realloc_free) {
 	ptr = realloc(ptr, 0);
 	uint64_t deallocated_after = deallocated();
 	assert_ptr_null(ptr, "Realloc didn't free");
-	if (config_stats) {
-		assert_u64_gt(deallocated_after, deallocated_before,
-		    "Realloc didn't free");
-	}
+	assert_u64_gt(deallocated_after, deallocated_before,
+	    "Realloc didn't free");
 }
 TEST_END
 
--- a/test/unit/zero_realloc_strict.c
+++ b/test/unit/zero_realloc_strict.c
@@ -2,9 +2,6 @@
 
 static uint64_t
 allocated() {
-	if (!config_stats) {
-		return 0;
-	}
 	uint64_t allocated;
 	size_t sz = sizeof(allocated);
 	assert_d_eq(mallctl("thread.allocated", (void *)&allocated, &sz, NULL,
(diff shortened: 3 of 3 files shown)


src/jemalloc.c
```diff
@@ -3116,7 +3116,11 @@ do_rallocx(void *ptr, size_t size, int flags, bool is_realloc) {
 	}
 	assert(alignment == 0 || ((uintptr_t)p & (alignment - 1)) == ZU(0));
 
-	*tsd_thread_deallocatedp_get(tsd) += old_usize;
+	if (p == ptr && usize == old_usize) {
+		thread_event_rollback(tsd, usize);
+	} else {
+		*tsd_thread_deallocatedp_get(tsd) += old_usize;
```
interwq: Let’s add a couple of test cases for the counters with realloc?

davidtgoldblatt:
> thread.allocated and thread.deallocated should not be increased
> when realloc() / rallocx() does not change the allocation.

I don't think this is correct; realloc is definitionally a new allocation (spec says "The realloc function deallocates the old object pointed to by ptr and returns a
pointer to a new object that has the size specified by size. The contents of the new
object shall be the same as that of the old object prior to deallocation, up to the lesser of
the new and old sizes. Any bytes in the new object beyond the size of the old object have
indeterminate values.").

The per-spec version also fits better into a conceptual framework IMO; it makes profiling decisions have a natural and easy interpretation (always do a new sample, regardless of whether the old allocation was sampled or not) that doesn't totally make sense in the case where the allocation is expanded in place.

davidtgoldblatt:
Or, put more succinctly: the stats reporting policies ought to be independent of our layout decisions. A given sequence of malloc/realloc/free operations should always change stats in the same way. Stats are user queries about their own behavior, not about ours.

For example, if we decide to start using realloc calls as an opportunity to move allocations to reduce fragmentation, then I don't think that should start affecting user `thread.allocated` queries; nothing they did changed, so we shouldn't tell them that something did.

yinan1048576:
I was not sure either whether this change is correct. In particular I also noticed that `realloc()` is always treated as `free()` and then `malloc()` from profiling perspective, which is actually what I liked from a logical point of view. The main questions I have if we don't make this change are:
- `xallocx()` would be different from `realloc()` and `rallocx()`, in the sense that:
  - `xallocx()` keeps the stats unchanged if the allocation is not moved. The manual says `xallocx()` does not move the allocation, but the manual also says that `realloc()` may move the allocation (where "may" means it's not always and it's natural for applications to relate the not-moved case to `xallocx()`)
  - `xallocx()` is also always equivalent to `free()` and then `malloc()` from profiling perspective, which again feels a pretty clean design to me but just kind of deviating from what the stats are recording.
- A real world example: say if an application does `p = realloc(p, same_size)` a million times, and then the application queries `thread.allocated`, it doesn't feel right to tell the application that the total allocation (and deallocation) size is a million times the size, when there's actually just one object being allocated. (This example may not be too crazy, because the application might not know that the same size is passed into `realloc()` repeatedly.)
So I think there must be something somewhere that should be changed.

davidtgoldblatt:
> xallocx() would be different from realloc() and rallocx()

I think that's fundamental though; the moving / nonmoving semantics are part of the API guarantee of xallocx. The stats don't change if the allocation isn't expanded because that's a failure case (similarly, if we're in the realloc failure case and it returns NULL, I also wouldn't expect that to change stats).

With realloc, they're an implementation detail. (The current profiling behavior feels sort of wrong to me in this respect; stack trace we grab if profiled should be associated with only the new part of the allocation rather than the old one I would think. That'd be very tricky to implement though, since the two allocations would share a profiling context).

> A real world example: say if an application does p = realloc(p, same_size) a million times, and then the application queries thread.allocated, it doesn't feel right to tell the application that the total allocation (and deallocation) size is a million times the size, when there's actually just one object being allocated.

The opposite feels much more intuitive to me; if the user reallocs an 8-byte allocation a million times, that's 8 million bytes of allocator activity. It would be weird if I could do that much allocator activity and have the allocation counters not change at all.

yinan1048576:
Your point on the difference between `xallocx()` and `realloc()` makes sense. On the real world example, I think the question is how the counter will be used by the application. One possible use is to remind the application how much allocation activity it did, and another is to tell the application the memory consumption of the thread. I bet the second possibility can be more dominant, especially given that we don't have any other counters like `thread.active` for each thread (like the `stats.active` count we have for the global memory consumption). If we keep things as is, I think we'd better clearly state about this caveat in the manual, e.g. "this number may deviate from the real memory consumption of the corresponding thread". Thoughts?

jasone:
@davidtgoldblatt summarizes my perspective exactly.

yinan1048576:
Just realized that applications can easily get the memory consumption by subtracting `thread.deallocated` from `thread.allocated`, so my concern becomes irrelevant.

PR #1694, opened 2019-12-04
Disable some spuriously-triggering warnings

davidtgoldblatt:
With recent versions of GCC, run_tests.sh will fail due to warnings. Disable them.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -250,6 +250,11 @@ if test "x$GCC" = "xyes" ; then
   JE_CFLAGS_ADD([-Wsign-compare])
   JE_CFLAGS_ADD([-Wundef])
   JE_CFLAGS_ADD([-Wno-format-zero-length])
+  dnl This warning triggers on the use of the universal zero initializer, which
+  dnl is a very handy idiom for things like the tcache static initializer (which
+  dnl has lots of nested structs).  See the discussion at.
+  dnl https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -250,6 +250,19 @@ if test "x$GCC" = "xyes" ; then
...
   JE_CFLAGS_ADD([-Wundef])
   JE_CFLAGS_ADD([-Wno-format-zero-length])
+  dnl For conditionally compiled features like profiling, we always include
+  dnl string buffers, even when the feature is disabled, in order to prevent us
+  dnl having to ifdef-away all the places where it touched (even though those
+  dnl places get compiled away).  To minimize the space overhead, we change the
+  dnl lengths to 1 when the feature is disabled.  Since we occasionally strncpy
+  dnl with a size of len - 1, this can lead to a call to strncpy with a size of
+  dnl 0, which triggers a string warning.
+  JE_CFLAGS_ADD([-Wno-stringop-truncation])
```
interwq: For the prof related string ops, maybe introduce a prof strncpy wrapper which bails when prof is off? I’m slightly hesitant here.

davidtgoldblatt: I don't think this suffices to fix the warning: https://gcc.godbolt.org/z/cm3ygz .

interwq: Yeah I remember now... let’s guard the strncpy with ifdef? Assuming it’s only one place, it feels a bit better than turning the check off.

davidtgoldblatt: Done.

PR #1696, opened 2019-12-05
A few more profiling tweaks

yinan1048576:


src/prof_data.c
```diff
@@ -374,6 +374,18 @@ prof_lookup(tsd_t *tsd, prof_bt_t *bt) {
 	return ret.p;
 }
 
+prof_tctx_t *
+prof_tctx_create(tsd_t *tsd) {
+	prof_tdata_t *tdata = prof_tdata_get(tsd, false);
+	if (tdata == NULL) {
+		return NULL;
```
yinan1048576: The `tdata` should have been initialized in the `prof_sample_accum_update()` call, but there might be cases where it can be destroyed before the logic reaches here?

src/prof_data.c
```diff
@@ -374,6 +374,18 @@ prof_lookup(tsd_t *tsd, prof_bt_t *bt) {
 	return ret.p;
 }
 
+prof_tctx_t *
+prof_tctx_create(tsd_t *tsd) {
+	prof_tdata_t *tdata = prof_tdata_get(tsd, false);
+	if (tdata == NULL) {
```
interwq: I don't think this can ever happen? We should have allocated tdata before reaching here.

src/prof_data.c
interwq: I don't think so. The tdata is owned by the current thread.

yinan1048576:
Use `assert()` rather than `not_reached()`.

yinan1048576:
Verified that `./run_tests.sh` passed.

PR #1699, opened 2019-12-10
Yet a few more profiling tweaks.

yinan1048576:


src/prof_data.c
```diff
@@ -1449,4 +1465,15 @@ prof_tctx_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
...
 
+void
+prof_tctx_try_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
+	malloc_mutex_assert_owner(tsd_tsdn(tsd), tctx->tdata->lock);
+	if (prof_tctx_should_destroy(tsd, tctx)) {
+		/* tctx->tdata->lock will be released in prof_tctx_destroy(). */
+		prof_tctx_destroy(tsd, tctx);
+	} else {
+		malloc_mutex_unlock(tsd_tsdn(tsd), tctx->tdata->lock);
+	}
```
interwq: nit: assert not owner of tdata->lock at the end of the function.

yinan1048576: The `tdata` may be gone in the `prof_tctx_destroy()` call, and in such cases we cannot assert on the `tdata->lock`.

PR #1702, opened 2019-12-14
Restructure setters for profiling info

yinan1048576:
Explicitly define three setters:
- `prof_tctx_set_not_sampled()`: set `prof_tctx` to `1U`, if we don't know in advance whether the allocation is large or not;
- `prof_tctx_reset()`: set `prof_tctx` to `1U`, if we already know in advance that the allocation is large;
- `prof_info_set()`: set a real `prof_tctx`, and also set other profiling info e.g. the allocation time.

Code structure wise, the prof level is kept as a thin wrapper, the large level only provides low level setter APIs, and the arena level carries out the main logic.

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -68,8 +68,7 @@ arena_prof_info_get(tsd_t *tsd, const void *ptr, alloc_ctx_t *alloc_ctx,
 }
 
 JEMALLOC_ALWAYS_INLINE void
-arena_prof_tctx_set(tsd_t *tsd, const void *ptr, alloc_ctx_t *alloc_ctx,
-    prof_tctx_t *tctx) {
+arena_prof_tctx_reset(tsd_t *tsd, const void *ptr, alloc_ctx_t *alloc_ctx) {
 	cassert(config_prof);
 	assert(ptr != NULL);
 
@@ -77,17 +76,17 @@ arena_prof_tctx_set(tsd_t *tsd, const void *ptr, alloc_ctx_t *alloc_ctx,
 	if (alloc_ctx == NULL) {
 		extent_t *extent = iealloc(tsd_tsdn(tsd), ptr);
 		if (unlikely(!extent_slab_get(extent))) {
-			large_prof_tctx_set(extent, tctx);
+			large_prof_tctx_reset(extent);
 		}
 	} else {
 		if (unlikely(!alloc_ctx->slab)) {
-			large_prof_tctx_set(iealloc(tsd_tsdn(tsd), ptr), tctx);
+			large_prof_tctx_reset(iealloc(tsd_tsdn(tsd), ptr));
 		}
 	}
 }
 
-static inline void
-arena_prof_tctx_reset(tsd_t *tsd, const void *ptr, prof_tctx_t *tctx) {
+JEMALLOC_ALWAYS_INLINE void
+arena_prof_tctx_reset_sampled(tsd_t *tsd, const void *ptr) {
 	cassert(config_prof);
 	assert(ptr != NULL);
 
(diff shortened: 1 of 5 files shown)


include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -68,26 +68,26 @@ arena_prof_info_get(tsd_t *tsd, const void *ptr, alloc_ctx_t *alloc_ctx,
...
 		if (unlikely(!alloc_ctx->slab)) {
-			large_prof_tctx_set(iealloc(tsd_tsdn(tsd), ptr), tctx);
+			large_prof_tctx_reset(iealloc(tsd_tsdn(tsd), ptr));
 		}
 	}
 }
 
 static inline void
-arena_prof_tctx_reset(tsd_t *tsd, const void *ptr, prof_tctx_t *tctx) {
+arena_prof_tctx_reset(tsd_t *tsd, const void *ptr) {
```
interwq: Just realized, the above two functions (reset and set not sampled) are doing the same thing, just one is more optimized with alloc_ctx. I’d say keep the reset name, with the alloc_ctx implementation.

src/large.c
```diff
@@ -383,6 +383,9 @@ large_prof_tctx_reset(extent_t *extent) {
 }
 
 void
-large_prof_alloc_time_set(extent_t *extent, nstime_t *t) {
-	extent_prof_alloc_time_set(extent, t);
+large_prof_info_set(extent_t *extent, prof_tctx_t *tctx) {
+	large_prof_tctx_set(extent, tctx);
+	nstime_t t = NSTIME_ZERO_INITIALIZER;
```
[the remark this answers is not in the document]
yinan1048576: The `nstime_update()` function will read the previous value to make sure that time increases monotonic, which is the reason why we need `NSTIME_ZERO_INITIALIZER`. I think there's a need to consolidate the logic into a utility function in `nstime`. Will do so in a separate PR.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -68,26 +68,26 @@ arena_prof_info_get(tsd_t *tsd, const void *ptr, alloc_ctx_t *alloc_ctx,
...
 	if (alloc_ctx == NULL) {
 		extent_t *extent = iealloc(tsd_tsdn(tsd), ptr);
 		if (unlikely(!extent_slab_get(extent))) {
-			large_prof_tctx_set(extent, tctx);
+			large_prof_tctx_reset(extent);
 		}
 	} else {
 		if (unlikely(!alloc_ctx->slab)) {
-			large_prof_tctx_set(iealloc(tsd_tsdn(tsd), ptr), tctx);
+			large_prof_tctx_reset(iealloc(tsd_tsdn(tsd), ptr));
```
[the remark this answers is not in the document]
yinan1048576: `iealloc()` still needs the `ptr` to fetch the extent?

include/jemalloc/internal/arena_inlines_b.h
yinan1048576: Right. Changed `prof_tctx_reset()` to `prof_tctx_reset_sampled()` and changed `prof_tctx_set_not_sampled()` to `prof_tctx_reset()`.

yinan1048576:

include/jemalloc/internal/arena_inlines_b.h
interwq: If they are doing the same thing, maybe merge the two? Or merge the underlying impl.

include/jemalloc/internal/arena_inlines_b.h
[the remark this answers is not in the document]
interwq: I’m misread, never mind.

include/jemalloc/internal/arena_inlines_b.h
yinan1048576: The optimized version just avoids some checking. I think merging the two would essentially end up reducing to the full version and getting rid of the optimization. Is this the intention?

interwq: We can keep the optimized one, i.e. `arena_prof_tctx_reset` still takes an additional alloc_ctx. `arena_prof_tctx_reset_sampled` can simply call the former and pass NULL alloc_ctx.

yinan1048576: Right, that's one of the checks. The other is the slab check, which the optimized version doesn't do. If we "squash" the layers, the optimized one basically just contain a single set operation, and it's used whenever the slab check can already been inferred.

interwq: Ah I see. Missed that part. Never mind.

yinan1048576:
Yes, so that there'll be no artificial "zero-init-then-update" construct.

yinan1048576:
Rebase.

yinan1048576:
Fix comments.

PR #1705, opened 2019-12-16
Extent module refactorings

davidtgoldblatt:
This goes much of the way towards splitting up the extent module into smaller subpieces.

- It pulls out hook invocation into its own module (with an eye towards an eventual V2 interface)
- Drops arena dependencies from the places where it's easy to.
- Makes the extent module header non-global.
- Pulls out implicit zeroing logic, functionality, and debug checks.
- Duplicates some small bits of state (e.g. indices and base allocators) to prevent those types from being "viral" throughout all code.
- Gives the arena-level caching of unused extent_t objects its own type.

The remaining extent-level dependencies on the arena are for stats (VM abandoning), and the oversized threshold purging logic. I don't think either of these are fundamental (and I think we'll have to change the latter for extent hooks V2), but they're a little bit of effort to kill off, and I'd rather get the easy stuff in before winter vacations rather than let them bitrot.


davidtgoldblatt:
run_tests.sh passes. Failing CI checks are transient packaging issues (i.e. unable to download dependencies).

src/edata_cache.c
```diff
@@ -0,0 +1,47 @@
...
+		return base_alloc_edata(tsdn, base);
+	}
+	edata_avail_remove(&edata_cache->avail, edata);
+	atomic_fetch_sub_zu(&edata_cache->count, 1, ATOMIC_RELAXED);
+	malloc_mutex_unlock(tsdn, &edata_cache->mtx);
+	return edata;
+}
+
+void
+edata_cache_dalloc(tsdn_t *tsdn, edata_cache_t *edata_cache, edata_t *edata) {
```
interwq: nit: I kind of want a name to indicate this is returning to the cache (dealloc can be a bit ambiguous), if you can find one (I tried but couldn't). Feel free to ignore.

src/ehooks.c
```diff
@@ -198,15 +241,12 @@ ehooks_default_merge_impl(void *addr_a, void *addr_b) {
 static bool
 ehooks_default_merge(extent_hooks_t *extent_hooks, void *addr_a, size_t size_a,
     void *addr_b, size_t size_b, bool committed, unsigned arena_ind) {
-	if (!maps_coalesce) {
```
interwq: The `maps_coalesce` was a static check; removing it means the non-Windows cases will do the two `iealloc` unconditionally. Maybe preserve the check for now?

src/extent2.c
```diff
@@ -194,6 +197,35 @@ extents_alloc(tsdn_t *tsdn, arena_t *arena, ehooks_t *ehooks, ecache_t *ecache,
 	return edata;
 }
 
+edata_t *
+extents_alloc_grow(tsdn_t *tsdn, arena_t *arena, ehooks_t *ehooks,
+    ecache_t *ecache, void *new_addr, size_t size, size_t pad, size_t alignment,
```
interwq: ecache isn't being used. Is that intentional?

include/jemalloc/internal/ehooks.h
```diff
@@ -1,4 +1,151 @@
...
+static inline bool
+ehooks_merge_will_fail(ehooks_t *ehooks) {
+	return ehooks_get_extent_hooks_ptr(ehooks)->merge == NULL;
+}
+
+static inline void *
+ehooks_alloc(ehooks_t *ehooks, void *new_addr, size_t size, size_t alignment,
+    bool *zero, bool *commit, unsigned arena_ind) {
+	extent_hooks_t *extent_hooks = ehooks_get_extent_hooks_ptr(ehooks);
+	return extent_hooks->alloc(extent_hooks, new_addr, size, alignment,
```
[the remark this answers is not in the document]
davidtgoldblatt: Currently, we disallow a NULL alloc hook (i.e. this version is the behavior-neutral one). I think it arguably makes sense to allow it, and treat an attempt to invoke it as a failed/declined request (the same way that a NULL merge call is treated as the hooks refusing to merge). This could be used (for example) to create an arena that can't consume any more memory than it has currently allocated. But at least in the short term I figured we'd minimize changes here (especially ones that commit us to more guarantees while we're thinking of changes in this area).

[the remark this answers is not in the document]
interwq: Makes sense. Maybe add an assertion when init the hooks.

src/ehooks.c
```diff
@@ -209,6 +209,23 @@ ehooks_default_merge(extent_hooks_t *extent_hooks, void *addr_a, size_t size_a,
...
+ehooks_default_zero_impl(void *addr, size_t size) {
+	/*
+	 * By default, we try to zero out memory using OS-provided demand-zeroed
+	 * pages.  If the user has specifically requested hugepages, though, we
+	 * don't want to purge in the middle of a hugepage (which would break it
+	 * up), so we act conservatively and use memset.
+	 */
+	bool needs_memset = true;
+	if (opt_thp != thp_mode_always) {
+		needs_memset = pages_purge_forced(addr, size);
```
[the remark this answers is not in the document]
davidtgoldblatt: These are logically at the same level of the stack; they're the implementation of a new "zero" hook. So it morally feels like they should call the underlying functions, rather than something else at the same level. Not a strong opinion though, I'll stop arguing if you disagree.

include/jemalloc/internal/ehooks.h
```diff
@@ -106,18 +106,54 @@ ehooks_merge_will_fail(ehooks_t *ehooks) {
...
+	assert(size > 0);
+	if (config_debug) {
+		/* Check the whole first page. */
+		size_t *p = (size_t *)addr;
+		for (size_t i = 0; i < PAGE / sizeof(size_t); i++) {
+			assert(p[i] == 0);
+		}
+		/* And 32 spots within. */
+		assert(PAGE >= sizeof(size_t) * 32);
+		for (size_t i = 0; i < 32; ++i) {
```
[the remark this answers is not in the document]
davidtgoldblatt: Will do; what do you think about checking 4 (or even 2) pages instead of 32? I can think of some non-crazy bug patterns that would only zero the first page or so.

src/edata_cache.c
davidtgoldblatt: acquire/release maybe? Or remove/insert?

I also like not naming this alloc/dalloc just because edata_cache_alloc and ecache_alloc is a recipe for confusion.

src/ehooks.c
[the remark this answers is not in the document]
interwq: No strong opinion too. Let’s keep it as is.

include/jemalloc/internal/ehooks.h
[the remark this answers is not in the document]
interwq: 2 or 4 pages sound fine to me.

src/edata_cache.c
interwq: I slightly lean towards remove/insert.

src/ehooks.c
davidtgoldblatt: The reason I figured I wasn't so concerned was that the lookups only ever happen if ehooks_merge_default is invoked directly. But by default, the ehooks_merge call will short-circuit into the `_impl` variant, avoiding the lookup. (So, the only way this function gets invoked is if the user sneaks out one of our internal function pointers and wraps it or something). What do you think about a middle ground where ehooks_merge checks if the merge hook specifically is non-default?

interwq: Sounds good!

src/extent2.c
davidtgoldblatt: No; follow-on diffs were going to try to break the arena dependency here, but I thought this stack of commits was already getting a little bit big. I'll remove it.

interwq: Ah I see. Feel free to keep it then.

src/edata_cache.c
davidtgoldblatt: Talked offline: get/put

include/jemalloc/internal/ehooks.h
```diff
@@ -0,0 +1,375 @@
...
+	if (config_debug) {
+		/* Check the whole first page. */
+		size_t *p = (size_t *)addr;
+		for (size_t i = 0; i < PAGE / sizeof(size_t); i++) {
+			assert(p[i] == 0);
+		}
+		/* And 4 spots within. */
+		const size_t nchecks = 4;
+		assert(PAGE >= sizeof(size_t) * 4);
+		for (size_t i = 0; i < 4; ++i) {
```
interwq: nit: replace 4 with nchecks

davidtgoldblatt: Done. I also beefed up the comment, since the debug-perf thing isn't totally obvious.

PR #1706, opened 2019-12-16
Initialization utilities for nstime

yinan1048576:


src/nstime.c
```diff
@@ -168,3 +170,14 @@ nstime_update_impl(nstime_t *time) {
 	return false;
 }
 nstime_update_t *JET_MUTABLE nstime_update = nstime_update_impl;
+
+void
+nstime_init_zero(nstime_t *time) {
```
interwq: nit: move this to header and make inline.

src/nstime.c
```diff
@@ -8,6 +8,8 @@
 #define BILLION	UINT64_C(1000000000)
 #define MILLION	UINT64_C(1000000)
 
+static const nstime_t zero = NSTIME_ZERO_INITIALIZER;
```
interwq: nit: move this to init_zero().

yinan1048576:

PR #1708, opened 2019-12-18
Refactor profiling headers

yinan1048576:
The goal is to denoise `prof_externs.h`, so that prof internal functions are only tied to places defining / using them without being exposed externally.

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -1119,6 +1119,7 @@ prof_prefork0(tsdn_t *tsdn) {
 		for (i = 0; i < PROF_NTDATA_LOCKS; i++) {
 			malloc_mutex_prefork(tsdn, &tdata_locks[i]);
 		}
+		malloc_mutex_prefork(tsdn, &log_mtx);
 		for (i = 0; i < PROF_NCTX_LOCKS; i++) {
 			malloc_mutex_prefork(tsdn, &gctx_locks[i]);
 		}
@@ -1150,6 +1151,7 @@ prof_postfork_parent(tsdn_t *tsdn) {
 		for (i = 0; i < PROF_NCTX_LOCKS; i++) {
 			malloc_mutex_postfork_parent(tsdn, &gctx_locks[i]);
 		}
+		malloc_mutex_postfork_parent(tsdn, &log_mtx);
 		for (i = 0; i < PROF_NTDATA_LOCKS; i++) {
 			malloc_mutex_postfork_parent(tsdn, &tdata_locks[i]);
 		}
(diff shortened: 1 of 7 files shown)


yinan1048576:
Moving some of the declarations to the `prof_inlines_*` files still expose them out though, so there's no effective difference.

include/jemalloc/internal/prof_internal_externs.h
```diff
@@ -0,0 +1,14 @@
+#ifndef JEMALLOC_INTERNAL_PROF_INTERNAL_EXTERNS_H
```
interwq: Maybe keep these still in prof_externs? It’s a bit strange to have an _internal_externs

yinan1048576: Got rid of `prof_internal_externs.h` by -
- Moved `prof_getpid()` and `prof_get_default_filename()` back to `prof_externs.h`. After all, they're quite general and may be externally needed.
- Moved the two `*_mutex_choose()` functions into `prof_data.c` and made them static. Also relocated the corresponding mutexes there but exposed them so that they can still be handled in `prof.c`.

yinan1048576:
Revised according to feedback. Additionally exposed the mutex in `prof_log.c` so that it can be handled during forking (which was previously missing).

include/jemalloc/internal/prof_inlines_b.h
```diff
@@ -5,6 +5,9 @@
 #include "jemalloc/internal/sz.h"
 #include "jemalloc/internal/thread_event.h"
 
+prof_tdata_t *prof_tdata_init(tsd_t *tsd);
+prof_tdata_t *prof_tdata_reinit(tsd_t *tsd, prof_tdata_t *tdata);
```
interwq: Why not keep them in prof_extern.h? I don’t see why moving them into _inlines.h is better.

include/jemalloc/internal/prof_inlines_a.h
```diff
@@ -3,6 +3,9 @@
 
 #include "jemalloc/internal/mutex.h"
 
+bool prof_idump_accum_impl(tsdn_t *tsdn, uint64_t accumbytes);
+void prof_idump_rollback_impl(tsdn_t *tsdn, size_t usize);
```
interwq: Similarly here.

src/prof.c
```diff
@@ -1147,6 +1119,7 @@ prof_prefork0(tsdn_t *tsdn) {
 		for (i = 0; i < PROF_NTDATA_LOCKS; i++) {
 			malloc_mutex_prefork(tsdn, &tdata_locks[i]);
 		}
+		malloc_mutex_prefork(tsdn, &log_mtx);
```
interwq: Split into a separate commit? This looks to be bug fix.

include/jemalloc/internal/prof_inlines_b.h
yinan1048576: As I mentioned in the earlier comment, there's no effective difference, in the sense that these functions are still exposed out. The only difference is for better understanding by first time readers: it appears clearer that these functions are only tied to these `prof_inlines_*` files rather than needed externally.

However, if it's not worthwhile to do so, I'm open to putting them back.

interwq: I worry about the order of the headers (e.g. externs are included first); also the inline functions may change. I don’t think there are other inline headers doing this, so let’s not create an exception here.

yinan1048576:

PR #1717, opened 2019-12-31
Disallow merge across mmap regions to preserve SN / first-fit

interwq:
Check the is_head state before merging two extents.  Disallow the merge if it's
crossing two separate mmap regions.  This enforces first-fit (by not losing the
SN) at a very small cost.


interwq:
@davidtgoldblatt -- I believe this is the right fix; also hopefully the logic looks a bit clearer now.

I'm still testing but want to get a stamp before your travel -- so far the results look neutral on the big binaries as expected. Still trying to measure if VM size has any change.

interwq:
Benchmark finished. No regression as expected; the vm size & metadata look to be slightly lower (a couple %).

PR #1719, opened 2020-01-02
Fix an assertion on extent head state with dss.

interwq:
Caught by the CI in https://github.com/jemalloc/jemalloc/pull/1717. The tests pass after rebasing on top of this.

### DIFF
--- a/src/ehooks.c
+++ b/src/ehooks.c
@@ -221,7 +221,8 @@ ehooks_no_merge_heads(tsdn_t *tsdn, void *addr_a, bool head_a, void *addr_b,
 		assert(!ehooks_same_sn(tsdn, addr_a, addr_b));
 		return true;
 	}
-	assert(ehooks_same_sn(tsdn, addr_a, addr_b));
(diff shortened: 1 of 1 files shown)


src/ehooks.c
```diff
@@ -221,7 +221,8 @@ ehooks_no_merge_heads(tsdn_t *tsdn, void *addr_a, bool head_a, void *addr_b,
 		assert(!ehooks_same_sn(tsdn, addr_a, addr_b));
 		return true;
 	}
-	assert(ehooks_same_sn(tsdn, addr_a, addr_b));
+	assert(ehooks_same_sn(tsdn, addr_a, addr_b) || (have_dss &&
+	    (extent_in_dss(addr_a) || extent_in_dss(addr_b))));
```
davidtgoldblatt: Should the inner condition not be `&&`? It doesn't seem like we should merge dss and non-dss extents.

interwq: I'm hesitant here too (not leaning either way really). Went conservative as this function was for permission checking; and it seems we don't necessary have to do in-dss checking before this one?

PR #1720, opened 2020-01-03
Fine-grained sharding: add default bins for auto arenas.

interwq:
Make the default bins live outside of the auto arenas.  This will help
implementing a thread-to-bin binding w/o relying on arena.

This still assumes the current bin-arena binding strategy, however the
`default_bins` will allow us to change that soon.


interwq:
Ready for review. The tests failure before was about the sharded bins unit test itself.

davidtgoldblatt:
Discussed offline -- we think we want to go a different direction with this. After #1804 it will be easier to remove page-level functionality out of the arena than to move the bins out. At that point, an arena can be "just a bucket of bins".

PR #1722, opened 2020-01-09
Make last-N dumping non-blocking

yinan1048576:
A few remarks:
- I'm chopping last-N dumping into batches, each under the `prof_recent_alloc_mtx`, so that sampled `malloc` and `free` can proceed between the batches, rather than being blocked until the entire dumping process finishes.
- I'm using the existing `prof_dump_mtx` to cover the entire dumping process, during which I first change the limit to unlimited (so that existing records can stay), then perform the dumping batches, and finally revert the limit back (and shorten the record list). `prof_dump_mtx` serves to only permit one thread at a time to dump, either the last-N records or the original stacktrace-based profiling information. An alternative approach is to use a separate mutex for the last-N records, so that the two types of dumping can take place concurrently. Thoughts?
- I'm additionally changing the `mallctl` logic for reading and writing the limit: they now need the `prof_dump_mtx`. For reading, this ensures that what's being read is always the real limit. For writing, this ensures that the application cannot change the limit during dumping. The downside is that the `mallctl` calls are blocked until the entire dumping process finishes, but I think it's fine, because the `mallctl` calls are very rare and only initiated by the application.
- I'm increasing the buffer size to be the same as the size used by stats printing and the original profiling dumping. I think I could even consolidate the last-N buffer with the original profiling buffer, especially since I'm already using `prof_dump_mtx`. Thoughts? I could have a separate commit for that, since that'd also need some refactoring of the original profiling dumping logic.
- The batch size is chosen to be 64. I figured making such a choice is quite tricky and here's how I get it -
  - The goal I'm pursuing is to find a batch size so that each batch can trigger at most one I/O procedural call: the worst case blocking time is always at least one I/O, so a smaller batch size cannot reduce the worst case blocking time, while a larger batch size can multiply the worst case blocking time.
  - The amount of output per record depends primarily on (a) the length of the stack trace and (b) whether the record has been released (two stack traces if released; one if not). I examined last-N dumps from production, 4 per service, and found that one of the services happened to have both the longest stack traces and the highest proportion of released records, and the average length per record for that service is in the order of 800-900 characters (in compact JSON format). So, if I set the batch size to be 64 records, each batch will at most output less than but close to 64K characters, which is the size of the buffer.


src/prof_recent.c
```diff
@@ -466,67 +479,95 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
+			emitter_json_kv(&emitter, "released",
+			    emitter_type_bool, &released);
+
+			emitter_json_kv(&emitter, "alloc_thread_uid",
+			    emitter_type_uint64, &n->alloc_tctx->thr_uid);
+			uint64_t alloc_time_ns = nstime_ns(&n->alloc_time);
+			emitter_json_kv(&emitter, "alloc_time",
+			    emitter_type_uint64, &alloc_time_ns);
+			emitter_json_array_kv_begin(&emitter, "alloc_trace");
+			dump_bt(&emitter, n->alloc_tctx);
```
interwq: It would be better to not do the actual dumping holding the recent lock. Even with a small batch size, if the dump involves IO there will be latency spikes for regular requests. Let’s try read out the information under the recent lock (keep the info on stack), and drop the lock when dump?

Also thanks for getting to this quickly!

yinan1048576: Do you first read out everything under the lock, and then drop the lock and write it out? We may need a significant amount of space to store all the information. One allocation record can output up to 800-900 characters.

yinan1048576: I see what you mean. Changed to: read out each batch under the lock, drop the lock and do I/O. Since the buffer is large enough to hold the entire batch, there should not be any I/O during the reading. (It's hard to make an absolute guarantee because there's no way to accurately predict the output size of the batch upfront.)

interwq: I'd say to push it a bit further, and guarantee non-blocking: we can add a `_nonblocking` version of the dump_bt, which attempts only append to the buffer w/o actually dumping, and returns whether it could do so. That way we don't really need a batch size, we can simply keep buffering the output, and only unlock and dump when the buffer is full.

yinan1048576: Discussed offline, will develop a different buffered writer to meet such needs.

yinan1048576:
Rebase, and improved the buffered writing design.

I haven't implemented the details for the new buffered writing functionality. Currently they're all `TODO` placeholders. Will fill them in later. A few things to note at this stage:
- I'm temporarily naming the new mode as "manual": in such a mode, the caller has to manually call dump; otherwise `write_cb()` will never call dump and will stop writing to the buffer as soon as the buffer would overflow.
- `write_cb()` doesn't return any error. Instead, I'll be defining a boolean field in `buf_write_arg_t` indicating whether `write_cb` was not able to write to the buffer, and the caller can call `buf_write_overflow()` to check that field. The benefit, in addition to complying with the `void (*write_cb)(void *, const char *)` signature we've been using in all places, is that it's more convenient: the caller side may be a chain (in our case: `dump()` -> `dump_bt()` -> `emitter_*()` -> `write_cb()`), so if `write_cb()` returns an error, we'd have to pass it through the chain; if `write_cb()` only sets the error field, then any caller along the chain can easily check the field without the need of any extra relaying logic.

Meanwhile, the change on the caller side is complete and is ready for review: the dumping logic now looks like the following:
- Gain the lock.
- Continue writing the records to the buffer until the point there's no more records or the buffer would overflow.
  - If there's no more records to write, simply release the lock.
  - If the buffer would overflow, discard the intermediate content of the "overflow-triggering" record in the buffer, release the lock, dump, and then repeat from the beginning (starting from the "overflow-triggering" record in the last round).
- Always dump one more time at the end.

yinan1048576:
Forgot about the case where a single record may overflow the entire buffer. Also, in the end, there should be some special treatments. Will revise.

yinan1048576:
Addressed both issues. Also realized that the beginning part needs special treatment as well.

yinan1048576:
Changed the design to be the following:
- Lock the dumping mutex.
- Lock the last-N mutex, copy the list head to local, reset the original list, and release the lock. Effectively what other threads would see from this point forward is an empty list.
- Perform exactly the same segmented dumping as before, but on the local copy.
- Release the dumping mutex.
- Delete the local copy.

The main difference in this design is the local copying, whereas the previous design first set the limit to unlimited, and then dump the original list. The main drawback of the previous design is that there can be an unbounded memory peak during dumping. The new design has a `2 * limit` bound. Moreover, the `mallctl` getter and setter no longer need to first gain the dumping lock - they can freely read and write since everything is isolated from the local copy.

There's one thing I'm not sure: when dumping finishes, should I simply delete the local copy, or should I insert it back to the front of the new list (if there's extra space)? Neither one is perfect (and that's the price we pay for getting a nice memory bound):
- Delete: the next dump won't contain any allocation being dumped last time. It's good in that there's no redundancy in the dumps. However, the application also loses the ability to record any new deallocation events on the allocations dumped last time. 
- Insert back: the application can see in the next dump any new deallocation events on the allocations dumped last time. However, the concatenated list may not be a true last N trailing window, because there can be missing allocations in the middle (e.g. if the extra space is a result from a decrease-then-increase of the limit).

Thoughts?

interwq:
re: after dumping -- I'd say, insert them back (but from the other end?) only if the curr # is less than the desired N. I'd also require the dump lock on the mallctl set / get (no need to optimize for them really), to avoid things change in the middle.

interwq:
The always-delete approach may cause confusion, as people may take two profiles and compare.

yinan1048576:
Revised according to feedback. Several things to note:
- I made a reservation for the copy of the dummy head in `init()`, so that there's no need to allocate or release in the dumping function. `init()` would report error if the reservation fails. Thoughts? If it looks good, I'll do some refactoring on the node creation logic.
- On inserting back after dumping: the implementation was more interesting than I thought:
  - On a high level, the logic is as follows:
    - Acquire the last-N lock.
    - (*) In the just-dumped list, pick the record, say `t`, so that the segment from `t` to the end should be inserted back to the front of the new list (without going over `max`). (The segment may be empty or the full just-dumped list.)
    - Do the insertion.
    - Release the last-N lock and the dumping lock.
    - (*) Handle the deletions for the records up to `t` in the just-dumped list.
  - After I wrote this logic, I realized it's equivalent to the following logic:
    - Acquire the last-N lock.
    - Insert the entire just-dumped list to the front of the new list.
    - Release the last-N lock.
    - Correct the length of the list if the insertion above caused the list to grow beyond `max`, by calling `prof_recent_alloc_max_write()`, which essentially take care of the two steps labelled by (*) above.
    - Release the dump lock.
  - The main advantage is that, by calling `prof_recent_alloc_max_write()`, the code logic is significantly simplified. (What remained were all trivial.) But there are some downsides:
    - The dump lock needs to be held longer.
    - `max` may change during the dumping process, so `prof_recent_alloc_max_ctl_read()` needs the dump lock.
  - I feel the benefit of logic simplification outweighs the downsides. Thoughts?

yinan1048576:
Rebase. It was great to observe that the new stress test can still pass (at least locally on my machine), with `prof_recent_alloc_dump()` being newly implemented.

No significant difference in this revision; the new buffered writer functionalities are still placeholders. Once we finalize the caller side logic in `prof_recent_alloc_dump()`, I'll be implementing the new buffered writer functionalities.

src/prof_recent.c
```diff
@@ -506,33 +519,106 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
 
-	max = prof_recent_alloc_max_update(tsd, max);
-	assert(max == 0);
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
+	prof_recent_alloc_assert_count(tsd);
+	bool need_adjust_back = false;
+	ssize_t new_max = 0;
+	if (dump_count > 0) {
+		assert(last != NULL && last->next == dump_dummy);
+		last->next = prof_recent_alloc_dummy->next;
```
interwq: Let’s add a merge utility function.

src/prof_recent.c
```diff
@@ -19,6 +19,7 @@ malloc_mutex_t prof_recent_alloc_mtx; /* Protects the fields below */
 static atomic_zd_t prof_recent_alloc_max;
 static ssize_t prof_recent_alloc_count = 0;
 static prof_recent_t *prof_recent_alloc_dummy = NULL;
+static prof_recent_t *prof_recent_alloc_dummy_reserve = NULL;
```
interwq: This could use some comments.

src/prof_recent.c
```diff
@@ -556,5 +642,18 @@ prof_recent_init() {
 	}
 	prof_recent_alloc_dummy->next = prof_recent_alloc_dummy;
 
+	assert(prof_recent_alloc_dummy_reserve == NULL);
```
interwq: Allocate the two nodes together is fine.

src/prof_recent.c
```diff
@@ -506,33 +519,106 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
 	    buf_writer_get_cbopaque(&buf_writer));
+
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_dump_mtx);
 	emitter_begin(&emitter);
 
 	malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
 	prof_recent_alloc_assert_count(tsd);
+	prof_recent_t *dump_dummy = prof_recent_alloc_dummy;
+	assert(dump_dummy != NULL);
+	prof_recent_alloc_dummy = prof_recent_alloc_dummy_reserve;
```
interwq: Add an init utility function.

src/prof_recent.c
```diff
@@ -506,33 +519,106 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
+		malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
+		if (node == NULL) {
+			node = dump_dummy->next;
+		}
+		assert(node != NULL);
+		bool need_more_batch = false;
+		while (node != dump_dummy) {
+			buf_writer_commit(&buf_writer);
+			prof_recent_alloc_dump_node(&emitter, node);
+			if (buf_writer_overflow(&buf_writer)) {
```
interwq: Maybe have the dump_node above return if overflowed?

include/jemalloc/internal/buf_writer.h
```diff
@@ -30,4 +30,15 @@ void *buf_writer_get_cbopaque(buf_writer_t *buf_writer);
 void buf_writer_flush(buf_writer_t *buf_writer);
 void buf_writer_terminate(tsdn_t *tsdn, buf_writer_t *buf_writer);
 
+/*
+ * Everything below are advanced APIs, only for scenarios where there's need to
+ * manually control the flushing operations.
```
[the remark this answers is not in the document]
yinan1048576: Will do, after we finalize the logic design and naming.

src/prof_recent.c
```diff
@@ -506,33 +519,106 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
+	emitter_json_array_kv_begin(&emitter, "recent_alloc");
 
-	emitter_json_kv(&emitter, "recent_alloc_max", emitter_type_ssize, &max);
+	buf_writer_set_manual(&buf_writer);
+	prof_recent_t *node = NULL;
+	prof_recent_t *last = NULL;
+	while (true) {
+		malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
+		if (node == NULL) {
+			node = dump_dummy->next;
```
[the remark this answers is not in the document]
yinan1048576: I put it here mainly because I wanted it to be under the `prof_recent_alloc_mtx` lock. Strictly speaking, this `dump_dummy->next` doesn't have to be under the lock, because we know that `dump_dummy` points to a list where only the content of the nodes rather than the nodes' linkages might be changed. But I'm just putting it under the lock to be rigorous, and to also be safe in case we might change any logic here in the future.

src/prof_recent.c
```diff
@@ -506,33 +519,106 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
+		while (node != dump_dummy) {
+			buf_writer_commit(&buf_writer);
+			prof_recent_alloc_dump_node(&emitter, node);
+			if (buf_writer_overflow(&buf_writer)) {
+				buf_writer_revert(&buf_writer);
+				if (buf_writer_empty(&buf_writer)) {
+					buf_writer_set_auto(&buf_writer);
+					prof_recent_alloc_dump_node(
+					    &emitter, node);
+					buf_writer_set_manual(&buf_writer);
```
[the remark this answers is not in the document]
yinan1048576: Actually, when I'm looking at it again, I agree that the auto/manual naming doesn't sound that intuitive. Perhaps let's get the naming right first? How about `set_auto_flush` and `set_manual_flush`?

src/prof_recent.c
yinan1048576: You mean to allocate a size-2 array?

src/prof_recent.c
yinan1048576: You mean `prof_recent_alloc_dump_node()`? By the design of the buffered writer, it doesn't know whether overflow happened or not, because the buffered writer doesn't propagate any error up. To tell so, `prof_recent_alloc_dump_node()` needs to call `buf_writer_overflow()`, so I'd rather put the call here, so that all buffered writer related logic live together.

yinan1048576:
Revised according to feedback. Also see inline comments.

src/prof_recent.c
interwq: Right. That way I assume we can simply switch between the two items?

src/prof_recent.c
interwq: Agreed. Thanks for explaining!

src/prof_recent.c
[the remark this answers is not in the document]
interwq: Is it true that manual flush doesn’t allow silent flushing, and needs to trigger flush explicitly? If so, maybe one function to take a bool input which is whether flushing is allowed? _allow_flush?

src/prof_recent.c
```diff
@@ -506,33 +575,70 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
+				} else {
+					need_more_batch = true;
```
interwq: If we do unlock-flush-relock right here, it should simplify the logic?

src/prof_recent.c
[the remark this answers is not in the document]
interwq: Sounds good. Reading under the lock is the right thing to do.

src/prof_recent.c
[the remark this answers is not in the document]
yinan1048576: But even after calling `buf_writer_set_allow_flush(false)`, as you said, we can still explicitly call `buf_writer_flush()`, which seems to invalidate the `buf_writer_set_allow_flush` naming.

How about `buf_writer_set_allow_auto_flush()`, or `buf_writer_set_allow_implicit_flush()`? I think `auto` is better than `implicit`, because `implicit_flush` makes me feel that what I write is all implicitly flushed, which is not true.

src/prof_recent.c
yinan1048576: It won't be easier to switch. There's still the same logic like:
```
prof_recent_t *dump_dummy = prof_recent_alloc_dummy;
prof_recent_alloc_dummy = prof_recent_alloc_dummy_reserve;
/* ... dump the dump_dummy list and then plug it back to the prof_recent_alloc_dummy list ... */
prof_recent_alloc_dummy_reserve = dump_dummy;
```
Any idea how it can be simplified?

Putting the two pointers in an array does simplify the initialization logic though, which can be a solid enough supporting argument itself.

src/prof_recent.c
yinan1048576: Right, though such a change would look exactly like a revert of your #1763? :)

interwq: Yeah however that one was to enable flat combining and reduce locking for the bin lock (which is heavily contended); here the recent lock is on a slow-slow path (i.e. dumping), and there won't be multiple threads going the same path (because of the dump lock). So no need to really pay the complexity.

src/prof_recent.c
interwq: Yeah we would still need to do the swapping, we can only maintain a single global pointer of the array though (i.e. no array[0] is always the recent records, while array[1] always points to the dumping one).

src/prof_recent.c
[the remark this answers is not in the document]
interwq: I see your point. To make it tiny bit shorter, `buf_writer_allow_auto_flush()`?

src/prof_recent.c
```diff
@@ -383,14 +388,18 @@ prof_recent_alloc(tsd_t *tsd, edata_t *edata, size_t size) {
...
+	malloc_mutex_unlock(tsd_tsdn(tsd), &prof_dump_mtx);
+	return max;
 }
 
-ssize_t
-prof_recent_alloc_max_ctl_write(tsd_t *tsd, ssize_t max) {
+static ssize_t
+prof_recent_alloc_max_write_locked(tsd_t *tsd, ssize_t max) {
 	assert(max >= -1);
+	malloc_mutex_assert_owner(tsd_tsdn(tsd), &prof_dump_mtx);
```
interwq: Move the recent_alloc lock out of this function as well? Usually when I see locked, I assume all locks are taken care of.

src/prof_recent.c
```diff
@@ -446,8 +455,35 @@ prof_recent_alloc_max_ctl_write(tsd_t *tsd, ssize_t max) {
...
+	ssize_t old_max = prof_recent_alloc_max_write_locked(tsd, max);
+	malloc_mutex_unlock(tsd_tsdn(tsd), &prof_dump_mtx);
+	return old_max;
+}
+
+static prof_recent_t *
+prof_recent_alloc_swap_and_clear_locked(tsd_t *tsd, ssize_t *dump_count,
+    ssize_t *dump_max) {
+	malloc_mutex_assert_owner(tsd_tsdn(tsd), &prof_dump_mtx);
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
```
interwq: Similarly, don't lock internally here. Move the call site to inside of the lock, and invoke this one under the `node == NULL` initialization branch.

src/prof_recent.c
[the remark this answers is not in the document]
yinan1048576: `buf_writer_allow_auto_flush()` sounds like returning a boolean of whether the buffered writer currently allows auto flush or not, which can actually be a useful utility function to have, not for this usage case but perhaps for some other cases in the future. How about still including the `set`?

src/prof_recent.c
yinan1048576: Right, but even in that case, I would macro define 0 and 1, so that it looks like `prof_recent_alloc_dummy[REAL]` and `prof_recent_alloc_dummy[RESERVE]`, rather than barebone `prof_recent_alloc_dummy[0]` and `prof_recent_alloc_dummy[1]`. It would end up looking pretty much the same if I just define two named pointers each pointing to one position in the array. Thoughts?

src/prof_recent.c
yinan1048576: It's true that only one thread goes through the dump logic at a time because of the dump lock, but the `prof_recent_alloc_mtx` can be hot, since every thread needs it whenever there's a sampled allocation / deallocation.

src/prof_recent.c
yinan1048576: Right. I was trying to make it similar to `prof_recent_alloc_max_write_locked()`, which must be put under the dump lock but must not be under the recent lock. Should I also change the `prof_recent_alloc_max_write_locked()` name?

src/prof_recent.c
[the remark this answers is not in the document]
interwq: SGTM.

src/prof_recent.c
interwq: Yeah it won't be simplified too much, the thing I want from it is to have only one global variable. We should also add a getter to wrap around the array.

src/prof_recent.c
interwq: Discussed offline. The dump path won't be flat-combined by other threads, so we don't care about the lock pattern here.

src/prof_recent.c
interwq: I'd say still keep this function _locked(), i.e. and move the recent_mtx to the call sites: 1) the mallctl one can simply have the locking around the call; 2) the dumping one should be refactored so that it's called before we drop the recent_mtx in the last iteration?

yinan1048576: You mean `prof_recent_alloc_max_write_locked()` should be entered with the recent lock already been locked? If so it would need to temporarily drop the lock and re-lock if it needs to delete nodes, and in such cases the lock would have to be taken twice, rather than only once in the current logic.

interwq: Ah I see. Then keep the current logic and only avoid the _locked in the name is fine.

src/prof_recent.c
```diff
@@ -446,8 +455,35 @@ prof_recent_alloc_max_ctl_write(tsd_t *tsd, ssize_t max) {
...
+ssize_t
+prof_recent_alloc_max_ctl_write(tsd_t *tsd, ssize_t max) {
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_dump_mtx);
+	ssize_t old_max = prof_recent_alloc_max_write_locked(tsd, max);
+	malloc_mutex_unlock(tsd_tsdn(tsd), &prof_dump_mtx);
+	return old_max;
+}
+
+static prof_recent_t *
+prof_recent_alloc_swap_and_clear_locked(tsd_t *tsd, ssize_t *dump_count,
```
interwq: nit: I see why you have `clear` in the name, however it's a bit confusing (we are not really releasing the records, just stashed them in a separate list). Maybe a _prepare and _finish pair?

src/prof_recent.c
yinan1048576: Realized that we cannot define an array for the two dummies, because they'll be individually dynamic: `prof_recent_alloc_dummy` can be shifted during prof allocation, and may also be deleted in `mallctl`. The same is true for `prof_recent_alloc_dummy_reserve`, when it's swapped to be used as `prof_recent_alloc_dummy`.

Extracted out the swapping operation into a utility function. Also avoided passing `dump_dummy` around, since it's always `prof_recent_alloc_dummy_reserve` (if the swap is fully completed in `prof_recent_alloc_dump_prepare_locked()` rather than having half of it left to `prof_recent_alloc_dump_finish_locked()`.

src/prof_recent.c
yinan1048576: Revised. The logic looks much simpler now!

yinan1048576:

src/prof_recent.c
```diff
@@ -506,33 +580,65 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
 	emitter_init(&emitter, emitter_output_json_compact,
 	    buf_writer_get_write_cb(&buf_writer),
 	    buf_writer_get_cbopaque(&buf_writer));
+
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_dump_mtx);
 	emitter_begin(&emitter);
 
 	malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
-	prof_recent_alloc_assert_count(tsd);
+	ssize_t dump_count = prof_recent_alloc_dump_prepare_locked(tsd);
```
interwq: Can these two lines be moved to right before the for loop? That way we don't need the lock/unlock here.

src/prof_recent.c
```diff
@@ -506,33 +580,65 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
-	assert(max == 0);
 	malloc_mutex_unlock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
+	buf_writer_set_allow_auto_flush(&buf_writer, true);
 
+	emitter_json_array_end(&emitter);
 	emitter_end(&emitter);
 	buf_writer_terminate(tsd_tsdn(tsd), &buf_writer);
+
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
+	ssize_t new_max = prof_recent_alloc_dump_finish_locked(tsd, dump_last,
```
interwq: Also, this one be move to right after the for loop?

src/prof_recent.c
```diff
@@ -556,5 +662,18 @@ prof_recent_init() {
 	}
 	prof_recent_alloc_dummy->next = prof_recent_alloc_dummy;
 
+	assert(prof_recent_alloc_dummy_reserve == NULL);
+	prof_recent_alloc_dummy_reserve = (prof_recent_t *)iallocztm(
```
interwq: Abstract to a function allocating the node? If the only difference between this and the regular node allocation is the TSDN_NULL, it can be an input arg.

src/prof_recent.c
```diff
@@ -506,33 +580,65 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
-	for (prof_recent_t *n = prof_recent_alloc_begin(tsd);
-	    n != prof_recent_alloc_end(tsd);
-	    n = prof_recent_alloc_next(tsd, n)) {
-		prof_recent_alloc_dump_node(&emitter, n);
+	buf_writer_set_allow_auto_flush(&buf_writer, false);
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
+	prof_recent_t *dump_dummy = prof_recent_alloc_dummy_reserve;
+	for (prof_recent_t *node = dump_dummy->next; node != dump_dummy;
+	    node = node->next) {
+		buf_writer_commit(&buf_writer);
```
interwq: Why commit at the beginning of the loop? It's a bit counter intuitive.

src/buf_writer.c
```diff
@@ -134,3 +134,33 @@ buf_writer_terminate(tsdn_t *tsdn, buf_writer_t *buf_writer) {
...
+}
+
+bool
+buf_writer_overflow(buf_writer_t *buf_writer) {
+	/* TODO: check overflow */
+	return false;
+}
+
+void
+buf_writer_revert(buf_writer_t *buf_writer) {
```
interwq: Question: should we make the revert implicit? meaning if it's overflowed, revert the uncommitted stuff automatically.

If we keep the revert call, maybe name it  `_revert_uncommitted`? Also the _overflow above might be better as _overflowed.

src/prof_recent.c
yinan1048576: The challenge is, we don't want any flushing under the lock (unless if it's absolutely necessary, like the `if (buf_writer_empty(&buf_writer))` case). It's possible to move these statements right before the loop, but we need to handle any overflow from `emitter_json_kv(&emitter, "recent_alloc_max", emitter_type_ssize, &dump_max);` and `emitter_json_array_kv_begin(&emitter, "recent_alloc");`. If we really want, we can make sure that the buffer size is always longer than the sum of these two parts, but it doesn't seem to me that there's a clean way to do so.

src/prof_recent.c
yinan1048576: Good point.

src/prof_recent.c
yinan1048576: Because we always want to have the closest point to revert to if needed. What would be an alternative?

src/buf_writer.c
yinan1048576: Good point. Will change. Will also implement the internals in `buf_writer` since the outer logic has pretty much been settled.

src/prof_recent.c
interwq: Maybe this way? It feels we should always have a "clean" state when starting a new iteration.

    if (buf_writer_overflowed(...)) {
        ...
    } else {
        commit();
    }

src/prof_recent.c
interwq: I see. Let's keep it as is.

src/prof_recent.c
```diff
@@ -469,66 +580,65 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
-		} else {
-			assert(nstime_equals_zero(&n->dalloc_time));
+	buf_writer_set_allow_auto_flush(&buf_writer, false);
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
+	prof_recent_t *dump_dummy = prof_recent_alloc_dummy_reserve;
+	for (prof_recent_t *node = dump_dummy->next; node != dump_dummy;
+	    node = node->next) {
+		buf_writer_commit(&buf_writer);
+		prof_recent_alloc_dump_node(&emitter, node);
+		if (buf_writer_overflow(&buf_writer)) {
```
interwq: If overflowed, the current node is not written, and we should continue w/o the node=node->next?

src/prof_recent.c
yinan1048576: One more subtle difference: I'm here using `arena_get(TSDN_NULL, 0, true)`, whereas for prof allocations I'm using `arena_get(tsd_tsdn(tsd), 0, false)`. In other words I force create the arena in boot but not in real traffic. Do we want to change anything here?

src/prof_recent.c
yinan1048576: I think the main difference is the handling for the first node: if it overflows, then where should we revert to? That's dependent on the implementation of the buffered writer. In my current implementation, the caller would have to explicitly call `commit()` to set the commit point.

src/prof_recent.c
yinan1048576: That's great catch! I was particularly paying attention to this case in the more complicated version, but forgot about it in this new version.

yinan1048576:
Completed an initial version for the buffered writer internal implementations. I'm currently having the simplest implementation, and I feel many places may need extra guards / checks / overwrites. Let me know what you think. Meanwhile I'll also write some unit tests.

include/jemalloc/internal/buf_writer.h
```diff
@@ -21,6 +21,13 @@ typedef struct {
 	size_t buf_size;
 	size_t buf_end;
 	bool internal_buf;
+	/*
+	 * The following fields are for advanced APIs - see explanations
+	 * below.
+	 */
+	bool allow_auto_flush;
+	size_t commit;
```
interwq: maybe committed?

src/prof_recent.c
```diff
@@ -495,7 +536,40 @@ prof_recent_alloc_dump_node(emitter_t *emitter, prof_recent_t *n) {
...
+	prof_recent_alloc_assert_count(tsd);
+	ssize_t new_max = dump_max;
+	prof_recent_t *dump_dummy = prof_recent_alloc_dummy_reserve;
+	if (dump_count > 0) {
+		assert(dump_last != NULL && dump_last->next == dump_dummy);
+		dump_last->next = prof_recent_alloc_dummy->next;
+		assert(dump_dummy->next != dump_dummy);
+		prof_recent_alloc_dummy->next = dump_dummy->next;
+		dump_dummy->next = dump_dummy;
+		prof_recent_alloc_count += dump_count;
```
interwq: nit: separate those to a merge function?

src/prof_recent.c
interwq: I see. Let's keep as is.

src/prof_recent.c
interwq: I think the `true` for force creating isn't really necessary for the boot case? prof_recent should be initialized late enough.

src/buf_writer.c
```diff
@@ -134,3 +144,25 @@ buf_writer_terminate(tsdn_t *tsdn, buf_writer_t *buf_writer) {
 		buf_writer_free_internal_buf(tsdn, buf_writer->buf);
 	}
 }
+
+void
+buf_writer_set_allow_auto_flush(buf_writer_t *buf_writer,
+    bool allow_auto_flush) {
+	buf_writer->allow_auto_flush = allow_auto_flush;
```
interwq: When switching auto_flush to false, also invoke _commit?

src/prof_recent.c
yinan1048576: Makes sense. Abstracted out the allocation logic. Also abstracted out the deallocation logic to be symmetric.

src/prof_recent.c
yinan1048576: Yeah, and also, if we put the `commit()` at the end, then the last `commit()` would not be useful. So `commit()` might not be a very good name: it may sound like we're committing what we have just done, but actually it's more for having a place to go back to before we do the next thing. Thoughts on a better name?

src/buf_writer.c
yinan1048576: Makes sense. That actually fixes a bug in my previous version: if a single item overflowed, then the item wouldn't be dumped at all even after the `buf_writer_set_allow_auto_flush(&buf_writer, true)` call, because the `overflowed` flag was still `true`.

yinan1048576:
Revised according to feedback. Also added more assertions in the buffered writer. I'll write some unit tests next, which might reveal some other bugs...

src/buf_writer.c
```diff
@@ -42,6 +43,10 @@ buf_writer_assert(buf_writer_t *buf_writer) {
 		assert(buf_writer->private_cbopaque == NULL);
 		assert(buf_writer->buf_size == 0);
 	}
+	if (buf_writer->allow_auto_flush) {
+		assert(buf_writer->committed == 0);
+		assert(!buf_writer->overflowed);
+	}
```
interwq: maybe add:

    else {
        assert(committed <= end);
    }

src/buf_writer.c
```diff
@@ -112,10 +118,19 @@ buf_writer_cb(void *buf_writer_arg, const char *s) {
...
 	assert(buf_writer->buf_end <= buf_writer->buf_size);
+	if (buf_writer->overflowed) {
+		return;
+	}
 	size_t i, slen, n, s_remain, buf_remain;
 	for (i = 0, slen = strlen(s); i < slen; i += n) {
 		if (buf_writer->buf_end == buf_writer->buf_size) {
-			buf_writer_flush(buf_writer);
+			if (buf_writer->allow_auto_flush) {
+				buf_writer_flush(buf_writer);
```
interwq: Question: when reached here, we are sure buf_size is not 0, is that right?

src/buf_writer.c
```diff
@@ -112,10 +118,19 @@ buf_writer_cb(void *buf_writer_arg, const char *s) {
...
+			} else {
+				buf_writer->buf_end = buf_writer->committed;
+				buf_writer->overflowed = true;
+				return;
```
interwq: nit: move the overflowed branch to the s_remain < buf_remain below, to avoid the memcpy of the last iteration.

src/buf_writer.c
```diff
@@ -134,3 +149,35 @@ buf_writer_terminate(tsdn_t *tsdn, buf_writer_t *buf_writer) {
 		buf_writer_free_internal_buf(tsdn, buf_writer->buf);
 	}
 }
+
+void
+buf_writer_set_allow_auto_flush(buf_writer_t *buf_writer,
+    bool allow_auto_flush) {
+	buf_writer_assert(buf_writer);
+	if (buf_writer->allow_auto_flush && !allow_auto_flush) {
```
interwq: I slightly prefer only guarding with `allow_auto_flush`. The only case there's a difference is when manual is set multiple times, to ensure calling _commit() once for each _set() call, so the implicit revert don't go "out of scope".

src/prof_recent.c
interwq: Agreed. Maybe `_set_checkpoint`? That way calling at the beginning is intuitive.

src/prof_recent.c
```diff
@@ -495,7 +546,47 @@ prof_recent_alloc_dump_node(emitter_t *emitter, prof_recent_t *n) {
 	emitter_json_object_end(emitter);
 }
 
-#define PROF_RECENT_PRINT_BUFSIZE 4096
+static void
+prof_recent_alloc_merge_locked(tsd_t *tsd, prof_recent_t *last) {
```
interwq: nit: better to make the swap and merge utility not assuming the lists internally, but rather pass them in. That way it would be possible to test them individually.

src/buf_writer.c
```diff
@@ -56,24 +61,25 @@ buf_writer_init(tsdn_t *tsdn, buf_writer_t *buf_writer, write_cb_t *write_cb,
...
 		buf_writer->public_write_cb = write_cb;
 		buf_writer->public_cbopaque = cbopaque;
 		buf_writer->private_write_cb = NULL;
 		buf_writer->private_cbopaque = NULL;
 		buf_writer->buf_size = 0;
-		buf_writer_assert(buf_writer);
-		return true;
 	}
+	buf_writer->buf_end = 0;
+	buf_writer->allow_auto_flush = true;
```
interwq: nit: let's do `_set_allow_auto_flush(..., false)` to make the default clear. We'd also need to move the `buf_writer_assert` in set_allow_auto_flush to its end.

src/buf_writer.c
```diff
@@ -112,10 +118,19 @@ buf_writer_cb(void *buf_writer_arg, const char *s) {
...
+	}
 	size_t i, slen, n, s_remain, buf_remain;
 	for (i = 0, slen = strlen(s); i < slen; i += n) {
 		if (buf_writer->buf_end == buf_writer->buf_size) {
-			buf_writer_flush(buf_writer);
+			if (buf_writer->allow_auto_flush) {
+				buf_writer_flush(buf_writer);
+			} else {
+				buf_writer->buf_end = buf_writer->committed;
+				buf_writer->overflowed = true;
```
interwq: nit: abstract these two operations out

src/prof_recent.c
```diff
@@ -495,7 +546,47 @@ prof_recent_alloc_dump_node(emitter_t *emitter, prof_recent_t *n) {
...
+	ssize_t new_max = dump_max;
+	if (dump_count > 0) {
+		prof_recent_alloc_merge_locked(tsd, dump_last);
+		prof_recent_alloc_count += dump_count;
+		assert(dump_max == prof_recent_alloc_max_get(tsd));
+		if (dump_max != -1 && prof_recent_alloc_count > dump_max) {
+			new_max = prof_recent_alloc_count;
+			assert(new_max > dump_max);
+			ssize_t old_max = prof_recent_alloc_max_update(tsd,
+			    new_max);
```
interwq: Let’s put the max adjustment logic together. Only merge the lists in _finish() is fine. Having the max_update and max_write separated feels wrong.

src/prof_recent.c
```diff
@@ -506,33 +597,62 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
+	    dump_count, dump_max);
 	malloc_mutex_unlock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
 
+	buf_writer_set_allow_auto_flush(&buf_writer, true);
+	emitter_json_array_end(&emitter);
 	emitter_end(&emitter);
 	buf_writer_terminate(tsd_tsdn(tsd), &buf_writer);
+
+	if (new_max > dump_max) {
+		ssize_t old_max = prof_recent_alloc_max_write(tsd, dump_max);
```
interwq: I realize the _finish() and this really needs to be done together holding the recent lock. If we drop the lock in the middle, we have discrepancy between the list length and the max.

Consider passing in a bool to _write(), indicating if the recent lock is taken. The write function then always release the lock.

include/jemalloc/internal/buf_writer.h
yinan1048576: Changed it to `checkpoint` to align with `buf_writer_set_checkpoint()`.

src/buf_writer.c
yinan1048576: Yes. That's guarded by `assert(buf_writer->buf != NULL);` (`buf_size == 0` if and only if `buf == NULL`.)

src/buf_writer.c
yinan1048576: We can only copy it but not move it. We still need to test at the beginning. Flushing is by design lazy in buffered writer, so if in a previous `write_cb()` call the buffer was exactly full, then the initial test is necessary. I think having a single test at the beginning outweighs the saving of one `memcpy`. Thoughts?

src/buf_writer.c
yinan1048576: Agree that we should only guard with `allow_auto_flush`.

Upon a second thought, I think `buf_writer_set_allow_auto_flush(false)` shouldn't auto commit. Especially since we're changing the name `commit` to `checkpoint`, the caller should explicitly call `buf_writer_set_checkpoint()` every time they anticipate a potential rollback; otherwise the API shouldn't make any guarantee. This is mainly to force the caller side code to be clearer: it has to strictly follow a "three-step cycle" - `buf_writer_set_checkpoint()` -> `buf_writer_overflowed()` -> `handle_overflow()`.

On the buffered writer side, I'm always setting `checkpoint = 0` in `buf_writer_set_allow_auto_flush()`, just as a minimal safe default.

The other thing I was thinking: in addition to not doing auto commit, consecutive `buf_writer_set_allow_auto_flush(false)` calls should not auto clear the `overflowed` flag either. In other words, the `overflowed` flag can only be cleared in `buf_writer_set_allow_auto_flush(true)` or `buf_writer_set_checkpoint()`; the caller can thus freely check whether there has been an overflow at any time, even after flushing or after another `buf_writer_set_allow_auto_flush(false)` call, as long as it's before they set the next checkpoint. This is mainly to give the caller more flexibility in code structure, as long as they strictly follow the three-step cycle.

Overall, the `buf_writer_set_allow_auto_flush()` function should be more transparent: it should just set `allow_auto_flush`, as its name suggests, with only minimal / necessary additional operations.

include/jemalloc/internal/buf_writer.h
```diff
@@ -21,6 +21,13 @@ typedef struct {
 	size_t buf_size;
 	size_t buf_end;
 	bool internal_buf;
+	/*
+	 * The following fields are for advanced APIs - see explanations
```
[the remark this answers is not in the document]
yinan1048576: I think adding `when allow_auto_flush is false` is somewhat misleading. It might be slightly better to put this `advanced APIs (i.e. when allow_auto_flush is false)` below the `bool allow_auto_flush;` line, but `allow_auto_flush` is also only meant for advanced APIs. Anyway, I think `see explanations below` should suffice.

src/buf_writer.c
yinan1048576: Defined a `buf_writer_set_allow_auto_flush_internal()` to avoid duplicating the logic without sacrificing the initial `buf_writer_assert()` in `buf_writer_set_allow_auto_flush()`, which I think is important for catching any bugs as early as possible and serving as a pre-condition verification.

Also I think having another `buf_writer_assert()` at the end of `buf_writer_set_allow_auto_flush()` makes sense. And I also added it to other similar places.

src/prof_recent.c
yinan1048576: See my main comment.

src/prof_recent.c
yinan1048576: See my main comment.

src/buf_writer.c
interwq: Agreed. Not auto setting sounds right.

yinan1048576:

Also changed the design around `prof_recent_alloc_max_write()`: it is now broken into `prof_recent_alloc_max_write_locked()` and `prof_recent_alloc_async_cleanup()`. The consequences are:
- `prof_recent_alloc_max_write_locked()` should be called under `prof_recent_alloc_mtx` while `prof_recent_alloc_async_cleanup()` is not. The caller needs to handle the locking logic around `prof_recent_alloc_mtx`.
- Revised `prof_recent_alloc_max_ctl_write()` and `prof_recent_alloc_dump()` accordingly. The benefit for `prof_recent_alloc_dump()` is that `prof_recent_alloc_max_write_locked()` can be inserted into `prof_recent_alloc_dump_finish_locked()`, which is called right after the main loop and before `prof_recent_alloc_mtx` is released, so that there's no discrepancies outside the `prof_recent_alloc_mtx` lock.
- Revised `prof_recent_alloc_max_ctl_read()` so that it only needs the `prof_recent_alloc_mtx` lock but not the `prof_dump_mtx` lock. It's not a huge deal but mainly makes the logic more rigorous.
- Decided to put `prof_recent_alloc_async_cleanup()` always under `prof_dump_mtx`, even though in theory it's not needed. The benefit is that the memory overhead as a consequence of the asynchronous cleanup is always bounded. Previously it was in theory unbounded.
- One compromise of the break-up is that it's no longer easy to do assertions on the number of nodes asynchronously deleted. Passing the count around can make the code slightly messier, so I decided it'd not be worthwhile.

Also added a lot more assertions all around.

src/buf_writer.c
interwq: Yeah keep the branch at beginning, but add another one right before memcpy, so that we can set overflow and return immediately.

interwq:
> Revised prof_recent_alloc_max_ctl_read() so that it only needs the prof_recent_alloc_mtx lock but not the prof_dump_mtx lock. It's not a huge deal but mainly makes the logic more rigorous.

I'd say always guard the read and write under both locks, in case we do anything trickier during dumping.

src/buf_writer.c
```diff
@@ -42,6 +43,23 @@ buf_writer_assert(buf_writer_t *buf_writer) {
...
+	}
+}
+
+static void
+buf_writer_set_allow_auto_flush_internal(buf_writer_t *buf_writer,
+    bool allow_auto_flush) {
+	buf_writer->allow_auto_flush = allow_auto_flush;
+	buf_writer->checkpoint = 0;
+	if (allow_auto_flush) {
+		buf_writer->overflowed = false;
```
interwq: Should be better to also clear overflowed; having (overflowed == true && checkpoint == 0) can cause confusion as well.

src/buf_writer.c
```diff
@@ -104,24 +120,43 @@ buf_writer_flush(buf_writer_t *buf_writer) {
...
 	    buf_writer->buf);
 	buf_writer->buf_end = 0;
+	buf_writer_assert(buf_writer);
+}
+
+static void
+buf_writer_overflow(buf_writer_t *buf_writer) {
+	buf_writer_assert(buf_writer);
+	assert(!buf_writer->allow_auto_flush);
+	assert(buf_writer->buf_end == buf_writer->buf_size);
```
interwq: This is only true because we have the unnecessary last iteration of memcpy, right? Let's remove if so.

src/prof_recent.c
```diff
@@ -414,38 +430,111 @@ prof_recent_alloc_max_ctl_write(tsd_t *tsd, ssize_t max) {
...
-	 * there's no simultaneous holdings of prof_recent_alloc_mtx and
-	 * tdata->lock.  In the worst case there can be slightly extra space
-	 * overhead taken by these nodes, but the total number of nodes at any
-	 * time is bounded by (max + sum(decreases)), where "max" means the
-	 * most recent prof_recent_alloc_max and "sum(decreases)" means the
-	 * sum of the deltas of all decreases in prof_recent_alloc_max in the
-	 * past.  This (max + sum(decreases)) value is completely transparent
-	 * to and controlled by application.
+	 * The tctx of the to-be-deleted nodes as well as the nodes themselves
+	 * will be asynchronously handled by prof_recent_alloc_async_cleanup().
```
interwq: Let's return a chain of the nodes to be deleted, instead of having the callsite figuring it out.

src/prof_recent.c
```diff
@@ -506,33 +595,73 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
-		prof_recent_alloc_dump_node(&emitter, n);
+	buf_writer_set_allow_auto_flush(&buf_writer, false);
+	malloc_mutex_lock(tsd_tsdn(tsd), &prof_recent_alloc_mtx);
+	prof_recent_t *dump_dummy = prof_recent_alloc_dummy_reserve;
+	for (prof_recent_t *node = dump_dummy->next; node != dump_dummy; ) {
+		buf_writer_set_checkpoint(&buf_writer);
+		prof_recent_alloc_dump_node(&emitter, node);
+		if (buf_writer_overflowed(&buf_writer)) {
+			if (buf_writer_empty(&buf_writer)) {
+				buf_writer_set_allow_auto_flush(
```
interwq: nit: add some comment for this case.

src/prof_recent.c
```diff
@@ -506,33 +595,73 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
+
+	/* Then release the nodes as well as their associated tctx. */
+	if (dump_count > 0) {
+		/*
+		 * dump_dummy is completely detached from the main
+		 * prof_recent_alloc_dummy list, so it can be safely accessed
+		 * outside the prof_recent_alloc_mtx lock.
+		 */
+		prof_recent_alloc_async_cleanup(tsd, dump_dummy->next,
+		    new_begin);
```
interwq: But the new_begin might not be safe to access? Either way, returning the starting node and maybe the count from the _write_max() should be cleaner.

yinan1048576:
First, fixed a bug: I shouldn't have added `assert(buf_writer->checkpoint <= buf_writer->buf_end)` in `buf_writer_assert()`, because we're not auto committing.

Then, I added a separate commit for the following comment:

> Yeah keep the branch at beginning, but add another one right before memcpy, so that we can set overflow and return immediately.

To be honest, I don't think the compromise on the code logic is worthwhile. We're only saving one `memcpy()`, and only for each time we come across the buffer limit. Also, in all cases our buffered writer is called under the `emitter` utility, which outputs very little content at a time.

src/prof_recent.c
yinan1048576: Right, it's not safe to access, and I'm not accessing it. It's used in the "iterator" fashion, or the "one-past-the-end" fashion. (All the iterations in this module follow this same manner.)

interwq:
Re: memcpy -- I'm convinced. Let's not include the extra branches then.

src/prof_recent.c
interwq: Well, having a dangling pointer even if only to check boundary is still suboptimal. Is it true that we should be able to avoid this, if we have the write_max() return the to-be-deleted chain?

src/buf_writer.c
yinan1048576: I thought about that but decided not to. `buf_writer_set_allow_auto_flush()` should do minimal work in addition to literally setting `allow_auto_flush`, so that I can force the caller side code to have a higher code quality, i.e. to exactly stick to the cycle of `buf_writer_set_checkpoint()` -> write something -> `buf_writer_overflowed()` -> handle overflow.

In particular, the only chance this `overflowed == true && checkpoint == 0` condition can cause an issue is:
- the user first calls `buf_writer_set_allow_auto_flush(false)` when `allow_auto_flush` is already `false`, and
- the user tries to write something without first setting a checkpoint.

Both these two parts should be prohibited.

yinan1048576:
> I'd say always guard the read and write under both locks, in case we do anything trickier during dumping.

I still think only guarding the read by `prof_recent_alloc_mtx` makes more sense. The read and write may sound like a pair, but I think the more correct way to see it is: the read belongs more to the original group of functions e.g. `prof_recent_alloc()`, `prof_recent_alloc_reset()`, etc., while the write and the dump should be a pair. The first group operates only under `prof_recent_alloc_mtx`, while the second group operates under `prof_dump_mtx` and intermittently under `prof_recent_alloc_mtx`, and in theory any member of the second group can freely call any member of the first group. This grouping is based more on the lock privilege/order rather than the literal names. After all, the usage of `prof_dump_mtx` is only introduced in this PR and is only for the purpose of:
- disallowing concurrent dumping, and
- disallowing the write while dumping.

Keeping the read under `prof_recent_alloc_mtx` only can help clarify the two-level hierarchy. I can have some more comments explaining this. Thoughts?

interwq:
On one hand not blocking a read mallctl on dumping helps a bit. On the other hand, reading the max during dumping is also misleading, since we are not respecting the max at that time. I don’t have a strong preference here. Feel free to go either way.

src/buf_writer.c
yinan1048576: Upon a second thought, I think when `buf_writer_set_allow_auto_flush(false)` is called, we should not clear either `overflowed` or `checkpoint`. It should just do what the name literally suggest: set `allow_auto_flush`, with minimal additional work to make things work. In other words, calling it again on a buffered writer already with `allow_auto_flush = false` should intuitively behave like a no-op, and should not change any internal state of the buffered writer.

If the caller is not careful, which means that the caller made both mistakes listed in my previous comment, then there's actually nothing dangerous:
- If `overflowed` has already been set, then the caller cannot write out anything, which is safe.
- If `overflowed` has not been set, then the buffer has not overflowed yet, so further writing to it, even if it triggers overflow, is also safe.

And on a higher level, the safety is guaranteed by that the second `buf_writer_set_allow_auto_flush(false)` behaves as a no-op so any subsequent writing simply behaves like continuation of the previous writing.

yinan1048576:
Revised according to feedback. Did two larger scale revisions:
- `prof_recent_alloc_max_write_locked()` returns the list of to-be-deleted nodes. Only the list head is returned (via a parameter to the function), and the `next` pointer of the tail is cleared.
- I realized that the confusions around the locks is a consequence of an "abuse" of the `prof_dump_mtx` lock. I didn't pick `prof_dump_mtx` for a particularly good reason after all. I had the intention of disallowing the last-N dumping and regular prof dumping to take place at the same time, but it's not necessary. My other intention was to someday allow the two dumping mechanisms to share the same buffer, but it's not yet there. So I changed the design to the following:
  - `prof_recent_alloc_max` is guarded and only guarded under `prof_recent_alloc_mtx`, and so are `prof_recent_alloc_count` and `prof_recent_alloc_dummy`, just as how we originally designed. Reading from and writing to `prof_recent_alloc_max` is always safe under the lock. Reading from it without the lock is also allowed (since the variable is atomic), but it only gives a hint and we don't guarantee anything further.
  - Dumping and `prof_recent_alloc_dummy_reserve` is additionally guarded under a new lock `prof_recent_dump_mtx`. The lock only allows one thread to dump at a time. In theory, multiple threads would be allowed to dump at the same time, if we had additional reserves for the dummy head.
  - As a consequence, the `mallctl()` write can take place while `prof_recent_dump_mtx` is in hold (but during the time `prof_recent_alloc_mtx` is intermittently released). It won't cause any issue, either to the dump or to the write.
  - Since the write only needs `prof_recent_dump_mtx`, it's more natural to see that the read also only needs it.
  - `prof_recent_alloc_async_cleanup()` is made to be completely async rather than under any lock. The memory overhead can be unbounded, but the memory overhead per dump / `mallctl()` call is bounded, and the caller can easily control the number of concurrent dump / `mallctl()` calls.

yinan1048576:
Bug fix.

yinan1048576:
Added an assertion to confirm that the (static) buffer size is enough.

Previously we had the idea of dynamically sizing the buffer size, if we observe that some record is too large for the buffer. Upon a second thought, I think it's not super important: we have `PROF_BT_MAX` set to be `128`, and each stack frame is represented as an integer, so the total bytes needed for one record has a strict upper limit, and according to my estimation it's well within 10K bytes, so having 64K bytes as the buffer size should be much more than enough.

I think it still makes sense to try best to avoid the case where a single record overflows the entire buffer, in which case the dumping logic would completely block other threads. To this end I'm simply adding an assertion to confirm that (in debug build). (Exactly computing the upper limit of the dump size per record is not trivial; otherwise I can directly assert on / compute the buffer size.) I think in reality, if the assertion fails, then it means something totally surprising is happening, so we have a good reason to abort and figure out why.

Meanwhile, it doesn't feel trivial to come up with an obviously correct design for the dynamic sizing logic. For example, sizing up the buffer costs more memory, and we'd better have some limit on the maximum size we can get to, but there doesn't seem to me to be an easy way to figure that out.

yinan1048576:
Rebase, and added unit tests for the new buffered writer functionalities, and extracted the changes on buffered writer as separate commits.

include/jemalloc/internal/buf_writer.h
```diff
@@ -30,4 +34,17 @@ void *buf_writer_get_cbopaque(buf_writer_t *buf_writer);
...
 void buf_writer_terminate(tsdn_t *tsdn, buf_writer_t *buf_writer);
 
+/*
+ * Everything below are advanced APIs, only for scenarios where there's need to
+ * manually control the flushing operations, e.g. if we want to only flush
+ * while not holding some critical lock.  Note that buf_writer_init() sets
+ * allow_auto_flush to true by default.  See prof_recent_alloc_dump() in
+ * src/prof_recent.c for example usage.
+ */
+void buf_writer_set_allow_auto_flush(buf_writer_t *buf_writer,
```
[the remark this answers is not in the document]
davidtgoldblatt: And similarly within the implementation stuff - some documentation about what the reclamation / locking strategy is, which functions do which parts of it, etc. would be very helpful.

yinan1048576:
Rebase, and added more documentations. Also did the following:
- Simplified the buffered writer API by removing the `buf_writer_empty()` function; this function not necessary on the caller side, and the caller logic is made to be more intuitive by not trying to call this function.
- Extracted some refactoring part out as separate commit to make reviewing slightly easier.

src/buf_writer.c
davidtgoldblatt: This seems like an argument for asserting that overflowed is false when we turn on auto dumping. It's never correct not to handle overflow if you've turned off auto-dump then caused it, right?

src/prof_recent.c
```diff
@@ -15,11 +15,19 @@
...
 static atomic_zd_t prof_recent_alloc_max;
 static ssize_t prof_recent_alloc_count = 0;
 static prof_recent_t *prof_recent_alloc_dummy = NULL;
 
+malloc_mutex_t prof_recent_dump_mtx; /* Protects dumping & the fields below. */
+/*
+ * prof_recent_alloc_dummy_reserve points to an empty list to be used during
+ * dumping, so that there's no need to do allocate / free there.
+ */
+static prof_recent_t *prof_recent_alloc_dummy_reserve = NULL;
```
davidtgoldblatt: I think this, and prof_recent_alloc_dummy above, could still use some more comments. This explains where it's used, but not what it is.

E.g.
- Who puts things into the list, and who takes them out?
- Why do those two things need to communicate through a global, instead of just one being the caller and one the callee?
- Why do we need both prof_recent_alloc_dummy and prof_recent_alloc_dummy_reserve? What's the difference between them?

yinan1048576: Added more comments. Actually, the more I think of it, the more I feel that a global `prof_recent_alloc_dummy_reserve` is not necessary. The comment I'm writing sounds defensive, where I try to enumerate all downsides of not defining it global, but I'm personally neutral now: defining it global has the design downside, and I'm not sure how that compares with the downsides I listed.

yinan1048576:

src/buf_writer.c
yinan1048576: You mean the case where:
- `allow_auto_flush` is turned off, and then
- the writing causes overflow, and then
- `allow_auto_flush` is turned back on without first trying to handle the overflow?

In the current design this is allowed. The handling for overflow is manual and is up to the caller. The caller may want to flush when an overflow takes place, but the caller do not have to. In particular, in the case of the last-N dumping, when overflow happens, the caller may either flush, or switch `allow_auto_flush` on without flushing (though we should avoid the scenario leading to the latter decision - see my comments).

src/prof_recent.c
yinan1048576: Leaning slightly towards keeping `prof_recent_alloc_dummy_reserve` as a global variable, mainly for implementation simplicity. Thoughts?

davidtgoldblatt: Ahh, I think I understand why the dummy node (/ reserve) here. This is really just part of a linked-list maintenance strategy IIUC.

I think probably, instead of hand-rolling the list-manipulation code, it would probably be better to use the ql module (our linked list code), which doesn't have a need for dummy nodes. (Possibly extended with a "pop all" functionality or something"). This would

I think this is probably more along the lines of what Qi meant down below -- rather than assigning through a `prof_recent_t **`, we can work in terms of a `prof_recent_list_t *to_delete` and just append onto it.

src/prof_recent.c
```diff
@@ -517,33 +637,98 @@ prof_recent_alloc_dump(tsd_t *tsd, void (*write_cb)(void *, const char *),
...
-	    n = prof_recent_alloc_next(tsd, n)) {
-		prof_recent_alloc_dump_node(&emitter, n);
+	/*
+	 * We dump the list in segments, where each segment:
+	 * - is dumped under one critical section, and
+	 * - does not trigger buffered writer flushing,
+	 * so that flushing always takes place outside of the critical sections
+	 * and thus would not block allocations / deallocations from other
+	 * threads.
+	 */
```
davidtgoldblatt: This might be a dumb question, but why do we need to hold any locks while dumping at all? The list is private at this point, since we popped it out of the global above.

yinan1048576: There are two locks we're holding:

`prof_recent_alloc_mtx`

The list is private only in the sense that new samples are not added to this list, but the list still holds existing samples, and if some of these samples are not freed yet, and if they're being freed during dumping, the freeing thread will need to modify the list (recording that the allocation is being freed as well as the freeing time and stack trace).

The `prof_recent_alloc_mtx` protects the list to which allocating threads add samples. When there's no dumping taking place, the freeing thread needs to acquire the `prof_recent_alloc_mtx` lock, because existing records might be being deleted, if another thread is adding new records to the list and the list has a finite limit.

In theory, while dumping, the freeing thread doesn't have to try to acquire the `prof_recent_alloc_mtx` lock, because it doesn't need to compete with allocating threads. However, there's still need for some lock that the dumping thread and the freeing thread cannot hold simultaneously; otherwise the dumping thread may dump a partial freeing stack trace. I just continue to use the `prof_recent_alloc_mtx` lock for this "some lock" for simplicity, since otherwise the freeing thread needs additional mechanism to determine whether there's another thread dumping or not.

`prof_recent_dump_mtx`

As I mentioned earlier, this lock strictly speaking is not necessary. It's only meant to permit only one thread to dump at a time. We can allow multiple threads to dump at the same time, but there can be a couple of issues:
- Even multiple threads can dump at the same time, they cannot simultaneously "check in" to the global list, because they need to compete for the `prof_recent_alloc_mtx` lock (and once they have it they'll pull the global list out). Therefore, each thread effectively only dumps the records that come in *after* the last "check in" time of another dumping thread. This is not perfectly intuitive from the application's perspective.
- If applications are not super careful, they likely would supply the same callback handle to all the dumping threads. If so, multiple concurrent dumping threads might invoke the callback handle simultaneously. Even if the logic underlying the callback handle is concurrency safe, the output would be interleaved with multiple pieces of dump content.
- (This point is implementation specific and may or may not be relevant if we decide to change the data structure.) The current implementation requires a dynamically allocated spare dummy head for each dumping thread. If we allow multiple threads to dump concurrently, then each thread needs to allocate itself, and if the allocation fails, then it's not able to dump. If we only allow one thread to dump at a time, then we can pre-allocate a spare dummy head at boot time, to make sure that all dumps will be successful.

yinan1048576: > However, there's still need for some lock that the dumping thread and the freeing thread cannot hold simultaneously; otherwise the dumping thread may dump a partial freeing stack trace.

Well, there can never be a partial freeing stack trace, because the stack trace is first completed and the record is then allowed to point to the stack trace. But, still, there can be possibilities of only dumping that the record is freed but not the freeing time / stack trace, or dumping the freeing time but not the stack trace, etc.

davidtgoldblatt: I think probably better would be to have an `atomic_b_t has_been_freed`. When the freeing thread comes in, it can fill in the stack trace at free and atomic_store(&ptr->has_been_freed, true, ATOMIC_RELEASE);`. The dumping thread, when dumping, can write `bool freed = atomic_load(&ptr->has_been_freed, ATOMIC_ACQUIRE); if (freed) { dump_free_stack_trace(ptr); }`.

Two reasons:
- This lock is the reason for the whole backup/rewind thing in the buffered writer, right? I'd love to be able to get rid of it if we can; it vastly increases the complexity of the buffered writer for a very edge-casey situation.
- I think that the release -> dump -> reacquire strategy here still has some tail latency problems. The underlying mutex types don't guarantee fairness, and in fact on most platforms, they're biased towards the last acquirer. So in practice, a thread constantly acquiring, dropping, and reacquiring a mutex can end up looking like it was holding the mutex for the entire time anyways.

yinan1048576: This fencing strategy is a great one.

> I'd love to be able to get rid of it if we can; it vastly increases the complexity of the buffered writer for a very edge-casey situation.

Agree. I'd also love to keep the buffered writer simple.

Here's what I get out of your suggestion:
- Allocating thread:
  - Acquire `prof_recent_alloc_mtx`.
  - Record (with possible deletion of the earliest record).
  - Release `prof_recent_alloc_mtx`.
- Freeing thread:
  - Acquire `prof_recent_alloc_mtx`.
  - Record (with the atomic boolean the last one to write, and with an `ATOMIC_RELEASE` fence).
  - Release `prof_recent_alloc_mtx`.
- Dumping thread:
  - Acquire `prof_recent_dump_mtx`.
  - Acquire `prof_recent_alloc_mtx`.
  - Pull the entire list out.
  - Release `prof_recent_alloc_mtx`.
  - Dump (first checking the atomic boolean with an `ATOMIC_ACQUIRE` fence before dumping any deallocation information).
  - Acquire `prof_recent_alloc_mtx`.
  - Insert the list back to the start.
  - Release `prof_recent_alloc_mtx`.
  - Release `prof_recent_dump_mtx`.

Looks good?

(BTW I already have an atomic "boolean" for this purpose: the pointer on the record pointing to the allocation.)

davidtgoldblatt: Yes, I think that looks great -- in particular, we moved the dumping thread from touching the hot prof_recent_alloc_mtx mutex from O(N) times to O(1).

davidtgoldblatt: Well, and don't hold it during any potentially long-running operations, like I/O

yinan1048576:

Get rid of the buffered writer changes. https://github.com/jemalloc/jemalloc/pull/1722/commits/98c50d89e986177dbb01758fdc8d8c86a9a5c942 is valuable in itself, and it is included as part of #1795. https://github.com/jemalloc/jemalloc/pull/1722/commits/9a819e7a25c3d84da6cf25d277ceddc415219658 is completely discarded.

Also figured a subtle bug unrelated to the dumping, and added the fix as a separate commit in the middle: when an allocating thread is trying to record a prof sample, if the allocation for the new record node fails, the previous logic would have the `edata` point to the dummy head, which is dangerous!

I haven't resolved the feedback regarding the data structure. It's logically separate from the core dumping design and is a lower layer decision. I'll think more and address it next week.

yinan1048576:
Revised. A few remarks:

After some thoughts, I think I'll leave the potential data structure change to a later PR. It's related to the whole last-N profiling mode, not just to the dumping part.

On the other hand, we only have a `qr` module, which is a circular doubly linked list, and a `ql` module, which is a non-circular doubly linked list (implemented as a thin wrapper on top of `qr`), but here I only need a circular singly linked list. It doesn't seem terribly clear to me whether we want to pay for the extra functionalities we don't need. Performance wise it's most likely fine. Memory consumption wise it would be one extra pointer per record, so the overhead would be proportional to the "N" in the "last-N".

BTW I also wonder if we ever needed a *doubly* linked list in our code base. If the answer is no, we might consider having a simplified list structure first.

Regarding the `to_delete` list: though I didn't change the data structure, I figured a good way to simplify the logic around it, and I updated the last commit accordingly. Basically, with the help of some restructuring tricks, `prof_recent_alloc_dump_finish_locked()` can be greatly simplified, in the part handling the case where the two lists merged together could become too long. Also, `to_delete` is no longer passed around via an extra parameter.

Also figured a minor correction, and put it as a separate commit: https://github.com/jemalloc/jemalloc/pull/1722/commits/7369a88e00d482097d189615cdbe99ed8ff5dcf9. This is a bit concerning though: I haven't dug deep, but it seems that functions called in unit tests are allowed to differ from their declarations by a `tsd` parameter.

yinan1048576:
Rebase onto #1808, which migrated the last-N list to use the `ql` module and also took care of https://github.com/jemalloc/jemalloc/pull/1722/commits/6452eeb1cba787f025840ca88cf842d9b19bc03f. The resulting code change becomes much simpler.

yinan1048576:
Rebase and minor correction - handling of the new mutex on forking.

yinan1048576:
The FreeBSD failure is well deserved: `arena_prefork7()` handles the bins, which are at the leaf level and should take place after the new `prof_recent_dump_mtx`.

Rather than splitting `prof_prefork1()` into two and putting the first one before `arena_prefork7()`, I decided to promote `prof_recent_dump_mtx` to right above all arena locks, so that it can go into `prof_prefork0()`. Such a high privilege is not necessary for `prof_recent_dump_mtx`, since it only needs to be higher than `prof_recent_alloc_mtx` which is at the leaf; it is just for making the pre-forking logic easier. And, given `prof_recent_dump_mtx`'s higher privilege now, I'm additionally putting the buffered writer init and terminate under it, with the benefit of not ever allowing two threads each holding a giant buffer simultaneously.

yinan1048576:
Rebase.

PR #1439, opened 2019-02-17
Improve slighty prof feature on FreeBSD

devnexen:
dumping process memory mappings.

PR #1726, opened 2020-01-10
Improving prof support on FreeBSD

devnexen:
Dumping process mapping in a native manner.
superseeds #1439

### DIFF
--- a/src/prof_data.c
+++ b/src/prof_data.c
@@ -8,6 +8,11 @@
 #include "jemalloc/internal/malloc_io.h"
 #include "jemalloc/internal/prof_data.h"
 
+#if defined(__FreeBSD__)
+#include <sys/sysctl.h>
+#include <sys/user.h>
+#endif
+
 /*
  * This file defines and manages the core profiling data structures.
  *
@@ -911,7 +916,7 @@ label_return:
 	return ret;
 }
 
-#ifndef _WIN32
+#if !defined(_WIN32) && !defined(__FreeBSD__)
 JEMALLOC_FORMAT_PRINTF(1, 2)
 static int
 prof_open_maps(const char *format, ...) {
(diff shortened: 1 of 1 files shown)


src/prof_data.c
```diff
@@ -939,13 +944,82 @@ prof_open_maps(const char *format, ...) {
...
+		goto label_return;
+	}
+
+	if (sysctl(mib, 4, NULL, &sz, NULL, 0) == -1) {
+		ret = true;
+		goto label_return;
+	}
+
+	/* Size for kinfo_vmentry */
+	sz = sz * 4 / 3;
```
davidtgoldblatt: What's the reasoning behind this computation? Can we have a comment explaining?

src/prof_data.c
```diff
@@ -939,13 +944,82 @@ prof_open_maps(const char *format, ...) {
...
+	buf = mmap(NULL, sz, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANON, -1, 0);
+
+	if (buf == MAP_FAILED) {
+		ret = true;
+		goto label_return;
+	}
+
+	if (sysctl(mib, 4, buf, &sz, NULL, 0) == -1) {
+		ret = true;
+		munmap(buf, sz);
```
davidtgoldblatt: Can we unify the munmaps with the mfd cleanup, down at label_return?

src/prof_data.c
```diff
@@ -939,13 +944,82 @@ prof_open_maps(const char *format, ...) {
...
+		goto label_return;
+	}
+
+	char *low = buf;
+	char *high = buf + sz;
+
+	while (low < high) {
+		/* paths + enough room for address ranges and permissions */
+		char map[PATH_MAX+68];
```
davidtgoldblatt: Where does 68 come from? Can we add a comment explaining?

src/prof_data.c
```diff
@@ -939,13 +944,82 @@ prof_open_maps(const char *format, ...) {
...
+			ret = true;
+			munmap(buf, sz);
+			goto label_return;
+		}
+
+		if (e->kve_path[0] == '\0') {
+			goto loop_cond;
+		}
+
+		malloc_snprintf(map, sizeof(map), "%.12lu-%.12lu %c%c%c %s\n",
```
davidtgoldblatt: Is 12 right? This must be at least nominally enough for the whole address space, right?

src/prof_data.c
devnexen: Well in fact this is to give enough room for kinfo_vmentry. FreeBSD libpvocstat thus underlyingly nvm does that too.

davidtgoldblatt: But like, what property of kinfo_vmentry requires this tweak to the reported size? Is it just to handle races in which a mapping is added in between the two sysctl calls?

src/prof_data.c
devnexen: Yes it takes not more than 12 large.

src/prof_data.c
devnexen: All FreeBSD code consumer which list proc maps use this workflow to give enough room for enough process entries in between calls indeed but barely more documented than that.

interwq:
The test failures on Windows look real, e.g. https://ci.appveyor.com/project/jasone/jemalloc/builds/30047453/job/nikr65egvqd1fu3h

src/prof_data.c
davidtgoldblatt: How come? E.g. if we have a 48-bit address space, it seems we could have as many as ceil(log(2^48)) == 15 digits.

davidtgoldblatt:
I think some variant of this still probably makes sense, but the CI breakages mean probably not this version.

PR #1728, opened 2020-01-13
Fallback to unbuffered printing if OOM

yinan1048576:
Also added stronger assertions in the buffered writer initializer.

### DIFF
--- a/include/jemalloc/internal/buf_writer.h
+++ b/include/jemalloc/internal/buf_writer.h
@@ -23,7 +23,9 @@ buf_write_init(buf_write_arg_t *arg, void (*write_cb)(void *, const char *),
     void *cbopaque, char *buf, size_t buf_len) {
 	arg->write_cb = write_cb;
 	arg->cbopaque = cbopaque;
+	assert(buf != NULL);
 	arg->buf = buf;
+	assert(buf_len >= 2);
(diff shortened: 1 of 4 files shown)


interwq:
LGTM. However I do wonder if we should make the effort to tolerate OOM (vs simply report failure).

@davidtgoldblatt opinions? If you are not against the change, we should find a way to test the OOM paths (minimal we can do is probably, define the PRINT_BUFSIZE to be huge to trigger OOM, and run CI in that case).

yinan1048576:
> However I do wonder if we should make the effort to tolerate OOM (vs simply report failure).

Right, I also was not totally sure. The reason why I chose not to report a failure is that we may not always have a good way to do so. For example, `malloc_stats_print()` doesn't have a way to return error.

> we should find a way to test the OOM paths

Agree. And in general we should have a way to test the OOM paths for a lot of other places in the code base. I think we could mock our internal allocator (using the technique of function pointer).

davidtgoldblatt:
Sorry, I missed this diff in my inbox. This seems reasonable to me. I agree about oom testing, as well.

yinan1048576:
Rebase and Tested the OOM case by manually set the buffer pointers to NULL. No build/test failures. (Unit tests contain calls to all three buffered writer usages.)

Testing OOM more generally doesn't seem trivial. Will need to think more at a later time.

PR #1729, opened 2020-01-14
Don't purge huge extents when decay is off.

interwq:


src/extent.c
```diff
@@ -1224,7 +1224,8 @@ extent_record(tsdn_t *tsdn, arena_t *arena, ehooks_t *ehooks, ecache_t *ecache,
 			    &arena->edata_cache, ehooks, rtree_ctx, ecache,
 			    edata, &coalesced, growing_retained);
 		} while (coalesced);
-		if (edata_size_get(edata) >= oversize_threshold) {
+		if (edata_size_get(edata) >= oversize_threshold &&
+		    arena_may_force_decay(arena)) {
```
interwq: Not quite happy with this getting more arbitrary. However it's still useful to avoid confusion (people reporting purging happens on regular arenas, instead of huge arena, even with decay off).

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -103,6 +103,12 @@ arena_prof_info_set(tsd_t *tsd, edata_t *edata, prof_tctx_t *tctx) {
 	large_prof_info_set(edata, tctx);
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+arena_may_force_decay(arena_t *arena) {
+	return !(arena_dirty_decay_ms_get(arena) == -1
```
davidtgoldblatt: Is this the right condition? The arena may force decay if either one is non-negative, not only if both are non-negative.

interwq: Actually, if either is -1, it won't force_purge. That means the extent cache will queue at either dirty or muzzy states, and never make to retained / unmapped state?

PR #1730, opened 2020-01-14
Rework and fix the assertions on malloc fastpath.

interwq:
The first half of the malloc fastpath may execute before malloc_init.  Make the
assertions work in that case.

The regression was introduced in https://github.com/jemalloc/jemalloc/commit/dd649c94859e2cdbe7b527cfb743b549c8d8bf50

### DIFF
--- a/include/jemalloc/internal/sz.h
+++ b/include/jemalloc/internal/sz.h
@@ -152,10 +152,15 @@ sz_size2index_compute(size_t size) {
 }
 
 JEMALLOC_ALWAYS_INLINE szind_t
-sz_size2index_lookup(size_t size) {
+sz_size2index_lookup_impl(size_t size) {
 	assert(size <= SC_LOOKUP_MAXCLASS);
-	szind_t ret = (sz_size2index_tab[(size + (ZU(1) << SC_LG_TINY_MIN) - 1)
-					 >> SC_LG_TINY_MIN]);
+	return sz_size2index_tab[(size + (ZU(1) << SC_LG_TINY_MIN) - 1)
+	    >> SC_LG_TINY_MIN];
+}
+
+JEMALLOC_ALWAYS_INLINE szind_t
+sz_size2index_lookup(size_t size) {
+	szind_t ret = sz_size2index_lookup_impl(size);
 	assert(ret == sz_size2index_compute(size));
 	return ret;
 }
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/sz.h
```diff
@@ -152,10 +152,15 @@ sz_size2index_compute(size_t size) {
...
-	szind_t ret = (sz_size2index_tab[(size + (ZU(1) << SC_LG_TINY_MIN) - 1)
-					 >> SC_LG_TINY_MIN]);
+	return sz_size2index_tab[(size + (ZU(1) << SC_LG_TINY_MIN) - 1)
+	    >> SC_LG_TINY_MIN];
+}
+
+JEMALLOC_ALWAYS_INLINE szind_t
+sz_size2index_lookup(size_t size) {
+	szind_t ret = sz_size2index_lookup_impl(size);
 	assert(ret == sz_size2index_compute(size));
```
interwq: W/o this PR, the assertion here fails on the first malloc, as it may happen before malloc_init().

yinan1048576: In that case, `ret` would be 0?

src/jemalloc.c
```diff
@@ -2362,20 +2362,26 @@ je_malloc(size_t size) {
...
 	 */
-	size_t usize = sz_index2size(ind);
-	/*
-	 * Fast path relies on size being a bin.
-	 * I.e. SC_LOOKUP_MAXCLASS < SC_SMALL_MAXCLASS
-	 */
+	size_t usize;
+
+	sz_size2index_usize_fastpath(size, &ind, &usize);
+	assert(usize > 0 || !malloc_initialized());
```
yinan1048576: Can we have `assert(usize > 0 || (!malloc_initialized() && ind == 0))` here?

Also see my comment below.

src/jemalloc.c
```diff
@@ -2384,6 +2390,8 @@ je_malloc(size_t size) {
 	 * 0) in a single branch.
 	 */
 	uint64_t allocated_after = allocated + usize;
+	assert((!malloc_initialized() && threshold == 0) ||
+	    (ind == sz_size2index(size) && usize == sz_index2size(ind)));
```
yinan1048576: Is it the case that the part `(ind == sz_size2index(size) && usize == sz_index2size(ind))` always holds, regardless of `malloc_initialized()`?

I think we can also consolidate the assertion here with the one above, into a big (verbose) one here, like:
```
assert(ind == sz_size2index(size) && usize == sz_index2size(ind));
if (malloc_initialized()) {
    assert(usize > 0);
} else {
    assert(ind == 0 && usize == 0 && threshold == 0);
}
```

include/jemalloc/internal/sz.h
interwq: Exactly.

src/jemalloc.c
interwq: I should better document this; we have to be conservative, as there may be concurrent threads doing malloc_init(), i.e. even if !initialized(), we may still observe non zero ind / usize (as in, another thread is doing malloc_init and writing to the lookup tables).

Let me polish and add comments.

interwq: The other thing is, we cannot really call `sz_size2index` / `sz_index2size` if not initialized, they have assertions that would fail.

src/jemalloc.c
interwq: Answered below, the ind==0 cannot be guaranteed as the initialization can be partially done.

PR #1732, opened 2020-01-14
No tdata creation for backtracing on dying thread

yinan1048576:
The allocation record in the last-N list will still be reset, but there's no deallocation time / `tctx` recorded.

### DIFF
--- a/src/prof_data.c
+++ b/src/prof_data.c
@@ -406,7 +406,7 @@ prof_lookup(tsd_t *tsd, prof_bt_t *bt) {
 
 prof_tctx_t *
 prof_tctx_create(tsd_t *tsd) {
-	if (tsd_reentrancy_level_get(tsd) > 0) {
+	if (!tsd_nominal(tsd) || tsd_reentrancy_level_get(tsd) > 0) {
 		return NULL;
 	}
 


src/prof_data.c
```diff
@@ -406,7 +406,7 @@ prof_lookup(tsd_t *tsd, prof_bt_t *bt) {
 
 prof_tctx_t *
 prof_tctx_create(tsd_t *tsd) {
-	if (tsd_reentrancy_level_get(tsd) > 0) {
+	if (!tsd_nominal(tsd) || tsd_reentrancy_level_get(tsd) > 0) {
```
interwq: Does the non nominal check covers the reentrancy one?

yinan1048576: I think checking `tsd_state_get(tsd) == tsd_state_nominal` would guarantee `tsd_reentrancy_level_get(tsd) == 0`, but `tsd_nominal(tsd)` which checks `tsd_state_get(tsd) <= tsd_state_nominal_max` may not?

Upon a second thought, I think checking `tsd_state_get(tsd) == tsd_state_nominal` feels safer, and it can be checked by `!tsd_fast(tsd)`.

Thoughts?

interwq: We should still allow the creation during `tsd_state_nominal_slow`; say if we added alloc hooks which switch to slow path, sampling should still be allowed. `tsd_nominal(tsd)` sounds right to me.

interwq: So let's get rid of the reentrancy condition?

yinan1048576: Can `tsd_nominal(tsd)` really guarantee `tsd_reentrancy_level_get(tsd) == 0`? I could not find any proof for either case...

interwq: You are right. No guarantee. Strict nominal isn’t right, e.g. if tcache is off, it’s nominal_slow which should still have profiling working.

Let’s keep it as is then.

PR #1734, opened 2020-01-15
Get rid of lock overlap in prof_recent_alloc_reset

yinan1048576:
The `decrement_recent_count()` call is made to be async, and there's no concurrent lock holdings at any time.

The issue was detected in a stress test I was writing. I'm also appending the stress test to this PR.

### DIFF
--- a/src/prof_data.c
+++ b/src/prof_data.c
@@ -218,8 +218,8 @@ prof_gctx_create(tsdn_t *tsdn, prof_bt_t *bt) {
 }
 
 static void
-prof_gctx_try_destroy(tsd_t *tsd, prof_tdata_t *tdata_self, prof_gctx_t *gctx,
-    prof_tdata_t *tdata) {
+prof_gctx_try_destroy(tsd_t *tsd, prof_tdata_t *tdata_self,
+    prof_gctx_t *gctx) {
 	cassert(config_prof);
 
 	/*
@@ -371,7 +371,7 @@ prof_lookup(tsd_t *tsd, prof_bt_t *bt) {
 		    arena_ichoose(tsd, NULL), true);
 		if (ret.p == NULL) {
 			if (new_gctx) {
-				prof_gctx_try_destroy(tsd, tdata, gctx, tdata);
+				prof_gctx_try_destroy(tsd, tdata, gctx);
 			}
 			return NULL;
 		}
@@ -389,7 +389,7 @@ prof_lookup(tsd_t *tsd, prof_bt_t *bt) {
 		malloc_mutex_unlock(tsd_tsdn(tsd), tdata->lock);
 		if (error) {
 			if (new_gctx) {
-				prof_gctx_try_destroy(tsd, tdata, gctx, tdata);
+				prof_gctx_try_destroy(tsd, tdata, gctx);
 			}
 			idalloctm(tsd_tsdn(tsd), ret.v, NULL, NULL, true, true);
 			return NULL;
@@ -767,7 +767,7 @@ prof_gctx_finish(tsd_t *tsd, prof_gctx_tree_t *gctxs) {
 		if (prof_gctx_should_destroy(gctx)) {
 			gctx->nlimbo++;
 			malloc_mutex_unlock(tsd_tsdn(tsd), gctx->lock);
-			prof_gctx_try_destroy(tsd, tdata, gctx, tdata);
+			prof_gctx_try_destroy(tsd, tdata, gctx);
 		} else {
 			malloc_mutex_unlock(tsd_tsdn(tsd), gctx->lock);
 		}
@@ -1367,8 +1367,7 @@ prof_tdata_expire(tsdn_t *tsdn, prof_tdata_t *tdata) {
 	malloc_mutex_lock(tsdn, tdata->lock);
 	if (!tdata->expired) {
 		tdata->expired = true;
-		destroy_tdata = tdata->attached ? false :
-		    prof_tdata_should_destroy(tsdn, tdata, false);
+		destroy_tdata = prof_tdata_should_destroy(tsdn, tdata, false);
 	} else {
 		destroy_tdata = false;
 	}
@@ -1492,8 +1491,7 @@ prof_tctx_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
 	}
 	malloc_mutex_unlock(tsd_tsdn(tsd), gctx->lock);
 	if (destroy_gctx) {
-		prof_gctx_try_destroy(tsd, prof_tdata_get(tsd, false), gctx,
-		    tdata);
+		prof_gctx_try_destroy(tsd, prof_tdata_get(tsd, false), gctx);
 	}
 
 	malloc_mutex_assert_not_owner(tsd_tsdn(tsd), tctx->tdata->lock);
--- a/src/prof_data.c
+++ b/src/prof_data.c
@@ -1315,6 +1315,7 @@ static void
 prof_tdata_destroy_locked(tsd_t *tsd, prof_tdata_t *tdata,
     bool even_if_attached) {
 	malloc_mutex_assert_owner(tsd_tsdn(tsd), &tdatas_mtx);
+	malloc_mutex_assert_not_owner(tsd_tsdn(tsd), tdata->lock);
 
 	tdata_tree_remove(&tdatas, tdata);
 
@@ -1432,10 +1433,6 @@ prof_tctx_should_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
 
 static void
 prof_tctx_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
-	prof_tdata_t *tdata = tctx->tdata;
-	prof_gctx_t *gctx = tctx->gctx;
-	bool destroy_tdata, destroy_tctx, destroy_gctx;
-
 	malloc_mutex_assert_owner(tsd_tsdn(tsd), tctx->tdata->lock);
 
 	assert(tctx->cnts.curobjs == 0);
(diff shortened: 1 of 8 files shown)


interwq:
Looks to be crashing (`--enable-debug --enable-prof`) with:
```
Bus error (core dumped)
Test harness error: test/unit/prof_recent w/ MALLOC_CONF="prof:true,lg_prof_sample:0,prof_recent_alloc_max:3"
```

yinan1048576:
The test failure is only observed in freebsd but not ubuntu. MSVC reported at compile time that the `usleep()` function was not known. My guess is that freebsd may not know it either, but it failed to report it at compile time.

I found a cross platform sleep utility, but figured that the computation for Windows is incorrect there - `Sleep()` takes in milliseconds according to Microsoft's documentation. I fixed that and also restructured the sleep utility, in a separate commit.

Let's see if both the MSVC compile error and the freebsd test failure can be resolved...

yinan1048576:
(Shouldn't have removed the assertion in the sleep function - putting it back.)

yinan1048576:
Got some fixes and improvements for the stress test. Still no clue why freebsd got the test failure.

FYI, I was able to reproduce the test failure, and according to gdb, the `SIGBUS` signal comes from:
https://github.com/jemalloc/jemalloc/blob/dab81bd315e3eee19552ab68d331f693b205866a/src/prof_data.c#L1499
where I found that `tdata` points to some garbage. It feels that `tdata` has been released by some other thread. No further clue. The stack trace is:
```
Program received signal SIGBUS, Bus error.
witness_assert_not_owner (witness_tsdn=0x8007609b8, witness=0x5a5a5a5a5a5a5ab2)
    at witness.h:220
220             if (witness->rank == WITNESS_RANK_OMIT) {
(gdb) bt
#0  witness_assert_not_owner (witness_tsdn=0x8007609b8,
    witness=0x5a5a5a5a5a5a5ab2) at witness.h:220
#1  0x00000000004b2ae9 in malloc_mutex_assert_not_owner (tsdn=0x80075f138,
    mutex=0x5a5a5a5a5a5a5a5a) at mutex.h:245
#2  0x00000000004a63f5 in prof_tctx_destroy (tsd=0x80075f138, tctx=0x8007a01e0)
    at src/prof_data.c:1499
#3  0x00000000004a5f32 in jet_prof_tctx_try_destroy (tsd=0x80075f138,
    tctx=0x8007a01e0) at src/prof_data.c:1515
#4  0x00000000004bbb93 in decrement_recent_count (tsd=0x80075f138,
    tctx=0x8007a01e0) at src/prof_recent.c:90
#5  0x00000000004bc495 in jet_prof_recent_alloc (tsd=0x80075f138,
    edata=0x801e6b9c0, size=1) at src/prof_recent.c:371
#6  0x000000000048c0ff in jet_prof_malloc_sample_object (tsd=0x80075f138,
    ptr=0x8018c88c0, size=1, usize=8, tctx=0x8007a0a00) at src/prof.c:165
#7  0x0000000000417ca7 in prof_malloc (tsd=0x80075f138, ptr=0x8018c88c0,
    size=1, usize=8, alloc_ctx=0x7fffde3efda8, tctx=0x8007a0a00)
    at prof_inlines_b.h:136
#8  0x0000000000417009 in imalloc_body (sopts=0x7fffde3efeb8,
    dopts=0x7fffde3efe80, tsd=0x80075f138) at src/jemalloc.c:2178
```

yinan1048576:
(Note that freebsd was the only platform where the failure was observed. The Windows compiler error was gone and there was no test failure there. Not sure if freebsd might be doing anything special.)

interwq:
The failure looks to be real. Possible corruption not related to the change in this diff. @yinan1048576 please do dig into it.

yinan1048576:
Figured that the garbage `tdata` was a result from a previous `tsd_cleanup()` at thread exit (executed by another thread). So it seems that the `tdata` is somehow shared across threads. Is this allowed in theory? Still need to dig deeper into how such sharing was possible.

interwq:
The memory of the tdata can be reused. But should be reinitialized.

interwq:
What happens when the tdata is gone but the recent record remains, e.g. here: 
https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/prof_structs.h#L61

interwq:
I just realized, #1732 might be the fix for the issue. @yinan1048576 can you test that?

yinan1048576:
I tried it before, but it could not fix the problem. Still debugging.

yinan1048576:
Finally identified the issue. It was more trivial than I thought -

I was trying hard to figure whether `tdata` or `tctx` could be double-freed in some weird cases, but so far I don't think neither could ever happen - our designs are pretty bullet-proof!

Then I realized the real cause: the crash was not caused by double-free at all - the crashing thread was not trying to free the `tdata` at all!

The crash-causing scenario is like the following:
- Thread 1 executes `prof_tctx_destroy()`: it checks `tdata` and finds it *shouldn't* delete it, and releases the `tdata` lock.
- Thread 2 then checks the same `tdata` and finds it *should* be deleted, and releases the lock and deletes it.
- Thread 1 jumps in after the `tdata` has been deleted, and it should not try to access the `tdata` lock at all (`prof_data.c:1499`).

A few notes:
- In the scenario described above, thread 2 can either be executing `prof_tctx_destroy()` or `prof_tdata_detach()`. I observed both cases while debugging.
- It's possible that multiple threads can try to delete the same `tdata` at the same time, because `prof_tctx_destroy()` can be executed by a thread other than the thread originally owning the `tctx` / `tdata`. Remote free is one example. Now with the last-N profiling mode on, whenever the head of the ring buffer is replaced by a new allocation, both the allocation and deallocation `tctx`s for the head allocation need to be handled, and they very likely belong to another thread. With the last-N profiling mode on, the crash causing scenario becomes much more likely.

I inserted a separate commit "Properly handle tdata deletion race" to fix the issue. While debugging, I also figured that a couple of places in the prof code can be cleaned up, and I inserted another separate commit for the cleanup.

I also expanded the new stress test so as to include random invocations of the `mallctl` reader / writer.

src/prof_data.c
```diff
@@ -1492,13 +1491,11 @@ prof_tctx_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
...
 	if (destroy_gctx) {
-		prof_gctx_try_destroy(tsd, prof_tdata_get(tsd, false), gctx,
-		    tdata);
+		prof_gctx_try_destroy(tsd, prof_tdata_get(tsd, false), gctx);
 	}
 
-	malloc_mutex_assert_not_owner(tsd_tsdn(tsd), tctx->tdata->lock);
-
 	if (destroy_tdata) {
+		malloc_mutex_assert_not_owner(tsd_tsdn(tsd), tdata->lock);
```
interwq: Good catch! Let’s refactor this a bit, to destroy tdata right after unlock? I’d say even make tdata a scoped (I.e. make line 1440-1450 scoped) variable and comment it cannot be accessed outside of the scope.

src/prof_data.c
```diff
@@ -1492,13 +1491,11 @@ prof_tctx_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
 	}
 	malloc_mutex_unlock(tsd_tsdn(tsd), gctx->lock);
 	if (destroy_gctx) {
-		prof_gctx_try_destroy(tsd, prof_tdata_get(tsd, false), gctx,
```
interwq: Also I couldn’t find where this (line 1480) happens: 

> Change state such that the dumping thread will complete destruction during a late dump iteration phase.

Yinan can you please verify that’s still true?

yinan1048576: Yes, it's true. See `prof_gctx_finish()` - it deletes any `purgatory` `tctx` at the end of the dumping process and also deletes the `gctx` if needed.

src/prof_data.c
yinan1048576: I don't think the comment is necessary after making `tdata` a scoped variable?

yinan1048576:

Also moved the crash-causing assertion into `prof_tdata_destroy_locked()`, since (a) the assertion is logically tied to the function, and (b) the assertion can be executed for all other places calling the function.

src/prof_data.c
interwq: Feel free to not do the comment, if the code is self explanatory enough.

src/prof_data.c
```diff
@@ -1444,9 +1441,20 @@ prof_tctx_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
...
+	prof_gctx_t *gctx = tctx->gctx;
+
+	{
+		prof_tdata_t *tdata = tctx->tdata;
+		ckh_remove(tsd, &tdata->bt2tctx, &gctx->bt, NULL, NULL);
+		bool destroy_tdata = prof_tdata_should_destroy(tsd_tsdn(tsd),
+		    tdata, false);
+		malloc_mutex_unlock(tsd_tsdn(tsd), tdata->lock);
+		if (destroy_tdata) {
+			prof_tdata_destroy(tsd, tdata, false);
```
interwq: If we end up destroying tdata here, does that imply tctx is guaranteed to be destroyed by end of this function? If so, let's assert that.

yinan1048576: No, it's not guaranteed. It's possible that this function ends but `tctx` is not destroyed, if `tctx` is being dumped by another thread, and in such a case the dumping thread will call `prof_gctx_finish()` in the end to destroy `tctx`.

interwq: Okay in that case, when the delayed destroying tctx happens, does it touch tdata? Either way, let's clear tctx->tdata so that it's easier to catch.

yinan1048576:
Revised according to feedback. Decided to unconditionally set `tctx->tdata = NULL` (as long as execution enters `prof_tctx_destroy()`): even if `tdata` will not be destroyed, `tctx` will be destroyed, either by this thread or another thread currently dumping, and there should no longer be any need to access the `tdata` field on `tctx`.

src/prof_data.c
```diff
@@ -1444,9 +1441,21 @@ prof_tctx_destroy(tsd_t *tsd, prof_tctx_t *tctx) {
...
 	assert(tctx->cnts.accumbytes == 0);
 
-	ckh_remove(tsd, &tdata->bt2tctx, &gctx->bt, NULL, NULL);
-	destroy_tdata = prof_tdata_should_destroy(tsd_tsdn(tsd), tdata, false);
-	malloc_mutex_unlock(tsd_tsdn(tsd), tdata->lock);
+	prof_gctx_t *gctx = tctx->gctx;
+
+	{
+		prof_tdata_t *tdata = tctx->tdata;
+		tctx->tdata = NULL;
```
interwq: Question: what if tctx and tdata are both not destroyed? Should we keep tctx->tdata around so that we can try destroy again? I think the answer is no need, since we did the ckh_remove above?

yinan1048576: Right. The `ckh_remove()` decouples the `tctx` from the `tdata`. If the `tdata` ends up not being destroyed in this function, it could be destroyed later, but definitely not through this particular `tctx`. The `tctx->tdata = NULL` and `ckh_remove()` as a pair serve to completely detach the two.

interwq: Thanks for the answers Yinan!

PR #1736, opened 2020-01-16
Set reentrancy to 1 for tsd_state_purgatory.

interwq:
Reentrancy is already set for other non-nominal tsd states (reincarnated and
minimal_initialized).  Add purgatory to be safe and consistent.

### DIFF
--- a/include/jemalloc/internal/tsd.h
+++ b/include/jemalloc/internal/tsd.h
@@ -410,7 +410,10 @@ tsd_fetch(void) {
 
 static inline bool
 tsd_nominal(tsd_t *tsd) {
-	return (tsd_state_get(tsd) <= tsd_state_nominal_max);
+	bool nominal = tsd_state_get(tsd) <= tsd_state_nominal_max;
+	assert(nominal || tsd_reentrancy_level_get(tsd) > 0);
+
+	return nominal;
 }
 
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
It's not clear to me that this is the right fix; there's two states now: nominal and non-nominal. The non-nominal ones are non-nominal for a variety of reasons (reentrancy, purgatory, bootstrapping). It seems the right thing to do would be to *not* set it in minimal_initialized and reincarnated, and change whatever safety check that currently relies on the reentrancy level to instead rely on nominality.

interwq:
I'm thinking we keep all the non-nominal states at reentracy-level 1, for sanity check purposes. Agree that we should switch the reentrancy level checks to nominal states. Does that sound reasonable? If so I'll search for the reentrancy level checks and update accordingly.

interwq:
Actually, I think we do want to keep the reentrancy level checks, as they cover a wider range of situations, e.g. for this one -- https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L2657

We want both the real reentrancy case and the non nominal case to follow that path.

interwq:
Offline discussion, during thread death, we want to have the same treatment as reentrancy. If things get more complicated in the future, we might need to add another explicit state, instead of having tsd_state_purgatory / reincarnated piggyback on reentrancy_level directly.

PR #1743, opened 2020-01-22
Add "thread.idle" mallctl.

davidtgoldblatt:
This can encapsulate various internal cleaning logic, and can be used to free up
resources before a long sleep.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -1451,9 +1451,19 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
 	}								\
 } while (0)
 
+/* Can read or write, but not both. */
 #define READ_XOR_WRITE()	do {					\
 	if ((oldp != NULL && oldlenp != NULL) && (newp != NULL ||	\
 	    newlen != 0)) {						\
+		ret = EPERM;						\
+		goto label_return;					\
+	}								\
+} while (0)
+
+/* Can neither read nor write. */
+#define NEITHER_READ_NOR_WRITE()	do {				\
+	if (oldp != NULL || oldlenp != NULL || newp != NULL ||		\
+	    newlen != 0) {						\
 		ret = EPERM;						\
 		goto label_return;					\
 	}								\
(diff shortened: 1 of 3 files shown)


davidtgoldblatt:
One thing I'm thinking about: there's really two sorts of idleness. One is the calling thread saying "I personally will be idle for the next chunk of time; don't leave resources around on my account"; it's sort of a local-idleness statement. This is what this diff adds. Another sort of idleness is "I am spare computing capacity; do whatever work you might find to be useful"; more of a global-idleness one.

I think we may eventually want both of these (i.e. the former should purge only on the local arena; perhaps the latter could do a more global purging, could fill in some arena-level caches in the background, do some sort of mesh-style background compaction, whatever). I don't want to add the global idleness mallctl, but I don't want to pick a name that would conflict with it either.

So, that being said, is `thread.idle` the right name for this? Another possibility would be some sort of hint namespace, where both idleness versions could live. We could have `hint.thread_idle` (or `hint.thread.idle`) or something for the local property, and `hint.background_work` or something for the global property.

Any opinions?

src/ctl.c
```diff
@@ -1900,6 +1902,12 @@ thread_tcache_flush_ctl(tsd_t *tsd, const size_t *mib,
 		goto label_return;
 	}
 
+	/*
+	 * Slightly counterintuitively, READONLY() really just requires that the
+	 * call isn't trying to write, and WRITEONLY() just requires that it
+	 * isn't trying to read; hence, adding both requires that the operation
+	 * is neither a read nor a write.
+	 */
```
interwq: nit: add another macro doing no-read-no-write?

src/ctl.c
```diff
@@ -1971,6 +1979,41 @@ thread_prof_active_ctl(tsd_t *tsd, const size_t *mib,
...
+
+	if (tcache_available(tsd)) {
+		tcache_flush(tsd);
+	}
+	/*
+	 * This heuristic is perhaps not the most well-considered.  But it
+	 * matches the only idling policy we have experience with in the status
+	 * quo.  Over time we should investigate more principled approaches.
+	 */
+	if (opt_narenas > ncpus * 2) {
```
interwq: I'd say use the thread # approach instead? I think that heuristic was more of a way saying "if the malloc_conf is tuned".

PR #1748, opened 2020-01-24
Encapsulate buffer allocation in buffered writer

yinan1048576:
This change tries to achieve the following goals:
- Encapsulate buffer allocation within the buffered writer module. I found this to be very useful in simplifying the caller side logic, as well as in supporting additions of more functionalities to the buffered writer module in the future. The buffered writer module is also made to be flexible in accepting a caller supplied buffer.
- Test OOM. This is another main advantage of the encapsulation: now OOM testing can be a unit test for the buffered writer module, rather than multiple tests each for one caller, which are not only redundant but also hard. I'll see later if I can further generalize the OOM testing so that other jemalloc components can benefit.

### DIFF
--- a/src/buf_writer.c
+++ b/src/buf_writer.c
@@ -6,31 +6,31 @@
 #include "jemalloc/internal/malloc_io.h"
 
 void
-buf_write_flush(buf_write_arg_t *arg) {
-	assert(arg->buf_end <= arg->buf_size);
-	arg->buf[arg->buf_end] = '\0';
-	if (arg->write_cb == NULL) {
-		arg->write_cb = je_malloc_message != NULL ?
+buf_writer_flush(buf_writer_t *buf_writer) {
+	assert(buf_writer->buf_end <= buf_writer->buf_size);
+	buf_writer->buf[buf_writer->buf_end] = '\0';
+	if (buf_writer->write_cb == NULL) {
+		buf_writer->write_cb = je_malloc_message != NULL ?
 		    je_malloc_message : wrtmessage;
 	}
-	arg->write_cb(arg->cbopaque, arg->buf);
-	arg->buf_end = 0;
+	buf_writer->write_cb(buf_writer->cbopaque, buf_writer->buf);
+	buf_writer->buf_end = 0;
 }
 
 void
-buf_write_cb(void *buf_write_arg, const char *s) {
-	buf_write_arg_t *arg = (buf_write_arg_t *)buf_write_arg;
+buf_writer_cb(void *buf_writer_arg, const char *s) {
+	buf_writer_t *buf_writer = (buf_writer_t *)buf_writer_arg;
 	size_t i, slen, n, s_remain, buf_remain;
-	assert(arg->buf_end <= arg->buf_size);
+	assert(buf_writer->buf_end <= buf_writer->buf_size);
 	for (i = 0, slen = strlen(s); i < slen; i += n) {
-		if (arg->buf_end == arg->buf_size) {
-			buf_write_flush(arg);
+		if (buf_writer->buf_end == buf_writer->buf_size) {
+			buf_writer_flush(buf_writer);
 		}
 		s_remain = slen - i;
-		buf_remain = arg->buf_size - arg->buf_end;
+		buf_remain = buf_writer->buf_size - buf_writer->buf_end;
 		n = s_remain < buf_remain ? s_remain : buf_remain;
-		memcpy(arg->buf + arg->buf_end, s + i, n);
-		arg->buf_end += n;
+		memcpy(buf_writer->buf + buf_writer->buf_end, s + i, n);
+		buf_writer->buf_end += n;
 	}
 	assert(i == slen);
 }
--- a/src/buf_writer.c
+++ b/src/buf_writer.c
@@ -5,23 +5,114 @@
 #include "jemalloc/internal/buf_writer.h"
 #include "jemalloc/internal/malloc_io.h"
 
-void
-buf_writer_flush(buf_writer_t *buf_writer) {
-	assert(buf_writer->buf_end <= buf_writer->buf_size);
-	buf_writer->buf[buf_writer->buf_end] = '\0';
-	if (buf_writer->write_cb == NULL) {
-		buf_writer->write_cb = je_malloc_message != NULL ?
-		    je_malloc_message : wrtmessage;
+static void *
+buf_writer_allocate_internal_buf(tsdn_t *tsdn, size_t buf_len) {
+#ifdef JEMALLOC_JET
+	if (buf_len > SC_LARGE_MAXCLASS) {
+		return NULL;
 	}
-	buf_writer->write_cb(buf_writer->cbopaque, buf_writer->buf);
+#else
+	assert(buf_len <= SC_LARGE_MAXCLASS);
+#endif
+	return iallocztm(tsdn, buf_len, sz_size2index(buf_len), false, NULL,
+	    true, arena_get(tsdn, 0, false), true);
+}
+
+static void
+buf_writer_free_internal_buf(tsdn_t *tsdn, void *buf) {
+	if (buf != NULL) {
+		idalloctm(tsdn, buf, NULL, NULL, true, true);
+	}
+}
+
+static write_cb_t buf_writer_cb;
+
+static void
+buf_writer_assert(buf_writer_t *buf_writer) {
+	if (buf_writer->buf != NULL) {
+		assert(buf_writer->public_write_cb == buf_writer_cb);
+		assert(buf_writer->public_cbopaque == buf_writer);
+		assert(buf_writer->private_write_cb != buf_writer_cb);
+		assert(buf_writer->private_cbopaque != buf_writer);
+		assert(buf_writer->buf_size > 0);
+	} else {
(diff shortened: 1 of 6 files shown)


src/buf_writer.c
```diff
@@ -5,8 +5,43 @@
...
+	idalloctm(TSDN_NULL, buf, NULL, NULL, true, true);
+}
+
+bool
+buf_writer_init(buf_writer_t *buf_writer, write_cb_t *write_cb, void *cbopaque,
+    char *buf, size_t buf_len) {
+	buf_writer->write_cb = write_cb;
+	buf_writer->cbopaque = cbopaque;
+	assert(buf_len >= 2);
+	if (buf != NULL) {
```
interwq: It looks the buffer is never passed in. Do you have any use case in mind where the buffer is external?

src/buf_writer.c
```diff
@@ -5,8 +5,43 @@
...
+		buf_writer->buf = buf;
+		buf_writer->internal_buf = false;
+	} else {
+		buf_writer->buf = buf_writer_allocate_internal_buf(buf_len);
+		buf_writer->internal_buf = true;
+	}
+	buf_writer->buf_size = buf_len - 1; /* Allowing for '\0' at the end. */
+	buf_writer->buf_end = 0;
+	return buf_writer->buf == NULL;
```
interwq: nit: if allocation failed, better return early instead of set up everything. Maybe alloc at the beginning.

src/buf_writer.c
```diff
@@ -5,8 +5,43 @@
 #include "jemalloc/internal/buf_writer.h"
 #include "jemalloc/internal/malloc_io.h"
 
+static void *
+buf_writer_allocate_internal_buf(size_t buf_len) {
+	if (buf_len > SC_LARGE_MAXCLASS) {
+		return NULL;
+	}
+	return iallocztm(TSDN_NULL, buf_len, sz_size2index(buf_len), false,
```
interwq: Let’s avoid TSDN NULL unless we have to. If it’s only init and finish, passing in tsd should work?

src/buf_writer.c
```diff
@@ -34,3 +70,11 @@ buf_writer_cb(void *buf_writer_arg, const char *s) {
 	}
 	assert(i == slen);
 }
+
+void
+buf_writer_finish(buf_writer_t *buf_writer) {
```
interwq: finish sounds not strong enough. Terminate or deactivate?

src/buf_writer.c
```diff
@@ -37,6 +37,18 @@ buf_writer_init(buf_writer_t *buf_writer, write_cb_t *write_cb, void *cbopaque,
...
+static write_cb_t buf_writer_cb;
+
+write_cb_t *
+buf_writer_get_write_cb(buf_writer_t *buf_writer) {
+	return buf_writer->buf == NULL ? buf_writer->write_cb : buf_writer_cb;
+}
+
+void *
+buf_writer_get_cbopaque(buf_writer_t *buf_writer) {
+	return buf_writer->buf == NULL ? buf_writer->cbopaque : buf_writer;
```
interwq: I’m lost here. Are we trying to provide some fall back options? Doesn’t seem right.

src/buf_writer.c
interwq: It looks the intention is to make best effort and keep going. In that case, at least make the buf size 0.

src/buf_writer.c
yinan1048576: Yes, when a static buffer is preferred. An example is profiling, where we statically allocate the buffer, and this buffered writer should be able to accept it. (#1527 is trying to migrate profiling to use the buffered writer.) Currently the stats printing and profiling last-N dumping request a dynamic buffer, but we may change it later.

In my unit test I'm testing three cases: a static buffer, a dynamic buffer, and OOM.

src/buf_writer.c
yinan1048576: Makes sense. Also added unit test to verify.

src/buf_writer.c
yinan1048576: How about `tsdn`? In stats printing, we may not have a `tsd`.

src/buf_writer.c
yinan1048576: Yes, that's part of the encapsulation: which call back function and which argument to use should be determined by the buffered writer module. If the caller requested a dynamic buffer and the buffered writer OOMed, the caller has a way to know it (from the return of `buf_writer_init()`), but the caller doesn't have to, and in such a case the caller can just safely use whatever function and argument the buffered writer gives back. See the usage examples as well as the unit tests.

yinan1048576:

src/buf_writer.c
interwq: Yes sounds good.

src/buf_writer.c
interwq: I guess the part I'm not quite follow, is why we return the `buf_writer` itself when it doesn't have any buffer (implies OOM?). Don't we always pass down the buf_writer always?

yinan1048576: We don't always pass down the buffered writer itself. In case of OOM, we fall back to the original (unbuffered) call back function and argument. The encapsulation means that this decision should be part of the buffered writer, and it's feasible but not required for the caller to handle it.

yinan1048576:
Differentiated the caller facing (denoted by `public_*`) and buffered writer internal (denoted by `private_*`) call back functions and arguments. More specifically:
```
|                  |        usual case        |           OOM            |
|------------------|--------------------------|--------------------------|
|  public_write_cb |       buf_writer_cb      | caller provided write_cb |
|  public_cbopaque |        buf_writer        | caller provided cbopaque |
| private_write_cb | caller provided write_cb |           NULL           |
| private_cbopaque | caller provided cbopaque |           NULL           |
```
Also added an assertion function and constantly checked it.

src/buf_writer.c
```diff
@@ -5,32 +5,128 @@
 #include "jemalloc/internal/buf_writer.h"
 #include "jemalloc/internal/malloc_io.h"
 
+static void *
+buf_writer_allocate_internal_buf(tsdn_t *tsdn, size_t buf_len) {
+	if (buf_len > SC_LARGE_MAXCLASS) {
```
interwq: Since it’s internally used, an assertion will be fine.

src/buf_writer.c
```diff
@@ -5,32 +5,128 @@
...
+		return NULL;
+	}
+	return iallocztm(tsdn, buf_len, sz_size2index(buf_len), false, NULL,
+	    true, arena_get(tsdn, 0, false), true);
+}
+
+static void
+buf_writer_free_internal_buf(tsdn_t *tsdn, void *buf) {
+	if (buf != NULL) {
```
interwq: not: Assert is internal.

test/unit/buf_writer.c
```diff
@@ -35,32 +37,84 @@ TEST_BEGIN(test_buf_write) {
...
+	void *cbopaque = buf_writer_get_cbopaque(&buf_writer);
+	assert_ptr_eq(cbopaque, &arg, "Should use arg");
+
+	char s[UNIT_MAX + 1];
+	ssize_t unit;
+
+	memset(s, 'a', UNIT_MAX);
+	arg = 4; /* Starting value of random argument. */
+	arg_store = arg;
+	for (unit = UNIT_MAX; unit >= 0; --unit) {
```
interwq: This is looping quite a bit. Maybe -= max/16 so it’s only 16 iterations

src/buf_writer.c
yinan1048576: This was actually my trick: I had `SC_LARGE_MAXCLASS + 1` in my unit test for the purpose of explicitly triggering a OOM.

src/buf_writer.c
yinan1048576: You mean to pass in the `buf_writer` pointer here? My intention was that `buf_writer_allocate_internal_buf()` and `buf_writer_free_internal_buf()` only deal with the buffer array itself, and their respective callers, `buf_writer_init()` and `buf_writer_terminate()`, operate at the `buf_writer` level.

test/unit/buf_writer.c
yinan1048576: Made sense. It's not necessary to test this much for the OOM case. Further reduced it to 4 iterations.

Also decided to enrich the OOM unit test a bit: the buffered writer module is made so that the caller doesn't have to know the OOM, so the test logic should also be intentionally "blind" about it.

yinan1048576:

src/buf_writer.c
interwq: ah got auto corrected. I meant: add an assert(buf_is_internal).

src/buf_writer.c
interwq: I see. Let's either do a comment, or an #ifdef to only trigger in tests (and assert on non-tests)?

yinan1048576: Good point.

src/buf_writer.c
yinan1048576: Discussed offline. Will not add the assert, since `buf_writer_terminate()` tested it before the call, and this function is more meant to be a utility just dealing with the buffer array.

yinan1048576:

PR #1749, opened 2020-01-29
 Enforce page alignment for sampled allocations

### LINKED ISSUE #1697: Enforce alignment for sampled allocations
Discussed with @yinan1048576 offline. Sampled allocations are bumped to large size (16k). We should also bump the alignment (e.g. also to 16k) because:

- We rely on page alignment (when cache index randomization is off) to quickly check if `is_sampled` on sized dealloc. Enforce alignment for sampled will make sized deallocation always "available" regardless of cache index randomization.
- currently the 4K alignment (w/o randomization) is too popular, e.g. 50% of size 2k and all of 4k size won't benefit from sized dealloc.

interwq:
This allows sampled allocations to be checked through alignment, therefore enable sized deallocation regardless of cache_oblivious.

Enables the speed up of sized deallocation, in `test/integration/stress`:
before:

    100000000 iterations, dallocx=1254009us, sdallocx=1222009us, ratio=1:1.026

with this PR:

    100000000 iterations, dallocx=1259009us, sdallocx=759005us, ratio=1:1.659

Resolves #1697

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2013,6 +2013,7 @@ imalloc_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
 	szind_t ind_large;
 	size_t bumped_usize = usize;
 
+	dopts->alignment = prof_sample_align(dopts->alignment);
 	if (usize <= SC_SMALL_MAXCLASS) {
 		assert(((dopts->alignment == 0) ?
 		    sz_s2u(SC_LARGE_MINCLASS) :
@@ -2029,6 +2030,7 @@ imalloc_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
 	} else {
 		ret = imalloc_no_sample(sopts, dopts, tsd, usize, usize, ind);
 	}
+	assert(prof_sample_aligned(ret));
 
 	return ret;
 }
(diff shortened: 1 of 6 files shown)


interwq:
Ready for review @yinan1048576 @davidtgoldblatt

src/jemalloc.c
```diff
@@ -2598,32 +2600,41 @@ isfree(tsd_t *tsd, void *ptr, size_t usize, tcache_t *tcache, bool slow_path) {
...
+			ctx = &alloc_ctx;
 			rtree_ctx_t *rtree_ctx = tsd_rtree_ctx(tsd);
 			rtree_szind_slab_read(tsd_tsdn(tsd), &extents_rtree,
-			    rtree_ctx, (uintptr_t)ptr, true, &dbg_ctx.szind,
-			    &dbg_ctx.slab);
-			assert(dbg_ctx.szind == alloc_ctx.szind);
-			assert(dbg_ctx.slab == alloc_ctx.slab);
+			    rtree_ctx, (uintptr_t)ptr, true, &ctx->szind,
+			    &ctx->slab);
+			assert(ctx->szind == sz_size2index(usize));
```
davidtgoldblatt: Let's make this a safety check failure? We've already done the slow part; might as well get some benefit out of it.

interwq: Good point. Let me add that.

interwq: Added a separate commit https://github.com/jemalloc/jemalloc/pull/1749/commits/8709112c2f2731ad874c7ac8d15e00488aa66678

src/jemalloc.c
```diff
@@ -2631,7 +2631,11 @@ isfree(tsd_t *tsd, void *ptr, size_t usize, tcache_t *tcache, bool slow_path) {
 			rtree_szind_slab_read(tsd_tsdn(tsd), &extents_rtree,
 			    rtree_ctx, (uintptr_t)ptr, true, &ctx->szind,
 			    &ctx->slab);
-			assert(ctx->szind == sz_size2index(usize));
+			bool sz_correct = (ctx->szind == sz_size2index(usize));
+			if (config_opt_safety_checks && !sz_correct) {
+				safety_check_fail_sized_dealloc();
+			}
+			assert(sz_correct);
```
davidtgoldblatt: config_opt_safety_checks is true whenever asserts are enabled; this can't ever fire.

src/safety_check.c
```diff
@@ -3,6 +3,13 @@
 
 static void (*safety_check_abort)(const char *message);
 
+void safety_check_fail_sized_dealloc(void) {
```
davidtgoldblatt: The two callers of this are slightly different; in the old one,  the stack trace when it fires is useless (i.e. it may not have been the source of the corruption). In the new one, the place where it fires gives the bug. That might justify a parameter or something to indicate which.

Not a strong opinion, disregard if you disagree.

src/jemalloc.c
interwq: It looks to be controlled separately, through `--enable-opt-safety-checks` (which is default off upstream https://github.com/jemalloc/jemalloc/blob/dev/configure.ac#L1475)?

src/safety_check.c
interwq: That's a valid concern. Let me tweak it a bit.

src/jemalloc.c
davidtgoldblatt: It's set with the bool rather than the macro -- https://github.com/jemalloc/jemalloc/blob/21cfe59ff7b10a61dabe26cd3dbfb7a255e1f5e8/include/jemalloc/internal/jemalloc_preamble.h.in#L171

interwq: Oh I see what you mean. Will update.

src/safety_check.c
```diff
@@ -3,6 +3,17 @@
 
 static void (*safety_check_abort)(const char *message);
 
+void safety_check_fail_sized_dealloc(bool current_dealloc) {
+	assert(config_opt_safety_checks);
+	char *src = current_dealloc ? "the current pointer being freed" :
+	    "in thread cache (possibly from previous deallocations). Suggest"
+	    "building with --enable-debug or address sanitizer for debugging";
```
interwq: Does this look reasonable @davidtgoldblatt ?

src/safety_check.c
```diff
@@ -3,6 +3,18 @@
...
 
+void safety_check_fail_sized_dealloc(bool current_dealloc) {
+	assert(config_opt_safety_checks);
+	char *src = current_dealloc ? "the current pointer being freed" :
+	    "in thread cache, possibly from previous deallocations";
+
+	safety_check_fail("<jemalloc>: size mismatch detected, likely caused by"
+	   " application sized deallocation bugs (source: %s). Suggest building"
+	    "with --enable-debug or address sanitizer for debugging. Abort.\n",
+	    src);
```
interwq: Does this look reasonable @davidtgoldblatt ?

davidtgoldblatt: LGTM

PR #1752, opened 2020-01-31
 NetBSD specific changes

zoulasc:
- NetBSD overcommits
- When mapping pages, use the maximum of the alignment requested and the compiled-in PAGE constant which might be greater than the current kernel pagesize, since we compile binaries with the maximum page size supported by the architecture (so that they work with all kernels).


davidtgoldblatt:
Looks good to me, thanks for the patch!

Could you squash this down to a single commit at jemalloc dev head? We keep a linear history for dev.

PR #1756, opened 2020-02-04
Abbreviate thread-event to te

interwq:


interwq:
@yinan1048576 this one should be straightforward for you.

interwq:
re: event_ctx. Good idea. Will update.

PR #1758, opened 2020-02-05
Add ASSURED_WRITE in mallctl

yinan1048576:
There are cases where we require the `mallctl()` caller to provide an input. The current `WRITE()` treats caller input as optional, so when an input is required, we'd have to artificially do some initialization ahead and some check afterwards. The new `MUST_WRITE` tries to make such scenarios less painful.

Currently I'm only migrating `tcache.flush` and `tcache.destroy` to the new `MUST_WRITE`. Let me know if any other `mallctl()` calls can be changed.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -1496,6 +1496,14 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
 	}								\
 } while (0)
 
+#define ASSURED_WRITE(v, t)	do {					\
+	if (newp == NULL || newlen != sizeof(t)) {			\
+		ret = EINVAL;						\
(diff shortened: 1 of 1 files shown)


src/ctl.c
```diff
@@ -1496,6 +1496,14 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
 	}								\
 } while (0)
 
+#define MUST_WRITE(v, t)	do {					\
```
interwq: ASSURED_WRITE? No strong opinion either way though.

yinan1048576:
Revise according to feedback.

PR #1759, opened 2020-02-06
Add mallctl for dumping last-N profiling records

yinan1048576:
I'm currently defining a `write_cb_packet_t` construct in `ctl.c` to represent the `[write_cb, cbopaque]` tuple passed in as input to the `mallctl()`.

Also added extensive unit tests.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -237,6 +237,7 @@ CTL_PROTO(experimental_utilization_batch_query)
 CTL_PROTO(experimental_arenas_i_pactivep)
 INDEX_PROTO(experimental_arenas_i)
 CTL_PROTO(experimental_prof_recent_alloc_max)
+CTL_PROTO(experimental_prof_recent_alloc_dump)
 
 #define MUTEX_STATS_CTL_PROTO_GEN(n)					\
 CTL_PROTO(stats_##n##_num_ops)						\
@@ -631,6 +632,7 @@ static const ctl_indexed_node_t experimental_arenas_node[] = {
 
 static const ctl_named_node_t experimental_prof_recent_node[] = {
 	{NAME("alloc_max"),	CTL(experimental_prof_recent_alloc_max)},
+	{NAME("alloc_dump"),	CTL(experimental_prof_recent_alloc_dump)},
 };
 
 static const ctl_named_node_t experimental_node[] = {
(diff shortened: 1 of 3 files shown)


src/ctl.c
```diff
@@ -3549,3 +3551,34 @@ experimental_prof_recent_alloc_max_ctl(tsd_t *tsd, const size_t *mib,
...
+experimental_prof_recent_alloc_dump_ctl(tsd_t *tsd, const size_t *mib,
+    size_t miblen, void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret;
+
+	if (!(config_prof && opt_prof)) {
+		ret = ENOENT;
+		goto label_return;
+	}
+
+	assert(sizeof(write_cb_packet_t) == sizeof(size_t) * 2);
```
interwq: Why is this required?

test/unit/prof_recent.c
```diff
@@ -381,6 +381,204 @@ TEST_END
...
+} confirm_record_t;
+
+#define ERROR "Dump output is wrong"
+
+static void confirm_record(const char *template,
+    const confirm_record_t *records, const size_t n_records) {
+	static const char *types[2] = {"alloc", "dalloc"};
+	static char buf[64];
+
+	assert_d_eq(memcmp(dump_out, template, strlen(template) - 2), 0, ERROR);
```
interwq: nit: explain why the -2 offset.

test/unit/prof_recent.c
```diff
@@ -381,6 +381,204 @@ TEST_END
...
+	for (record = records; record < records + n_records; ++record) {
+		if (record != records) {
+			assert_true(start < end, ERROR);
+			assert_c_eq(*start++, ',', ERROR);
+		}
+
+		assert_true(start < end, ERROR);
+		assert_c_eq(*start++, '{', ERROR);
+
+		{
```
interwq: Abstract the three blocks below into a function?

test/unit/prof_recent.c
```diff
@@ -381,6 +381,204 @@ TEST_END
...
+			malloc_snprintf(buf, sizeof(buf), "%s",
+			    record->released ? "true" : "false");
+			assert_d_eq(memcmp(start, buf, strlen(buf)), 0, ERROR);
+			start += strlen(buf);
+			assert_c_eq(*start++, ',', ERROR);
+		}
+
+		const char **type = types;
+		while (true) {
+			assert_true(start < end, ERROR);
```
interwq: Similarly, the three blocks here.

src/ctl.c
yinan1048576: It's similar to what we had for the memory utilization APIs. See https://github.com/jemalloc/jemalloc/blob/7014f81e172290466e1a28118b622519bbbed2b0/src/ctl.c#L3442.

The input structure is not public, so the caller would have to make up a `sizeof(size_t) * 2` input array, but internally I'm using this `write_cb_packet_t` structure, so I'd better assert that the two are equivalent.

yinan1048576:
Decided to abstract at one layer lower.

src/ctl.c
interwq: Here probably better to say `== sizeof(void *) * 2`

yinan1048576: Oh I see. Yes, you're right.

yinan1048576:

yinan1048576:
Noticed that my `ERROR` macro clashed with some internal macro in some environment. Changed it to `DUMP_ERROR`.

PR #1760, opened 2020-02-06
Differentiate between expect_* and assert_* in tests

yinan1048576:
We have two different needs from test assertions:
- Assess whether something is as expected, and continue regardless of the outcome;
- Assess whether something is as expected, and immediately stop if not.

The `assert_*` macros we currently have can only satisfy the first need but not the second. The second need is critical for cases where the subsequent test logic depends on the assertion.

I realized the second need while developing for #1759: while scanning through a long string to verify its components, it doesn't make much sense to continue scanning, if it's detected that something is missing in the middle of the string and the subsequent verification logic has positional assumptions.

This kind of need is somewhat similar to why we need our `assert()`, but it's slightly different: `assert()` should be used to guard against issues that are directly relevant to the current code, e.g. bugs in the test code itself; the new need applies to what we want to test for, but there are just dependencies from subsequent test logic.

I searched around and found that such a differentiation already exists. One example is https://github.com/google/googletest/blob/master/googletest/docs/primer.md#assertions . I think their terminology also makes sense: the first need should be `expect_*` while the second need should be `assert_*`.

I made the change accordingly in our codebase. I also modified some test files I'm more familiar with, and leave other files to the future.

### DIFF
--- a/test/include/test/btalloc.h
+++ b/test/include/test/btalloc.h
@@ -25,6 +25,6 @@ btalloc_##n(size_t size, unsigned bits) {				\
 		}							\
 	}								\
 	/* Intentionally sabotage tail call optimization. */		\
-	assert_ptr_not_null(p, "Unexpected mallocx() failure");		\
+	expect_ptr_not_null(p, "Unexpected mallocx() failure");		\
 	return p;							\
 }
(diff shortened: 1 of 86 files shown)


yinan1048576:
Improved the bash script.

yinan1048576:
Minor formatting improvements.

yinan1048576:
Rebase, and made use of `assert_*` in more places.

yinan1048576:
Rebase.

yinan1048576:
Merging, ignoring the installation failures in `FreeBSD:freebsd-11-2-release-amd64` and `FreeBSD:freebsd-12-0-release-amd64`.

PR #1763, opened 2020-02-07
Rework the bin locking around tcache refill / flush.

interwq:
Previously, tcache fill/flush (as well as small alloc/dalloc on the arena) may
    potentially drop the bin lock for slab_alloc and slab_dalloc.  This commit
    refactors the logic so that the slab calls happen in the same function / level
    as the bin lock / unlock.  The main purpose is to be able to use flat combining
    without having to keep track of stack state.
    
    In the meantime, this change reduces the locking, especially for slab_dalloc
    calls, where nothing happens after the call.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -60,8 +60,6 @@ static void arena_decay_to_limit(tsdn_t *tsdn, arena_t *arena,
     size_t npages_decay_max, bool is_background_thread);
 static bool arena_decay_dirty(tsdn_t *tsdn, arena_t *arena,
     bool is_background_thread, bool all);
-static void arena_dalloc_bin_slab(tsdn_t *tsdn, arena_t *arena, edata_t *slab,
-    bin_t *bin);
 static void arena_bin_lower_slab(tsdn_t *tsdn, arena_t *arena, edata_t *slab,
     bin_t *bin);
 
@@ -996,7 +994,7 @@ arena_decay(tsdn_t *tsdn, arena_t *arena, bool is_background_thread, bool all) {
 	arena_decay_muzzy(tsdn, arena, is_background_thread, all);
 }
 
-static void
+void
 arena_slab_dalloc(tsdn_t *tsdn, arena_t *arena, edata_t *slab) {
 	arena_nactive_sub(arena, edata_size_get(slab) >> LG_PAGE);
 
@@ -1252,101 +1250,55 @@ arena_slab_alloc(tsdn_t *tsdn, arena_t *arena, szind_t binind, unsigned binshard
 	return slab;
 }
 
-static edata_t *
-arena_bin_nonfull_slab_get(tsdn_t *tsdn, arena_t *arena, bin_t *bin,
-    szind_t binind, unsigned binshard) {
-	edata_t *slab;
-	const bin_info_t *bin_info;
+/*
+ * Before attempting the _with_fresh_slab approaches below, the _no_fresh_slab
+ * variants (i.e. through slabcur and nonfull) must be tried first.
+ */
+static void
+arena_bin_refill_slabcur_with_fresh_slab(tsdn_t *tsdn, arena_t *arena,
+    bin_t *bin, szind_t binind, edata_t *fresh_slab) {
+	malloc_mutex_assert_owner(tsdn, &bin->lock);
+	/* Only called after slabcur and nonfull both failed. */
+	assert(bin->slabcur == NULL);
+	assert(edata_heap_first(&bin->slabs_nonfull) == NULL);
+	assert(fresh_slab != NULL);
 
-	/* Look for a usable slab. */
-	slab = arena_bin_slabs_nonfull_tryget(bin);
-	if (slab != NULL) {
-		return slab;
+	/* A new slab from arena_slab_alloc() */
+	assert(edata_nfree_get(fresh_slab) == bin_infos[binind].nregs);
+	if (config_stats) {
+		bin->stats.nslabs++;
+		bin->stats.curslabs++;
 	}
-	/* No existing slabs have any space available. */
-
-	bin_info = &bin_infos[binind];
-
-	/* Allocate a new slab. */
-	malloc_mutex_unlock(tsdn, &bin->lock);
-	/******************************/
-	slab = arena_slab_alloc(tsdn, arena, binind, binshard, bin_info);
-	/********************************/
-	malloc_mutex_lock(tsdn, &bin->lock);
-	if (slab != NULL) {
-		if (config_stats) {
-			bin->stats.nslabs++;
-			bin->stats.curslabs++;
-		}
-		return slab;
-	}
-
-	/*
-	 * arena_slab_alloc() failed, but another thread may have made
-	 * sufficient memory available while this one dropped bin->lock above,
-	 * so search one more time.
-	 */
-	slab = arena_bin_slabs_nonfull_tryget(bin);
-	if (slab != NULL) {
-		return slab;
-	}
-
-	return NULL;
+	bin->slabcur = fresh_slab;
 }
 
-/* Re-fill bin->slabcur, then call arena_slab_reg_alloc(). */
+/* Refill slabcur and then alloc using the fresh slab */
 static void *
-arena_bin_malloc_hard(tsdn_t *tsdn, arena_t *arena, bin_t *bin,
-    szind_t binind, unsigned binshard) {
+arena_bin_malloc_with_fresh_slab(tsdn_t *tsdn, arena_t *arena, bin_t *bin,
+    szind_t binind, edata_t *fresh_slab) {
+	malloc_mutex_assert_owner(tsdn, &bin->lock);
+	arena_bin_refill_slabcur_with_fresh_slab(tsdn, arena, bin, binind,
+	    fresh_slab);
+
+	return arena_slab_reg_alloc(bin->slabcur, &bin_infos[binind]);
+}
+
+static bool
+arena_bin_refill_slabcur_no_fresh_slab(tsdn_t *tsdn, arena_t *arena,
+    bin_t *bin) {
+	malloc_mutex_assert_owner(tsdn, &bin->lock);
+	/* Only called after arena_slab_reg_alloc[_batch] failed. */
+	assert(bin->slabcur == NULL || edata_nfree_get(bin->slabcur) == 0);
 
 	if (bin->slabcur != NULL) {
-		/* Only attempted when current slab is full. */
-		assert(edata_nfree_get(bin->slabcur) == 0);
-	}
-
-	const bin_info_t *bin_info = &bin_infos[binind];
-	edata_t *slab = arena_bin_nonfull_slab_get(tsdn, arena, bin, binind,
-	    binshard);
-	if (bin->slabcur != NULL) {
-		if (edata_nfree_get(bin->slabcur) > 0) {
-			/*
-			 * Another thread updated slabcur while this one ran
-			 * without the bin lock in arena_bin_nonfull_slab_get().
-			 */
-			void *ret = arena_slab_reg_alloc(bin->slabcur,
-			    bin_info);
-			if (slab != NULL) {
-				/*
-				 * arena_slab_alloc() may have allocated slab,
-				 * or it may have been pulled from
-				 * slabs_nonfull.  Therefore it is unsafe to
-				 * make any assumptions about how slab has
-				 * previously been used, and
-				 * arena_bin_lower_slab() must be called, as if
-				 * a region were just deallocated from the slab.
-				 */
-				if (edata_nfree_get(slab) == bin_info->nregs) {
-					arena_dalloc_bin_slab(tsdn, arena, slab,
-					    bin);
-				} else {
-					arena_bin_lower_slab(tsdn, arena, slab,
-					    bin);
-				}
-			}
-			return ret;
-		}
-
 		arena_bin_slabs_full_insert(arena, bin, bin->slabcur);
-		bin->slabcur = NULL;
 	}
 
-	if (slab == NULL) {
-		return NULL;
-	}
-	bin->slabcur = slab;
-	assert(edata_nfree_get(bin->slabcur) > 0);
+	/* Look for a usable slab. */
+	bin->slabcur = arena_bin_slabs_nonfull_tryget(bin);
+	assert(bin->slabcur == NULL || edata_nfree_get(bin->slabcur) > 0);
 
-	return arena_slab_reg_alloc(slab, bin_info);
+	return (bin->slabcur == NULL);
 }
 
 /* Choose a bin shard and return the locked bin. */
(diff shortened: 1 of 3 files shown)


interwq:
Benchmark done. As expected the lock rate is reduced. Malloc-related mutex cycles improvement is between 5% to 9%; malloc_stats show ~10% drop in # of lock operations.

stats from two hosts running similar workload: (top half is with this PR)

size | ind | allocated | nmalloc | (#/sec) | ndalloc | (#/sec) | nrequests | (#/sec) | n_lock_ops | (#/sec) | n_waiting | (#/sec) | n_spin_acq | (#/sec)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
8 | 0 | 2966584 | 2918150548 | 532509 | 2917779725 | 532441 | 22365312311 | 4081261 | 438905727 | 80092 | 1019 | 0 | 113779 | 20
16 | 1 | 5607872 | 1931885498 | 352533 | 1931535006 | 352469 | 5496740403 | 1003054 | 247622150 | 45186 | 504 | 0 | 47531 | 8
32 | 2 | 716609152 | 12912357639 | 2356269 | 12889963603 | 2352183 | 31369458400 | 5724353 | 1324026292 | 241610 | 23413 | 4 | 1319563 | 240
48 | 3 | 530338752 | 5128429948 | 935844 | 5117381224 | 933828 | 16267612083 | 2968542 | 703582075 | 128390 | 4645 | 0 | 368756 | 67
64 | 4 | 421403584 | 1590349408 | 290209 | 1583764977 | 289008 | 10318096431 | 1882864 | 242432144 | 44239 | 1469 | 0 | 51475 | 9
80 | 5 | 208579040 | 1537485280 | 280563 | 1534878042 | 280087 | 2991800233 | 545948 | 296032628 | 54020 | 215 | 0 | 21788 | 3
96 | 6 | 306388512 | 1111423770 | 202814 | 1108232223 | 202232 | 2102679147 | 383700 | 232407261 | 42410 | 591 | 0 | 32531 | 5
  |   |   |   |   |   |   |   |   |   |   |   |   |   |  
size | ind | allocated | nmalloc | (#/sec) | ndalloc | (#/sec) | nrequests | (#/sec) | n_lock_ops | (#/sec) | n_waiting | (#/sec) | n_spin_acq | (#/sec)
8 | 0 | 2968264 | 2908992400 | 533075 | 2908621367 | 533007 | 22221415882 | 4072093 | 542922980 | 99491 | 1157 | 0 | 152828 | 28
16 | 1 | 5575392 | 1925318982 | 352816 | 1924970520 | 352752 | 5466140008 | 1001674 | 299114792 | 54813 | 610 | 0 | 48388 | 8
32 | 2 | 715060032 | 12832031109 | 2351480 | 12809685483 | 2347386 | 31171596013 | 5712222 | 1359917426 | 249206 | 37972 | 6 | 1807015 | 331
48 | 3 | 530310672 | 5105233949 | 935538 | 5094185810 | 933513 | 16110483548 | 2952260 | 749799031 | 137401 | 6930 | 1 | 486420 | 89
64 | 4 | 421145216 | 1589383501 | 291255 | 1582803107 | 290050 | 10257344259 | 1879667 | 303726995 | 55658 | 1576 | 0 | 74366 | 13
80 | 5 | 199936240 | 1530659778 | 280494 | 1528160575 | 280036 | 2970430829 | 544334 | 352006310 | 64505 | 332 | 0 | 40737 | 7
96 | 6 | 310570944 | 1111769306 | 203732 | 1108534192 | 203139 | 2088863329 | 382786 | 256626392 | 47027 | 823 | 0 | 47886 | 8

yinan1048576:
Quick question while I'm reading through the code: is it worthwhile to also rework `arena_bin_reset()`? The logic around the bin lock there seems similar.

interwq:
Re: arena reset. I can give it a try. Probably better in a separate PR though.

yinan1048576:
Yeah, probably not in this PR. I guess it wouldn't affect the numbers much, since `arena_reset()` is rarely called(?)

interwq:
Right. In the workloads we test, none does that.

yinan1048576:
Another quick thought: maybe it's beneficial to add `malloc_mutex_assert_owner(tsdn, &bin->lock);` to the functions along the `malloc` path, just to improve code interpretability? (The old code would actually need it even more, since the `lock` and `unlock` pairs were often not at the same level.)

interwq:
Re: assert lock owner -- good point. Added https://github.com/jemalloc/jemalloc/pull/1763/commits/76f8977c484409b8de4348bc16d9a15469e2222b for the ease of review.

src/arena.c
```diff
@@ -1369,63 +1319,108 @@ arena_bin_choose_lock(tsdn_t *tsdn, arena_t *arena, szind_t binind,
...
+
+		if (terminate) {
+			break;
+		}
+		assert(!cur_iter_alloc && fresh_slab == NULL);
+		fresh_slab = arena_slab_alloc(tsdn, arena, binind, binshard,
+		    bin_info);
+		cur_iter_alloc = true;
+		/* fresh_slab NULL case handled in the for loop. */
+	} while (true);
```
yinan1048576: This two-level loop is one of the most mysterious logic I've ever read! :) I think in any case it would be helpful to add some more comments.

On a high level, I can see that you're getting rid of the `arena_bin_malloc_hard()` call in the middle, and to fulfill its purpose you're calling `arena_slab_alloc()` at the end, outside of the bin lock.

I'm not quite sure if I'm understanding your usage of the two booleans correctly. Let me confirm:

`terminate`: on a higher level, I think `terminate = true` if either all `nfill` slots are fulfilled, or if there's no hope to believe that the next `arena_slab_alloc()` may succeed; otherwise `terminate = false`.

`cur_iter_alloc`: it's `true` only in the first iteration of the inner `for` loop of all non-first iterations of the outer `do`-`while` loop, and if it's `true`, you would lose hope and `terminate` whenever you cannot immediately find any available memory when under the bin lock.

I wonder: say if the `arena_slab_alloc()` call in a particular iteration of the `do`-`while` loop already OOMed, then in the next iteration of the `do`-`while` loop, how is it possible for the `for` loop to ever enter the second iteration? Probably only if it just happened that some other thread returned some available memory back during the time the bin lock is released but after `arena_slab_alloc()` returned `NULL`? If we ignore this possibility (if it's reasonably small), then we don't need this `cur_iter_alloc` at all, but rather just test for OOM on `arena_slab_alloc()` and break the `do`-`while` loop if it OOMed?

I understand that you're trying to keep the logic equivalent to what we had before: previously, `arena_bin_nonfull_slab_get()` calls `arena_bin_slabs_nonfull_tryget()` again even if `arena_slab_alloc()` OOMed, but at that time it was trivial to do so and it was pretty easy to understand. But now to do so, the logic becomes a puzzle to interpret. :) I think if we can get rid of the `cur_iter_alloc`, the logic can be much simpler (if the cost we're paying is negligible).

yinan1048576: Well, the possibility may not be negligible, but I guess, even if we double check after `arena_slab_alloc()` OOMed and it turns out that we still cannot get any available memory, it might be possible that some other thread may bring back some available memory before `arena_tcache_fill_small()` ends, so the double checking cannot provide us with any stronger guarantee.

yinan1048576:
> Re: assert lock owner -- good point. Added 76f8977 for the ease of review.

Looks great!

src/arena.c
interwq: What you described is pretty much accurate. Indeed the `cur_iter_alloc` is to track, if the slab_alloc was attempted in this exact iteration of refill. You are totally right that this is the source of the complexity.

The refill can take quite some operations (think about refilling 100 items with mostly full slabs, which means dozens of slabs are used). One particular case the `cur_iter_alloc` solves is: 
1) current thread finds nothing in the bin, unlocks and tries slab_alloc;
2) slab_alloc returns nothing (I'll explain later why this still matters nowadays);
3) after relock, the thread is able to find available slabs in the bin, and keeps refilling;
4) after a while, the bin local slabs are all consumed, then the thread tries to see if `fresh_slab` has anything.

In this case, if we don't have `cur_iter_alloc` and conclude "tried allocating but didn't get fresh_slab; it must OOM", then it's based on things happened a while before (i.e. step 2). That's why I had the `cur_iter_alloc` to indicate if we just tried slab_alloc. Perhaps it should be named `recent_allocated` or `just_allocated`?

Then the benefit of doing is not quite obvious. If we consider only the OOM case on the powerful servers w/ overcommit, it doesn't matter at all. It comes down to two other scenarios:
1) if the current arena is hosting huge pages, in which case its limited resource and we should prioritize using it. Also it will be a contented resource (e.g. the idea David is trying, always prefer the dedicated arena w/ transparent huge pages). Here try harder before bailing out, makes sure the huge pages are tightly packed / utilized.
2) on small devices with limited memory and no overcommit, it can help avoid OOM.

Completely agree this needs better comments. However I'm still inclined to pay the complexity here because of the two cases above. WDYT?

src/arena.c
```diff
@@ -1252,101 +1250,53 @@ arena_slab_alloc(tsdn_t *tsdn, arena_t *arena, szind_t binind, unsigned binshard
...
+
+	/* A new slab from arena_slab_alloc() */
+	assert(edata_nfree_get(fresh_slab) == bin_infos[binind].nregs);
+	if (config_stats) {
+		bin->stats.nslabs++;
+		bin->stats.curslabs++;
 	}
-
-	return NULL;
+	assert(edata_nfree_get(fresh_slab) > 0);
```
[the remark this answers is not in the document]
interwq: That's true. Will remove this one.

src/arena.c
```diff
@@ -1369,63 +1319,108 @@ arena_bin_choose_lock(tsdn_t *tsdn, arena_t *arena, szind_t binind,
 void
 arena_tcache_fill_small(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
```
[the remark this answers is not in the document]
interwq: Yeah I'll try moving it in a separate PR too.

src/arena.c
interwq: One thing forgot to mention was, arena_slab_alloc requires the `eset` lock, so it can potentially block and take a while, which is why retrying bin local helps.

yinan1048576: How about keeping the logic as is but trying a better way to illustrate the code logic? For example, how about the following strategy:
- Name variables *literally* stating their meaning,
- Define multiple variables each for a single meaning, and
- Put everything else into comments.

For example, one scheme I can think of is the following:
```
bool tried_arena_slab_alloc = false;
do {
    bool should_try_arena_slab_alloc = false;
    /* ... lock a bin ... */
    for (bool got_some_memory = false; filled < nfill; filled += cnt, got_some_memory = true) {
        /* ... try increasingly harder to find some available memory ... */
        /* ... and if we got nothing ... */
        if (!tried_arena_slab_alloc || got_some_memory) {
            /* ... explain ... */
            should_try_arena_slab_alloc = true;
            break;
        }
        /* ... handle OOM ... */
        break;
    }
    /* ... unlock the bin ... */
    if (!should_try_arena_slab_alloc) {
        break;
    }
    /* ... try arena_slab_alloc() ... */
    tried_arena_slab_alloc = true;
} while (true);
```
Essentially I'm using `should_try_arena_slab_alloc` for `!terminate`, and breaking `cur_iter_alloc` into the two separate components it was trying to capture, `tried_arena_slab_alloc` and `got_some_memory`. The specific wordings may not be the best, but I think it's along the right direction to make the logic more interpretable. Thoughts?

interwq: Yeah let me try refactoring. One correction though, we may actually need slab_alloc multiple times. For example if a refill needs more items than a single slab holds; or if slab_alloc failed first but bin local resource became available (in which case, slab_alloc should be retried after exhausting bin resources, same as mention in my last comment). So the cur_iter_alloc actually means, was slab_alloc attempted very recently (will think about a better name as well).

yinan1048576: > So the cur_iter_alloc actually means, was slab_alloc attempted very recently

Right, but this "very recently" actually meant two possibilities:
- Directly at the end of the `do`-`while` loop;
- Indirectly afterwards.

Which was the reason why I thought it'd be easier to explicitly break it.

interwq: I'm not following here. The "just_allocated" is true only if this current `for` iteration break out and does slab_alloc, and is false for the iterations afterwards. What do you mean by "Indirectly afterwards"?

interwq: The problem with the code segment above is, `tried_arena_slab_alloc` is set to true and never reset, which means the slab_alloc can only be reached once.

interwq: Discussed offline. I misunderstood the `got_some_memory` part. I think what Yinan suggested makes sense. Just updated.

interwq:
@yinan1048576 -- can you help to take a look again? If you agree I'll start adding comments.

interwq:
Added another version in https://github.com/jemalloc/jemalloc/pull/1763/commits/8b23bfc17f27706e917f4b1aeac5d1fdfe20b0f4

I think I slightly prefer this one (relying on solely `made_progress`).

src/arena.c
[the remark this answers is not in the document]
interwq: Did some trail, but realized that this is more coupled with the arena. If we move it to tcache.c, we'd have to expose at least 4 current internal functions: arena_slab_reg_alloc_batch, arena_bin_refill_slabcur_no_fresh_slab, arena_bin_refill_slabcur_with_fresh_slab, and arena_slab_alloc. I'll say let's keep it here for now, until there is more motivation.

src/arena.c
yinan1048576: Now looks great!

interwq:
@yinan1048576 : need your help to take a look again. Had some more refactoring -- mainly replaced the external do-while loop with a goto; and comments.
https://github.com/jemalloc/jemalloc/pull/1763/commits/088a260f59f981db8bf6609779e43d5ac11a497b
https://github.com/jemalloc/jemalloc/pull/1763/commits/d7b597853e3b348f70e8ccf11eeda55e81254f8b

src/tcache.c
```diff
@@ -165,7 +165,7 @@ tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, cache_bin_t *tbin,
 
 	bool merged_stats = false;
 	unsigned dalloc_count = 0;
-	VARIABLE_ARRAY(edata_t *, dalloc_slabs, nflush);
+	VARIABLE_ARRAY(edata_t *, dalloc_slabs, nflush + 1);
```
yinan1048576: Why `nflush + 1`?

src/arena.c
```diff
@@ -1369,63 +1321,139 @@ arena_bin_choose_lock(tsdn_t *tsdn, arena_t *arena, szind_t binind,
...
+	 */
+	bool made_progress = true;
+	edata_t *fresh_slab = NULL;
+	bool alloc_and_retry = false;
+	unsigned filled = 0;
+
+	bin_t *bin;
 	unsigned binshard;
-	bin_t *bin = arena_bin_choose_lock(tsdn, arena, binind, &binshard);
+label_refill:
```
yinan1048576: What is the rationale for having `goto` and `label_refill`, instead of a loop?

src/tcache.c
interwq: It's not really necessary but to avoid an UB warning. See https://github.com/jemalloc/jemalloc/issues/1767.

src/arena.c
interwq: Good question. I'd say the main purpose, is to solve the indent / 80-char length crisis of this function (it's now one less level indented).

PR #1766, opened 2020-02-13
Fix Undefined Behavior in hash.h

krytarowski:
hash.h:200:27, left shift of 250 by 24 places cannot be represented in type 'int'


interwq:
Thanks for the patch!

PR #1771, opened 2020-02-21
Cirrus-CI: test on new freebsd releases.

interwq:


interwq:
It looks 11.3 is not available yet. For now only keep 12.1 then.

PR #1772, opened 2020-02-24
Tcache: Unify bin flush logic.

davidtgoldblatt:
The small and large pathways share most of their big-picture logic, even if some of the specifics are different (e.g. both have a trick to merge the stats locking with object locking if possible, even though the specific stats kept for small and large allocations are different). Unifying them makes the similarities and differences clearer.

While we're here, beef up the comments.

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -139,21 +139,44 @@ tbin_edatas_lookup_size_check(tsd_t *tsd, cache_bin_t *tbin, szind_t binind,
 	}
 }
 
-void
-tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, cache_bin_t *tbin,
-    szind_t binind, unsigned rem) {
-	assert(binind < SC_NBINS);
+JEMALLOC_ALWAYS_INLINE bool
+tcache_bin_flush_match(edata_t *edata, unsigned cur_arena_ind,
+    unsigned cur_binshard, bool small) {
+	if (small) {
+		return edata_arena_ind_get(edata) == cur_arena_ind
+		    && edata_binshard_get(edata) == cur_binshard;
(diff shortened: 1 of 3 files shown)


src/tcache.c
```diff
@@ -139,21 +139,44 @@ tbin_edatas_lookup_size_check(tsd_t *tsd, cache_bin_t *tbin, szind_t binind,
...
+	arena_t *tcache_arena = tcache->arena;
+	assert(tcache_arena != NULL);
 
-	arena_t *arena = tcache->arena;
-	assert(arena != NULL);
 	unsigned nflush = ncached - rem;
-	/* Variable length array must have > 0 length. */
+	/*
+	 * Variable length array must have > 0 length; the last element is never
+	 * touched.
```
yinan1048576: May be helpful to briefly explain why the last element is never touched but we still reserve space for it.

src/tcache.c
```diff
@@ -165,173 +188,154 @@ tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, cache_bin_t *tbin,
 		}
 	}
 
-	bool merged_stats = false;
+	/*
+	 * The slabs where we freed the last remaining object in the slab (and
+	 * so need to free the slab itself.
```
yinan1048576: nit: right parenthesis.

yinan1048576:
... and still haven't figured out why MSVC doesn't like `JEMALLOC_ALWAYS_INLINE`.

davidtgoldblatt:
Made the suggested changed. The MSVC breakage wasn't an always-inline thing; it was for the much funnier reason that windows in some build modes will `#define small char`. I put up a version that just `#undef`s small in the prelude.

PR #1776, opened 2020-03-03
Correction on geometric sampling

yinan1048576:
This is a minor correction, mainly for the purpose of being rigorous. The sampling result should mostly be unaffected. The couple of things I'm correcting:
- In the previous logic `u` is in theory in the range of `[0, 1)`. An exact `0` would not be valid for `log(u)` to work. So I'm shifting it to `(0, 1]`, by changing an exact `r = 0` to an exact `u = 1`. (Directly changing `r = 0` to `r = 2^53` would be risky because the floating point computation might result in something slightly above `1`, which would make `log(u)` positive.)
- Changed the comment so that `bytes_until_sample` is computed as 1 plus the floor, rather than the ceiling. This is also what we're actually computing. It is also the correct approach if `u` is in the range of `(0, 1]`; directly taking the ceiling would generate a geometric sample of `0` when `u` is exactly `1`.

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -468,9 +468,16 @@ prof_sample_threshold_update(tsd_t *tsd) {
 	 *   Springer-Verlag, New York, 1986
 	 *   pp 500
 	 *   (http://luc.devroye.org/rnbookindex.html)
+	 *
+	 * In the actual computation, there's a non-zero probability that our
+	 * pseudo random number generator generates an exact 0, and to avoid
+	 * log(0), we set u to 1.0 in case r is 0.  Therefore u effectively is
+	 * uniformly distributed in (0, 1] instead of [0, 1).  Further, rather
+	 * than taking the ceiling, we take the floor and then add 1, since
+	 * otherwise bytes_until_sample would be 0 if u is exactly 1.0.
 	 */
 	uint64_t r = prng_lg_range_u64(tsd_prng_statep_get(tsd), 53);
-	double u = (double)r * (1.0/9007199254740992.0L);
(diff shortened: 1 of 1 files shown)


src/prof.c
```diff
@@ -455,22 +455,25 @@ prof_sample_threshold_update(tsd_t *tsd) {
 	 * Compute sample interval as a geometrically distributed random
 	 * variable with mean (2^lg_prof_sample).
 	 *
-	 *                      __        __
-	 *                      |  log(u)  |                     1
-	 * bytes_until_sample = | -------- |, where p = ---------------
-	 *                      | log(1-p) |             lg_prof_sample
-	 *                                              2
+	 *                          |          |
+	 *                          |  log(u)  |                  1
```
davidtgoldblatt: I think this is a little cleaner if we leave the math as is, and mention the zero->1 correction for the prng-output-is-zero case explicitly. Right now our logic seems sort of unmotivated if you only read the section from the book cited.

davidtgoldblatt: Mention it in english down below I mean. E.g. "We actually apply a slight tweak to this; to handle the case in which the prng call outputs 0, we translate those values to be 1, effectively converting the distribution from [0, 1) to (0, 1]".

yinan1048576: Makes sense. Added explanations rather than tweaking the math, though I think the ideal way would be to explain in / change the book: the formula in the book is only correct if `u` is truly continuous, in which case `prob(u = 0) = prob(u = 1) = 0`; practitioners failing to pay attention to either one of the two ends would be in trouble.

yinan1048576:
Merge ignoring the FreeBSD installation failure.

Issue #1751, opened 2020-01-29
Change randomization mechanism for profiling sample interval


yinan1048576:
Here's our current approach for determining the next profiling sample interval:

https://github.com/jemalloc/jemalloc/blob/ea351a7b52430de88007bf16f354a132da311c5b/src/prof.c#L438-L476

According to the comment, we draw a random number from a geometric distribution with `p = 1/2^lg_prof_sample`. If we denote the random number as `I`, then the expectation is `E(I) = 1/p = 2^lg_prof_sample`, so on average, we sample every `2^lg_prof_sample` bytes.

The randomization is great for the purpose of avoiding any undesired synchronization between sampling and application's allocation pattern. However, I find that this particular randomization mechanism has a couple of limitations.

### Limitations

- Sampling guarantee

One key design principle of this byte-based sampling approach, in contrast to a frequency-based sampling approach, is: we want to catch all the large allocations. In other words, we want to make a guarantee in the form of "if an allocation is `K` bytes or larger, then we always sample it". However, we cannot make such a guarantee no matter how large `K` is, even though we often thought that we can use `2^lg_prof_sample` for `K`.

- Distribution skewness

The probability mass function of a geometric distribution is strictly decreasing. For example, let `M = E(I) = 2^lg_prof_sample`, then the probabilities of having `I` falling in `[0, 0.5*M]`, `[0.5*M, M]`, `[M, 1.5*M]`, `[1.5*M, 2*M]` and `[2*M, ∞]` are 39%, 24%, 14%, 9% and 14%, respectively. I find these numbers disturbing, especially the part that there's a 40% chance that the sampling interval is below half of `2^lg_prof_sample`; applications would likely expect that the distribution be relatively more centered on `2^lg_prof_sample`.

- Sampling ratio estimation

We've seen needs for estimating the overall bytes allocated per stack trace from the sampled bytes, and I find it very hard. An example is our own leak checking facility:

https://github.com/jemalloc/jemalloc/blob/ea351a7b52430de88007bf16f354a132da311c5b/src/prof_data.c#L999-L1028

The computation relies on the claim that the probability of an allocation of size `X` being sampled is `1 - exp(-X/M)` (where `M = 2^lg_prof_sample`). I don't know how exactly we derived this. A likely source can be:

https://github.com/google/pprof/blob/be90f332299a1d03fbb09bdbebdf28118ae300fb/profile/legacy_profile.go#L650-L674

[...]

PR #1777, opened 2020-03-05
Direct allocation attribution in profiling

yinan1048576:
This change is not completed yet, but I think it has reached a stage suitable for seeking some early feedback.

The context is in #1751. To enable correct inference, we need to record both the number of sampled bytes and the number of unsampled bytes for each stack trace. The number of sampled bytes is equivalent to `curobjs`, so we only need to record the number of unsampled bytes, which I denote as `surplus` - the amount beyond what's needed for triggering sampling.

`surplus` is relayed in two routes:
- `thread_alloc_event()` -> `prof_malloc_sample_object()`, via a new thread local field `prof_sample_event_surplus`, which is needed because the sampling logic is currently separate from the thread event logic;
- `prof_malloc_sample_object()` -> `prof_free_sampled_object()`, via a new field `e_prof_surplus` on `edata`, which is needed because `free()` needs to roll back the allocation's `surplus` from the stack trace total.

My remaining work involves the following:
- Add unit tests.
- Add an option to actually use the additional information i.e. the `surplus`. (I figured that computing the `surplus` is fairly cheap so I don't really need to guard the computation under any option.)
  - Print the raw `surplus` value as well as some other convenient estimation(s) to the profiling dump.
  - Compute leak checking differently.

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -104,11 +104,11 @@ arena_prof_tctx_reset_sampled(tsd_t *tsd, const void *ptr) {
 }
 
 JEMALLOC_ALWAYS_INLINE void
-arena_prof_info_set(edata_t *edata, prof_tctx_t *tctx) {
+arena_prof_info_set(edata_t *edata, prof_tctx_t *tctx, size_t surplus) {
 	cassert(config_prof);
 
 	assert(!edata_slab_get(edata));
-	large_prof_info_set(edata, tctx);
+	large_prof_info_set(edata, tctx, surplus);
 }
 
 JEMALLOC_ALWAYS_INLINE void
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -104,7 +104,7 @@ arena_prof_tctx_reset_sampled(tsd_t *tsd, const void *ptr) {
 }
 
 JEMALLOC_ALWAYS_INLINE void
-arena_prof_info_set(tsd_t *tsd, edata_t *edata, prof_tctx_t *tctx) {
+arena_prof_info_set(edata_t *edata, prof_tctx_t *tctx) {
 	cassert(config_prof);
 
 	assert(!edata_slab_get(edata));
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -226,12 +226,56 @@ te_ctx_get(tsd_t *tsd, te_ctx_t *ctx, bool is_alloc) {
 	}
 }
 
+/*
+ * The lookahead functionality facilitates events to be able to lookahead, i.e.
+ * without touching the event counters, to determine whether an event would be
+ * triggered.  The event counters are not advanced until the end of the
+ * allocation / deallocation calls, so the lookahead can be useful if some
+ * preparation work for some event must be done early in the allocation /
+ * deallocation calls.
+ *
+ * Currently only the profiling sampling event needs the lookahead
+ * functionality, so we don't yet define general purpose lookahead functions.
+ *
+ * Surplus is a terminology referring to the amount of bytes beyond what's
+ * needed for triggering an event, which can be a useful quantity to have in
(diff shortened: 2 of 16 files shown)


yinan1048576:
Rebase, and rewrite. The main change in the rewrite is that I'm not creating the new thread local field any more, thanks to #1779.

include/jemalloc/internal/thread_event.h
```diff
@@ -217,12 +217,44 @@ te_ctx_get(tsd_t *tsd, te_ctx_t *ctx, bool is_alloc) {
...
+
+JEMALLOC_ALWAYS_INLINE bool
+te_prof_sample_event_lookahead_surplus(tsd_t *tsd, size_t usize,
+    size_t *surplus) {
+	/* The subtraction is intentionally susceptible to underflow. */
+	size_t accumbytes = tsd_thread_allocated_get(tsd) + usize -
+	    tsd_thread_allocated_last_event_get(tsd);
+	size_t sample_wait = tsd_prof_sample_event_wait_get(tsd);
+	if (accumbytes < sample_wait) {
+		/* Don't bother to set surplus - it will never be read. */
```
yinan1048576: This may trigger some warning. Let's see...

yinan1048576: As expected. Adding an initializer.

yinan1048576:
Fix compiler warnings, as well as one other place revealed by the FreeBSD unit tests: `te_prof_sample_event_lookahead_surplus()` would better be short-circuited in edge cases e.g. non-nominal TSD / reentrancy.

yinan1048576:
Rebase on top of #1787 to get rid of the warning.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -101,11 +101,11 @@ arena_prof_tctx_reset_sampled(tsd_t *tsd, const void *ptr) {
 }
 
 JEMALLOC_ALWAYS_INLINE void
-arena_prof_info_set(edata_t *edata, prof_tctx_t *tctx) {
+arena_prof_info_set(edata_t *edata, prof_tctx_t *tctx, uint64_t surplus) {
```
davidtgoldblatt: Surplus should be a size_t, no?

include/jemalloc/internal/thread_event.h
```diff
@@ -217,12 +217,56 @@ te_ctx_get(tsd_t *tsd, te_ctx_t *ctx, bool is_alloc) {
...
+    size_t *surplus) {
+	if (surplus != NULL) {
+		/*
+		 * This is a dead store: the surplus will be overwritten before
+		 * any read.  The initialization suppresses compiler warnings.
+		 * Meanwhile, using usize to initialize is good for debugging
+		 * purpose, because a valid surplus value is strictly less than
+		 * usize.
+		 */
+		*surplus = usize;
```
davidtgoldblatt: If I might suggest: I'm not sure I would spot that `*surplus == usize` is invalid while debugging. Some other garbage constant (e.g. `0x99999999` or `0xdeadbeef` or something) OTOH, I would.

yinan1048576: Makes sense. Use `SIZE_MAX` instead. (Strictly speaking `0x99999999` and `0xdeadbeef` can be valid `surplus` values.)

yinan1048576:
Revise according to feedback.

yinan1048576:
Rebase, and added another field `accumsurplus`. It's analogous to `cursurplus`, but records the accumulative surplus, and is computed whenever `opt_prof_accum` is on.

My next step is to define an opt-in run-time option and, when it is turned on, to ingest the surplus values in the dump content.

yinan1048576:
Rebase.

yinan1048576:
Rebase, and added:
- Unit tests.
- An option to print total surplus rather than total bytes in prof dumps. Users don't need the total bytes and can derive whatever inference they need from
  - the number of sampled objects,
  - total surplus, and
  - `lg_prof_sample`.

yinan1048576:
Closing. #1897 serves to correct the bias. Will reopen if there's any need for inference, in addition to a single point estimate.

PR #1779, opened 2020-03-06
Rewrite profiling thread event

yinan1048576:
There are a few subtle things here:
- The main change is to always reset the sampling threshold in `prof_alloc_rollback()`. Previously, we only reset when `updated == true`. I think we should also reset when `updated == false`, which is actually more important - otherwise the next allocation would always be sampled because the threshold would still be observed as having been hit. If my reasoning is correct, I feel that the `updated` flag is redundant (in the entire profiling logic), and I'll try to add another commit to remove it.
- A related fix in `xallocx()`. The issue is revealed in `test_xallocx_overflow` in `test/unit/safety_check`. The issue was hidden before my main change in `prof_alloc_rollback()`: when an `xallocx()` does not resize, the next `malloc()` is always sampled, which happens to be what we want in the unit test, only except that the sampling was triggered by the last allocation, rather than by itself. Now `prof_alloc_rollback()` correctly resets the threshold, but a later `te_alloc_rollback()` call further extends the wait time so that the next `malloc()` (of the same size) cannot hit the threshold any more. Extending the wait time for events not yet triggered is the right thing to do in `te_alloc_rollback()`; for events already been hit, `te_alloc_rollback()` shouldn't attempt anything, because it's not equipped with the ability to revert the events. The right way is to call `te_alloc_rollback()` first and then call the rollback functions for each triggered event. To make that happen for `xallocx()`, I also slightly restructured the logic. I think the ideal way is to call the event rollback functions within `te_alloc_rollback()`; I'll try to have another commit later to do that.
- Another separate change, which I put as an individual commit for clarity: I don't think `prof_alloc_rollback()` should create a `tdata` if the `tdata` hasn't been created yet. I think `tdata` is not created only if `prof_alloc_rollback()` is not really rolling back a sampled allocation, and in such a case it'd just better do nothing and exit.

### DIFF
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -220,6 +220,7 @@ te_ctx_get(tsd_t *tsd, te_ctx_t *ctx, bool is_alloc) {
 
 JEMALLOC_ALWAYS_INLINE bool
 te_prof_sample_event_lookahead(tsd_t *tsd, size_t usize) {
+	assert(usize == sz_s2u(usize));
 	return tsd_thread_allocated_get(tsd) + usize -
 	    tsd_thread_allocated_last_event_get(tsd) >=
 	    tsd_prof_sample_event_wait_get(tsd);
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -218,6 +218,13 @@ te_ctx_get(tsd_t *tsd, te_ctx_t *ctx, bool is_alloc) {
 	}
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+te_prof_sample_event_lookahead(tsd_t *tsd, size_t usize) {
+	return tsd_thread_allocated_get(tsd) + usize -
+	    tsd_thread_allocated_last_event_get(tsd) >=
+	    tsd_prof_sample_event_wait_get(tsd);
+}
+
 JEMALLOC_ALWAYS_INLINE void
 te_event_advance(tsd_t *tsd, size_t usize, bool is_alloc) {
 	te_assert_invariants(tsd);
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -33,7 +33,6 @@ typedef struct te_ctx_s {
 
 void te_assert_invariants_debug(tsd_t *tsd);
 void te_event_trigger(tsd_t *tsd, te_ctx_t *ctx, bool delay_event);
-void te_alloc_rollback(tsd_t *tsd, size_t diff);
 void te_event_update(tsd_t *tsd, bool alloc_event);
 void te_recompute_fast_threshold(tsd_t *tsd);
 void tsd_te_init(tsd_t *tsd);
--- a/include/jemalloc/internal/prof_externs.h
+++ b/include/jemalloc/internal/prof_externs.h
@@ -53,7 +53,7 @@ void prof_idump_rollback_impl(tsdn_t *tsdn, size_t usize);
 prof_tdata_t *prof_tdata_init(tsd_t *tsd);
 prof_tdata_t *prof_tdata_reinit(tsd_t *tsd, prof_tdata_t *tdata);
 
-void prof_alloc_rollback(tsd_t *tsd, prof_tctx_t *tctx, bool updated);
+void prof_alloc_rollback(tsd_t *tsd, prof_tctx_t *tctx);
 void prof_malloc_sample_object(tsd_t *tsd, const void *ptr, size_t size,
     size_t usize, prof_tctx_t *tctx);
 void prof_free_sampled_object(tsd_t *tsd, size_t usize, prof_info_t *prof_info);
--- a/include/jemalloc/internal/prof_inlines_b.h
+++ b/include/jemalloc/internal/prof_inlines_b.h
@@ -85,11 +85,11 @@ prof_info_set(tsd_t *tsd, edata_t *edata, prof_tctx_t *tctx) {
 }
 
 JEMALLOC_ALWAYS_INLINE bool
-prof_sample_should_skip(tsd_t *tsd, size_t usize) {
+prof_sample_should_skip(tsd_t *tsd, bool sample_event) {
 	cassert(config_prof);
 
 	/* Fastpath: no need to load tdata */
-	if (likely(!te_prof_sample_event_lookahead(tsd, usize))) {
+	if (likely(!sample_event)) {
 		return true;
 	}
 
@@ -106,12 +106,11 @@ prof_sample_should_skip(tsd_t *tsd, size_t usize) {
 }
 
 JEMALLOC_ALWAYS_INLINE prof_tctx_t *
-prof_alloc_prep(tsd_t *tsd, size_t usize, bool prof_active) {
+prof_alloc_prep(tsd_t *tsd, bool prof_active, bool sample_event) {
 	prof_tctx_t *ret;
 
-	assert(usize == sz_s2u(usize));
-
-	if (!prof_active || likely(prof_sample_should_skip(tsd, usize))) {
+	if (!prof_active ||
+	    likely(prof_sample_should_skip(tsd, sample_event))) {
 		ret = (prof_tctx_t *)(uintptr_t)1U;
 	} else {
 		ret = prof_tctx_create(tsd);
@@ -137,7 +136,7 @@ prof_malloc(tsd_t *tsd, const void *ptr, size_t size, size_t usize,
 JEMALLOC_ALWAYS_INLINE void
 prof_realloc(tsd_t *tsd, const void *ptr, size_t size, size_t usize,
     prof_tctx_t *tctx, bool prof_active, const void *old_ptr, size_t old_usize,
-    prof_info_t *old_prof_info) {
+    prof_info_t *old_prof_info, bool sample_event) {
 	bool sampled, old_sampled, moved;
 
 	cassert(config_prof);
@@ -145,7 +144,7 @@ prof_realloc(tsd_t *tsd, const void *ptr, size_t size, size_t usize,
 
 	if (prof_active && ptr != NULL) {
 		assert(usize == isalloc(tsd_tsdn(tsd), ptr));
-		if (prof_sample_should_skip(tsd, usize)) {
+		if (prof_sample_should_skip(tsd, sample_event)) {
 			/*
 			 * Don't sample.  The usize passed to prof_alloc_prep()
 			 * was larger than what actually got allocated, so a
--- a/include/jemalloc/internal/prof_inlines_b.h
+++ b/include/jemalloc/internal/prof_inlines_b.h
@@ -85,11 +85,11 @@ prof_info_set(tsd_t *tsd, edata_t *edata, prof_tctx_t *tctx) {
 }
 
 JEMALLOC_ALWAYS_INLINE bool
-prof_sample_accum_update(tsd_t *tsd, size_t usize, bool update) {
+prof_sample_should_skip(tsd_t *tsd, size_t usize) {
 	cassert(config_prof);
 
 	/* Fastpath: no need to load tdata */
-	if (likely(prof_sample_event_wait_get(tsd) > 0)) {
+	if (likely(!te_prof_sample_event_lookahead(tsd, usize))) {
 		return true;
 	}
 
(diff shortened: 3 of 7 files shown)


yinan1048576:
Discussed offline. The more correct way to fix the rollback issue is to obtain the original wait time rather than drawing a new wait time. This is better handled with the help of #1777. I'll close this PR and include the (correct) fix there.

yinan1048576:
Reopening. Here's what's going on -

This PR started with my effort trying to fix an issue in `prof_alloc_rollback()`. The issue is: if `!updated` (e.g. after a OOM'ed `realloc()`/`rallocx()` or a non-resized `xallocx()`), the profiling wait time would stay at `0` and the next event threshold would be set as according to other events. The next event threshold could either be too long or too short, relative to the next profiling wait time (had we set it): if it's too long, then we're losing samples; if it's too short, then whichever event coming next prematurely triggers profiling.

The idea behind my earlier solution was to draw another profiling wait time sample in `prof_alloc_rollback()`. This was actually right on expectation: `prof_alloc_rollback()` is called regardless of whether the allocation was sampled or not, so a new geometric sample is equivalent to the previous wait time, both on expectation and in full distribution.

This correctness is not trivial, and the more obvious way to guarantee correctness is to just restore the original wait time. I first thought that the solution was to retain a copy of the original wait time before it's cleared, and then grab the copy in `prof_alloc_rollback()`; in such a case the new thread local `surplus` field being introduced in #1777 would be handy. After a while I realized that it's not necessary - the best way to restore the original wait time is not to touch it at all.

So I've got a larger scale redesign around profiling event. Now the logic looks like the following:
- Without touching any thread event counter, look ahead to determine whether the profiling event would be triggered for the current allocation. The size used for the look-ahead is the `usize` in case of `malloc()`, and is the best guess in case of `realloc()` / `rallocx()` / `xallocx()`. If we end up anticipating a profiling event, try to get a `tctx`.
- Allocate.
- If OOM'ed, call `prof_alloc_rollback()` to cleanup the `tctx`, if there is one. `prof_alloc_rollback()` doesn't touch any thread event counter.
- For `realloc()` / `rallocx()` / `xallocx()`, correct the best guess with the real `usize` of the allocation.
- Call `prof_malloc()`, in case of `malloc()`, or `prof_realloc()`, in case of `realloc()` / `rallocx()` / `xallocx()`. `prof_malloc()` is straightforward since there has been no correction. `prof_realloc()` needs to look ahead again with the corrected `usize`, and would need to call `prof_alloc_rollback()` to clean up the `tctx` if the best guess triggered sampling but the corrected `usize` does not. Note that if the best guess didn't trigger sampling but the corrected `usize` does, there's nothing we can do.
- Advance the thread event counters (right before the allocation call completes).

A few remarks about the look-ahead:
- It does not appear on `malloc()` fast path, where the usual thread event counter trick still works as before.
- It may be generalized to other events, and we may have a general thread event look-ahead functionality for all events. At this moment only the profiling sampling event needs to look ahead, so I don't do any generalization yet, and may consider doing so later if there's a stronger need.

The benefits of the new logic are:
- Thread event module: there's no longer the need for a rollback functionality. The interface to the caller is much cleaner: only a "look-ahead" plus a final "commit".
- Profiling module: there's no longer the back-and-forth adjustment on the profiling wait time, and thus both `prof_alloc_prep()` and `prof_alloc_rollback()` got rid of the confusing boolean flag.
- Profiling event: previously the event triggering was in the thread event module, but the event update was in the profiling module, making the logic very error prone; now both take place in thread event (at the "commit" step), and within a single function, which is finally as easy as how the other types of events are handled.
- I should also be able to simplify the design for #1777. In particular, I don't think there's the need for the new thread local `surplus` field. Will do some revisions on top of this PR.

davidtgoldblatt:
I haven't looked at the code yet, but one question as I do:

> Note that if the best guess didn't trigger sampling but the corrected usize does, there's nothing we can do.

Is this possible? The corrected usize is always less than or equal to the guessed usize I would think. For realloc and rallocx, the guessed usize is always correct. For xallocx, the guess is size + extra, and the returned value is in between size and size + extra. So it would seem like we can only ever err in the direction of allocating a profiling context unnecessarily, not ever in the direction of failing to sample when we should have.

include/jemalloc/internal/thread_event.h
```diff
@@ -218,6 +218,13 @@ te_ctx_get(tsd_t *tsd, te_ctx_t *ctx, bool is_alloc) {
 	}
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+te_prof_sample_event_lookahead(tsd_t *tsd, size_t usize) {
```
davidtgoldblatt: I think with a little bit of refactoring, we can get rid of this function: when the thread event module calls prof_alloc_prep, in addition to usize, it can also pass `tsd_thread_allocated_get(tsd) - tsd_thread_allocated_last_event_get(tsd)` (say, in a parameter named `bytes_allocated_since_last_event` or something).

That lets us get rid of the loop where thread event calls prof calls thread event; we sort of decouple how tightly tied together they are. (I also think the resulting condition is a little cleaner; `if (usize + bytes_since_last_event >= tsd_prof_sample_event_wait_get(tsd))` reads a little bit more obviously to me as to what the intent is, although I guess that's debatable).

yinan1048576:
> > Note that if the best guess didn't trigger sampling but the corrected usize does, there's nothing we can do.

> Is this possible? The corrected usize is always less than or equal to the guessed usize I would think. For realloc and rallocx, the guessed usize is always correct. For xallocx, the guess is size + extra, and the returned value is in between size and size + extra. So it would seem like we can only ever err in the direction of allocating a profiling context unnecessarily, not ever in the direction of failing to sample when we should have.

To be honest, I wasn't sure either. I put that statement more for starting a discussion. :)

I didn't drill all the way down. I tried to read the code again. Still haven't understood everything, but here's what I saw:

For `rallocx()` / `realloc()`: I think we explicitly passed in `&usize` for a reason. What I can see is:
https://github.com/jemalloc/jemalloc/blob/305b1f6d962c5b5a76b7ddb4b55b14d88bada9ba/src/jemalloc.c#L3093-L3103

My understanding is that the actual `usize` may be smaller than the original `usize` passed to the internal `iralloct()`, if there is enough space at the original location and the alignment luckily satisfies the alignment requirement? If this is the only chance for the actual `usize` being different, then I think you're still right regarding the erring direction.

For `xallocx()`: I remember this is something I bumped into long ago - it's possible that `size + extra` is smaller than `old_usize`, and the internal `ixalloc()` may just return the original allocation, and in such a acase `usize` is larger than `usize_max` (which is confusing given the naming). If this is the only chance for the actual `usize` being larger than `usize_max`, then I think you're still right regarding the erring direction, in some sense: I'm only calling `prof_realloc()` when `usize != old_usize`, and whenever `usize != old_usize`, it's guaranteed that `usize <= usize_max`.

I may have missed some critical cases. If so, let me know; otherwise, I'll add some assertions right before the `prof_realloc()` calls, to verify / make the code look more clearly.

include/jemalloc/internal/thread_event.h
yinan1048576: I think there are two things here:

First, there's only `prof_alloc_prep()` -> te look-ahead, rather than te -> `prof_alloc_prep()` -> te. The way of my current design is: te only provides a single write API, the commit, which should always be called at the end and with the right amount of increment; before that, there is some handy event-specific read-only look-ahead API, if the caller is in need (not if te is in need).

Second, in case you're talking about the logic loop, not actual function-call loop, then you're right that te depends on prof and prof also depends on te (the look-ahead), on the module level. But I don't see an easy way to chop either dependency. The te -> prof dependency is by design how te works. The prof -> te look-ahead dependency is also how the te look-ahead works. It's not obvious to me why passing down `bytes_since_last_event` could cut the dependency; the caller just ends up trying to get this `bytes_since_last_event` every time before invoking `prof_alloc_prep()`, which feels in need of a code consolidation.

Thoughts?

davidtgoldblatt: It's the logic loop that's the concern. The ideal end stat of module dependencies is a DAG.

Stuff in TSD is effectively global.
```
data = f();
g(data);
```
has some amount of extra code relative to:
```
f();
g(); // g and f are secretly communicating through global data
```
But it's sort of a "good" redundancy in that it's really just making explicit what would otherwise be a hidden communication channel. If the computation code is the issue, we can always pull out the `bytes_since_last_event` computation into a helper function.

davidtgoldblatt: Or put more succinctly: functional purity is a really useful property for testing and reasoning about our internals. We shouldn't give it up without a really big benefit.

yinan1048576: Makes sense. Made the revision as a separate commit, since it's kind of a refactoring of the refactoring.

Rather than using `bytes_since_last_event` to communicate, I'm still encapsulating the look-ahead inside of the thread event module, and let the caller directly get a boolean `sample_event` and then use it to communicate with prof. I think separating out the `tsd_thread_allocated_get(tsd) - tsd_thread_allocated_last_event_get(tsd)` part and the `if (usize + bytes_since_last_event >= tsd_prof_sample_event_wait_get(tsd))` part is unnecessary; they are thread event specific computations and would better be kept inside of the thread event module.

After doing the DAG conversion, I actually figured that the outcome looks more desired despite the extra function call, due to the following:
- I realized that `prof_alloc_prep()`'s only dependency on `usize` is for the look-ahead. So rather than passing `usize` all the way in, I'm now only passing the `sample_event` flag.
- The caller is now more explicit in obtaining the `sample_event` flag, in that whenever there's a possible `usize` correction, the caller explicitly invokes the look-ahead again (and pass the result to `prof_realloc()`).
- I realized that I'd need to relocate the look-ahead call to the caller anyway in #1777 - at that time I'd need the `surplus` value, and the most convenient way to obtain it is through the look-ahead; if the look-ahead call is buried deep down in `prof_alloc_prep()`, then `surplus` needs to be surfaced all the way back so that `prof_malloc()` / `prof_realloc()` can use it. This is a good example of the practical benefit of the DAG design.

yinan1048576:
Rebase, and revise according to feedback.

I think we're most likely right about our claim that the `usize` could only become smaller in case of a correction in `rallocx()` / `realloc()` / `xallocx()` before passing to `prof_realloc()`. I'm therefore adding assertions to verify that. If any of them breaks, then there's something interesting happening.

Also, I realized that my earlier claim of
> Note that if the best guess didn't trigger sampling but the corrected usize does, there's nothing we can do.

is nonsense even if there's chance the condition is true - we surely are able to get a `tctx` in `prof_realloc()` if we want to.

yinan1048576:
Two additional remarks -

The first remark is a behavioral change as a consequence of this PR: before this PR, if a sampled allocation also triggers interval-based dumping, the dump won't contain this last allocation. I couldn't find any relevant issue reported on GitHub. Perhaps it was internally reported somewhere a while ago. But, it's a bug, and this PR fixes it, because now dumping is not called until the final "commit", by the time of which the sample has already been recorded. To verify, I just tested locally and confirmed that the dump now includes the last sample.

The second remark is that I realized that the following I said

> Also, I realized that my earlier claim of

> > Note that if the best guess didn't trigger sampling but the corrected usize does, there's nothing we can do.

> is nonsense even if there's chance the condition is true - we surely are able to get a `tctx` in `prof_realloc()` if we want to.

is not correct. Even though it's feasible to get a `tctx` in `prof_realloc()`, we would not have an easy solution to handle a `tctx` OOM, i.e. if it turns out that we need to create a new `tctx` but the memory allocation for it fails.

I realize that we have always been carefully sticking to a profiling *contract* in the following form:

If the requested allocation size triggers the sampling threshold, then as long as jemalloc returns the allocation to the caller, it's guaranteed that the allocation has got its corresponding profiling record inside jemalloc; otherwise, jemalloc doesn't even try to allocate, and directly indicates a failure to the caller.

Even though the profiling threshold is probabilistic, this contract is essential in guaranteeing the correctness of the attribution estimation, and it applies to all jemalloc APIs involving an allocation, e.g. `malloc()` / `realloc()` / `xallocx()`. This is the very reason why we always call `prof_alloc_prep()` as a preparation *before* the actual allocation. Otherwise, it would have made this PR even cleaner - I would have just put the entire profiling logic under the "commit" call at the end of the allocations and there would be no need for the look-ahead.

So back to `prof_realloc()`: if we try to get a `tctx` there but it fails, then if we want to keep the contract, we must revert the allocation, which is not trivial to do.

Therefore, as long as we still want to keep the contract, we'd have to make sure that the `usize` passed to `prof_realloc()` never exceeds the `usize` passed to `prof_alloc_prep()`, so adding the assertions (as I did) is the right thing to do.

PR #1782, opened 2020-03-08
Edata cache and inspect fixes

davidtgoldblatt:
This adds:
- Unit tests for the edata_cache_t (which was previously tested only via the extent tests).
- A slight performance improvement to it (an atomic fetch-add which can just be a load-then-store).
- An unsynchronized edata_cache_small_t, which can amortize locking and allocations. (This is not actually used anywhere yet, but I've got several branches that all could use this and an early merge would be helpful from a workflow perspective).
- A minor fix to the inspect unit test (some of its assumptions stop being correct if profiling is enabled).

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -195,6 +195,7 @@ TESTS_UNIT := \
 	$(srcroot)test/unit/counter.c \
 	$(srcroot)test/unit/decay.c \
 	$(srcroot)test/unit/div.c \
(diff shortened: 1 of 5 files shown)


davidtgoldblatt:
I think probably the long term end state has *only* the unsynchronized small cache, and some backing synchronized_edata_cache_t; I think waiting until then to split things up then may save some effort.

PR #1784, opened 2020-03-09
Rewrite cache bin module

davidtgoldblatt:
This rewrites the cache bin module to track more of its low water, full, and empty states in the bin itself. This has a few advantages
- It makes the logic easier to reason about
- It unifies the 32-bit and 64-bit implementations
- It makes it easy for the fast paths to change the low-water mark if necessary. This results in around a 4% improvement in cycles-spent-in-malloc in prod workloads (for around a .1% end-to-end improvement).

This also pulls the maintenance logic out of the tcache and arena modules and into the cache_bin module, and documents it; the resulting code is a lot easier to read and maintain.

### DIFF
--- a/src/cache_bin.c
+++ b/src/cache_bin.c
@@ -102,3 +102,8 @@ cache_bin_init(cache_bin_t *bin, cache_bin_info_t *info, void *alloc,
 	assert(cache_bin_ncached_get(bin, info) == 0);
 	assert(cache_bin_empty_position_get(bin, info) == empty_position);
 }
+
+bool
+cache_bin_still_zero_initialized(cache_bin_t *bin) {
+	return bin->cur_ptr.ptr == NULL;
+}
--- a/src/cache_bin.c
+++ b/src/cache_bin.c
@@ -8,7 +8,7 @@ cache_bin_info_init(cache_bin_info_t *info,
     cache_bin_sz_t ncached_max) {
 	size_t stack_size = (size_t)ncached_max * sizeof(void *);
 	assert(stack_size < ((size_t)1 << (sizeof(cache_bin_sz_t) * 8)));
-	info->stack_size = (cache_bin_sz_t)stack_size;
+	info->ncached_max = (cache_bin_sz_t)ncached_max;
 }
 
 void
@@ -23,23 +23,14 @@ cache_bin_info_compute_alloc(cache_bin_info_t *infos, szind_t ninfos,
 	 */
 	*size = sizeof(void *) * 2;
 	for (szind_t i = 0; i < ninfos; i++) {
-		*size += infos[i].stack_size;
+		*size += infos[i].ncached_max * sizeof(void *);
 	}
 
 	/*
-	 * 1) Align to at least PAGE, to minimize the # of TLBs needed by the
+	 * Align to at least PAGE, to minimize the # of TLBs needed by the
 	 * smaller sizes; also helps if the larger sizes don't get used at all.
-	 * 2) On 32-bit the pointers won't be compressed; use minimal alignment.
 	 */
-	if (LG_SIZEOF_PTR < 3 || *size < PAGE) {
-		*alignment = PAGE;
-	} else {
-		/*
-		 * Align pow2 to avoid overflow the cache bin compressed
-		 * pointers.
-		 */
-		*alignment = pow2_ceil_zu(*size);
-	}
+	*alignment = PAGE;
 }
 
 void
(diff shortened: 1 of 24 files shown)


test/unit/cache_bin.c
```diff
@@ -112,10 +112,10 @@ TEST_BEGIN(test_cache_bin) {
...
 		/*
-		 * This should fail -- the reduced version can't change low
-		 * water.
+		 * This should fail -- the easy variant can't change the low
+		 * water mark.
 		 */
-		ptr = cache_bin_alloc_easy_reduced(&bin, &success);
+		ptr = cache_bin_alloc_easy(&bin, &success);
 		expect_ptr_null(ptr, "");
 		expect_false(success, "");
```
yinan1048576: nit: these seem like something we missed to fill in.

src/jemalloc.c
```diff
@@ -81,6 +81,17 @@ const char *zero_realloc_mode_names[] = {
...
 };
 
+static void default_junk_alloc(void *ptr, size_t usize) {
+	memset(ptr, 0xa5, usize);
+}
+
+static void default_junk_free(void *ptr, size_t usize) {
+	memset(ptr, 0x5a, usize);
+}
+
```
yinan1048576: Explicitly define the two junk constants?

src/jemalloc.c
```diff
@@ -2444,17 +2455,22 @@ je_malloc(size_t size) {
...
-			bin->tstats.nrequests++;
-		}
-
-		LOG("core.malloc.exit", "result: %p", ret);
-
-		/* Fastpath success */
+		fastpath_success_finish(tsd, allocated_after, bin, ret);
+		return ret;
+	}
+	ret = cache_bin_alloc(bin, &tcache_success);
```
yinan1048576: So basically this extra `cache_bin_alloc()` check is how you got the performance win (so that this "second-order" common case won't need to be routed to the slow path)? And an essential part of why it can work is that the extra check can now be accomplished without costing extra registers, which is credited to the restructuring? Otherwise, logic wise, the current computation is pretty much equivalent to what was the case before https://github.com/jemalloc/jemalloc/commit/05681e387a3202567ff95528dbc460e92e031a3c?

src/cache_bin.c
```diff
@@ -0,0 +1,97 @@
...
+
+void
+cache_bin_postincrement(cache_bin_info_t *infos, szind_t ninfos, void *alloc,
+    size_t *cur_offset) {
+	/* Note: a7 vs. 7a above -- this tells you which pointer leaked. */
+	*(uintptr_t *)((uintptr_t)alloc + *cur_offset) =
+	    (uintptr_t)0xa7a7a7a7a7a7a7a7ULL;
+	*cur_offset += sizeof(void *);
+}
+
```
yinan1048576: Explicitly define the two junk constants?

src/jemalloc.c
```diff
@@ -2745,6 +2770,14 @@ bool free_fastpath(void *ptr, size_t size, bool size_hint) {
 
 	tcache_t *tcache = tsd_tcachep_get(tsd);
 	cache_bin_t *bin = tcache_small_bin_get(tcache, alloc_ctx.szind);
+
+	/*
+	 * We "should" junk here, if set to.  But junking being enabled means
+	 * that we're on the fast path, which means we failed our fast-path
+	 * check above.
+	 */
+	assert(!opt_junk_free);
```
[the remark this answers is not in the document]
everschen: Hello yinan,
from your id, i just curious are you from VMW*** company?  :)
actually i also quite interested in jemalloc, I had written some learning note in 2018, it can be accessed from https://github.com/everschen/tools/blob/master/DOC/Jemalloc.pdf, 
especially about the unit test architecture, it attracted me quite a lot.
maybe later this year, i would like to join you to continue to learn the unit test.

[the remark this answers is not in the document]
davidtgoldblatt: Reworded.

test/unit/cache_bin.c
davidtgoldblatt: You mean the strings? I mostly don't add a message for things that are contextually clear. I can if you like, but I think I'd just be copying and pasting "expected failure" into the message for `expect_false(success)`. I'd just as soon refactor the test macros so that they don't have a required message parameter.

src/jemalloc.c

src/jemalloc.c
davidtgoldblatt: Right -- I think that the actual pointer reserved and the layout of pointers in the cache bin array should be almost completely unchanged with this refactor.

src/cache_bin.c
davidtgoldblatt: Done.

test/unit/cache_bin.c
yinan1048576: Makes sense. I also often didn't like having to have the message, and relied on some macro to auto-fill it in.

yinan1048576:
(You might want to restart the test that's being stuck.)

yinan1048576:
If possible, can you also put what the `malloc()` fast path assembly looks like after this change?

davidtgoldblatt:
Current fast path: https://gist.github.com/davidtgoldblatt/e73abe54749d43140bb8463c17cd56bb
Fast path + low-water-mark change with this PR: https://gist.github.com/davidtgoldblatt/17973981e2b09b13e427de9d96bfea9f

PR #1787, opened 2020-03-13
Reorganize cpp APIs and suppress unused function warnings

yinan1048576:
Get rid of
```
src/jemalloc_cpp.cpp:116:1: warning: unused function 'sizedDeleteImpl' [-Wunused-function]
sizedDeleteImpl(void* ptr, std::size_t size) noexcept {
^
1 warning generated.
```
And reorganize the APIs so that they follow a more intuitive order.


yinan1048576:
BTW I don't think we have any tests for the cpp APIs, do we?

davidtgoldblatt:
We've got some in the integration test directory, but yeah we could definitely use more.

PR #1795, opened 2020-03-20
Add piping API to buffered writer

yinan1048576:
The first commit cleans up the current buffered writer API -
- Previously, buffered writer has two getter functions respectively for the callback function and for the callback argument: if the buffer allocation failed, the getter functions just return the callback function and argument the buffered writer was originally given; otherwise, the getter functions return the buffered writer's internal write callback function and the buffered writer object itself.
- Now, regardless of whether the buffer allocation failed or not, the caller always calls buffered writer's internal write callback function with the buffered writer object itself, and the internal callback function decides to whether route the write call directly to the original callback function or buffer the write. There is thus no longer the need for the getter functions, and the caller side code is simplified.

The second commit adds a piping API to the buffered writer, for the usage case where what's given to the write callback function may be a read callback function instead of a simple string. This is useful for prof dumping, and it may also be a generally useful API to have. The signature of the read callback function is current designed to be analogous to the `read()` system call (as well as our `malloc_read_fd()` wrapper). It looks pretty general to me, but it can be changed if needed.

### DIFF
--- a/include/jemalloc/internal/buf_writer.h
+++ b/include/jemalloc/internal/buf_writer.h
@@ -27,4 +27,8 @@ void buf_writer_flush(buf_writer_t *buf_writer);
 write_cb_t buf_writer_cb;
 void buf_writer_terminate(tsdn_t *tsdn, buf_writer_t *buf_writer);
 
+typedef ssize_t (read_cb_t)(void *read_cbopaque, void *buf, size_t limit);
+void buf_writer_pipe(buf_writer_t *buf_writer, read_cb_t *read_cb,
+    void *read_cbopaque);
+
 #endif /* JEMALLOC_INTERNAL_BUF_WRITER_H */
--- a/include/jemalloc/internal/buf_writer.h
+++ b/include/jemalloc/internal/buf_writer.h
@@ -13,10 +13,8 @@
 typedef void (write_cb_t)(void *, const char *);
 
 typedef struct {
-	write_cb_t *public_write_cb;
-	void *public_cbopaque;
-	write_cb_t *private_write_cb;
-	void *private_cbopaque;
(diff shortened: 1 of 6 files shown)


davidtgoldblatt:
I'm on board with the first commit. Can you give a little motivation for the second? It seems like it sort of inverts control flow.

I.e. in previous buf_writer use cases, there's some other code that's "pushing" text into the buf_writer, whereas with the second commit we have the buf_writer "pulling" it. I think it's plausible for this to be useful functionality, but I'm unsure that it ought to live in the buf_writer itself (i.e. why not have some external function call the generator and then push into the buf_writer).

yinan1048576:
Yes, that's doable, if you mean that we have something like this:
```
while not_finished(source) {
    flush_if_full(buf_writer);
    n = pull_but_do_not_explode(buf_writer, source);
    buf_end += n;
}
```

and this `source` is a wrapper of either a string or a read callback function?

I started trying to implement something like that, but I then realized that the extra code needed (mainly for implementing the `source` layer) would be more than simply duplicating the logic for the two cases, i.e. what I have now. But I'm actually open to the code complication, because I don't like duplicating code logic.

Or, do you mean to have a wrapper that pulls content into a string, and the caller still calls the original write callback handle provided by the buffered writer? I don't think that is a good idea. There are two possibilities:
- the string directly uses the buffer in the buffered writer as storage: in such a case the caller logic has to peek into the internals of the buffered writer, or the buffered writer needs to provide lower level APIs.
- the string is separate storage: in such a case there would be extra memory cost, and extra `memcpy()` calls needed to copy the content from the string to the buffer.

And in either case, the caller needs to implement its own version of something like the `while()` loop I had earlier, which is also effectively duplicating code.

Thoughts?

Issue #1797, opened 2020-03-24
Allow periodic grabbing of thread name


davidtgoldblatt:
When dumping sampled allocations, it would often be handy to include the thread name of the allocator from the perspective of pthread_getname_np. We can add an option to re-set prof.thread.name to the current thread name when we sample. This avoid forcing the user to call a mallctl on threads they don't necessarily control.

yinan1048576:
Just had an offline chat with @binliu19. The default behavior, if this option is turned on, should be to try to fetch the thread name via system call at each sample until we have a valid thread name; afterwards we no longer try to fetch. The "until we have a valid thread name" part means one of the following two cases:
- We successfully fetched the thread name from system call and we successfully allocated a buffer to store it internally; or
- The thread name has successfully been set via an explicit `mallctl` call (which can overwrite a previously existing valid thread name, either fetched via system call or set via a `mallctl` call).

We can optionally keep trying to get the thread name via system call even after we have a valid thread name, but it's not ideal, because:
- thread name rarely changes after being set, and
- it risks overwriting a thread name previously set via an explicit `mallctl` call.

So let's just stop as soon as we've got a valid thread name?

davidtgoldblatt:
Resolved in https://github.com/jemalloc/jemalloc/pull/1798

PR #1798, opened 2020-03-25
Add option to fetch system thread name on each prof sample

yinan1048576:
This is solution to #1797.

The implementation is different from my comment on the issue: the system call is always invoked at sampling time. The purpose is to guarantee that we don't miss any thread name update set by the caller via system calls.

This would overwrite any thread name set via `mallctl`, so I'm explicitly disallowing the `mallctl` setter call when the new option is turned on. Logically the new option means: use the system thread name *instead of* the `mallctl` thread name.

Also added platform independent unit tests.

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -48,6 +48,7 @@ bool opt_prof_final = false;
 bool opt_prof_leak = false;
 bool opt_prof_accum = false;
 char opt_prof_prefix[PROF_DUMP_FILENAME_LEN];
+bool opt_prof_experimental_use_sys_thread_name = false;
 
 /* Accessed via prof_idump_[accum/rollback](). */
 static counter_accum_t prof_idump_accumulated;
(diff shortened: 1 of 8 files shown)


src/prof.c
```diff
@@ -799,42 +869,11 @@ prof_thread_name_get(tsd_t *tsd) {
...
-
-	tdata = prof_tdata_get(tsd, true);
-	if (tdata == NULL) {
-		return EAGAIN;
-	}
-
-	/* Validate input. */
-	if (thread_name == NULL) {
+	if (opt_prof_experimental_use_sys_thread_name) {
 		return EFAULT;
```
davidtgoldblatt: EFAULT is probably the wrong thing to return here now. Maybe EINVAL or ENOENT, depending on the degree to which you regard the "use thread name" bool as an implicit argument to the function, or something that removes this mallctl from existence.

src/prof.c
```diff
@@ -133,9 +134,101 @@ prof_alloc_rollback(tsd_t *tsd, prof_tctx_t *tctx) {
...
+	}
+	return 0;
+}
+
+static int
+prof_read_sys_thread_name_impl(char *buf, size_t limit) {
+#ifdef JEMALLOC_HAVE_PTHREAD_SETNAME_NP
+	return pthread_getname_np(pthread_self(), buf, limit);
+#else
+	return -1;
```
davidtgoldblatt: Probably should be `ENOSYS` if we're going to use posix error codes (which are positive in error cases).

test/unit/prof_use_sys_thread_name.c
```diff
@@ -0,0 +1,75 @@
...
+
+static int
+test_prof_read_sys_thread_name_error(char *buf, size_t limit) {
+	return -1;
+}
+
+static int
+test_prof_read_sys_thread_name(char *buf, size_t limit) {
+	assert(strlen(test_thread_name) < limit);
+	strcpy(buf, test_thread_name);
```
davidtgoldblatt: A raw strcpy will trigger warnings in some of our CI setups IIRC. Probably better is strncpy, even with the assert.

src/prof.c
yinan1048576: `ENOENT` feels better to me.

yinan1048576:

PR #1800, opened 2020-03-27
Make prof_tctx_t pointer a true prof atomic fence

yinan1048576:
This is more of a question than a PR. I don't quite get the reason why we need the strong `ATOMIC_RELEASE` and `ATOMIC_ACQUIRE` for the `prof_tctx_t` pointer on `edata`. I might have missed something important, and I want to better understand it. Otherwise we shouldn't impose stronger than necessary requirements.

### DIFF
--- a/src/large.c
+++ b/src/large.c
@@ -360,9 +360,9 @@ large_prof_tctx_reset(edata_t *edata) {
 
 void
 large_prof_info_set(edata_t *edata, prof_tctx_t *tctx) {
-	large_prof_tctx_set(edata, tctx);
 	nstime_t t;
 	nstime_init_update(&t);
 	edata_prof_alloc_time_set(edata, &t);
 	edata_prof_recent_alloc_init(edata);
+	large_prof_tctx_set(edata, tctx);
 }


davidtgoldblatt:
There needs to be some sort of synchronization between the setting of the `prof_tctx_t *` and the reading of it (because we're not only interested in it's value; we actually want to dereference it, and so need protection to prevent the reads of the `prof_tctx_t` fields from racing with the writes). Making the store release and the load acquire is the easiest way to ensure that.

I don't know the locking regimen of the profiling code terribly well; it's possible that both accesses are protected by some lock; but in that case they could both be non-atomics. I can't think of a case where we'd want this kind of pattern to be atomic but also relaxed. (This sort of pointer chasing is the canonical release-acquire example).

yinan1048576:
> I can't think of a case where we'd want this kind of pattern to be atomic but also relaxed. (This sort of pointer chasing is the canonical release-acquire example).

One example for such a setup is: write is always under a lock but read may or may not be, for performance reasons. In such a case the `ATOMIC_RELAXED` can be useful for guaranteeing that the read outside the lock is safe.

> I don't know the locking regimen of the profiling code terribly well; it's possible that both accesses are protected by some lock; but in that case they could both be non-atomics.

According to my understanding, the read and write of the `prof_tctx_t` field on `edata` is not protected by any lock. However, the part I was more curious about is: in which situations will two threads try to access the `prof_tctx_t` field on the very same `edata` at the same time, especially one of them reading and the other writing? The only cases I can think of are the following:
- One thread is allocating while another is freeing (the same object). I don't think this is possible. There must be some fencing so that, by the time the application get the pointer back from the allocation call, all operations in the allocation are guaranteed to have completed.
- One thread is doing ordinary allocation / free business and another thread is doing some kind of `edata` cleanup operations. I don't think this will be an issue, because the cleanup thread'd better use some lock to prevent the possibility of the `edata`s involved in the cleanup still being actively accessed by other threads.
- One thread is freeing and another thread is re-allocating (the same object).  I think this is the only place we may worry about. This is not quite a sensible application logic, but the application is allowed to do so if it wants to.

So my conclusion here is: we seem to have designed the atomic fencing around the `prof_tctx_t` field on `edata` only for an illegal type of application behavior. I feel I might be missing something, but even this is true, I think it suffices to make a legitimate reason for the design.

So, rather than removing the atomic fencing, I'll still keep it, and I examined the related code base to verify that we're making the right use of the fencing. The correct mechanism should be: by the time a thread writes to the `prof_tctx_t` field, not only the content the pointer points to, but also all the prof related information about the object, should have already been written, so that any thread having fetched a valid `prof_tctx_t` pointer can safely assure the availability of all those information. I found that the only exceptions to this mechanism are the following:
- https://github.com/jemalloc/jemalloc/commit/b664bd79356d7f6da6f413023f9aef014b85c145 set the allocation time *after* writing the `prof_tctx`, and
- https://github.com/jemalloc/jemalloc/commit/4afd709d1f3ae7a727f144a96d8b834157d31e17 kept the same order.

So I'm changing this PR to fix this order issue.

davidtgoldblatt:
> One example for such a setup is: write is always under a lock but read may or may not be, for performance reasons. In such a case the ATOMIC_RELAXED can be useful for guaranteeing that the read outside the lock is safe.

This pattern isn't safe by itself; even if the writer holds a lock, unless the reader also acquires it, there's nothing to establish synchronization. (In some situations, you can be sure that the reader has synchronized externally somehow, but that's fairly uncommon). Here, since it's pointer data, you don't just need the value of the pointer; you're going to dereference it. The synchronization protects the dereferenced data rather than the raw pointer value itself.

> So my conclusion here is: we seem to have designed the atomic fencing around the prof_tctx_t field on edata only for an illegal type of application behavior. I feel I might be missing something, but even this is true, I think it suffices to make a legitimate reason for the design.

This feels compelling to me -- I don't think that the prof_tctx_field needs to be atomic at all. For a given edata_t, it's either protected by a mutex (in the ecache or emap), or owned exclusively by some thread, with no other threads allowed to access it[1], so there's not actually any concurrency here.

[1] This is not actually completely true -- there are some (benign, given current compilers) races on e_bits that happen while merging (one thread may be setting szind while another is checking state to see if a merge is possible). I've got a refactor planned to move some of that metadata into the rtree though, which fixes the issue.

yinan1048576:
Discussed offline. The fencing, as in the current state of this PR, can be seen more as a "self-protection" for the profiling module: there's no way we can prevent applications from making calls in an illegal fashion (e.g. a concurrent `realloc()` and a `free()` on the same object), and when that happens, we can run into serious issues; the "self-protection" makes sure that the crashing stack trace doesn't contain anything related to the profiling data on the allocation object. So still merging.

> This pattern isn't safe by itself; even if the writer holds a lock, unless the reader also acquires it, there's nothing to establish synchronization. (In some situations, you can be sure that the reader has synchronized externally somehow, but that's fairly uncommon). Here, since it's pointer data, you don't just need the value of the pointer; you're going to dereference it. The synchronization protects the dereferenced data rather than the raw pointer value itself.

Agree, and thanks for the explanations! (The scenario I was thinking of was mainly like: there's no de-referencing, just getting the pointer value, say testing whether the `prof_tctx_t` value equals `1U` or not.)

PR #1802, opened 2020-03-29
Add malloc_conf_2_conf_harder

davidtgoldblatt:
This comes in handy when you're just a user of a canary system who wants to change settings set by the configuration system itself (and so you control the binary but not the environment variables, which can override your settings).

The silly name is deliberate -- it's a hint that this shouldn't really be present in "serious" code.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -209,6 +209,7 @@ TESTS_UNIT := \
 	$(srcroot)test/unit/junk_free.c \
 	$(srcroot)test/unit/log.c \
 	$(srcroot)test/unit/mallctl.c \
+	$(srcroot)test/unit/malloc_conf_2.c \
 	$(srcroot)test/unit/malloc_io.c \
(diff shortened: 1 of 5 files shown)


yinan1048576:
This is a handy feature. I had to manually overwrite the options before.

I also wonder: how do you deal with the jemalloc tests? What I was doing was: if I had to manually overwrite some option, some tests may fail which would block the canary runs, so I had to manually surround my manual overwrite with something like `ifndef JEMALLOC_JET`.

Implementation wise, I think you'd better put a fifth element in `opts_explain` (in `malloc_conf_init_helper()`) as well as in `opts_cache` (in `malloc_conf_init()`).

davidtgoldblatt:
Updated per comments. Changed the name to `malloc_conf_2_conf_harder`, which both better captures the intent and also sounds more like the [namesake](https://itunes.apple.com/us/movie/die-hard-2-die-harder/id578539479).

Issue #325, opened 2016-02-05
Implement decay-based unused dirty page purging.


jasone:
The current ratio-based mechanism for unused dirty page purging is effective at controlling accumulation of unused dirty pages, except that its conservatism can cause extreme performance degradation if the application's memory usage constantly fluctuates more than e.g. 1/9 with the default ```lg_dirty_mult:3``` setting.  In fact, we would prefer a much more stringent ratio, perhaps ```lg_dirty_mult:6```, but an unacceptable fraction of applications' performance would suffer with such a setting.

Part of the problem is that our internal clock can only measure passage of time in terms of allocation event ticks.  Introduce a wall clock and track how long unused dirty pages have existed, so that unused dirty page count decays toward zero over time.

jasone:
A year ago I did some design work for this feature as part of my [Applicative conference talk](http://applicative.acm.org/2015/applicative.acm.org/speaker-JasonEvans.html), and this week I wrote some simple simulation code to see how the decay algorithms behaved.

Specifically I simulated various malloc/free/noop traces and tracked the moving maximum ```active``` memory over a fixed-size trailing window of ticks.  Based on the moving maximum, at each tick I determined how much unused dirty memory to purge such that ```active + unused``` memory would intersect the moving maximum after a fixed number of ticks.  This algorithm requires no memory other than the moving maximum, which conveniently avoids tracking ages of unused dirty pages.  Overall the algorithm worked well.  An oversight jumped out at me though.  Since the amount of memory to purge is discretely computed at each tick (rather than being based on a pre-computed purging schedule), the decay approximates a piecewise curve that asymptotically approaches the moving maximum.  I had imagined this decay as linear, but it is not.  As indicated in my Applicative presentation, the ideal decay curve is sigmoidal, and I was intending a piecewise linear approximation, but this algorithm is effectively a delayed exponential curve.

[...]

PR #1804, opened 2020-04-01
Extract a page allocation ("pa") module

davidtgoldblatt:
This (extremely long -- I'm sorry Yinan!) PR gathers together all the assorted page-level accessors, stats, and functionality, and consolidates it into one place.

We along the way pull out and test various bits of helper functionality (like decay), which previously only got touched by integration tests. We also simplify much of the stats code, and split up the page-level stats from the allocation-level ones.

This is a fairly nice cleanup on its own (compare e.g. the slab and large alloc pathways before and after this change, and their similarity). But the real benefit comes in just having the pa module interface itself. We can now have a place to interpose arbitrary alternate allocation strategies; it enables the hugepage-backed page allocator coming in a later PR. It's even testable in isolation. (Even in a program that calls malloc! You can spin up an entirely separate base and emap for testing).

Unfortunately, some of the details of that later PR snuck into this one in ways that are sort of hard to rebase out of (like the "ranged" bit added to the edata). They aren't terribly well-motivated at the moment, but I promise they're useful down the road.

This gets a big chunk of the way towards de-arenafication. The pa shard an arena contains is logically separate from it; there's no reason that, in the future, the arena->shard mapping can be dynamic (say, with multiple auto arenas sharing a single pa_shard, while manual arenas get their own).

This is not as complete as it could be. I think a future world has the page allocator know even less about how the extents it returns get used (i.e. the values of slab and szind are the business of the arena and emap, not the page allocator), but this PR is sprawling enough as it is.


yinan1048576:
Will review next week.

davidtgoldblatt:
> Out of curiosity, I haven't looked all the way into the decay module, but is there anything e.g. a paper that explains why we should use the smoothstep model for decaying?

This was a little bit before my time, so I couldn't say for sure the motivations there. Two things I'd suggest are https://github.com/jemalloc/jemalloc/issues/325 and https://www.youtube.com/watch?v=RcWp5vwGlYU , which give the motivation for time-based decay generally but not smoothstep specifically. I think it's mostly an engineering decision overall; i.e. there's not some subtle mathematical justification for smooth-step over ratio-based decay or something else entirely. Over the years we've talked about alternative or mixed strategies.

> A minor comment regarding forking handling: it feels to me that we can have a perhaps centralized module for that. I'm having trouble trying to always remember to handle each new mutex. (This doesn't need to be part of this PR.)

Indeed -- see https://github.com/jemalloc/jemalloc/issues/943 . There was a PR for this a while back, but we failed to follow up (unfortunately. I still think it's a good idea): https://github.com/jemalloc/jemalloc/pull/1153 .

> I'm a bit lost while reading your last stage of commits where the rtree is involved (starting from 70de0cd). Can you explain on a high level the set of goals you're trying to achieve?

I'm sorry; these are a little bit jumbled (some rebases got a little bit messy here). I think there are a few buckets:
- The rtree changes are just simplifications of metadata management to make that code easier to wrangle, in the service of later changes to come.
- The emap changes make it stop modify the edata_t itself. This moves the emap more towards its goal of only handling mapping, rather than some combination of mapping and edata management.
- The extent changes are similar. The extent/pa modules should only concern themselves with page-level allocation; the way that the caller is going to use the returned value shouldn't matter to them.
- The parameterization changes are purely in the service of improving testability.
- Pulling emap_global out of emap does likewise. After this change, we can test the page-level allocator on its own, without touching the same emap that the test program uses. (I don't actually have such a test in this PR, but I've got one written).

The addition of the edata ranged bit is for an incoming hugepages PR. The history here is that I wrote it first, did some of the other refactorings in that PR, and then realized that they ought to go in this one. But that means that all the refactorings that touch edatas assume that it has a ranged-bit initializer, and pulling it out only to put it back in in the next PR seemed like a pain.

yinan1048576:
Thank you for the pointers. I think I know better now why we need a sigmoidal curve. I'll think more about the fork handling module.

I'll look at the last section on Friday. Meanwhile, I forgot another nit: I liked that you extracted out the lockedstat module, and I think there should be a way to further unify it with the counter module (see https://github.com/jemalloc/jemalloc/commit/d71a145ec1bb8153c3d69be27eea5b076d59abfe). The counter module has the same `#ifdef JEMALLOC_ATOMIC_U64` business, only except that it assumes a built-in mutex and it has a fancier pair of add/subtract. Possible consolidation can be done in a later PR, but for now, it seems the name "locked stat" is not general enough to include scenarios like the prof / stat dumping "stopwatch". Perhaps "locked counter" or something?

davidtgoldblatt:
What about just naming the module `lockedint`, and the functions/types `locked_u64_t` and `locked_zu_t`?

davidtgoldblatt:
(Just because "counter" is a little wordy).

yinan1048576:
Yes, sounds good.

davidtgoldblatt:
^--- lockedstat renaming

davidtgoldblatt:
Oh absolutely (in fact, I think one of the nice features of this PR is that it allows that testing without going through all the machinery in between realloc and extent.c).

PR #1807, opened 2020-04-02
Add some advanced functionalities to ql module

yinan1048576:


yinan1048576:
Add a rotate functionality.

yinan1048576:
Add final couple of handy shorthand notations. Ready for review.

include/jemalloc/internal/ql.h
```diff
@@ -58,6 +58,15 @@ struct {								\
 	ql_first(a_head) = qr_next((a_elm), a_field);			\
 } while (0)
 
+#define ql_concat(a_head_a, a_head_b, a_field) do {			\
+	if (ql_first(a_head_a) == NULL) {				\
+		ql_first(a_head_a) = ql_first(a_head_b);		\
+	} else if (ql_first(a_head_b) != NULL) {			\
+		qr_meld(ql_first(a_head_a), ql_first(a_head_b),		\
+		    a_field);						\
+	}								\
```
davidtgoldblatt, line 74: Let's null out a_head_b here? This makes this logically a "pop all" function. I think in most reasonable use cases the very next thing we'll want to do is reinitialize b to the empty list anyways (or never touch it again, in which case it doesn't end up mattering).

yinan1048576: Great point. Also added unit tests to verify.

yinan1048576:
Rebase, and revised according to feedback. Added one more commit to better structure the `ql` module on top of the new utilities.

PR #1812, opened 2020-04-10
Add documentation to the linked list modules.

davidtgoldblatt:


include/jemalloc/internal/ql.h
```diff
@@ -116,6 +176,21 @@ struct {								\
 	ql_first(a_head) = qr_next(ql_first(a_head), a_field);		\
 } while (0)
 
+/*
+ * Helper macro to iterate over each element in a list in order, starting from
+ * the head (or in reverse order, starting from the tail).  The usage is
+ * (assuming my_t defined as above).
+ *
+ * int sum(my_t *item) {
```
yinan1048576, line 184: `item` should be of the list type?

include/jemalloc/internal/ql.h
```diff
@@ -3,69 +3,119 @@
...
+/* A pointer to the first element in the list, or NULL if the list is empty. */
 #define ql_first(a_head) ((a_head)->qlh_first)
 
+/* Dynamically initializes a list. */
 #define ql_new(a_head) do {						\
 	ql_first(a_head) = NULL;					\
 } while (0)
 
+/* Removes all elements from the list. */
 #define ql_clear(a_head) ql_new(a_head)
```
yinan1048576, line 50: This is something I thought about earlier but didn't have a clear conclusion. This `ql_clear()` is mainly meant to be used internally (within this header), and "responsibly", in the sense that the content of the list has always been safely relocated elsewhere before it's called, rather than risking being lost if the caller is not careful. Perhaps changing the name to `ql_clear_internal()` or `_ql_clear()` makes more sense in this case. We can also expect the caller to be responsible themselves and state this in the comment, but I couldn't find an intuitive example why the caller may want to clear the list first and then manipulate the nodes themselves. I think I'm now slightly leaning towards keeping it internal, but either way works.

davidtgoldblatt: TBH I think the fact that we have both `ql_clear` and `ql_new` is an oversight. You do need to clear nonempty lists sometimes (e.g. arena tcache list after forking), but I don't know that that was the intent (and practically, we use ql_new there).

davidtgoldblatt: Put up a version that just only has ql_new.

yinan1048576: I originally put `ql_new()`s in these places but felt that the name sounded a bit "intrusive", because we're not creating new lists there, so I added the `ql_clear()` as a "rephrase". Now it seems `ql_new()` feels better to me. I think it's fine.

PR #1819, opened 2020-04-15
[Seeking early feedback] Do not rollback prof idump counter in arena_prof_promote()

yinan1048576:
First, a question: when we bump a small allocation to a large allocation when it's sampled, do we count the original `usize` or the bumped one in the events?

I'm leaning towards that we should count the original `usize` - the bumped `usize` is jemalloc internal and should not affect user observable behaviors.

In any case, we don't need to rollback `usize` when we're bumping. If we only count `usize`, then we do nothing in bumping, and the final event counter update can take care of it, and this is what's currently the case in this PR; if we count the bumped `usize`, then we can count the difference between the two in bumping, and the final event counter update can complete it.

After we reach an agreement (which is the early feedback I'm seeking), I'll further do some simplifications.

### DIFF
--- a/include/jemalloc/internal/counter.h
+++ b/include/jemalloc/internal/counter.h
@@ -51,31 +51,6 @@ counter_accum(tsdn_t *tsdn, counter_accum_t *counter, uint64_t accumbytes) {
 	return overflow;
 }
 
-JEMALLOC_ALWAYS_INLINE void
-counter_rollback(tsdn_t *tsdn, counter_accum_t *counter, uint64_t bytes) {
-	/*
-	 * Cancel out as much of the excessive accumbytes increase as possible
-	 * without underflowing.  Interval-triggered events occur slightly more
-	 * often than intended as a result of incomplete canceling.
-	 */
-	uint64_t a0, a1;
-#ifdef JEMALLOC_ATOMIC_U64
-	a0 = atomic_load_u64(&counter->accumbytes,
-	    ATOMIC_RELAXED);
-	do {
-		a1 = (a0 >= bytes) ? a0 - bytes : 0;
-	} while (!atomic_compare_exchange_weak_u64(
-	    &counter->accumbytes, &a0, a1, ATOMIC_RELAXED,
-	    ATOMIC_RELAXED));
-#else
-	malloc_mutex_lock(tsdn, &counter->mtx);
-	a0 = counter->accumbytes;
-	a1 = (a0 >= bytes) ?  a0 - bytes : 0;
-	counter->accumbytes = a1;
-	malloc_mutex_unlock(tsdn, &counter->mtx);
-#endif
-}
-
 bool counter_accum_init(counter_accum_t *counter, uint64_t interval);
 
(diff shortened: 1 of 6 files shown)


davidtgoldblatt:
>First, a question: when we bump a small allocation to a large allocation when it's sampled, do we count the original usize or the bumped one in the events?

I think we use the original usize, right? We don't compute the bumped one until we're in the profiling pathways, which is definitionally after we're in the profiling event. (I think?).

> I'm leaning towards that we should count the original usize - the bumped usize is jemalloc internal and should not affect user observable behaviors.

This makes sense to me. Are we not doing that now?

> In any case, we don't need to rollback usize when we're bumping. If we only count usize, then we do nothing in bumping, and the final event counter update can take care of it, and this is what's currently the case in this PR; if we count the bumped usize, then we can count the difference between the two in bumping, and the final event counter update can complete it.

My strong suspicion is that all the recent refactorings of internal counters have rendered a lot of the promotion/demotion stuff more complicated than it needs to be. This all sounds right to me.

yinan1048576:
> I think we use the original usize, right? We don't compute the bumped one until we're in the profiling pathways, which is definitionally after we're in the profiling event. (I think?).

Yes, the bumped usize is encapsulated within profiling, and anything else can (and should) only get the original usize.

> This makes sense to me. Are we not doing that now?

Kind of, except that we're rolling back and then accumulating the same usize, ending up not doing anything. I suspected that the rollback call in `arena_prof_promote()` seemed to have the intention of trying to "swap" for another usize - and the bumped usize is the only possibility, which was the reason why I asked the question.

> My strong suspicion is that all the recent refactorings of internal counters have rendered a lot of the promotion/demotion stuff more complicated than it needs to be. This all sounds right to me.

I think the promotion stuff existed long before the counter changes, which is the reason why I guessed that, even though things don't look right, there might have been some legitimate intention.

Anyway, it seems that neither of us can figure out any legitimate intention underlying the rollback call. Let me proceed with a couple of final touches.

yinan1048576:
Added two more commits and the PR is ready for final review.

The first one makes `prof_idump_accum()` non-inline: it doesn't need to because it's only accessed on slow path (when prof sampling is triggered).

The second one is a bug fix: the bug is introduced in https://github.com/jemalloc/jemalloc/commit/198f02e7972023d10c9e4c4c6ab162738d103707, which consolidated the prof idump counters from one per tcache and one per arena to a single global counter, but failed to pay attention to the init part. The bug would allow manual arena creations to mistakenly clear the counter.

yinan1048576:
Ah, and https://github.com/jemalloc/jemalloc/commit/198f02e7972023d10c9e4c4c6ab162738d103707 also failed to remove the counter field on arena. Removing it.

yinan1048576:
Realized that there is now no longer the need for a separate `prof_inlines_a.h`: I believe the reason why we needed it before was that the arena/tcache depended on the `prof_idump` counter adjustment logic, but now only the thread event depends on it.

Also spotted some places needing a cleanup in the tcache allocation logic, while I was verifying that there was no dependency on prof.

davidtgoldblatt:
New commits look good too.

PR #1820, opened 2020-04-15
Add missing forking handling for users of counter module

yinan1048576:


yinan1048576:
I'm also curious if/how the other users of `lockedint` are handling their locks during forking.

PR #1823, opened 2020-04-17
Unify write callback signature

yinan1048576:
The write callback signature is relocated from buffered writer to `malloc_io`, which it more naturally belongs to. Unified the write callback declarations in the code base, so that they become more simplified and less error prone.

This is also a step needed for rewriting prof dumping.


yinan1048576:
Good point. Revise.

PR #1828, opened 2020-04-22
Batch allocation API

yinan1048576:


yinan1048576:
Implemented a batch allocation only from fresh slabs.

yinan1048576:
Better split the bin shard selection logic, and improve around the stats update.

yinan1048576:
Generalize the arena level API to `size_t`, and implemented a user-facing API on top of it that handles the profiling logic. Need to steal the commit from #1777 that adds the thread event surplus reading functionality.

yinan1048576:
Minor improvement, and added more comments.

I was also tempted to try an alternative approach, where I pre-compute how many prof samples I will draw in the batch, and then only call the arena level batch allocation function once, rather than calling it once per prof sample. The benefit is maximal locality for all the non prof samples, as much as what the underlying PA module can possibly achieve.

I ended up giving up this approach, mainly because it's very complicated to handle the OOM case well. In order to guarantee the contract that `thread.allocated` always faithfully represent the allocation amount, we'd have to handle the following complications:
- the possibility of either freeing a portion of the prof samples, or batch-freeing a segment of the batch-allocated memory, depending on which gets allocated first,
- supporting either "multi-hop" thread event look-ahead, or thread event rollback.

None of these feel well justified to invest (significantly more) effort for.

So I'm still sticking to my previous design. The limitation is that the non prof samples separated by prof samples would not be located together, the locality effect of which should be negligible.

yinan1048576:
Minor comment clarifications.

yinan1048576:
Rebase and added a commit to add a zero version to `arena_slab_reg_alloc_batch()`. Performance matters for this function, so decided to create a new function rather than asking the original function to support the zero flag. Want to have some early feedback on it.

yinan1048576:
Revised:
- Added `mallocx` styled flag support. All flags except the `tcache` flag are supported.
- The API will always fetch from fresh slabs to maximize locality. No attempt is made to fetch from `tcache`, `slabcur` or non-full slabs. Simplified the commit stack accordingly.

yinan1048576:
Corrected a few minor issues:
- Do not assert `usize == sz_s2u(usize)` in thread event handler: `usize` can be the sum of any number of the real `usize`, so the assertion doesn't make sense.
- Better handling of non-prof case in the batch allocation wrapper function.
- Correct flag handling for the sampled allocations within the batch.

My next step is to add mallctl as well as unit tests (which can be tricky for prof build but we'll see). The PR at the current stage is ready for a first round (for the core logic).

yinan1048576:
Added unit tests. There are two sets of tests, one for profiling-off case and one for profiling-on case. The profiling-off case is relatively straightforward: the tests are for contiguity of the allocations returned.

The profiling-on case is tricky. To test for the new API's behavior in the general scenario, I'm not setting `lg_prof_sample:0` as in the other profiling tests, but rather I'm setting it realistically, so that allocations are truly probabilistically sampled. Relying on the fact that the sampled allocations as well as the first allocations in fresh slabs are page-aligned, I made the test independent of the sampling decisions.

My final step is to write a `mallctl` handle to expose the API (given that all tests pass).

yinan1048576:
It seems GitHub is not able to sync the newest changes here for some reason. They can be viewed at https://github.com/yinan1048576/jemalloc/commits/batch-alloc .

yinan1048576:
Slightly improved the tests and added mallctl handle.

yinan1048576:
Changed the tests so that they call the new mallctl handle.

test/unit/batch_alloc.c
```diff
@@ -0,0 +1,96 @@
...
+			j = 0;
+		}
+		void *p = ptrs[i];
+		expect_zu_eq(isalloc(tsdn, p), usize, "");
+		if (zero) {
+			expect_d_eq(memcmp(p, ref, usize), 0, "");
+		}
+		free(p);
+		if (j == 0) {
+			expect_true(prof_sample_aligned(p), "");
```
davidtgoldblatt, line 24: We should be clearer about our expectations for prof_sample alignment vs. merely page alignment. We should be able to turn up the alignment of profiled objects without breaking tests.

src/ctl.c
```diff
@@ -3627,3 +3629,30 @@ experimental_prof_recent_alloc_dump_ctl(tsd_t *tsd, const size_t *mib,
...
+    size_t miblen, void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret;
+
+	batch_alloc_packet_t batch_alloc_packet;
+	ASSURED_WRITE(batch_alloc_packet, batch_alloc_packet_t);
+
+	size_t filled = arena_batch_alloc(batch_alloc_packet.ptrs,
+	    batch_alloc_packet.num, batch_alloc_packet.size,
+	    batch_alloc_packet.flags);
+	READ(filled, size_t);
```
davidtgoldblatt, line 3652: We should do a test read of 0 or something first (or manually validate input parameters). If the size the user passes is incorrect, we'll return EINVAL; we don't want to actually do the allocations then.

test/unit/batch_alloc.c
```diff
@@ -59,10 +59,28 @@ verify_batch(void **ptrs, size_t batch, size_t usize, bool zero) {
...
+typedef struct batch_alloc_packet_s batch_alloc_packet_t;
+struct batch_alloc_packet_s {
+	void **ptrs;
+	size_t num;
+	size_t size;
+	int flags;
+};
+
+static size_t
+batch_alloc(void **ptrs, size_t num, size_t size, int flags) {
```
davidtgoldblatt, line 71: I don't think it's necessary now, but I think we'll eventually have to have a version of this API that returns a function pointer to the implementation. The mallctl pathways are pretty slow. But it's fine to wait until we have benchmark data or something, too.

test/unit/batch_alloc.c
```diff
@@ -0,0 +1,96 @@
...
+	const size_t size = 7;
+	const size_t a = 16;
+	size_t filled = arena_batch_alloc(ptrs, BATCH, size, MALLOCX_ALIGN(a));
+	assert_zu_eq(filled, BATCH, "");
+	verify_batch(ptrs, BATCH, sz_sa2u(size, a), false);
+}
+TEST_END
+
+int
+main(void) {
```
davidtgoldblatt, line 109: We should also have tests of stats behavior under calls to batch_alloc (things like malloc counts, nslabs, etc.).

src/arena.c
```diff
@@ -883,6 +883,59 @@ arena_cache_bin_fill_small(tsdn_t *tsdn, arena_t *arena,
...
+			batch = nfill - filled;
+			if (batch > nfree) {
+				batch = nfree;
+			}
+			assert(batch > 0);
+			arena_slab_reg_alloc_batch(slab, bin_info,
+			    (unsigned)batch, &ptrs[filled]);
+			filled += batch;
+		}
+		malloc_mutex_lock(tsdn, &bin->lock);
```
davidtgoldblatt, line 993: I think it's fine for now, but in the common case in which we don't use a manual arena and don't pick sample an allocation, this does a lot of unnecessary lock acquisitions and releases. (We can instead just wait until the end, and update the stats and min slab).

Probably a premature optimization for now, but worth leaving a comment.

test/unit/batch_alloc.c
davidtgoldblatt: And probably zeroing behavior, too.

test/unit/batch_alloc.c
yinan1048576: Agree on the function pointer part. (Regarding this particular change, I was just adding this convenient function in the test to avoid code duplication.)

src/arena.c
```diff
@@ -883,6 +883,59 @@ arena_cache_bin_fill_small(tsdn_t *tsdn, arena_t *arena,
...
+	const bin_info_t *bin_info = &bin_infos[binind];
+	const size_t nfree = bin_info->nregs;
+	assert(nfree > 0);
+	unsigned binshard;
+	bin_t *bin = arena_bin_choose(tsdn, arena, binind, &binshard);
+	size_t nslab = 0;
+	size_t filled = 0;
+	edata_t *slab;
+
+	do {
```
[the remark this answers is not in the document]
davidtgoldblatt: Ahhhh, I understand the logic now. I didn't see the only-the-last-iteration semantics. I think this code layout is fine in that case (but could you mind adding a comment explaining).

I could go either way on the optimization; I can make an argument I buy in both directions:
- Argument for doing it: this is the API people are calling because they're doing hot-path constant factor optimizations; we should try to make it as fast as we can.
- Argument against doing it: it's sort of a pain. This path is already faster than just calling malloc() N times, so it's not like anyone has a right to complain.

I think I lean towards doing the optimization, but I won't fight you on it if you think it's too much trouble for now.

I think the most reasonable place to get storage is probably to use an edata_list_active_t. There's an optimization within the optimization though -- we don't track full slabs for auto arenas (since auto arenas can't be destroyed, which is the only time we need to track them). So we can even avoid iterating through allocated slabs in the common cases (though we still need to track non-full slabs and stick them into slabs-nonfull). In the non-auto-arena case, we can just do a batch list merge of the new full slabs into the arena bin slabs_full list, so in either case we only really need to touch a few unusual slabs.

src/arena.c
```diff
@@ -260,6 +260,80 @@ arena_slab_reg_alloc_batch(edata_t *slab, const bin_info_t *bin_info,
...
+		}
+		slab_data->bitmap[group] = g;
+	}
+#endif
+	memset((void *)low, 0, (size_t)(expect - low));
+	edata_nfree_sub(slab, cnt);
+}
+
+static inline void
+arena_slab_reg_alloc_batch(edata_t *slab, const bin_info_t *bin_info,
```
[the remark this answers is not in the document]
davidtgoldblatt: The first option makes sense to me.

test/unit/batch_alloc.c
davidtgoldblatt: Oh no, the existing test looks good. I guess I managed to forget about it in the 1 minute gap between those two comments.

src/jemalloc.c
```diff
@@ -3916,6 +3916,107 @@ je_malloc_usable_size(JEMALLOC_USABLE_SIZE_CONST void *ptr) {
...
+		bool prof_sample_event = config_prof && opt_prof
+		    && te_prof_sample_event_lookahead_surplus(tsd,
+		    batch * usize, &surplus);
+
+		if (prof_sample_event) {
+			/*
+			 * Adjust so that the batch does not trigger prof
+			 * sampling.
+			 */
+			batch -= surplus / usize + 1;
```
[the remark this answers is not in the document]
davidtgoldblatt: What's the meaning of surplus? I'm not sure my mental model of it is correct.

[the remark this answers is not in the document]
yinan1048576: Surplus is the amount of bytes beyond what's needed for triggering an event. In the new attribution model, it's the "unsampled bytes". 0 means the current size exactly hits the sample threshold.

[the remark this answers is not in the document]
davidtgoldblatt: OK I think I agree that it's right as is.

yinan1048576:
Addressed all comments except the following:
- Testing for stats: I was trying to get the bin stats and assess whether the difference was correct, but the tests ended up being a trivial repetition of the relevant logic in `arena_fill_small_fresh()`, so I didn't keep the tests. Also, the stats update logic is now simple enough and does not really need a test I think.
- Guarding around mis-specified output size for the mallctl handle - will update after #1885 is resolved.

Also made some other improvements, including:
- Added some useful facilities in the typed list module.
- Added branch hints in the outer `batch_alloc()` wherever appropriate, since we somewhat care about its performance.
- Added more batch sizes in the tests to verify some edge cases.
- Added simple tests for large allocations (to verify the fallback logic).

yinan1048576:
I assumed in the tests that the slabs are of a fixed size, but the assumption turned out to be false, e.g. with the `--with-lg-page=16` flag on. Correcting it...

src/arena.c
```diff
@@ -883,6 +883,64 @@ arena_cache_bin_fill_small(tsdn_t *tsdn, arena_t *arena,
...
+		if (batch == nregs) {
+			if (manual_arena) {
+				edata_list_active_append(&fulls, slab);
+			}
+			slab = NULL;
+		}
+	}
+
+	malloc_mutex_lock(tsdn, &bin->lock);
+	if (slab != NULL) {
```
davidtgoldblatt, line 930: Can you add a comment that only the last slab can be nonempty here?

davidtgoldblatt:
> Testing for stats: I was trying to get the bin stats and assess whether the difference was correct, but the tests ended up being a trivial repetition of the relevant logic in arena_fill_small_fresh(), so I didn't keep the tests. Also, the stats update logic is now simple enough and does not really need a test I think.

I think this sort of test is still useful as for regression-detection functionality, even if we're convinced the code as it exists today is bug free. I think one day we might want to break out batch-allocs and non-batch allocs into separate stats. At that point, the `allocs` stat will become derived -- if we forget to do that change, it's a bug our test suite won't catch, and which the future version of ourselves might not think to either.

yinan1048576:
Corrected the error. Also made the following improvements:
- Added a utility to determine whether an allocation is sampled or not.
- Refactored the tests to make dynamic batch size determination easier.
- Made the assertions in the tests much more precise.

Let's see all tests can pass.

> I think this sort of test is still useful as for regression-detection functionality, even if we're convinced the code as it exists today is bug free. I think one day we might want to break out batch-allocs and non-batch allocs into separate stats. At that point, the `allocs` stat will become derived -- if we forget to do that change, it's a bug our test suite won't catch, and which the future version of ourselves might not think to either.

Makes sense. Before testing the stats, can I confirm with you that the current handling of the stats is correct? The current logic is a possible increment of `bin->stats.nonfull_slabs` encapsulated in `arena_bin_slabs_nonfull_insert()` under `arena_bin_lower_slab()`, if the last slab is not full, together with the following:
```
    bin->stats.nslabs += nslab;
    bin->stats.curslabs += nslab;
    bin->stats.nmalloc += filled;
    bin->stats.nrequests += filled;
    bin->stats.curregs += filled;
```
Anything I'm missing?

davidtgoldblatt:
Yes, I think that's right.

yinan1048576:
Rebase, and
- Added tests for stats.
- Added tests for manual arena case.
- Guarded against wrong output size in mallctl.

yinan1048576:
Fixed a few easy errors. The harder errors should be more visible...

yinan1048576:
It's a long but interesting experience debugging these issues. Luckily there isn't a real issue, mostly artifacts. Haven't double-checked every single failure, so the list I summarize below has a slight chance of not being exhaustive. Will see how this round of testing goes...
- The `arena_decay_tick()` call at the end of `arena_fill_small_fresh()` may bootstrap the `arena_tdata` if it's not there, which involves an allocation that can change the stats. I added a "dummy" allocation before the real tests start to resolve the issue. (I can alternatively make the tests less exact, but I really want to preserve the tests' precision.)
- The batch allocation call invokes the ordinary set of events, which include the tcache GC event that can change the stats. I increased the tcache GC interval to be a huge value to resolve the issue. (I can alternatively disable the tcache, but I want to keep it because that's the typical setting where the batch allocation API is invoked by real users.)
- The default for tcache flag is auto, so even when an arena flag is present, the tcache will be checked first and the top allocation there will be returned regardless of its source. Therefore, the sampled allocations in the batch can come from a different arena. This is fine from locality perspective since the sampled allocations won't live together with non-sampled ones anyway, but it's not fine from "arena grouping" perspective. I have a more systematic fix in #1892, but while it's being discussed, I'm doing some hacks in this PR. One is to assume `MALLOCX_TCACHE_NONE` whenever a manual arena is specified from the caller. Another is to assume `MALLOCX_TCACHE_NONE` in the "dummy" allocation described above.

yinan1048576:
> Another is to assume `MALLOCX_TCACHE_NONE` in the "dummy" allocation described above.

Well, this is not a hack, in the sense that it'd be no longer needed if we have the systematic fix; it's always needed even if the arena flag is not present.

yinan1048576:
All failures are gone. I'm temporarily adding a CI-only commit to see if the stats tests can now also pass in reentrancy cases. I suspect that the solutions to the failures earlier should also automatically take care of the reentrancy cases, and let me check just in case...

yinan1048576:
No failure as expected. Merging the change into the previous commit. Now testing another thing: the unreachable code error was previously observed but there's no longer any issue after a seemingly equivalent code change. Trying to restore to the previous state. Curious what happened.

yinan1048576:
Fixed a minor issue in the new CI-only commit. See how it goes...

yinan1048576:
Confirmed that the equivalent code change is indeed equivalent. The previous errors might be a consequence of some unexpected interactions between the unreachable code assertion in the test and the `cassert()` assertion in the code. Let me try to restore the errors and dig a bit into it and see what comes out... (Meanwhile the main body of the PR is ready for review.)

yinan1048576:
The previously observed issue cannot be restored. Got rid of the CI-only commit.

PR #1830, opened 2020-04-23
Add a high resolution timestamp option for profiling

tyroguru:
The approach taken is to provide a distinct nstime_update() for profiling code which can provide a high or coarse (normal) resolution resolution timestamp. The choice to provide a new and distinct nstime_update()  Which timestamp is provided depends upon the setting of the new profiling parameter, 'prof_time_resolution', which currently has two settings:

  prof_time_resolution: default
  prof_time_resolution: high

A setting of 'high' gives a high resolution timestamp which on modern Linux platforms has a nonosecond resolution. The default setting is currently 1 millisecond.

Note that this feature is currently only enabled for the Linux platform. The current implementation will allow a non-Linux platform to specify 'prof_time_resolution:default' as a parameter but
'prof_time_resolution:high' will be generate an abort at runtime with an appropriate message.

@davidtgoldblatt (and anyone else!) - can you let me know if this approach is viable? Thanks.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -14,6 +14,7 @@
 #include "jemalloc/internal/log.h"
 #include "jemalloc/internal/malloc_io.h"
 #include "jemalloc/internal/mutex.h"
+#include "jemalloc/internal/nstime.h"
 #include "jemalloc/internal/rtree.h"
 #include "jemalloc/internal/safety_check.h"
 #include "jemalloc/internal/sc.h"
@@ -1497,6 +1498,26 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				CONF_HANDLE_BOOL(
 				    opt_prof_experimental_use_sys_thread_name,
 				    "prof_experimental_use_sys_thread_name")
+				if (CONF_MATCH("prof_time_resolution")) {
+					if (CONF_MATCH_VALUE("default")) {
+						opt_prof_time_res =
+						    prof_time_res_default;
+					} else if (CONF_MATCH_VALUE("high")) {
+						if (!config_high_res_timer) {
+							CONF_ERROR(
+							    "No high resolution"
+							    " timer support",
+							    k, klen, v, vlen);
+						} else {
+							opt_prof_time_res =
+							    prof_time_res_high;
+						}
+					} else {
+						CONF_ERROR("Invalid conf value",
+						    k, klen, v, vlen);
+					}
+				}
+				CONF_CONTINUE;
 			}
 			if (config_log) {
 				if (CONF_MATCH("log")) {
--- a/src/nstime.c
+++ b/src/nstime.c
@@ -152,25 +152,65 @@ nstime_monotonic_impl(void) {
 }
 nstime_monotonic_t *JET_MUTABLE nstime_monotonic = nstime_monotonic_impl;
 
-static bool
+prof_time_res_t opt_prof_time_res =
+	prof_time_res_default;
+
+const char *prof_time_res_mode_names[] = {
+	"default",
+	"high",
+};
+
+
+static void
+nstime_get_realtime(nstime_t *time) {
+#if defined(JEMALLOC_HAVE_CLOCK_REALTIME) && !defined(_WIN32)
+	struct timespec ts;
+
+	clock_gettime(CLOCK_REALTIME, &ts);
+	nstime_init2(time, ts.tv_sec, ts.tv_nsec);
+#else
+	unreachable();
+#endif
+}
+
+static void
+nstime_prof_update_impl(nstime_t *time) {
+	nstime_t old_time;
+
+	nstime_copy(&old_time, time);
+
+	if (opt_prof_time_res == prof_time_res_high) {
+		nstime_get_realtime(time);
+	} else {
+		nstime_get(time);
+	}
+}
+nstime_prof_update_t *JET_MUTABLE nstime_prof_update = nstime_prof_update_impl;
+
+static void
 nstime_update_impl(nstime_t *time) {
 	nstime_t old_time;
 
 	nstime_copy(&old_time, time);
-	nstime_get(time);
+  nstime_get(time);
 
 	/* Handle non-monotonic clocks. */
 	if (unlikely(nstime_compare(&old_time, time) > 0)) {
 		nstime_copy(time, &old_time);
-		return true;
 	}
-
-	return false;
 }
 nstime_update_t *JET_MUTABLE nstime_update = nstime_update_impl;
 
-bool
+void
 nstime_init_update(nstime_t *time) {
 	nstime_init_zero(time);
-	return nstime_update(time);
+	nstime_update(time);
 }
+
+void
+nstime_prof_init_update(nstime_t *time) {
+	nstime_init_zero(time);
+	nstime_prof_update(time);
+}
+
+
--- a/configure.ac
+++ b/configure.ac
@@ -1776,6 +1776,18 @@ if test "x${je_cv_mach_absolute_time}" = "xyes" ; then
   AC_DEFINE([JEMALLOC_HAVE_MACH_ABSOLUTE_TIME])
 fi
 
+dnl check for CLOCK_REALTIME (always should be available on Linux)
+JE_COMPILABLE([clock_gettime(CLOCK_REALTIME, ...)], [
+#include <time.h>
+], [
+	struct timespec ts;
+
+	clock_gettime(CLOCK_REALTIME, &ts);
+], [je_cv_clock_realtime])
+if test "x${je_cv_clock_realtime}" = "xyes" ; then
+  AC_DEFINE([JEMALLOC_HAVE_CLOCK_REALTIME])
+fi
+
 dnl Use syscall(2) (if available) by default.
 AC_ARG_ENABLE([syscall],
   [AS_HELP_STRING([--disable-syscall], [Disable use of syscall(2)])],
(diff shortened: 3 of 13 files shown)


src/jemalloc.c
```diff
@@ -104,6 +104,22 @@ const char *zero_realloc_mode_names[] = {
...
 
+prof_time_resolution_t opt_prof_time_resolution =
+	prof_time_resolution_default;
+
+const char *prof_time_resolution_mode_names[] = {
+	"default",
+	"high",
+};
+
+bool opt_high_resolution_timer =
```
davidtgoldblatt, line 115: This should live in `jemalloc_preamble.h.in` and be `static const bool config_high_resolution_timer` ("opt" is for MALLOC_CONF-style configuration options; compiler / build time settings are under "config").

src/jemalloc.c
```diff
@@ -1459,6 +1475,21 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				CONF_HANDLE_BOOL(
 				    opt_prof_experimental_use_sys_thread_name,
 				    "prof_experimental_use_sys_thread_name")
+				if (CONF_MATCH("prof_time_resolution")) {
+					if (CONF_MATCH_VALUE("default")) {
+						opt_prof_time_resolution = prof_time_resolution_default;
+					} else if (CONF_MATCH_VALUE("high")) {
+						if (!opt_high_resolution_timer) {
```
davidtgoldblatt, line 1482: We should probably also do a runtime-check, since it's possible that the build-time machine has CLOCK_REALTIME support but the run-time machine doesn't (i.e. actually try a gettime call and see if it fails with EINVAL).

src/nstime.c
```diff
@@ -152,12 +152,41 @@ nstime_monotonic_impl(void) {
 }
 nstime_monotonic_t *JET_MUTABLE nstime_monotonic = nstime_monotonic_impl;
 
+static void
+nstime_get_realtime(nstime_t *time) {
+  struct timespec ts;
+
+  clock_gettime(CLOCK_REALTIME, &ts);
```
davidtgoldblatt, line 159: I think this needs to be ifdef-guarded; CLOCK_REALTIME isn't guaranteed to be defined.

src/nstime.c
```diff
@@ -152,12 +152,41 @@ nstime_monotonic_impl(void) {
 }
 nstime_monotonic_t *JET_MUTABLE nstime_monotonic = nstime_monotonic_impl;
 
+static void
+nstime_get_realtime(nstime_t *time) {
+  struct timespec ts;
```
davidtgoldblatt, line 157: Nit -- we use freebsd kernel style, and tabs for indentation.

src/nstime.c
```diff
@@ -152,12 +152,41 @@ nstime_monotonic_impl(void) {
...
+  nstime_init2(time, ts.tv_sec, ts.tv_nsec);
+}
+
+static bool
+nstime_prof_update_impl(nstime_t *time) {
+  nstime_t old_time;
+
+	nstime_copy(&old_time, time);
+
+  if (unlikely(opt_prof_time_resolution == prof_time_resolution_high)) {
```
davidtgoldblatt, line 169: Is this unlikely? I suspect we'll turn it on by default.

src/nstime.c
```diff
@@ -152,12 +152,41 @@ nstime_monotonic_impl(void) {
...
+    nstime_get_realtime(time);
+  } else {
+    nstime_get(time);
+  }
+
+	/* Handle non-monotonic clocks. */
```
davidtgoldblatt, line 175: Given that this is prof-only, I sort of think we should allow time to go backwards; for the internal clocks, we need monotonicity for correctness (i.e. we algorithmically need a nondecreasing sequence). But if the user asks for timestamps per some particular clock, we might as well give it to them.

src/jemalloc.c
tyroguru: Just to note that any system that implements a clock_gettime() interface as defined by POSIX.1 has to implement a CLOCK_REALTIME clock_id. A cursory reading of the spec (https://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_getres.html) seems to indicate that CLOCK_REALTIME is probably the only mandatory clock_id that has to be implemented.

Would we expect any system not to implement a minimal posix compaint clock_gettime()?

src/nstime.c
tyroguru: Let me know what you think about the posix comment above and I'll act accordingly.

src/jemalloc.c
davidtgoldblatt: Ah, makes sense to me, I don't think we need a runtime check then.

src/nstime.c
davidtgoldblatt: (I think this one still applies, since e.g. windows doesn't have these).

tyroguru: I may well be missing something on platform support so please bear with me. I don't see how this code could be called on a Windows platform.

This code can only be called on a platform where 'opt_prof_time_resolution == prof_time_resolution_high' and the opt processing code code in jemalloc.c ensures that this can only be set if the compile time code detected that  clock_gettime(CLOCK_REALTIME) was available on this platform, i.e., config_high_resolution_timer = true: this should never be the case for a Windows compilation environment. 

Like I say, I may be missing something.

src/nstime.c
tyroguru: OK - I'll drop the unlikely().

src/nstime.c
tyroguru: I had included the non-monotonic check just to keep the previous behaviour exactly the same. I agree with you completely though in that we (jemalloc) shouldn't be interfering with timestamps for profiling and that should be left up to the consumer. If this monotonic guarantee for profiling isn't documented I'm happy to remove it.

tyroguru: @davidtgoldblatt - I just copied nstime_update_impl() here and I just noticed what a strange interface definition it has. A return of 'false' indicates the work was done successfully because their isn;t a possible failure really. A return of 'true' indicates that a clock went backwards and that was corrected! Hmm. It looks like nobody checks the return values anywhere for this interface.

I'm going to not return anything in my next revisiopn so let me know if you prefer nstime_prof_update_impl() to keep the same return type as nstime_update_impl() .

davidtgoldblatt: > If this monotonic guarantee for profiling isn't documented I'm happy to remove it.

I'd say go ahead and remove it. We only call the `_prof` variants down prof pathways, so it won't affect non-prof logic.

> I'm going to not return anything in my next revision so let me know if you prefer nstime_prof_update_impl() to keep the same return type as nstime_update_impl() .

We might as well just have all of these return void; I think it's probably only historical cruft that makes us not do so.

tyroguru:
Looking for advice on linkage issues for the test suites.

I have introduced a reference in nstime.c to a variable (`opt_prof_time_resolution`) that is allocated globally in jemalloc.c and therefore this variable sits in the BSS of jemalloc.o. This is causing problems for the integration test suite as it links in nstime.o but doesn't link in jemalloc.o and therefore we get an undefined symbol at link time, i.e., 

```
$ gcc   -o test/integration/aligned_alloc test/integration/aligned_alloc.o test/src/btalloc.integration.o test/src/btalloc_0.integration.o test/src/btalloc_1.integration.o test/src/math.integration.o test/src/mtx.integration.o test/src/sleep.integration.o test/src/SFMT.integration.o test/src/test.integration.o test/src/thd.integration.o test/src/timer.integration.o src/nstime.integration.o src/malloc_io.integration.o -Wl,-rpath,lib lib/libjemalloc.so  -pthread -lm
src/nstime.integration.o: In function `nstime_prof_update_impl':
/data/users/jonhaslam/jemalloc/run_tests.out.orig/run_test_0.out/../../src/nstime.c:171: undefined reference to `je_opt_prof_time_resolution'
/data/users/jonhaslam/jemalloc/run_tests.out.orig/run_test_0.out/../../src/nstime.c:171: undefined reference to `je_opt_prof_time_resolution'
```

I have tried:

- Adding "-Wl,--unresolved-symbols=ignore-in-object-files" to the link flags for the integration test suite as this is understood by gcc and clang but the MWING* Windows environments don't understand that linker option.
- linking jemalloc.o in the integration test suite (as the jet test suite does) and it, not surprisingly, have a massive list of dependencies itself - we'd have to link in a ton of extra stuff for that.

So, any help would be appreciated as I'm not sure which way to go with this. Maybe I've missed something obvious but I can't see what.

src/jemalloc.c
```diff
@@ -1497,6 +1498,21 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				CONF_HANDLE_BOOL(
 				    opt_prof_experimental_use_sys_thread_name,
 				    "prof_experimental_use_sys_thread_name")
+				if (CONF_MATCH("prof_time_resolution")) {
+					if (CONF_MATCH_VALUE("default")) {
+						opt_prof_time_resolution = prof_time_resolution_default;
```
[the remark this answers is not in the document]
tyroguru: That 8 space shiftwidth/tabstop sure do make it difficult to get into 80 characters. I've refactored all my variable names so that "resolution" has been contracted to "res" so that lines fit in the 80 chars.

src/nstime.c
tyroguru: > Just a couple of style nits now. If you could squash down to 1 commit, I think this is good to go.

I had a few problems with my squash but I hopefully everything is right now. Apologies up front if I missed anything.

davidtgoldblatt:
Looks good!

PR #1831, opened 2020-04-23
Android build fix proposal.

devnexen:
These are detected at configure time while they are glibc
specifics. the bionic equivalent is not api compatible
and dlopen is restricted in this platform.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -649,10 +649,11 @@ case "${host}" in
   *-*-bitrig*)
 	abi="elf"
 	;;
-  *-*-linux-android)
+  *-*-linux-android*)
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
I'm not sure I completely understand the underlying cause; is this necessary because the compilation test is getting run against glibc, but the actual generated code is getting run against bionic?

davidtgoldblatt:
I'm also curious if (+cc) @cferris1000 has any thoughts.

devnexen:
Because the bionic hooks signature are incompatibles (ie the caller in addition).

davidtgoldblatt:
I think this was overlooked by mistake -- it seems reasonable to me.

PR #1837, opened 2020-04-29
Move opt.zero zeroing decisions earlier in the allocation life-cycle

davidtgoldblatt:
This is both cleaner (it makes it so that the core internals don't know about opt.zero; they only know about the specific requests made of them), and makes the opt.zero case faster (since the extent-level allocator can madvise-zero the pages, avoiding having to memset them).

Tested the opt.zero improvement using the new benchmark with opt.zero on and off.

We take an extra branch now on realloc pathways, but it should mostly be correctly predicted (and those aren't hot paths to begin with).

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2165,7 +2165,9 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd) {
 	}
 
 	/* This is the beginning of the "core" algorithm. */
-
+	if (config_fill && sopts->slow && opt_zero) {
+		dopts->zero = true;
+	}
 	if (dopts->alignment == 0) {
 		ind = sz_size2index(size);
 		if (unlikely(ind >= SC_NSIZES)) {
@@ -2263,12 +2265,9 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd) {
 
 	assert(usize == isalloc(tsd_tsdn(tsd), allocation));
 
-	if (config_fill && sopts->slow && !dopts->zero) {
-		if (unlikely(opt_junk_alloc)) {
-			junk_alloc_callback(allocation, usize);
-		} else if (unlikely(opt_zero)) {
-			memset(allocation, 0, usize);
-		}
+	if (config_fill && sopts->slow && !dopts->zero
+	    && unlikely(opt_junk_alloc)) {
+		junk_alloc_callback(allocation, usize);
 	}
 
 	if (sopts->slow) {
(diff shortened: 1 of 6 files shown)


test/stress/large_microbench.c
```diff
@@ -0,0 +1,35 @@
+#include "test/jemalloc_test.h"
+#include "test/bench.h"
+
+static void
+large_mallocx_free(void) {
+	void *p = mallocx(128 * 1024, 0);
```
yinan1048576, line 6: nit: use the smallest large size macro to be more general?

test/stress/large_microbench.c
```diff
@@ -0,0 +1,35 @@
...
+	if (p == NULL) {
+		test_fail("Unexpected mallocx() failure");
+		return;
+	}
```
yinan1048576, line 10: nit: `assert_ptr_not_null()` to be more compact?

src/jemalloc.c
```diff
@@ -3275,14 +3278,11 @@ do_rallocx(void *ptr, size_t size, int flags, bool is_realloc) {
 	UTRACE(ptr, size, p);
 	check_entry_exit_locking(tsd_tsdn(tsd));
 
-	if (config_fill && malloc_slow && !zero && usize > old_usize) {
+	if (config_fill && unlikely(opt_junk_alloc) && usize > old_usize
+	    && !zero) {
```
yinan1048576, line 3282: Why there is no `malloc_slow`?

src/jemalloc.c
```diff
@@ -3561,16 +3565,11 @@ je_xallocx(void *ptr, size_t size, size_t extra, int flags) {
...
-			size_t excess_len = usize - old_usize;
-			void *excess_start = (void *)((uintptr_t)ptr + old_usize);
-			if (unlikely(opt_junk_alloc)) {
-				junk_alloc_callback(excess_start, excess_len);
-			} else if (unlikely(opt_zero)) {
-				memset(excess_start, 0, excess_len);
-			}
-		}
+	if (config_fill && unlikely(opt_junk_alloc) && usize > old_usize &&
+	    !zero) {
```
yinan1048576, line 3569: Same here.

test/stress/large_microbench.c
davidtgoldblatt: Will do.

test/stress/large_microbench.c
davidtgoldblatt: Will do.

src/jemalloc.c
davidtgoldblatt: opt_junk_alloc implies malloc_slow, but we'd still need to check it inside the branch (previously, checking it was helpful, since we folded both branches behind malloc_slow).

src/jemalloc.c
davidtgoldblatt: Ditto.

davidtgoldblatt:
Benchmark results of `MALLOC_CONF="zero:true" test/stress/large_microbench` before the second commit
```
1000000 iterations, large mallocx=8208006us, small mallocx=29000us, ratio=1:283.5
```
with the second commit:
```
1000000 iterations, large mallocx=948000us, small mallocx=28000us, ratio=1:33.57
```

PR #1840, opened 2020-05-01
LQ_QUANTUM should be 4 on mips64 hardware.

brooksdavis:
This matches the ABI stack alignment requirements.


brooksdavis:
We've been carrying this in CTSRD-CHERI/cheribsd and and to re-add it after merging 5.2.1 when updating our FreeBSD base.

davidtgoldblatt:
This looks good, thanks for the patch! I'm going to wait a bit before merging because of some internal release cadence issues, but should be able to in a week or two.

davidtgoldblatt:
Thanks!

PR #1847, opened 2020-05-13
Allow narenas:default.

davidtgoldblatt:
This can be useful when you know you want to override some lower-priority
configuration setting with its default value, but don't know what that value
would be.


davidtgoldblatt:
The reason narenas in particular is giving me trouble is that it's machine-dependent; i.e. I can just manually override (say) a different decay time, since the default is static. narenas is harder without knowing which machine I'm on.

PR #1849, opened 2020-05-14
Simplify flag reading and processing

yinan1048576:
This PR aims at removing redundant logic for flag reading and processing, and making the resulting code more readable and less error prone. It's also a preparation for #1828.

Also fixed a few issues along the way.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2105,6 +2105,10 @@ imalloc_no_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
 		arena = NULL;
 	} else {
 		arena = arena_get(tsd_tsdn(tsd), dopts->arena_ind, true);
+		if (unlikely(arena == NULL) &&
+		    dopts->arena_ind >= narenas_auto) {
+			return NULL;
+		}
 	}
 
 	if (unlikely(dopts->alignment != 0)) {
@@ -3252,7 +3256,7 @@ do_rallocx(void *ptr, size_t size, int flags, bool is_realloc) {
 	if (unlikely((flags & MALLOCX_ARENA_MASK) != 0)) {
 		unsigned arena_ind = MALLOCX_ARENA_GET(flags);
 		arena = arena_get(tsd_tsdn(tsd), arena_ind, true);
-		if (unlikely(arena == NULL)) {
+		if (unlikely(arena == NULL) && arena_ind >= narenas_auto) {
 			goto label_oom;
 		}
 	} else {
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2046,6 +2046,15 @@ aligned_usize_get(size_t size, size_t alignment, size_t *usize, szind_t *ind,
 	return false;
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+zero_get(bool guarantee, bool slow) {
+	if (config_fill && slow && unlikely(opt_zero)) {
+		return true;
+	} else {
+		return guarantee;
+	}
+}
+
 /* ind is ignored if dopts->alignment > 0. */
 JEMALLOC_ALWAYS_INLINE void *
 imalloc_no_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
@@ -2196,9 +2205,7 @@ imalloc_body(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd) {
 	}
 
 	/* This is the beginning of the "core" algorithm. */
-	if (config_fill && sopts->slow && opt_zero) {
-		dopts->zero = true;
-	}
+	dopts->zero = zero_get(dopts->zero, sopts->slow);
 	if (aligned_usize_get(size, dopts->alignment, &usize, &ind,
 	    sopts->bump_empty_aligned_alloc)) {
 		goto label_oom;
@@ -3230,10 +3237,7 @@ do_rallocx(void *ptr, size_t size, int flags, bool is_realloc) {
 	tsd = tsd_fetch();
 	check_entry_exit_locking(tsd_tsdn(tsd));
 
-	bool zero = flags & MALLOCX_ZERO;
-	if (config_fill && unlikely(opt_zero)) {
-		zero = true;
-	}
+	bool zero = zero_get(MALLOCX_ZERO_GET(flags), /* slow */ true);
 
 	if (unlikely((flags & MALLOCX_ARENA_MASK) != 0)) {
 		unsigned arena_ind = MALLOCX_ARENA_GET(flags);
@@ -3499,11 +3503,7 @@ je_xallocx(void *ptr, size_t size, size_t extra, int flags) {
 	tsd_t *tsd;
 	size_t usize, old_usize;
 	size_t alignment = MALLOCX_ALIGN_GET(flags);
-
-	bool zero = flags & MALLOCX_ZERO;
-	if (config_fill && unlikely(opt_zero)) {
-		zero = true;
-	}
+	bool zero = zero_get(MALLOCX_ZERO_GET(flags), /* slow */ true);
 
 	LOG("core.xallocx.entry", "ptr: %p, size: %zu, extra: %zu, "
 	    "flags: %d", ptr, size, extra, flags);
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2085,30 +2085,37 @@ tcache_get_from_ind(tsd_t *tsd, unsigned tcache_ind, bool slow, bool is_alloc) {
 	return tcache;
 }
 
-/* ind is ignored if dopts->alignment > 0. */
-JEMALLOC_ALWAYS_INLINE void *
-imalloc_no_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
-    size_t size, size_t usize, szind_t ind) {
-	arena_t *arena;
-
-	/* Fill in the tcache. */
-	tcache_t *tcache = tcache_get_from_ind(tsd, dopts->tcache_ind,
-	    sopts->slow, /* is_alloc */ true);
-
-	/* Fill in the arena. */
-	if (dopts->arena_ind == ARENA_IND_AUTOMATIC) {
+/* Return true if a manual arena is specified and arena_get() OOMs. */
+JEMALLOC_ALWAYS_INLINE bool
+arena_get_from_ind(tsd_t *tsd, unsigned arena_ind, arena_t **arena_p) {
+	if (arena_ind == ARENA_IND_AUTOMATIC) {
 		/*
 		 * In case of automatic arena management, we defer arena
 		 * computation until as late as we can, hoping to fill the
 		 * allocation out of the tcache.
 		 */
-		arena = NULL;
+		*arena_p = NULL;
 	} else {
-		arena = arena_get(tsd_tsdn(tsd), dopts->arena_ind, true);
-		if (unlikely(arena == NULL) &&
-		    dopts->arena_ind >= narenas_auto) {
-			return NULL;
+		*arena_p = arena_get(tsd_tsdn(tsd), arena_ind, true);
+		if (unlikely(*arena_p == NULL) && arena_ind >= narenas_auto) {
+			return true;
 		}
+	}
+	return false;
+}
+
+/* ind is ignored if dopts->alignment > 0. */
+JEMALLOC_ALWAYS_INLINE void *
+imalloc_no_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
+    size_t size, size_t usize, szind_t ind) {
+	/* Fill in the tcache. */
+	tcache_t *tcache = tcache_get_from_ind(tsd, dopts->tcache_ind,
+	    sopts->slow, /* is_alloc */ true);
+
+	/* Fill in the arena. */
+	arena_t *arena;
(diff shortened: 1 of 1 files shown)


src/jemalloc.c
```diff
@@ -2776,22 +2794,15 @@ free_default(void *ptr) {
...
-			}
+		bool fast = tsd_fast(tsd);
+		tcache_t *tcache = tcache_get_from_ind(tsd,
+		    TCACHE_IND_AUTOMATIC, !fast, /* is_alloc */ false);
+		if (!fast) {
 			uintptr_t args_raw[3] = {(uintptr_t)ptr};
 			hook_invoke_dalloc(hook_dalloc_free, ptr, args_raw);
-			ifree(tsd, ptr, tcache, true);
 		}
+		ifree(tsd, ptr, tcache, !fast);
```
davidtgoldblatt, line 2818: We need to repeat the `ifree` call twice -- once with `fast` being statically true, and the other with `fast` being statically false (i.e. inside the if branches). This is key to folding away all the slow-guarded condition checks on the fast path. Likewise -- dropping the `likely` here is potentially optimization-affecting.

src/jemalloc.c
```diff
@@ -2105,6 +2105,9 @@ imalloc_no_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
 		arena = NULL;
 	} else {
 		arena = arena_get(tsd_tsdn(tsd), dopts->arena_ind, true);
+		if (unlikely(arena == NULL)) {
```
davidtgoldblatt: I'm kinda neutral on this. There are two cases:
- The arena ID is a user-created arena, in which case it's a usage error if the user didn't already initialize it, and we are free to assert-fail, misbehave, etc.
- The arena ID is an auto-arena that simply hasn't been initialized yet. We already reserve the right to re-route auto arena requests (like for the huge arena or for reentrant allocations), and will probably stop listening to them entirely at some point.

No real preference either way though.

src/jemalloc.c
```diff
@@ -2147,6 +2147,9 @@ imalloc_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
 		arena_prof_promote(tsd_tsdn(tsd), ret, usize);
 	} else {
 		ret = imalloc_no_sample(sopts, dopts, tsd, usize, usize, ind);
+		if (unlikely(ret == NULL)) {
```
davidtgoldblatt, line 2158: Just as a commit title nit, this isn't really a fix so much as a clarification. Falling through is still correct.

src/jemalloc.c
```diff
@@ -2055,27 +2055,45 @@ zero_get(bool guarantee, bool slow) {
...
+		} else {
+			tcache = NULL;
 		}
-	} else if (dopts->tcache_ind == TCACHE_IND_NONE) {
-		tcache = NULL;
 	} else {
-		tcache = tcaches_get(tsd, dopts->tcache_ind);
+		/*
+		 * Not allowed to be reentrant and specify a custom tcache on
+		 * the free path.
```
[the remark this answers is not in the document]
yinan1048576: What might be the possible reasoning to support this claim?

src/jemalloc.c
yinan1048576: Right, makes sense. Also repeating the `tcache_get_from_ind()` call. I think now it should work exactly as before.

src/jemalloc.c
yinan1048576: Makes sense. I spotted this because we're not handling it consistently: in `malloc()` we fall back but in `realloc()` we OOM. I think we at least need to make them consistent.

I'm changing it so that (by checking the id against `narenas_auto`) we fail if it's a manually created arena and we fallback if it's an auto arena. Does that sound better?

src/jemalloc.c
yinan1048576: Right. Just found that `NULL` can also pass `prof_sample_aligned()`.

src/jemalloc.c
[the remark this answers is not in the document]
davidtgoldblatt: I'm remembering more. I think we originally just wanted to support the bare minimum amount of reentrancy required (libc functions), and no libc functions pass flags in. So we banned them, just out of a desire to minimize the amount of reentrancy we might accidentally allow.

src/jemalloc.c
davidtgoldblatt: TBH I think anything is fine; I was just giving a little context on the history. Maybe `assert(arena != NULL || arena_ind < opt_narenas)` and whatever's easiest after that. Surfacing the OOM earlier than strictly necessary is fine IMO.

src/jemalloc.c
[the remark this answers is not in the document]
yinan1048576: Got it. That makes sense. Let me revise the comment.

src/jemalloc.c
yinan1048576: As you were mentioning about the fast path, I realized that I could also unify the fast path so that it calls this centralized function. (And in that case I'll make sure that the assembly stays equally efficient.)

src/jemalloc.c
yinan1048576: Please let me know what you think of my new revision. BTW I think it'd be `narenas_auto` rather than `opt_narenas` to be more rigorous, though the two should rarely be different.

yinan1048576:
I'll do a more thorough revision to improve on the performance aspect.

yinan1048576:
Well I soon got into an "over-optimization" mood, so I decided to exercise some self-control and stay calm and back out a bit. Here's the principle I'm following at this moment:
- Alignment flag and zero flag: we have not been doing branch predictions in most places, presumably because they are more like standard utilities for ordinary callers to use. So in these cases, I try to prioritize code simplicity over performance.
- tcache flag and arena flag: we have been doing branch predictions in all places, presumably because they are more tied to jemalloc internals and are only for advanced users. So in these cases, I tilt to performance whenever applicable, while still trying best to simplify code.

Other changes include:
- Avoid duplicating branch predictions.
- Invoke the tcache selector on malloc / free fast path.

src/jemalloc.c
```diff
@@ -2015,6 +2015,37 @@ dynamic_opts_init(dynamic_opts_t *dynamic_opts) {
...
+		if (bump_empty_aligned_alloc && unlikely(size == 0)) {
+			size = 1;
+		}
+		*usize = sz_sa2u(size, alignment);
+	}
+	if (unlikely(*usize == 0 || *usize > SC_LARGE_MAXCLASS)) {
+		return true;
+	}
+	return false;
+}
```
yinan1048576, line 2047: Branch prediction is only used for the out of range case and in the `bump_empty_aligned_alloc` handling.

src/jemalloc.c
```diff
@@ -3099,9 +3113,7 @@ je_mallocx(size_t size, int flags) {
 	dopts.num_items = 1;
 	dopts.item_size = size;
 	if (unlikely(flags != 0)) {
-		if ((flags & MALLOCX_LG_ALIGN_MASK) != 0) {
-			dopts.alignment = MALLOCX_ALIGN_GET_SPECIFIED(flags);
-		}
+		dopts.alignment = MALLOCX_ALIGN_GET(flags);
```
yinan1048576, line 3159: This can be slightly more costly but the overhead should be small thanks to the `unlikely(flags != 0)` guard above.

src/jemalloc.c
```diff
@@ -3253,9 +3265,7 @@ do_rallocx(void *ptr, size_t size, int flags, bool is_realloc) {
 	hook_ralloc_args_t hook_args = {is_realloc, {(uintptr_t)ptr, size,
 		flags, 0}};
 	if (config_prof && opt_prof) {
-		usize = (alignment == 0) ?
-		    sz_s2u(size) : sz_sa2u(size, alignment);
-		if (unlikely(usize == 0 || usize > SC_LARGE_MAXCLASS)) {
+		if (aligned_usize_get(size, alignment, &usize, NULL, false)) {
```
yinan1048576, line 3283: No `unlikely` here, to avoid duplicated branch prediction.

src/jemalloc.c
```diff
@@ -3663,13 +3665,9 @@ je_dallocx(void *ptr, int flags) {
...
 	check_entry_exit_locking(tsdn);
-
 	size_t usize;
-	if (likely((flags & MALLOCX_LG_ALIGN_MASK) == 0)) {
-		usize = sz_s2u(size);
-	} else {
-		usize = sz_sa2u(size, MALLOCX_ALIGN_GET_SPECIFIED(flags));
-	}
+	/* In case of out of range, let the user see it rather than fail. */
+	aligned_usize_get(size, MALLOCX_ALIGN_GET(flags), &usize, NULL, false);
```
yinan1048576, line 3657: The original branch prediction is effectively lost, and the direct use of `MALLOCX_ALIGN_GET()` further adds to the overhead, but I assume callers of `inallocx()` may not care that much about performance anyway.

src/jemalloc.c
```diff
@@ -2046,6 +2046,15 @@ aligned_usize_get(size_t size, size_t alignment, size_t *usize, szind_t *ind,
...
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+zero_get(bool guarantee, bool slow) {
+	if (config_fill && slow && unlikely(opt_zero)) {
+		return true;
+	} else {
+		return guarantee;
+	}
+}
```
yinan1048576, line 2056: No branch prediction for zero flag.

src/jemalloc.c
```diff
@@ -2055,27 +2055,45 @@ zero_get(bool guarantee, bool slow) {
...
+		assert(is_alloc || tsd_reentrancy_level_get(tsd) == 0 ||
+		    tsd_state_nocleanup(tsd));
+		if (tcache_ind == TCACHE_IND_NONE) {
+			tcache = NULL;
+		} else {
+			tcache = tcaches_get(tsd, tcache_ind);
+		}
 	}
+	return tcache;
+}
```
yinan1048576, line 2086: Branch prediction is only used for testing fast path and testing reentrancy (on deallocation path). Do not do branch prediction on `tcache_ind == TCACHE_IND_AUTOMATIC`, but instead leave it to `mallocx_tcache_get()`, to avoid duplicated branch predictions.

src/jemalloc.c
```diff
@@ -3124,16 +3144,7 @@ je_mallocx(size_t size, int flags) {
...
-			if ((flags & MALLOCX_TCACHE_MASK)
-			    == MALLOCX_TCACHE_NONE) {
-				dopts.tcache_ind = TCACHE_IND_NONE;
-			} else {
-				dopts.tcache_ind = MALLOCX_TCACHE_GET(flags);
-			}
-		} else {
-			dopts.tcache_ind = TCACHE_IND_AUTOMATIC;
-		}
+		dopts.tcache_ind = mallocx_tcache_get(flags);
```
yinan1048576, line 3161: In this case, the branch prediction inside `mallocx_tcache_get()` becomes redundant to some extent, given the `unlikely(flags != 0)` guard above, but it likely would not do harm.

src/jemalloc.c
```diff
@@ -2085,30 +2085,37 @@ tcache_get_from_ind(tsd_t *tsd, unsigned tcache_ind, bool slow, bool is_alloc) {
...
+		 */
+		*arena_p = NULL;
+	} else {
+		*arena_p = arena_get(tsd_tsdn(tsd), arena_ind, true);
+		if (unlikely(*arena_p == NULL) && arena_ind >= narenas_auto) {
+			return true;
+		}
+	}
+	return false;
+}
```
yinan1048576, line 2105: Branch prediction is only used for testing the out of range case. No prediction on `arena_ind == ARENA_IND_AUTOMATIC`, but instead leave it to `mallocx_arena_get()`, to avoid duplications.

src/jemalloc.c
```diff
@@ -3145,13 +3157,9 @@ je_mallocx(size_t size, int flags) {
...
 	if (unlikely(flags != 0)) {
 		dopts.alignment = MALLOCX_ALIGN_GET(flags);
-
 		dopts.zero = MALLOCX_ZERO_GET(flags);
-
 		dopts.tcache_ind = mallocx_tcache_get(flags);
-
-		if ((flags & MALLOCX_ARENA_MASK) != 0)
-			dopts.arena_ind = MALLOCX_ARENA_GET(flags);
+		dopts.arena_ind = mallocx_arena_get(flags);
```
yinan1048576, line 3162: Likewise, the branch prediction in `mallocx_arena_get()` is an overkill to some extent, but there should be no harm.

src/jemalloc.c
```diff
@@ -3253,14 +3261,9 @@ do_rallocx(void *ptr, size_t size, int flags, bool is_realloc) {
...
-	if (unlikely((flags & MALLOCX_ARENA_MASK) != 0)) {
-		unsigned arena_ind = MALLOCX_ARENA_GET(flags);
-		arena = arena_get(tsd_tsdn(tsd), arena_ind, true);
-		if (unlikely(arena == NULL) && arena_ind >= narenas_auto) {
-			goto label_oom;
-		}
-	} else {
-		arena = NULL;
+	unsigned arena_ind = mallocx_arena_get(flags);
+	if (arena_get_from_ind(tsd, arena_ind, &arena)) {
```
yinan1048576, line 3265: No `unlikely` so as to avoid duplication.

yinan1048576:
Also confirmed that the assembly of `malloc()`, `free()`, and `sdallocx_noflags()` stays the same in optimized build.

PR #1852, opened 2020-05-27
Add a small program to print data structure sizes.

davidtgoldblatt:


davidtgoldblatt:
I'm fairly ambivalent about names. I don't think there's a meaningful risk of confusion if we internally redefine "stress" to just mean "helper binaries with non-pass/fail output", but I don't oppose some sort of alternate name either.

Issue #1850, opened 2020-05-15
Tracking the peak memory used by individual threads ?


fwyzard:
Hi,
I would be interested in resurrecting the idea discussed in #1176, that is, implement the possibility of measuring the peak memory usage within a section of code that is running on a single task.

From #1176:

> In the context of profiling a modular application, I am interested in the peak memory used by each thread, within each module. Here "module" could be a function call, a TBB task, _etc._ that is going to run within a given thread.
> 
> Something like
>   - at the begin of each module, check the memory usage and reset the peak value;
>   - while the module is running, track each memory allocation and deallocation and update the peak memory usage accordingly;
>   - at the end of the module, read the peak memory usage.
> 
> `jemalloc` can already track each allocation and deallocation; if we define the "memory usage" as "allocated memory - deallocated memory" (and let the client code take care of the offset) a possible solution is to add one more field to the `tsd` and track the peak "allocated - deallocated memory".

On the other hand, I see that the relevant code has seen quite some changes after jemalloc `5.2.1`.

Answering to #1176 
  - @interwq [mentioned](https://github.com/jemalloc/jemalloc/pull/1176#issuecomment-381698379) allocation hooks as a coming experimental feature. Did that happen ? Is it something I can try to use for this ?
  - @jasone [mentioned](https://github.com/jemalloc/jemalloc/pull/1176#issuecomment-382070521) the possibility of implementing a different approach based on thread heap profiling. Is that still a viable option ?  If yes, could you share some pointers of where to start looking for trying to implement it ?

Thanks,
.Andrea

davidtgoldblatt:
I think https://github.com/jemalloc/jemalloc/pull/1853 should solve this issue accurately enough for most reasonable use cases without forcing the allocation fast path to maintain extra tracking state.

fwyzard:
Thanks @davidtgoldblatt , that looks pretty interesting, and should cover our use case.

PR #1853, opened 2020-05-28
Add (approximate) per-thread peak memory tracking

davidtgoldblatt:
This resolves https://github.com/jemalloc/jemalloc/issues/1850 and https://github.com/jemalloc/jemalloc/pull/1176 .

Note that this doesn't (or at least, shouldn't) change fast paths at all -- we share the thread-event counter except during updates.

### DIFF
--- a/include/jemalloc/internal/peak.h
+++ b/include/jemalloc/internal/peak.h
@@ -1,0 +1,37 @@
+#ifndef JEMALLOC_INTERNAL_PEAK_H
+#define JEMALLOC_INTERNAL_PEAK_H
+
+typedef struct peak_s peak_t;
+struct peak_s {
+	/* The highest recorded peak value, after adjustment (see below). */
+	uint64_t cur_max;
+	/*
+	 * The difference between alloc and dalloc at the last set_zero call;
+	 * this lets us cancel out the appropriate amount of excess.
+	 */
+	uint64_t adjustment;
+};
+
+#define PEAK_INITIALIZER {0, 0}
+
+static inline uint64_t
+peak_max(peak_t *peak) {
+	return peak->cur_max;
+}
+
+static inline void
+peak_update(peak_t *peak, uint64_t alloc, uint64_t dalloc) {
+	int64_t candidate_max = (int64_t)(alloc - dalloc - peak->adjustment);
+	if (candidate_max > (int64_t)peak->cur_max) {
+		peak->cur_max = candidate_max;
+	}
+}
+
+/* Resets the counter to zero; all peaks are now relative to this point. */
+static inline void
+peak_set_zero(peak_t *peak, uint64_t alloc, uint64_t dalloc) {
+	peak->cur_max = 0;
+	peak->adjustment = alloc - dalloc;
+}
+
+#endif /* JEMALLOC_INTERNAL_PEAK_H */
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -53,10 +53,12 @@ void tsd_te_init(tsd_t *tsd);
  *  E(event,		(condition), is_alloc_event)
  */
 #define ITERATE_OVER_ALL_EVENTS						\
-    E(tcache_gc,	(opt_tcache_gc_incr_bytes > 0), true)		\
-    E(prof_sample,	(config_prof && opt_prof), true)	    	\
-    E(stats_interval,	(opt_stats_interval >= 0), true)	    	\
-    E(tcache_gc_dalloc,	(opt_tcache_gc_incr_bytes > 0), false)
+    E(tcache_gc,		(opt_tcache_gc_incr_bytes > 0), true)	\
+    E(prof_sample,		(config_prof && opt_prof), true)  	\
+    E(stats_interval,		(opt_stats_interval >= 0), true)   	\
+    E(tcache_gc_dalloc,		(opt_tcache_gc_incr_bytes > 0), false)	\
+    E(peak_alloc,		config_stats, true)			\
+    E(peak_dalloc,		config_stats, false)
 
 #define E(event, condition_unused, is_alloc_event_unused)		\
     C(event##_event_wait)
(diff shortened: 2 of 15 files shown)


fwyzard:
Thanks @davidtgoldblatt , this looks pretty interesting, and should cover our use case.

Looking at the tests, I'm not sure I understand this comment: https://github.com/jemalloc/jemalloc/blob/b67604d2aa9ea1cffa09743106e1d2639beb8738/test/unit/mallctl.c#L987-L990 

Does it mean that a deallocation should always trigger an update of the peak measurement ?

davidtgoldblatt:
Good point, I updated the comment. All I'm really trying to say there is, the peak-tracking shouldn't depend on us doing a mallctl at the peak (as the previous testing had). If the net bytes allocated increased and then decreased, we should remember the increase without having to do a mallctl in between.

fwyzard:
Understood :+1:

include/jemalloc/internal/peak.h
```diff
@@ -0,0 +1,32 @@
...
+peak_max(peak_t *peak) {
+	return peak->cur_max;
+}
+
+static inline void
+peak_update(peak_t *peak, uint64_t alloc, uint64_t dalloc) {
+	int64_t candidate_max = (int64_t)(alloc - dalloc - peak->adjustment);
+	if (candidate_max > (int64_t)peak->cur_max) {
+		peak->cur_max = candidate_max;
+	}
```
yinan1048576, line 22: Casting to `int64_t` might risk underflow / overflow issue. Can the logic here be implemented differently?

include/jemalloc/internal/thread_event.h
```diff
@@ -53,10 +53,12 @@ void tsd_te_init(tsd_t *tsd);
...
-    E(tcache_gc,	(opt_tcache_gc_incr_bytes > 0), true)		\
-    E(prof_sample,	(config_prof && opt_prof), true)	    	\
-    E(stats_interval,	(opt_stats_interval >= 0), true)	    	\
-    E(tcache_gc_dalloc,	(opt_tcache_gc_incr_bytes > 0), false)
+    E(tcache_gc,		(opt_tcache_gc_incr_bytes > 0), true)	\
+    E(prof_sample,		(config_prof && opt_prof), true)  	\
+    E(stats_interval,		(opt_stats_interval >= 0), true)   	\
+    E(tcache_gc_dalloc,		(opt_tcache_gc_incr_bytes > 0), false)	\
+    E(peak_alloc,		config_stats, true)			\
+    E(peak_dalloc,		config_stats, false)
```
yinan1048576, line 61: `thread.allocated` and `thread.deallocated` should now be updated even when `stats` is off, but it also makes sense if you want to turn off the peak tracking functionalities when `stats` is off.

src/peak_event.c
```diff
@@ -0,0 +1,69 @@
...
+peak_alloc_new_event_wait(tsd_t *tsd) {
+	return TE_MIN_START_WAIT;
+}
+
+uint64_t
+peak_alloc_postponed_event_wait(tsd_t *tsd) {
+	return TE_MIN_START_WAIT;
+}
+
+uint64_t
```
yinan1048576, line 49: Did you see any warning? It should at least warn because the signature of the event handler function differ from the declarations in `src/thread_event.c`. The event handler function is not supposed to return the wait time.

src/peak_event.c
```diff
@@ -0,0 +1,69 @@
...
+peak_alloc_event_handler(tsd_t *tsd, uint64_t elapsed) {
+	peak_event_update(tsd);
+	return PEAK_EVENT_WAIT;
+}
+
+uint64_t
+peak_dalloc_new_event_wait(tsd_t *tsd) {
+	return TE_MIN_START_WAIT;
```
yinan1048576, line 57: `TE_MIN_START_WAIT` is basically the next allocation - so do you mean to trigger the update event at every allocation? Or, you actually meant `PEAK_EVENT_WAIT`?

include/jemalloc/internal/peak.h
```diff
@@ -0,0 +1,32 @@
+#ifndef JEMALLOC_INTERNAL_PEAK_H
+#define JEMALLOC_INTERNAL_PEAK_H
+
+typedef struct peak_s peak_t;
+struct peak_s {
+	uint64_t cur_max;
+	uint64_t adjustment;
+};
```
yinan1048576, line 8: Could you add some comments on the meaning of the two fields?

include/jemalloc/internal/peak.h
davidtgoldblatt: This is actually pretty safe -- the only underflow/overflow issues occur when the net bytes allocated by a thread exceed INT64_MAX between reset calls. (I.e. we're resilient to thread.allocated/deallocated wraparound outside of that case).

include/jemalloc/internal/thread_event.h
davidtgoldblatt: Yeah; I think we don't want to allow the mallctl while stats are disabled (even though it's free today, it might not be in perpetuity). So given that, no sense in the extra updates (rare and cheap though they may be).

src/peak_event.c
davidtgoldblatt: No warnings. Those declarations aren't exposed in the header, so there's no way they could trigger a warning (and of course, it links and runs fine). Will fix.

src/peak_event.c
davidtgoldblatt: Yeah, should be PEAK_EVENT_WAIT. Fixed.

include/jemalloc/internal/peak.h
davidtgoldblatt: Done.

src/peak_event.c
yinan1048576: Yeah I see. Do you think I should have structured the declarations in a different way? It's kind of like specifying an interface, but it's now obvious that I wasn't able to make it do its job. Perhaps putting the declarations in `thread_event.h` instead of in `thread_event.c`? There could be some risk of circular dependencies though. For example for the case of your `peak` functionality, there'd be a `tsd` - `thread_event` - `peak` - `tsd` dependency loop, though I think in theory we are able to resolve the loop.

PR #1856, opened 2020-06-11
Introduce a couple more PA-named modules.

davidtgoldblatt:
This introduces, PAI "page allocator interface", which abstracts over the underlying page-level operations that interest us. It then pulls out the extent selection functionality in PA into PAC, "page allocator classic".

Right now neither of these is terribly well motivated; they're part of a larger-scale plan:

The PAI interface will allow adding a cache of small extents, that can be used to take pressure off the central extent allocator's lock; the interface allows testing the cache independently of all the underlying allocator functionality.

Pulling out the PA logic into PAC will make it easier to shim in a hugepage-allocator.

### DIFF
--- a/include/jemalloc/internal/pa.h
+++ b/include/jemalloc/internal/pa.h
@@ -7,7 +7,16 @@
 #include "jemalloc/internal/edata_cache.h"
 #include "jemalloc/internal/emap.h"
 #include "jemalloc/internal/lockedint.h"
+#include "jemalloc/internal/pac.h"
 #include "jemalloc/internal/pai.h"
+
+/*
+ * The page allocator; responsible for acquiring pages of memory for
+ * allocations.  It picks the implementation of the page allocator interface
+ * (i.e. a pai_t) to handle a given page-level allocation request.  For now, the
+ * only such implementation is the PAC code ("page allocator classic"), but
+ * others will be coming soon.
+ */
 
 enum pa_decay_purge_setting_e {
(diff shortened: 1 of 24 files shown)


include/jemalloc/internal/pa.h
```diff
@@ -7,46 +7,17 @@
...
 #include "jemalloc/internal/lockedint.h"
-
-enum pa_decay_purge_setting_e {
-	PA_DECAY_PURGE_ALWAYS,
-	PA_DECAY_PURGE_NEVER,
-	PA_DECAY_PURGE_ON_EPOCH_ADVANCE
-};
-typedef enum pa_decay_purge_setting_e pa_decay_purge_setting_t;
+#include "jemalloc/internal/pac.h"
+#include "jemalloc/internal/pai.h"
```
yinan1048576, line 11: Not sure if I'm understanding it correctly, but at the end of the refactoring `pa` should not need `pai` in theory, right? `pa` only depends on each `pai` implementation it needs, and the implementations each include `pai`?

davidtgoldblatt:
> Curious: are real interfaces in other languages internally implemented like this, except for some more compile time checks?

Mostly; more typically, the instance of the object would just have a pointer to the struct containing function pointers, rather than embedding them in the object itself. I'm doing it here just because we expect instances of arenas to be rare, but calls to the page allocator to be somewhat common; this lets us avoid dirtying another page.

PR #1866, opened 2020-06-17
Some cleanups for source files

yinan1048576:
This is an answer to https://github.com/jemalloc/jemalloc/pull/1794#discussion_r438383165. It goes beyond #1794 so separate it out.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -117,7 +117,6 @@ C_SRCS := $(srcroot)src/jemalloc.c \
 	$(srcroot)src/extent.c \
 	$(srcroot)src/extent_dss.c \
 	$(srcroot)src/extent_mmap.c \
-	$(srcroot)src/hash.c \
 	$(srcroot)src/hook.c \
 	$(srcroot)src/inspect.c \
 	$(srcroot)src/large.c \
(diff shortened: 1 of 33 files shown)


davidtgoldblatt:
And, just to double-check: after this diff, there's nothing that matches `JEMALLOC_.*_C_`, right?

yinan1048576:
Right. What I did was exactly the `sed` command I put into the commit summary.

davidtgoldblatt:
Right, but what I mean is, we want to make sure that there's no `#ifdef ARENA_C_` or anything left (which the ifdef thing wouldn't catch). Checked though, looks OK.

davidtgoldblatt:
I think we can also remove hash.c, prng.c, and ticker.c.

yinan1048576:
> Right, but what I mean is, we want to make sure that there's no `#ifdef ARENA_C_` or anything left (which the ifdef thing wouldn't catch). Checked though, looks OK.

That's great point. Also checked, and as you mentioned, there isn't anything like that.

> I think we can also remove hash.c, prng.c, and ticker.c.

Makes sense. Added a commit.

PR #1868, opened 2020-06-18
Witness: Make ranks an enum.

davidtgoldblatt:
This lets us avoid having to increment a bunch of values manually every time we
add a new sort of lock.

### DIFF
--- a/include/jemalloc/internal/witness.h
+++ b/include/jemalloc/internal/witness.h
@@ -7,62 +7,65 @@
 /* LOCK RANKS */
 /******************************************************************************/
 
-/*
- * Witnesses with rank WITNESS_RANK_OMIT are completely ignored by the witness
- * machinery.
- */
+enum witness_rank_e {
+	/*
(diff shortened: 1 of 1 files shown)


include/jemalloc/internal/witness.h
```diff
@@ -7,62 +7,65 @@
...
+	WITNESS_RANK_EXTENTS,
+	WITNESS_RANK_EDATA_CACHE,
+
+	WITNESS_RANK_EMAP,
+	WITNESS_RANK_RTREE,
+	WITNESS_RANK_BASE,
+	WITNESS_RANK_ARENA_LARGE,
+	WITNESS_RANK_HOOK,
+
+	WITNESS_RANK_LEAF=0x1000,
```
yinan1048576, line 56: Any reason why we wouldn't have over 4096 lock types some day? :)

davidtgoldblatt: Gonna introduce a new comment type, the TODONT

PR #1869, opened 2020-06-19
Improve naming for prof system thread name option

yinan1048576:
A more thorough answer to https://github.com/jemalloc/jemalloc/pull/1794#discussion_r414054227.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -127,7 +127,7 @@ CTL_PROTO(opt_prof_final)
 CTL_PROTO(opt_prof_leak)
 CTL_PROTO(opt_prof_accum)
 CTL_PROTO(opt_prof_recent_alloc_max)
-CTL_PROTO(opt_prof_experimental_use_sys_thread_name)
+CTL_PROTO(opt_prof_sys_thread_name)
 CTL_PROTO(opt_prof_time_res)
 CTL_PROTO(opt_zero_realloc)
 CTL_PROTO(tcache_create)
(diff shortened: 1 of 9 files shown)


src/ctl.c
```diff
@@ -382,11 +382,10 @@ static const ctl_named_node_t opt_node[] = {
 	{NAME("prof_final"),	CTL(opt_prof_final)},
 	{NAME("prof_leak"),	CTL(opt_prof_leak)},
 	{NAME("prof_accum"),	CTL(opt_prof_accum)},
-	{NAME("prof_recent_alloc_max"), CTL(opt_prof_recent_alloc_max)},
-	{NAME("prof_experimental_use_sys_thread_name"),
-	    CTL(opt_prof_experimental_use_sys_thread_name)},
-	{NAME("zero_realloc"),	CTL(opt_zero_realloc)},
-	{NAME("prof_time_resolution"),	CTL(opt_prof_time_res)}
+	{NAME("prof_recent_alloc_max"),	CTL(opt_prof_recent_alloc_max)},
+	{NAME("prof_sys_thread_name"),	CTL(opt_prof_sys_thread_name)},
```
davidtgoldblatt, line 386: Any reason to drop `experimental` here?

I don't think it *really* matters, since we're not documenting this for a while and have warned all known users about possible semantics changes, but it doesn't hurt either.

yinan1048576: No particular reason, other than that "experimental" is a pretty long word.

Also, like you said, anything we're not documenting is kind of automatically "experimental". I don't think this system thread name feature is strikingly "more experimental" than other "experimental" features e.g. the high resolution timer, so we don't really have to stress on "experimental" for it.

PR #1872, opened 2020-06-26
Sample every allocation by default in tests

yinan1048576:
This is what we've been doing all the time, which makes sense because a non-zero `lg_prof_sample` brings randomness and is thus not good for tests. This PR serves to save some typing.


davidtgoldblatt:
I think this is probably a net-negative. Introducing behavior differences between test and non-test modes is sort of a bad pattern in general, especially when we have an existing knob for the behavior we'd want to control. We mostly use JET functionality only to make logically private interfaces public, or to allow mocking out of system-provided behavior.

(In fact, I'd like to go the other direction if we can, and kill off the jet build entirely -- it's a big contributor to long CI run times).

PR #1874, opened 2020-07-04
enabling mpss on solaris/illumos.

devnexen:
reusing slighty linux configuration as possible, aligning the
 address range to HUGEPAGE.


davidtgoldblatt:
Looks good, thanks!

PR #1875, opened 2020-07-06
Make high resolution timer default for profiling

yinan1048576:


davidtgoldblatt:
Is this correct? I don't know about the high-res timer guarantees on all the platforms we support (maybe POSIX guarantees realtime clocks too)?

yinan1048576:
I haven't looked into the code. Won't it fall back to the default timer?

davidtgoldblatt:
I think as a question of accuracy; CLOCK_REALTIME can go backwards, while CLOCK_MONOTONIC can't (I think?). So we can incorrectly (rather than just coarsely) report allocation lifetimes with this change.

yinan1048576:
Got it. Then let's leave the default timer as default.

> So we can incorrectly (rather than just coarsely) report allocation lifetimes with this change.

nit: the time should be of higher resolution rather than being coarse with this change?

PR #1876, opened 2020-07-06
Corrections for prof_log_start()

yinan1048576:
Before this change, the `prof_log_start()` call at init time is before the mutex and the timestamp are initialized. Also, the hash table creation can take place at the first freed sample rather than at init time.

### DIFF
--- a/src/prof_log.c
+++ b/src/prof_log.c
@@ -87,7 +87,7 @@ struct prof_alloc_node_s {
 };
 
 /*
- * Created on the first call to prof_log_start and deleted on prof_log_stop.
+ * Created on the first call to prof_try_log and deleted on prof_log_stop.
  * These are the backtraces and threads that have already been logged by an
  * allocation.
  */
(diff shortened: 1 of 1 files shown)


src/prof_log.c
```diff
@@ -673,36 +672,26 @@ prof_log_stop(tsdn_t *tsdn) {
...
 			abort();
 		}
-	}
-
-	if (malloc_mutex_init(&log_mtx, "prof_log",
-	    WITNESS_RANK_PROF_LOG, malloc_mutex_rank_exclusive)) {
-		return true;
-	}
-
-	if (ckh_new(tsd, &log_bt_node_set, PROF_CKH_MINITEMS,
```
davidtgoldblatt, line 693: Do we just not need to initialize these hash tables? How come?

yinan1048576: We can initialize them here but we don't have to. `prof_try_log()` checks about these tables and initializes them if needed - and it has to do so because there can be multiple segmented logging period. I think it's a good idea in itself to delay the table initialization until the first freed sample comes in, which is I believe also the intention of the original design (which I'm still learning via reading the code).

yinan1048576:
Rebase, and slight modification: `prof_log_stop_final()` is always registered by `atexit()` so that applications can start logging in the middle of the program run and collect the information at exit (which is the same as the behavior before).

PR #1885, opened 2020-07-22
Verify output space before doing heavy work in mallctl

yinan1048576:
Added the `VERIFY_READ()` utility in `ctl.c` to verify that the output space is enough, and use it in a few places where the work is heavy and is not read-only, e.g. creating tcache / arena.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -1494,6 +1494,7 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
 			size_t	copylen = (sizeof(t) <= *oldlenp)	\
 			    ? sizeof(t) : *oldlenp;			\
 			memcpy(oldp, (void *)&(v), copylen);		\
+			*oldlenp = copylen;				\
 			ret = EINVAL;					\
 			goto label_return;				\
 		}							\
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -1488,6 +1488,15 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
 	}								\
 } while (0)
 
+/* Verify that the space provided is enough. */
+#define VERIFY_READ(t)	do {						\
+	if (oldp == NULL || oldlenp == NULL || *oldlenp != sizeof(t)) {	\
+		*oldlenp = 0;						\
+		ret = EINVAL;						\
+		goto label_return;					\
+	}								\
+} while (0)
+
 #define READ(v, t)	do {						\
 	if (oldp != NULL && oldlenp != NULL) {				\
 		if (*oldlenp != sizeof(t)) {				\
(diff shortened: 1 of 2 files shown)


src/ctl.c
```diff
@@ -1494,6 +1494,7 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
 			size_t	copylen = (sizeof(t) <= *oldlenp)	\
 			    ? sizeof(t) : *oldlenp;			\
 			memcpy(oldp, (void *)&(v), copylen);		\
+			*oldlenp = sizeof(t);				\
```
davidtgoldblatt, line 1505: Huh, you know, I always sort of assumed this was already the semantics. But looking through the rest of ctl, it looks like we never actually write to `*oldlenp`, and the APIs its based on use it more for stuff like variable-sized data instead of the size hinting done here.

@jasone, can we summon you to give an authorial intent on what to do with too-small values in `*oldlenp`? `sysctlbyname` and friends will write as much as they can, and might write less than asked for and store the bytes written. For stuff like `arenas.create`, I'm not sure we can do anything super reasonable in trying to give the user (say) only the first byte of the arena id created.

jasone: The `mallctl` interface was initially modeled on BSD's [`sysctl(3)`](https://www.freebsd.org/cgi/man.cgi?query=sysctl&apropos=0&sektion=3&manpath=FreeBSD+12.1-stable&arch=default&format=html). However, the jemalloc man page precisely documents current behavior (we read as much data as possible, and we return `EINVAL` rather than `ENOMEM` as `sysctl` does). What we've lost by deviating from the original inspiration is a mechanism for querying how much space the caller should provide via `oldp`. I don't recall intentionally omitting such functionality.

As this diff is implemented, querying would become possible, but with non-NULL `oldp` and `*oldlenp == 0` -- a contorted combination in my opinion. I suggest that if the desire is to enable querying, we instead add code predicated by `if (oldp == NULL && oldlenp != NULL)`. And any semantic change should be documented in the man page.

As for fixing the existing code path, I think the change should instead be:
```suggestion
			*oldlenp = copylen;				\
```
The man page doesn't say what happens to `*oldlenp` in the `EINVAL` case, but I think setting it to `copylen` is consistent with intent (and we should document this).

davidtgoldblatt: Thanks! If the output parameter isn't big enough for the result to be meaningful (e.g. arenas.create with a 1-byte `*oldlenp`), would you rather still take the action (and leak the arena)? Or just return EINVAL and set `*oldlenp` to 0 without doing anything further?

(I guess with arenas things are a little tricky since you can plausibly guess an arena id through an out-of-band mechanism; really the context in which this came up is with an experimental batch-alloc API though, where not getting all the data we want to give is definitely a mistake).

jasone: Tough call. In practice I've always used the `mallctl` API as if `EINVAL` is a hard error, for which case it doesn't matter whether we try harder to avoid leaking arenas or tcaches. My slight preference would be to document tighter constraints on `oldlenp` for those particular cases that really matter, but make the default behavior follow the `sysctl` model. My reasoning is that `sysctl` was designed to support variable outputs, and we would eliminate that as a future possibility for `mallctl` if we universally prohibited partial copy-outs.

yinan1048576:
Revise according to feedback.

PR #1888, opened 2020-07-24
Add a flat bitmap type

davidtgoldblatt:
A coming hugepage allocator will need a fuller-featured set of operations (finding both set and unset bits in both directions, setting ranges of bits).

Some of these are difficult to implement on top of bitmap_t. In particular, searching for both bit types would require a parallel set of higher-level groups when tree-based bitmaps are turned on (one tracking set bits, one tracking unset bits). This is slower than bitmap_t for very large bitmaps, but in all the places where we need the advanced functionality we can straightforwardly bound bitmap sizes.

We require fls primitives in addition to ffs ones for this functionality. Once we have fls, much of the bit_util library can be more simply implemented. I cleaned it up since we're gaining it anyways.


test/unit/bit_util.c
```diff
@@ -147,9 +145,69 @@ TEST_BEGIN(test_ffs_zu) {
...
+TEST_BEGIN(test_fls_llu_slow) {
+	TEST_FLS(unsigned long long, llu_slow, qd, "llu");
+}
+TEST_END
+
+
 int
 main(void) {
-	return test(
+	return test_no_reentrancy(
```
yinan1048576, line 210: We don't do reentrancy tests here because they're not necessary, not because they'll have different test outcomes, right?

include/jemalloc/internal/flat_bitmap.h
```diff
@@ -0,0 +1,222 @@
+#ifndef JEMALLOC_INTERNAL_FB_H
```
yinan1048576, line 1: Nice abbreviation.

test/unit/bit_util.c
davidtgoldblatt: Right; none of the bit_utils do (or plausibly could do in later refactorings) things like complex libc calls. They're also among our longest-running tests, so not running them three times saves space.

include/jemalloc/internal/flat_bitmap.h
davidtgoldblatt: lol. It felt weird to keep writing it, but it also felt weird to deliberately pick what seemed like a worse name to avoid the abbreviation.

test/unit/bit_util.c
davidtgoldblatt: Err, time, not space.

PR #1892, opened 2020-07-29
Bypass tcache when a manual arena is specified

yinan1048576:
This is to ensure that the allocation always comes from the specified arena. This also means that the arena flag can "overwrite" the tcache flag.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2114,7 +2114,17 @@ zero_get(bool guarantee, bool slow) {
 }
 
 JEMALLOC_ALWAYS_INLINE tcache_t *
-tcache_get_from_ind(tsd_t *tsd, unsigned tcache_ind, bool slow, bool is_alloc) {
+tcache_get_from_ind(tsd_t *tsd, unsigned tcache_ind, bool slow, bool is_alloc,
+    arena_t *arena) {
+	if (unlikely(arena != NULL)) {
+		/*
+		 * In cases where a manual arena is specified, we bypass tcache
+		 * to guarantee that the allocation comes from the designated
+		 * arena.
+		 */
+		assert(is_alloc);
+		return NULL;
+	}
 	tcache_t *tcache;
 	if (tcache_ind == TCACHE_IND_AUTOMATIC) {
 		if (likely(!slow)) {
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
I think that this change is morally very right, but that this is probably not the right time to do it.

The interaction between tcaches and custom arenas is something that we (or at least, me and @interwq; I think most of the discussions were before you started) think need to be reworked in the next backwards-incompatible release, but should probably be done systematically rather than piecemeal. For example: this prevents one type of arena leak, but not others (e.g. tying the allocating thread to a manual arena, then freeing from a non-manual arena). I think the systematic thing may be something even stronger: something like requiring specifying an arena on free if you do so on alloc (as a practical matter, 100% of the violations of this rule have been bugs in my experience).

yinan1048576:
I don't think I fully follow your reasoning. Can you clarify a bit?

According to my understanding, the arena flag only works on the allocation path. The free path doesn't look at the arena flag, nor does it need to.

The key part here is the tcache, which by nature can contain allocations from different arenas even though it's officially tied to only one arena. Without the change in this PR, the tcache flag is a contract whereas the arena flag is a best effort; with this change, it becomes the reverse. I think if there has to be a choice, I would go for acknowledging the arena flag more, because I think this aligns better with the caller's intention.

The other reason why I'm putting up this PR is, for the batch allocation API, the tcache flag is not relevant (at least for now), but according to the current setup, the arena flag is treated as a best effort, which doesn't feel right. For example, the sampled allocations can come from the tcache so they can belong to a different arena. I'd either need to make this change, or specifically hack the logic for the batch allocation API, so that whenever the arena flag is present, we bypass the tcache there.

yinan1048576:
> I think the systematic thing may be something even stronger: something like requiring specifying an arena on free if you do so on alloc (as a practical matter, 100% of the violations of this rule have been bugs in my experience).

By "specifying an arena", do you really mean "bypassing the tcache", which is the same as my intention here?

yinan1048576:
Discussed offline. I'm putting some notes here for reference. Here's what I've got in my own language -

This PR ensures that whenever a manual arena is specified in an allocation request, the allocation being returned is always coming from the designated arena. However, the two caveats are:
- It comes with a significant performance overhead, which may not always be tolerable.
- The other side of the problem is still not solved: when a manual arena is not specified in an allocation request, the allocation being returned can still come from a manual arena, which is an equally serious issue.

Probably one approach for the long term would be to implement some kind of "manual arena dedicated" tcache, potentially with an additional requirement that any free call with a "manual arena sourced" allocation must have the same arena flag.

PR #1895, opened 2020-07-30
[CI only] Batch allocation investigate

yinan1048576:


yinan1048576:
The previously observed errors are no longer there on the same stack of commits. The cause might either be some new commit in dev branch, or the difference between assert and expect, which I don't want to switch due to possible OOM crashes. There's little value in continuing the investigation. Stopping here.

PR #1898, opened 2020-08-04
Add a debug mode: sized-deallocation checking

davidtgoldblatt:
This is fairly expensive on the sized deallocation pathways, so it can't be part of the regular opt safety checks. But it's handy to have around in cases where those checks have found that a bug exists but haven't pinpointed it.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2783,6 +2783,27 @@ ifree(tsd_t *tsd, void *ptr, tcache_t *tcache, bool slow_path) {
 	thread_dalloc_event(tsd, usize);
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+maybe_check_alloc_ctx(tsd_t *tsd, void *ptr, emap_alloc_ctx_t *alloc_ctx) {
+	if (config_opt_size_checks) {
+		emap_alloc_ctx_t dbg_ctx;
+		emap_alloc_ctx_lookup(tsd_tsdn(tsd), &arena_emap_global, ptr,
+		    &dbg_ctx);
+		if (alloc_ctx->szind != dbg_ctx.szind) {
+			safety_check_fail_sized_dealloc(
+			    /* curent_dealloc */ true);
+			return true;
+		}
+		if (alloc_ctx->slab != dbg_ctx.slab) {
+			safety_check_fail(
+			    "Internal heap corruption detected: "
+			    "mismatch in slab bit");
+			return true;
+		}
+	}
+	return false;
+}
+
 JEMALLOC_ALWAYS_INLINE void
 isfree(tsd_t *tsd, void *ptr, size_t usize, tcache_t *tcache, bool slow_path) {
 	if (!slow_path) {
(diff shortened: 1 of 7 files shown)


src/jemalloc.c
```diff
@@ -2783,6 +2783,29 @@ ifree(tsd_t *tsd, void *ptr, tcache_t *tcache, bool slow_path) {
 	thread_dalloc_event(tsd, usize);
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+maybe_check_alloc_ctx(tsd_t *tsd, void *ptr, emap_alloc_ctx_t *alloc_ctx) {
+	if (config_opt_size_checks) {
+		emap_alloc_ctx_t dbg_ctx;
+		emap_alloc_ctx_lookup(tsd_tsdn(tsd), &arena_emap_global, ptr,
+		    &dbg_ctx);
+		if (config_opt_size_checks) {
```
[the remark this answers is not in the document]
davidtgoldblatt: Whoops. Just refactoring noise. I'll fix it.

src/safety_check.c
```diff
@@ -12,7 +12,6 @@ void safety_check_fail_sized_dealloc(bool current_dealloc) {
 	   " application sized deallocation bugs (source: %s). Suggest building"
 	    "with --enable-debug or address sanitizer for debugging. Abort.\n",
 	    src);
-	abort();
```
yinan1048576, line 15: Should we `assert(config_opt_safety_checks);` at the beginning of this function? Perhaps it should be an or of the two options?

test/unit/size_check.c
```diff
@@ -0,0 +1,59 @@
...
+}
+
+TEST_BEGIN(test_invalid_size_sdallocx) {
+	test_skip_if(!config_opt_size_checks);
+	safety_check_set_abort(&fake_abort);
+
+	fake_abort_called = false;
+	void *ptr = malloc(SC_SMALL_MAXCLASS);
+	assert_ptr_not_null(ptr, "Unexpected failure");
+	sdallocx(ptr, SC_LARGE_MINCLASS, 0);
```
yinan1048576, line 18: Well, to be super picky here, I think checking for a different small size rather than a large size might be slightly better. One reason is that in most of the real cases, there will be a mismatch within the small size classes. Another reason is that these tests can still pass if there isn't the `alloc_ctx->szind != dbg_ctx.szind` branch at all in the `maybe_check_alloc_ctx()` function, which is the more important branch, and checking for small size mismatch can improve the test coverage.

src/safety_check.c
davidtgoldblatt: Eh; it seems this might be a pain if the number of checks we allow grows. I'm not sure it buys us much.

test/unit/size_check.c
davidtgoldblatt: Will do.

src/safety_check.c
yinan1048576: But, is it even correct? I mean, if `config_opt_safety_checks` is turned off, this `safety_check_fail_sized_dealloc()` function can still be called whenever `config_opt_size_checks` is turned on?

davidtgoldblatt: Oh, I see what you're saying; I missed the assert at the beginning of this function. Yep, you're absolutely right.

PR #1899, opened 2020-08-05
Record alignment and usize explicitly in prof recent records

yinan1048576:
A few reasons behind this PR:
- Computing `usize` via `sz_s2u(size)` when dumping the prof recent records is simply wrong, if there was an alignment specified.
- To be complete, the record should contain all three of request size, alignment, and `usize`.
- Trying to compute the `usize` from the request size and alignment may be correct but not ideal. On the one hand, the formula might not be that simple, e.g. considering a `realloc()` with a lower alignment specified. On the other hand, the prof modules should not have knowledge about size class computations. So I end up passing all three values around.

### DIFF
--- a/include/jemalloc/internal/prof_externs.h
+++ b/include/jemalloc/internal/prof_externs.h
@@ -49,7 +49,7 @@ prof_tdata_t *prof_tdata_reinit(tsd_t *tsd, prof_tdata_t *tdata);
 
 void prof_alloc_rollback(tsd_t *tsd, prof_tctx_t *tctx);
 void prof_malloc_sample_object(tsd_t *tsd, const void *ptr, size_t size,
-    size_t usize, prof_tctx_t *tctx);
+    size_t alignment, size_t usize, prof_tctx_t *tctx);
 void prof_free_sampled_object(tsd_t *tsd, size_t usize, prof_info_t *prof_info);
 prof_tctx_t *prof_tctx_create(tsd_t *tsd);
 void prof_idump(tsdn_t *tsdn);
--- a/include/jemalloc/internal/prof_inlines.h
+++ b/include/jemalloc/internal/prof_inlines.h
@@ -142,23 +142,24 @@ prof_alloc_prep(tsd_t *tsd, bool prof_active, bool sample_event) {
 }
 
 JEMALLOC_ALWAYS_INLINE void
-prof_malloc(tsd_t *tsd, const void *ptr, size_t size, size_t usize,
-    emap_alloc_ctx_t *alloc_ctx, prof_tctx_t *tctx) {
+prof_malloc(tsd_t *tsd, const void *ptr, size_t size, size_t alignment,
+    size_t usize, emap_alloc_ctx_t *alloc_ctx, prof_tctx_t *tctx) {
 	cassert(config_prof);
 	assert(ptr != NULL);
 	assert(usize == isalloc(tsd_tsdn(tsd), ptr));
 
 	if (unlikely((uintptr_t)tctx > (uintptr_t)1U)) {
-		prof_malloc_sample_object(tsd, ptr, size, usize, tctx);
+		prof_malloc_sample_object(tsd, ptr, size, alignment, usize,
+		    tctx);
 	} else {
 		prof_tctx_reset(tsd, ptr, alloc_ctx);
 	}
 }
 
 JEMALLOC_ALWAYS_INLINE void
-prof_realloc(tsd_t *tsd, const void *ptr, size_t size, size_t usize,
-    prof_tctx_t *tctx, bool prof_active, const void *old_ptr, size_t old_usize,
-    prof_info_t *old_prof_info, bool sample_event) {
+prof_realloc(tsd_t *tsd, const void *ptr, size_t size, size_t alignment,
+    size_t usize, prof_tctx_t *tctx, bool prof_active, const void *old_ptr,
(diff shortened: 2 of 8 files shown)


davidtgoldblatt:
I'm not opposed per se to recording usize explicitly if it's convenient, but the third bullet point seems wrong to me:

> On the one hand, the formula might not be that simple, e.g. considering a realloc() with a lower alignment specified.

The usize <-> (size, alignment) mapping is ironclad. After a successful realloc/rallocx/xallocx, the result can be passed to sdallocx with the same size and alignment parameters. If lowering the alignment changes the usize we'd compute from first principles, the allocation has to move. `sz_sa2u` should always be correct.

> On the other hand, the prof modules should not have knowledge about size class computations.

The size class modules are kind of at the bottom of the dependency graph (except for the libraries of purely self-contained functionality, like bit utilities and the data structure implementations). Basically anything can depend on them, and prof in particular sits near the top of the hierarchy (e.g. it does its own internal allocations, which definitely rely on size information; unbiasing will need it too).

Again, not active opposition or anything; just pointing out that it'd be fine either way. Especially here, where it's a cold path and we don't care so much about size, I think keeping usize around is reasonable if it's convenient.

yinan1048576:
> The usize <-> (size, alignment) mapping is ironclad. After a successful realloc/rallocx/xallocx, the result can be passed to sdallocx with the same size and alignment parameters. If lowering the alignment changes the usize we'd compute from first principles, the allocation has to move. `sz_sa2u` should always be correct.

Well, I meant `rallocx()`, and I meant a higher alignment. But your reasoning makes sense, and anything violating it should be a bug. I'll dig a bit into it.

> The size class modules are kind of at the bottom of the dependency graph (except for the libraries of purely self-contained functionality, like bit utilities and the data structure implementations). Basically anything can depend on them, and prof in particular sits near the top of the hierarchy (e.g. it does its own internal allocations, which definitely rely on size information; unbiasing will need it too).

Well, what I meant was more around the knowledge of the `sz_sa2z()` function. For example, in theory, we may not have to compute the `usize` based on a purely formulaic approach - that's a decision beyond what profiling should know.

yinan1048576:
> Well, I meant `rallocx()`, and I meant a higher alignment. But your reasoning makes sense, and anything violating it should be a bug. I'll dig a bit into it.

The result is #1900.

yinan1048576:
I have a tricky question around `xallocx()`: what should be the correct request size to record in profiling, e.g. in the recent records as well as the computation of internal fragmentation in the future?

Currently I'm putting in the request size, but I figure it may not be right.

One alternative is the following:
```
if (usize >= size + extra) {
    correct_request_size = size + extra;
} else if (usize >= size) {
    correct_request_size = size;
} else {
    correct_request_size = 0;
}
```

The reasoning behind the last branch is that the allocation will be of no use to the user, so that the range the user will actually use is 0 (and the user will release the allocation).

However, we might also argue that the `better_request_size` is always `usize`: since `usize` is returned to the user, the user can freely choose to utilize the entire range even though it was not the previous intention.

Thoughts?

davidtgoldblatt:
After some thinking, I think I have a unifying theory: for functions that return their usize, we should treat internal fragmentation as definitionally 0. This really just means xallocx and smallocx, but should probably be expanded to bump the reporting for sampled allocations that have malloc_usable_size or the like called on them. My reasoning is that users of those APIs are explicitly optimizing for using all the space we give them, rather than relying solely on their arguments.

This might not be theoretically true given just the man page, but I think it's practically true given how xallocx is used (i.e. vector expansion scenarios). If we return a usize larger than the small end of size, the user is going to end up filling up all the bytes up to usize.

In terms of what to record for recent records: why not just record everything? I don't think we're particularly space constrained there.

yinan1048576:
I agree on the reasoning behind treating internal fragmentation as 0 for cases where the caller has obtained information about the usable size. However I'm stuck with the following -

It feels to me that the intention of
```
void *p = mallocx(1234, 0);
size_t usable_size = malloc_usable_size(p); /* or sallocx(p, 0) */
```
and
```
size_t usable_size = nallocx(1234, 0);
void *p = mallocx(1234, 0);
```
should be the same. However, it looks like that it's doable to correct the internal fragmentation on the first case but not the second. Thoughts?

yinan1048576:
BTW, something I just want to note before I forget: The flags argument of `sallocx()` is not used at all and is also misleading to callers IMO. We should do something about it.

davidtgoldblatt:
> However, it looks like that it's doable to correct the internal fragmentation on the first case but not the second. Thoughts?

davidtgoldblatt:
> However, it looks like that it's doable to correct the internal fragmentation on the first case but not the second. Thoughts?

This is true, but I think unavoidable. Internal fragmentation reporting is fundamentally best-effort in the presence of nallocx. I wouldn't worry about malloc_useable_size or sallocx for now.

The time where that's a bigger issue is with large size classes, but I've got some longer-term plans there that I think make it less of an issue.

> The flags argument of sallocx() is not used at all and is also misleading to callers IMO. We should do something about it.

Not used now, but could be at some point in the future. This sort of thing is reasonably common for systems-y APIs.

yinan1048576:
The part of this PR on fixing the recording for `usize` is pulled out to #1917. The other part of it, on recording alignment (and potentially the `extra` parameter for `xallocx()`) is great but there isn't a real need at this moment. I'm temporarily closing this PR; will reopen it if real needs arise.

PR #1900, opened 2020-08-05
Fix size miscalculation bug in reallocation

yinan1048576:
Before this change, the snippet

```
tsdn_t *tsdn = tsdn_fetch();
void *p = mallocx(7, MALLOCX_ALIGN(64));
malloc_printf("isalloc(tsdn, p) = %zu\n", isalloc(tsdn, p));
p = rallocx(p, 57, MALLOCX_ALIGN(128));
malloc_printf("isalloc(tsdn, p) = %zu\n", isalloc(tsdn, p));
free(p);
```

yields

```
isalloc(tsdn, p) = 64
isalloc(tsdn, p) = 64
```

After this change, it yields

```
isalloc(tsdn, p) = 64
isalloc(tsdn, p) = 128
```

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -1297,7 +1297,7 @@ void *
 arena_ralloc(tsdn_t *tsdn, arena_t *arena, void *ptr, size_t oldsize,
     size_t size, size_t alignment, bool zero, tcache_t *tcache,
     hook_ralloc_args_t *hook_args) {
-	size_t usize = sz_s2u(size);
+	size_t usize = alignment == 0 ? sz_s2u(size) : sz_sa2u(size, alignment);
 	if (unlikely(usize == 0 || size > SC_LARGE_MAXCLASS)) {
 		return NULL;
 	}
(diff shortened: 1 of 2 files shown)


davidtgoldblatt:
Actually, nit: could we have a test that exercises this functionality too? Say, in test/integration/rallocx, using rallocx and then comparing the values from nallocx and sallocx?

yinan1048576:
Fixing test failures.

> Actually, nit: could we have a test that exercises this functionality too? Say, in test/integration/rallocx, using rallocx and then comparing the values from nallocx and sallocx?

Good point. Will try to write one. The tricky part is: I think, even if I had tried my best to write a comprehensive test, there could have been a good chance that it failed to cover cases like this one; along the same line of reasoning, the tests I'm going to write has a good chance of not being able to cover the case of the next to-be-discovered bug. But anyway I'll try to come up with some tests.

yinan1048576:
Added an integration stress test, to enumerate over the space of ("original usable size" X "new request size" X "new alignment"). The test failed miserably before the fix and passed after the fix.

yinan1048576:
Reduced the range of the test sizes, and see if the failure is gone.

PR #1901, opened 2020-08-06
Stress tests: cleanups, and a mallctl test.

davidtgoldblatt:
The backstory here is that I was curious of how fast mallctl interfaces were. I wrote a little stress test to determine how fast it was. Then I got frustrated that I couldn't easily read that speed, so I added the ns/iter functionality. Doing that requires the ability to 0-pad.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -293,14 +293,15 @@ else
 CPP_SRCS :=
 TESTS_INTEGRATION_CPP :=
 endif
-TESTS_ANALYZE := $(srcroot)test/analyze/rand.c \
-	$(srcroot)test/analyze/sizes.c \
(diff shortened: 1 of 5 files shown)


davidtgoldblatt:
The answers, for the curious:

```
10000000 iterations, mallctl_short=257998us (25.799 ns/iter), mallctlbymib_short=105999us (10.599 ns/iter), ratio=1:2.434
test_mallctl_vs_mallctlbymib_short (non-reentrant): pass
10000000 iterations, mallctl_long=3094985us (309.498 ns/iter), mallctlbymib_long=602997us (60.299 ns/iter), ratio=1:5.133
test_mallctl_vs_mallctlbymib_long (non-reentrant): pass
```

PR #1905, opened 2020-08-12
Add narenas_ratio.

davidtgoldblatt:
This allows setting arenas per cpu dynamically, rather than forcing the user to
know the number of CPUs in advance if they want a particular CPU/space tradeoff.

Practically, I want this mostly for running my own tuning experiments, rather than because I think end-users should be tweaking it on their own. Leaving it undocumented for now; we can always declare it to be supported later.

The way this sets things up, an explicit narenas setting (of a value other than "default") will still override an narenas_ratio setting (basically saying that the more specific attempt to set wins). This feels right to me; narenas_ratio is setting the default ratio. You can still get the equivalent behavior by setting "narenas:default" at the same time you set the ratio if you don't want this behavior.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -127,6 +127,7 @@ bool	opt_utrace = false;
 bool	opt_xmalloc = false;
 bool	opt_zero = false;
 unsigned	opt_narenas = 0;
+unsigned	opt_narenas_ratio = 4;
 
 unsigned	ncpus;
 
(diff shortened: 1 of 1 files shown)


src/jemalloc.c
```diff
@@ -1294,6 +1295,12 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 					    /* clip */ false)
 				}
 			}
+			if (CONF_MATCH("narenas_ratio")) {
+				CONF_HANDLE_UNSIGNED(opt_narenas_ratio,
+				    "narenas_ratio", 1, UINT_MAX,
```
yinan1048576, line 1300: It would be better if there's a more reasonable max, but it's fine if the option will only be internally used.

PR #1908, opened 2020-08-13
Some optimizations for mallctl

yinan1048576:
My original goal was to facilitate `malloc_stats_print()`, which needs to iterate through a giant `mallctl` command tree. Currently everything is done via an inefficient `mallctl()` call. The `mallctlbymib()` handle was designed to improve efficiency, but it's unfortunately difficult to use in `malloc_stats_print()`. This stack of commits makes it possible for `malloc_stats_print()` to easily enjoy the performance close to `mallctlbymib()`. I'm somewhat lazy to work on the actual change for `malloc_stats_print()` right now; will try to do it at a later time.

### DIFF
--- a/include/jemalloc/internal/ctl.h
+++ b/include/jemalloc/internal/ctl.h
@@ -102,6 +102,8 @@ int ctl_bymib(tsd_t *tsd, const size_t *mib, size_t miblen, void *oldp,
     size_t *oldlenp, void *newp, size_t newlen);
 int ctl_mibnametomib(tsd_t *tsd, size_t *mib, size_t miblen, const char *name,
     size_t *miblenp);
+int ctl_bymibname(tsd_t *tsd, size_t *mib, size_t miblen, const char *name,
+    size_t *miblenp, void *oldp, size_t *oldlenp, void *newp, size_t newlen);
 bool ctl_boot(void);
 void ctl_prefork(tsdn_t *tsdn);
 void ctl_postfork_parent(tsdn_t *tsdn);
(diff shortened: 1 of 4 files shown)


yinan1048576:
Rebase, and actually changed the stats printing code to make use of the new `mallctl` optimizations.

Did not try to optimize every `mallctl`; instead, optimized those costing the most CPU, which I decided to mean all those being called in loops.

I also added assertions to all places I changed, for the easiness of verification. I'll add another commit to remove those after I confirm that the assertions can all pass.

Will also assess the performance difference of the final version. In many cases, the callback function supplied to stats printing is simply a string pending function, so the performance bottleneck is the internal cost of stats printing rather than I/O. Therefore I expect to see a significant performance boost in such cases.

yinan1048576:
Changed the assertions so that they'll always be triggered.

davidtgoldblatt:
Am I right that the last commit is intended to change before landing? The prior ones look reasonable, assuming that we're building towards using them without the redundant checks in the last commit.

yinan1048576:
Yes, will add another commit after I confirm that the assertions can all pass. (I don't think our tests have complete test coverage, so I'm doing some internal checking as well.)

yinan1048576:
Completed the following:
- Confirmed internally that all assertions pass, for both the table format and the JSON format. (We do different things for each format.)
- Added another commit to remove all the assertions, as well as all the code logic related to either the assertions or the old `mallctl` way.
- Measured internally: the change resulted in 65% performance improvement for table format and 60% performance improvement for JSON format, both statistically significant.

davidtgoldblatt:
Nit: just rebase the last two commits into a single one (every commit in a stack should leave the repo in a "good" state, for some fuzzy definition of good; this is helpful both for blame purposes and in bisecting). Otherwise, looks good.

yinan1048576:
Merged the last two commits.

davidtgoldblatt (approved):

PR #1909, opened 2020-08-14
Add a component for a hugepage-centric allocator

davidtgoldblatt:
This is another step towards the implementation of a design for a hugepage allocator. It introduces the hpa_central_t; a global extent allocator intended to work in tandem with the psset allocator in https://github.com/jemalloc/jemalloc/pull/1904.

That allocator can serve requests of up to a few hundred pages. Another source of pages is still necessary, though, both to supply the psset allocator as well as to serve larger allocations. This serves that purpose.

This is not a complete implementation of the page allocation interface yet; it can't handle aligned allocations, return zeroed memory, or shrink or expand allocations. These may come with time, or we may simply rely on the PAC for those use cases. I'd rather be able to iterate more quickly than try to be feature-complete in an initial version.

### DIFF
--- a/include/jemalloc/internal/hpa_central.h
+++ b/include/jemalloc/internal/hpa_central.h
@@ -1,0 +1,47 @@
+#ifndef JEMALLOC_INTERNAL_HPA_CENTRAL_H
+#define JEMALLOC_INTERNAL_HPA_CENTRAL_H
+
+#include "jemalloc/internal/base.h"
+#include "jemalloc/internal/emap.h"
+
+typedef struct hpa_central_s hpa_central_t;
+struct hpa_central_s {
+	/* The emap we use for metadata operations. */
+	emap_t *emap;
+
+	edata_cache_t *edata_cache;
+	eset_t eset;
+
+	size_t sn_next;
+};
+
+void hpa_central_init(hpa_central_t *central, edata_cache_t *edata_cache,
+    emap_t *emap);
+/*
+ * Tries to satisfy the given allocation request with an extent already given to
+ * central.
+ */
+edata_t *hpa_central_alloc_reuse(tsdn_t *tsdn, hpa_central_t *central,
+    size_t size_min, size_t size_goal);
+/*
+ * Adds the given edata to the central allocator as a new allocation.  The
+ * intent is that after a reuse attempt fails, the caller can allocate a new
+ * extent using whatever growth policy it prefers and allocate from that, giving
+ * the excess to the hpa_central_t (this is analogous to the
+ * extent_grow_retained functionality; we can allocate address space in
+ * exponentially growing chunks).
+ *
+ * The edata_t should come from the same base that this hpa was initialized
+ * with.  Only complete extents should be added (i.e. those for which the head
+ * bit is true, and for which their successor is either not owned by jemalloc
+ * or also has a head bit of true).  It should be active, large enough to
+ * satisfy the requested allocation, and not already in the emap.
+ *
+ * If this returns true, then we did not accept the extent, and took no action.
+ * Otherwise, modifies *edata to satisfy the allocation.
+ */
+bool hpa_central_alloc_grow(tsdn_t *tsdn, hpa_central_t *central,
+    size_t size, edata_t *to_add);
+void hpa_central_dalloc(tsdn_t *tsdn, hpa_central_t *central, edata_t *edata);
+
+#endif /* JEMALLOC_INTERNAL_HPA_CENTRAL_H */
--- a/include/jemalloc/internal/prng.h
+++ b/include/jemalloc/internal/prng.h
@@ -1,7 +1,6 @@
 #ifndef JEMALLOC_INTERNAL_PRNG_H
 #define JEMALLOC_INTERNAL_PRNG_H
 
-#include "jemalloc/internal/atomic.h"
 #include "jemalloc/internal/bit_util.h"
 
 /*
@@ -59,66 +58,38 @@ prng_state_next_zu(size_t state) {
 
 /*
  * The prng_lg_range functions give a uniform int in the half-open range [0,
- * 2**lg_range).  If atomic is true, they do so safely from multiple threads.
- * Multithreaded 64-bit prngs aren't supported.
+ * 2**lg_range).
  */
 
 JEMALLOC_ALWAYS_INLINE uint32_t
-prng_lg_range_u32(atomic_u32_t *state, unsigned lg_range, bool atomic) {
-	uint32_t ret, state0, state1;
-
+prng_lg_range_u32(uint32_t *state, unsigned lg_range) {
 	assert(lg_range > 0);
 	assert(lg_range <= 32);
 
-	state0 = atomic_load_u32(state, ATOMIC_RELAXED);
-
-	if (atomic) {
-		do {
-			state1 = prng_state_next_u32(state0);
-		} while (!atomic_compare_exchange_weak_u32(state, &state0,
-		    state1, ATOMIC_RELAXED, ATOMIC_RELAXED));
-	} else {
-		state1 = prng_state_next_u32(state0);
-		atomic_store_u32(state, state1, ATOMIC_RELAXED);
-	}
-	ret = state1 >> (32 - lg_range);
+	*state = prng_state_next_u32(*state);
+	uint32_t ret = *state >> (32 - lg_range);
 
 	return ret;
 }
 
 JEMALLOC_ALWAYS_INLINE uint64_t
 prng_lg_range_u64(uint64_t *state, unsigned lg_range) {
-	uint64_t ret, state1;
-
 	assert(lg_range > 0);
 	assert(lg_range <= 64);
 
-	state1 = prng_state_next_u64(*state);
-	*state = state1;
-	ret = state1 >> (64 - lg_range);
+	*state = prng_state_next_u64(*state);
+	uint64_t ret = *state >> (64 - lg_range);
 
 	return ret;
 }
 
 JEMALLOC_ALWAYS_INLINE size_t
-prng_lg_range_zu(atomic_zu_t *state, unsigned lg_range, bool atomic) {
-	size_t ret, state0, state1;
-
+prng_lg_range_zu(size_t *state, unsigned lg_range) {
 	assert(lg_range > 0);
 	assert(lg_range <= ZU(1) << (3 + LG_SIZEOF_PTR));
 
-	state0 = atomic_load_zu(state, ATOMIC_RELAXED);
-
-	if (atomic) {
-		do {
-			state1 = prng_state_next_zu(state0);
-		} while (atomic_compare_exchange_weak_zu(state, &state0,
-		    state1, ATOMIC_RELAXED, ATOMIC_RELAXED));
-	} else {
-		state1 = prng_state_next_zu(state0);
-		atomic_store_zu(state, state1, ATOMIC_RELAXED);
-	}
-	ret = state1 >> ((ZU(1) << (3 + LG_SIZEOF_PTR)) - lg_range);
+	*state = prng_state_next_zu(*state);
+	size_t ret = *state >> ((ZU(1) << (3 + LG_SIZEOF_PTR)) - lg_range);
 
 	return ret;
 }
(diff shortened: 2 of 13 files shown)


davidtgoldblatt:
Build fixes, all on integer conversion warnings.

I ended up finishing the maybe-atomic prng -> non-atomic prng conversion. The warnings were motivated by the fact that I was using the 64-bit prng functionality for size_t output, which was the case only because the other prng functions are more of a pain to use.

davidtgoldblatt:
Forgot to add the msvc project files.

include/jemalloc/internal/prng.h
```diff
@@ -64,61 +64,34 @@ prng_state_next_zu(size_t state) {
  */
 
 JEMALLOC_ALWAYS_INLINE uint32_t
-prng_lg_range_u32(atomic_u32_t *state, unsigned lg_range, bool atomic) {
-	uint32_t ret, state0, state1;
-
+prng_lg_range_u32(uint32_t *state, unsigned lg_range) {
```
yinan1048576, line 67: So we can also delete the `#include "jemalloc/internal/atomic.h"` at the beginning of this header file?

include/jemalloc/internal/hpa_central.h
```diff
@@ -0,0 +1,41 @@
+#ifndef JEMALLOC_INTERNAL_HPA_CENTRAL_H
```
yinan1048576, line 1: Similar to my comment in #1904, it would be great if you could write some design doc on how these APIs are supposed to be used in the HPA world.

src/hpa_central.c
```diff
@@ -0,0 +1,198 @@
...
+	assert(edata_base_get(edata) == edata_addr_get(edata));
+	edata_state_set(edata, extent_state_active);
+	return edata;
+}
+
+bool
+hpa_central_alloc_grow(tsdn_t *tsdn, hpa_central_t *central,
+    size_t size, edata_t *edata) {
+	assert((size& PAGE_MASK) == 0);
+	assert(edata_base_get(edata) == edata_addr_get(edata));
```
interwq, line 83: Is this assertion saying, no address randomization on this extent? If so, I'd say add a wrapper (in a future PR).

src/hpa_central.c
```diff
@@ -0,0 +1,198 @@
...
+	assert(edata_state_get(edata) == extent_state_dirty);
+	assert(edata_base_get(edata) == edata_addr_get(edata));
+	edata_state_set(edata, extent_state_active);
+	return edata;
+}
+
+bool
+hpa_central_alloc_grow(tsdn_t *tsdn, hpa_central_t *central,
+    size_t size, edata_t *edata) {
+	assert((size& PAGE_MASK) == 0);
```
interwq, line 82: nit: spacing

src/hpa_central.c
```diff
@@ -0,0 +1,198 @@
...
+	    == base_ind_get(central->edata_cache->base));
+	assert(edata_is_head_get(edata));
+	assert(edata_state_get(edata) == extent_state_active);
+	assert(edata_pai_get(edata) == EXTENT_PAI_HPA);
+	assert(edata_slab_get(edata) == false);
+	assert(edata_szind_get_maybe_invalid(edata) == SC_NSIZES);
+
+	/* edata shouldn't be mapped anywhere; it should be a new alloc. */
+	if (config_debug) {
+		emap_full_alloc_ctx_t context1 = {0};
```
interwq, line 95: nit: let's abstract the 4 lines to an inline function? It looks useful in other places as well.

include/jemalloc/internal/hpa_central.h
```diff
@@ -0,0 +1,41 @@
...
+ * edata_t should come from the same base that this hpa was initialized with.
+ * Only complete extents should be added (i.e. those for which the head bit is
+ * true, and for which their successor is either not owned by jemalloc or also
+ * has a head bit of true).  It should be active, large enough to satisfy the
+ * requested allocation, and not already in the emap.
+ *
+ * If this returns true, then we did not accept the extent, and took no action.
+ * Otherwise, modifies *edata to satisfy the allocation.
+ */
+bool hpa_central_alloc_grow(tsdn_t *tsdn, hpa_central_t *central,
```
interwq, line 37: nit: It's a bit unclear to me when / under what condition this is called

src/hpa_central.c
```diff
@@ -0,0 +1,198 @@
...
+	 */
+	edata_addr_set(edata, edata_base_get(edata));
+	edata_zeroed_set(edata, false);
+
+	edata_t *lead = NULL;
+	if (!edata_is_head_get(edata)) {
+		lead = hpa_central_dalloc_get_merge_candidate(tsdn, central,
+		    edata_before_get(edata));
+	}
+	if (lead != NULL) {
```
interwq, line 185: nit: merge this branch, and the `edata_t *lead` declaration into the branch above

src/hpa_central.c
```diff
@@ -0,0 +1,198 @@
...
+hpa_central_dalloc_get_merge_candidate(tsdn_t *tsdn, hpa_central_t *central,
+    void *addr) {
+	edata_t *edata = emap_lock_edata_from_addr(tsdn, central->emap, addr,
+	    /* inactive_only */ true);
+	if (edata == NULL) {
+		return NULL;
+	}
+	extent_pai_t pai = edata_pai_get(edata);
+	extent_state_t state = edata_state_get(edata);
+	emap_unlock_edata(tsdn, central->emap, edata);
```
interwq, line 145: Is it safe to unlock here? Should we hold the lock for the eset_remove and dalloc, or these's another higher level lock covers that?

src/hpa_central.c
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/* Merges b into a, freeing b back to the edata cache.. */
+static void
+hpa_central_dalloc_merge(tsdn_t *tsdn, hpa_central_t *central, edata_t *a,
+    edata_t *b) {
+	emap_prepare_t prepare;
+	emap_merge_prepare(tsdn, central->emap, &prepare, a, b);
+	edata_size_set(a, edata_size_get(a) + edata_size_get(b));
+	/* Don't need any locking here -- the range checking is internal. */
```
interwq, line 164: Not following here. Does a higher level lock protects this?

test/unit/hpa_central.c
```diff
@@ -0,0 +1,452 @@
...
+}
+TEST_END
+
+TEST_BEGIN(test_first_fit_large_goal) {
+	/*
+	 * Fragmentation avoidance is more important in the HPA than giving the
+	 * user their preferred amount of space, since we expect the average
+	 * unused extent to be more costly (PAC extents can get purged away
+	 * easily at any granularity; HPA extents are much more difficult to
+	 * purge away if they get stranded).
```
interwq, line 134: The explanation here is pretty helpful; maybe move it to the source instead?

include/jemalloc/internal/hpa_central.h
```diff
@@ -0,0 +1,41 @@
+#ifndef JEMALLOC_INTERNAL_HPA_CENTRAL_H
+#define JEMALLOC_INTERNAL_HPA_CENTRAL_H
+
+#include "jemalloc/internal/base.h"
+#include "jemalloc/internal/emap.h"
+
+typedef struct hpa_central_s hpa_central_t;
+struct hpa_central_s {
+	/* The emap this ra uses. */
```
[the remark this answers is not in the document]
davidtgoldblatt: Typo. (Well, really, this code was pulled out of a previous experiment where the abbreviation made sense). Fixed.

src/extent.c
```diff
@@ -809,6 +809,10 @@ extent_can_coalesce(ecache_t *ecache, const edata_t *inner,
 		return false;
 	}
 
+	if (edata_pai_get(inner) != edata_pai_get(outer)) {
```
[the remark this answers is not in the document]
davidtgoldblatt: Yep. Added a comment.

src/hpa_central.c
```diff
@@ -0,0 +1,198 @@
...
+void
+hpa_central_init(hpa_central_t *central, edata_cache_t *edata_cache,
+    emap_t *emap) {
+	central->emap = emap;
+	central->edata_cache = edata_cache;
+	eset_init(&central->eset, extent_state_dirty);
+	central->sn_next = 0;
+}
+
+/* Returns the trail, or NULL in case of failure. */
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

src/hpa_central.c
```diff
@@ -0,0 +1,198 @@
...
+	if (trail == NULL) {
+		eset_insert(&central->eset, edata);
+		return NULL;
+	}
+	eset_insert(&central->eset, trail);
+
+label_success:
+	emap_assert_mapped(tsdn, central->emap, edata);
+	assert(edata_size_get(edata) >= size_min);
+	assert(edata_state_get(edata) == extent_state_dirty);
```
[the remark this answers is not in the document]
davidtgoldblatt: Right; for now, we're not dealing with purging at all. Eventually, we should be cleverer (I think I know how this design should look, but I want to try to get some usage data first). Added a comment.

src/hpa_central.c
davidtgoldblatt: Right; this should always be a "fresh" extent, analogous to the grow_retained pathway in extent.c; we should be getting a raw OS page.

(Really, the page-level allocators shouldn't really know about cache index randomization at all; both this and extent.c set addr to base immediately right away. I figured I'd do that in a refactor at some point).

src/hpa_central.c
davidtgoldblatt: Fixed.

src/hpa_central.c
davidtgoldblatt: Will do. Also, fixed the redundancy (meant to check that the first and last page were unmapped; unintentionally, it just checked the first page twice).

include/jemalloc/internal/hpa_central.h
davidtgoldblatt: Cleaned up the comment a bit, lmk if it's clearer.

src/hpa_central.c
davidtgoldblatt: Done.

src/hpa_central.c
davidtgoldblatt: The latter. The hpa_central_t is single-threaded; concurrent access needs external locking. The emap locking is because the emap API requires it (practically, this is only necessary to protect against races with the extent.c allocator; none of this locking is actually necessary if we move more metadata into the emap in the way we've sort of planned).

test/unit/hpa_central.c
davidtgoldblatt: Will do.

src/hpa_central.c
davidtgoldblatt: I'm honestly not sure what I meant here. I think I was trying to draw a comparison with the extent.c implementation, where (since the ecache lock is released during the merge), you need to worry about adjacent merge attempts from the same ecache.

davidtgoldblatt: I'll rewrite.

davidtgoldblatt: Ended up just doing the locking. This is a fairly slow path anyways, and the correctness reasoning is nontrivial (and not actually correct depending on how strict a view we take on edata access guarantees).

include/jemalloc/internal/prng.h
```diff
@@ -133,7 +133,9 @@ prng_range_u32(atomic_u32_t *state, uint32_t range, bool atomic) {
 	uint32_t ret;
 	unsigned lg_range;
 
-	assert(range > 1);
+	if (range == 1) {
```
interwq, line 112: nit: the commit msg probably better as `the range argument itself is generated`

include/jemalloc/internal/prng.h
interwq, line 67: nit: also update the comment above mentioning atomic.

interwq (approved):
LGTM!

davidtgoldblatt:
Only comment / include changes from last successful run -- appveyor failures are from the issue fixed in #1949.

PR #1910, opened 2020-08-14
Move the grow_retained helpers out of PAC

davidtgoldblatt:
We'll need the same functionality soon for a hugepage allocator elsewhere.


yinan1048576:
Sorry, the Wikipedia link is supposed to be [this one](https://en.wikipedia.org/wiki/Geometric_growth).

davidtgoldblatt:
Talked off line -- will change the name, but after some stacked diffs are merged, to avoid a painful rebase.

PR #1912, opened 2020-08-17
support for PPC64LE

ezeeyahoo:
Here are changes to add support PPC64LE. Please review it and do let me know if any issue is there?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -194,6 +194,15 @@ matrix:
       env: CC=gcc CXX=g++ COMPILER_FLAGS="" CONFIGURE_FLAGS="--with-malloc-conf=dss:primary,background_thread:true" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
     - os: linux
       env: CC=gcc CXX=g++ COMPILER_FLAGS="" CONFIGURE_FLAGS="--with-malloc-conf=percpu_arena:percpu,background_thread:true" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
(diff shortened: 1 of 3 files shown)


davidtgoldblatt:
I really like the idea of adding Power tests now that travis supports them. One substantive change, and one nit:
- Substantive: Rather than embedding the appropriate lines directly, we should add power as one of the "unusual" options, and do any special treatment we need for it in the format_job function. (This will make sure we don't just test a single Power config, but also Power + debug mode, Power + unusual page sizes, etc.).
- Nit: We keep a linear history; rather than having a marge commit, this should be something that can be rebased on top of the fix in #1907.

ezeeyahoo:
Should I make  MAX_UNUSUAL_OPTIONS = 3 ?
with
```
all_unusuals = (
    [os_unusual] + [arch_unusual] + [compilers_unusual] + compiler_flag_unusuals
    + configure_flag_unusuals + malloc_conf_unusuals
)
```

davidtgoldblatt:
I think leaving it at 2 is probably fine; Travis runs already take an unpleasantly long amount of time.

davidtgoldblatt:
> I am little bit struggling with addons: part, where I have to keep both *gcc_multilib and add compiler packages. Would you recommend to add before_install: instead of trying to manage both packages and *gcc_multilib mapping(correct me!) together?

I'm not completely sure I follow, but the way I'd suggest handling build-specific config is to treat it specially in `format_job` (i.e. tweaking the compiler, setup steps, etc.).

PR #1913, opened 2020-08-17
Mallctl stress test: fix a type.

davidtgoldblatt:
The mallctlbymib_long helper was copy-pasted from mallctlbymib_short, and
incorrectly used its output variable (a char *) rather than the output variable
of the mallctl call it was using (a uint64_t), causing breakages when
sizeof(char *) differed from sizeof(uint64_t).

### DIFF
--- a/test/stress/mallctl.c
+++ b/test/stress/mallctl.c
@@ -44,9 +44,9 @@ size_t mib_long[6];
 static void
 mallctlbymib_long(void) {
 	size_t miblen = sizeof(mib_long)/sizeof(mib_long[0]);
-	const char *version;
-	size_t sz = sizeof(version);
-	int err = mallctlbymib(mib_long, miblen, &version, &sz, NULL, 0);
+	uint64_t nmalloc;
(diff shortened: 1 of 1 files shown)


sharkcz:
looks good now, thanks
````
=== test/stress/mallctl ===
10000000 iterations, mallctl_short=3731504us (373.150 ns/iter), mallctlbymib_short=1087146us (108.714 ns/iter), ratio=1:3.432
test_mallctl_vs_mallctlbymib_short (non-reentrant): pass
10000000 iterations, mallctl_long=35304772us (3530.477 ns/iter), mallctlbymib_long=4621624us (462.162 ns/iter), ratio=1:7.639
test_mallctl_vs_mallctlbymib_long (non-reentrant): pass
--- pass: 2/2, skip: 0/2, fail: 0/2 ---
````

davidtgoldblatt:
This was in a stress test, which doesn't run as part of CI (and, when I was running manually, it was on a 64-bit architecture).

PR #1915, opened 2020-08-20
Extract malloc_conf as separate module

yinan1048576:
All here is rearranging. Also attempted to extract out the init module, but things were more complicated and may change a lot when we rewrite the arena module.


davidtgoldblatt:
Closing this out as no longer mergeable. It's not clear to me that we'd see the benefit of splitting things up if we don't get some composability benefits out of it.

PR #1916, opened 2020-08-21
Do not create size class tables for non-prof builds

yinan1048576:
Save a few KB for non-prof builds. I'm planning to add some more size class related tables for internal fragmentation tracking, so the saving can be more.

### DIFF
--- a/include/jemalloc/internal/prof_data.h
+++ b/include/jemalloc/internal/prof_data.h
@@ -10,6 +10,9 @@ extern malloc_mutex_t prof_dump_mtx;
 extern malloc_mutex_t *gctx_locks;
 extern malloc_mutex_t *tdata_locks;
 
+extern size_t prof_unbiased_sz[SC_NSIZES];
+extern size_t prof_shifted_unbiased_cnt[SC_NSIZES];
+
 void prof_bt_hash(const void *key, size_t r_hash[2]);
 bool prof_bt_keycomp(const void *k1, const void *k2);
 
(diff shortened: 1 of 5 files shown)


yinan1048576:
> The reason I did it to begin with was that I thought some libcs would lazily allocate some tables during the first exp call

What do you mean by "the first exp call", and by "some tables"?

davidtgoldblatt:
It's not uncommon for the implementation of the C `exp` function to use lookup tables for part of its implementation. But for those programs that don't call `exp`, you could imagine an optimization in which those tables are lazily initialized the first time `exp` is called, and stored into memory returned from malloc.

davidtgoldblatt:
(I.e. which would then trigger a reentrancy scenario -- necessitating bootstrap initialization where we know more about the state we're in and have more protections).

yinan1048576:
I have a slightly better understanding of what you were trying to say. Basically you're comparing the current approach, i.e. the tables, with an alternative approach where the unbiasing numbers are computed at sampling time, right? It's interesting to know the reentrancy part of the story behind the current approach, but I thought a main motivation for having the tables was for improving performance. From what you are now saying, are you implying that you don't think this is critical?

davidtgoldblatt:
I think I was worried about that too, although running a quick microbenchmark, that math looks pretty cheap, especially given that it happens only on the sampling pathway.

yinan1048576:
Added some extra `cassert(config_prof)` guards, which are also sensible in themselves.

PR #1917, opened 2020-08-21
Correct usize in prof last-N record

yinan1048576:
This is the minimal subset of #1899 for bug fixing purpose.


yinan1048576:
Rebase and add unit test.

PR #1918, opened 2020-08-23
Avoid running some unnecessary code when prof_active is off

umanwizard:


yinan1048576:
The intention is very reasonable. However, the profiling stack has been designed so that, when `config_prof` and `opt_prof` are both on, either `prof_malloc()` or `prof_realloc()` has to be called for each allocation, regardless of whether it's sampled or not. This is a contract the entire stack follows and assumes. The test failures reflect what would happen otherwise.

With the change you're proposing, whenever `prof_active` is off, what's being saved is fairly cheap. If you look into the additional functions being called in the `if()` branch, I suspect that, aside from the essential `prof_malloc()` call, the total should be no more than 10-20 assembly instructions.

Having said that, I like your intention of trying to optimize the `imalloc()` function. As David was also suggesting, we haven't spent much time into it, and we know it matters. If you have other ideas, please don't hesitate to put up a PR!

yinan1048576:
> The intention is very reasonable. However, the profiling stack has been designed so that, when `config_prof` and `opt_prof` are both on, either `prof_malloc()` or `prof_realloc()` has to be called for each allocation, regardless of whether it's sampled or not. This is a contract the entire stack follows and assumes. The test failures reflect what would happen otherwise.

Sorry - I don't think this reason is correct. Let me get back to you with the right reason a bit later.

yinan1048576:
The correct version should be: either `prof_malloc()` or `prof_realloc()` has to be called for each *large* allocation, regardless of whether it's sampled or not.

umanwizard:
Thanks for the explanation! Since it seems this is wrong, based on your explanation, I will close it.

PR #1921, opened 2020-08-24
Omit stats rows with no data

yinan1048576:
The output for bin stats does not have rows with no data any more; before this change, such rows exist in the output and there would be `---` lines after them whenever the next row was not empty. The output for large sizes stays the same; this change just serves to short-circuit early.


davidtgoldblatt (approved):

yinan1048576:
Rebase, and did the following:
- Added proper handling for ending gap in HPA stats.
- Decided not to bother with the short-circuiting at all. There can be a lot more to do if we really want to short-circuit as much as possible. People can work on it in the future if interested. Also, in theory, a sufficiently smart compiler should be able to perform the short-circuiting automatically for us anyway.

davidtgoldblatt:
New commit looks fine too.

PR #1922, opened 2020-08-25
Do not repeat reentrancy test in profiling

yinan1048576:
The saving is tiny; more for correctness, since reentrancy is handled in the thread event module.


include/jemalloc/internal/prof_inlines.h
```diff
@@ -115,9 +115,7 @@ prof_sample_should_skip(tsd_t *tsd, bool sample_event) {
 		return true;
 	}
 
-	if (tsd_reentrancy_level_get(tsd) > 0) {
-		return true;
-	}
+	assert(tsd_reentrancy_level_get(tsd) == 0);
```
davidtgoldblatt, line 118: Can you add a comment explaining why the assert should never trigger?

yinan1048576:

PR #1924, opened 2020-08-25
Do not bump to large size for page aligned request

yinan1048576:
I started this change with a much greater ambition, to make some significant improvement to the `sz_sa2u()` computations, but through some trials and errors, I came to better understand how things work and ended up with this tiny change.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -1049,10 +1049,17 @@ arena_palloc(tsdn_t *tsdn, arena_t *arena, size_t usize, size_t alignment,
     bool zero, tcache_t *tcache) {
 	void *ret;
 
-	if (usize <= SC_SMALL_MAXCLASS
-	    && (alignment < PAGE
-	    || (alignment == PAGE && (usize & PAGE_MASK) == 0))) {
+	if (usize <= SC_SMALL_MAXCLASS) {
 		/* Small; alignment doesn't require special slab placement. */
+
+		/* usize should be a result of sz_sa2u() */
+		assert((usize & (alignment - 1)) == 0);
+
+		/*
+		 * Small usize can't come from an alignment larger than a page.
+		 */
+		assert(alignment <= PAGE);
+
 		ret = arena_malloc(tsdn, arena, usize, sz_size2index(usize),
 		    zero, tcache, true);
 	} else {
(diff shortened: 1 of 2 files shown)


src/arena.c
```diff
@@ -1041,9 +1041,7 @@ arena_palloc(tsdn_t *tsdn, arena_t *arena, size_t usize, size_t alignment,
     bool zero, tcache_t *tcache) {
 	void *ret;
 
-	if (usize <= SC_SMALL_MAXCLASS
-	    && (alignment < PAGE
-	    || (alignment == PAGE && (usize & PAGE_MASK) == 0))) {
+	if (usize <= SC_SMALL_MAXCLASS && alignment <= PAGE) {
```
davidtgoldblatt, line 1044: Add a comment and make the (usize & PAGE_MASK) == 0 an assert; the correctness here is nontrivial.

davidtgoldblatt (commented):
One nit, otherwise looks good.

src/arena.c
yinan1048576: I'm not sure what are you referring to. Do you mean to assert `alignment < PAGE || (usize & PAGE_MASK) == 0` instead? If so, from what I can see, this assertion is a bit redundant or probably not suitable for here, because it's just checking whether `sz_sa2u()` functioned properly (since that's how the `usize` was obtained). Do you mean something else?

davidtgoldblatt: I think I mean `assert((usize & (alignment - 1)) == 0)`. In general, usize and alignment need not be simply related the way the logic here might imply; it falls out as a consequence of alignment handling for small size classes, which isn't obvious at first glance.

E.g. When I read this, my first thought was "this can't possibly be right, because if usize == PAGE*3/2 and alignment == PAGE, we'll return just any old PAGE*3/2 region, which need not be PAGE-aligned". But really, that's not a valid combination of combination to pass to this function; I want to assert that the combination we got was valid (i.e. has been appropriately sz_sa2u'd).

yinan1048576: It feels that, possibly, the logic here was originally not meant to be dependent on the knowledge that `usize` was coming out of an `sz_sa2u()` call, but it doesn't seem valid either, because simply making that `arena_malloc()` call below would not have worked for all combinations of `usize <= SC_SMALL_MAXCLASS` and `alignment < PAGE`. I think the logic here just cannot be made independent of the size computations, and, given that, I'm making some even further simplification changes here - the `if` condition only needs to be `usize <= SC_SMALL_MAXCLASS` check, and I added comments to explain. I'm 99% sure that this is correct, and I'll see if the tests agree...

yinan1048576 (commented):
Rebase and revise.

davidtgoldblatt (approved):

PR #1927, opened 2020-09-01
Dependency generation commands to use compiler flags

klimkin:
Compilers like Clang may require `--gcc-toolchain=` to find std headers. If jemalloc only uses preprocessor flags, dependency generation fails.


klimkin:
Works when adding `--gcc-toolchain=$GCC_DIR` to `CPPFLAGS`

PR #1928, opened 2020-09-01
[Seeking early feedback] Enable profiling for large internal fragmentation

yinan1048576:
This is not meant to be used widely. It's mainly for people who'd want to understand more about a particular type of allocation requests - those with a small request size plus a large alignment. This type of allocation requests results in serious internal fragmentation, given how jemalloc sizing currently works. I want to get some early feedback on where I'm going before adding unit tests.

The current implementation builds on top of profiling - only the "bad" requests get profiled. The "good" allocation requests still need to go through the medium path even though they won't be sampled, because we don't want to afford the "goodness" check on fast path. Profiling is sampled, and it might be beneficial to slightly tune up the profiling frequency when this new mode is turned on.

(I had to borrow #1923 because otherwise the logic would be very different for `rallocx()`.)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -249,6 +249,7 @@ TESTS_UNIT := \
 	$(srcroot)test/unit/prof_mdump.c \
 	$(srcroot)test/unit/prof_recent.c \
 	$(srcroot)test/unit/prof_reset.c \
+	$(srcroot)test/unit/prof_select.c \
 	$(srcroot)test/unit/prof_stats.c \
 	$(srcroot)test/unit/prof_tctx.c \
 	$(srcroot)test/unit/prof_thread_name.c \
--- a/include/jemalloc/internal/prof_externs.h
+++ b/include/jemalloc/internal/prof_externs.h
@@ -30,6 +30,15 @@ extern bool opt_prof_sys_thread_name;
 /* Whether to record per size class counts and request size totals. */
 extern bool opt_prof_stats;
 
+/* If not 0, profiling will only select allocations of the designated usize. */
+extern size_t opt_prof_select_usize;
+
+/*
+ * If opt_prof_select_usize is not 0, profiling will only select allocations
+ * wasting at least the designated amount of memory in internal fragmentation.
+ */
+extern size_t opt_prof_select_waste;
+
 /* Accessed via prof_active_[gs]et{_unlocked,}(). */
 extern bool prof_active;
 
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -138,6 +138,8 @@ CTL_PROTO(opt_prof_final)
 CTL_PROTO(opt_prof_leak)
 CTL_PROTO(opt_prof_accum)
 CTL_PROTO(opt_prof_recent_alloc_max)
+CTL_PROTO(opt_prof_select_usize)
+CTL_PROTO(opt_prof_select_waste)
 CTL_PROTO(opt_prof_stats)
 CTL_PROTO(opt_prof_sys_thread_name)
 CTL_PROTO(opt_prof_time_res)
@@ -425,6 +427,8 @@ static const ctl_named_node_t opt_node[] = {
 	{NAME("prof_leak"),	CTL(opt_prof_leak)},
 	{NAME("prof_accum"),	CTL(opt_prof_accum)},
 	{NAME("prof_recent_alloc_max"),	CTL(opt_prof_recent_alloc_max)},
+	{NAME("prof_select_usize"),	CTL(opt_prof_select_usize)},
+	{NAME("prof_select_waste"),	CTL(opt_prof_select_waste)},
 	{NAME("prof_stats"),	CTL(opt_prof_stats)},
 	{NAME("prof_sys_thread_name"),	CTL(opt_prof_sys_thread_name)},
 	{NAME("prof_time_resolution"),	CTL(opt_prof_time_res)},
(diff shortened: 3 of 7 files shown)


yinan1048576:
When I'm looking at this PR again, I think it might be worthwhile to step back a bit and have some more discussions upfront.

The underlying intention here is to help users understand about the main contributors to the memory being wasted in their program that's due to internal fragmentation. More specifically, when they face a report like what's generated from #1914, and if the goal is not for jemalloc to adjust the size classes, users need to know where the "problematic" request sizes are coming from, and then explore ways to adjust them to better fit jemalloc size classes. The report itself doesn't tell the sources of the "problematic" request sizes, so there needs to be a way to facilitate that.

On a higher level, the goal of telling where allocations come from aligns with what profiling is built for. This is the reason why I'm thinking about creating the functionality inside of profiling. One key downside is that we only get information about the samples, but overcoming it would most likely means some additional branching on the malloc fast path, as well as some additional attribution logic wired outside of profiling. I'm leaning towards having the functionality within profiling at the cost of missing a portion of the allocations; users can tune up the sample rate if they want more comprehensive coverage.

The next question is how to make the allocation attribution in profiling work for internal fragmentation understanding purpose instead of memory consumption understanding purpose. This is where I've got the idea of filtering, since the profiling information as a whole is an aggregate of both the "problematic" and "non-problematic" allocations. One thing we need to make sure here is that users should not interpret the profiling dump content the usual way, but instead see it as a "filtered" attribution report.

The filter I wrote earlier was for selecting the allocations that result in worse internal fragmentation than the upper bounds as set by the jemalloc size class spacings. It can be a useful filter, but I'm now seeing that it has a few issues:
- The profiling attribution report aggregates over allocations and all size classes. It can be hard for users to know how to adjust the request sizes at places in code where the user doesn't know or is not able to know the request size(s) being emitted. These can also be places where internal fragmentation optimization is hard anyway though.
- The filtering threshold may not always be what users want. For example, some user may want to be even stricter and reduce internal fragmentation to a level well below what can be achieved given the jemalloc size class spacings.

What I can see as an alternative is to accept a callback function pointer from the user, which serves for filtering purpose. In the function, users may specify any particular usable size(s) they're interested in, as well as any customized internal fragmentation threshold according to their own level of sensitivity.

Any thoughts, either on the solution, or on any alternatives to any part of my reasoning?

yinan1048576:
I'm temporarily proceeding with the following plan -

The caller can specify a particular size class, as well as the level of tolerance for internal fragmentation, i.e. the maximum memory waste. Then, profiling will only select allocations that are of the specified size class and exceeding the waste tolerance.

The benefit over the function pointer is simplicity. The cost is some flexibility, but I feel it may not be that serious. Specifically, if the caller wants to investigate the internal fragmentation of multiple size classes, they'd need to invoke multiple runs each with a different size class. This might be fine, since a profiling dump with multiple size classes mixed up may be less helpful anyway for internal fragmentation understanding purpose, as described earlier.

Thoughts?

davidtgoldblatt:
My initial inclination is that this strategy doesn't seem to carry its weight in terms of added complexity relative to the benefit. I don't think we have evidence that this is a "real" problem to begin with; and it seems like finding some very simple stats may be an easier way to begin. The profiling selection mechanism feels wrong as well; picking out size classes to sample as part of a manual process triggers some of my "this is the wrong way to do this" senses.

It seems like if we wanted to tackle this problem systematically, the right way would be a parallel set of sampling counters driven by sums of `sa2u(size, alignment) - s2u(size)` (not suggesting we try this).

yinan1048576:
> My initial inclination is that this strategy doesn't seem to carry its weight in terms of added complexity relative to the benefit. I don't think we have evidence that this is a "real" problem to begin with; and it seems like finding some very simple stats may be an easier way to begin. The profiling selection mechanism feels wrong as well; picking out size classes to sample as part of a manual process triggers some of my "this is the wrong way to do this" senses.

Well, could you elaborate a bit, in a way to someone who does not naturally carry your sense? For example, if the user sees a list of size classes each with an amount of internal fragmentation waste, and if the user wants to do something, how can we help? Imagine what the user sees is the following (which is roughly what the internal fragmentation stats would look like):
```
size | waste
------------
   8 | 2MB
  16 | 1MB
  32 | 1GB
 ... | ...
```

If I'm the user, I'd be interested in knowing how I can do something with the 1GB waste in the 32B size class. How can jemalloc help me with that? (BTW I believe this kind of optimization attempts was what led to `fbvector` and some other similar stuffs.)

> It seems like if we wanted to tackle this problem systematically, the right way would be a parallel set of sampling counters driven by sums of `sa2u(size, alignment) - s2u(size)` (not suggesting we try this).

Does this feel like what I had in my initial version? I was basically trying to zoom into the allocations where `s2u(size) < usize`.

davidtgoldblatt:
> Well, could you elaborate a bit, in a way to someone who does not naturally carry your sense?

Just that, in general, we should try to avoid having profiling "modes", where the meaning and interpretation of the data depend on complicated combinations of options. We should, as much as possible, try to avoid workflows that involve this sort of iterative drilldown. The special case of small objects with large alignment is fairly edge-casey, and would also show up in regular profiling. This complicates the profiling logic even further, for what I'm not sure we have reason to believe is a real problem.

yinan1048576:
Any other thoughts, especially constructive feedback on how to proceed?

davidtgoldblatt:
I don't think we should pursue this sort of strategy; perhaps eventually a generalized profiling framework will allow us to push these sorts of things into the user-side. For now, though, it's narrowly scoped to an uncommon use case, but comes at an intellectual cost.

yinan1048576:
Rebase and added `mallctl` and unit tests. It is fine if this new feature does not go into dev; I understand the complications. Given that there doesn't seem to be any straightforward alternative for now, I'd like to keep a record of the current change, since people are in need of this feature and they can at least have something to help with internal fragmentation attribution locally.

PR #1930, opened 2020-09-09
RFC: allow to specify larger slab sizes

hliu202:
We may get better performance from larger slab sizes. Currently, there is a malloc_conf `slab_sizes` to set the default slab sizes. But it is limited by `lg-slab-maxregs`, which means the maximum number of regions in a slab.

This patch adds a new configure `--with-lg-slab-maxregs`, which allows to override the default `lg-slab-maxregs`, which is `<lg-page> - <lg-tiny-min>`. The value should not be less than the default value. The max value of this option is related to `LG_BITMAP_MAXBITS` (see more in bitmap.h).

For example, on a 4k page size system, if we:
  1) configure jemalloc with with `--with-lg-slab-maxregs=12`.
  2) `export MALLOC_CONF="slab_sizes:9-16:4"`

The slab size of 16 bytes is set to 4 pages. Previously, the default lg-slab-maxregs is `9 (i.e. 12 - 3)`. The max slab size of 16 bytes is 2 pages (i.e. `(1<<9) * 16` bytes). By increasing the value from 9 to 12, the max slab size can be set by `MALLOC_CONF` is 16 pages (i.e. `(1<<12) * 16` bytes).

---
BTW. another way to increase slab size is by specifying larger `--with-lg-page`, but I think it is not flexible. For example:
1. it increases the slab size of all small size classes, and much more memory usage
2. it makes some previous large size classes to be treated as small size classes.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -250,6 +250,13 @@ any of the following arguments (not a definitive list) to 'configure':
     configuration, jemalloc will provide additional size classes that are not
     16-byte-aligned (24, 40, and 56).
 
+* `--with-lg-slab-maxregs=<lg-slab-maxregs>`
+
+    Specify the maximum number of regions in a slab, which is
+    (<lg-page> - <lg-tiny-min>) by default. This increases the limit of slab
+    sizes specified by "slab_sizes" in malloc_conf. This should never be less
+    than the default value.
+
 * `--with-lg-vaddr=<lg-vaddr>`
 
     Specify the number of significant virtual address bits.  By default, the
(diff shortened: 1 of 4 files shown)


davidtgoldblatt:
I sat on this for a while because I think probably better would be to dynamically expand the array at the end of the edata_t if called for (so that any slab size is always valid), but I don't think it's reasonable to block this on that. I do want to note two caveats, though:
- This functionality is still experimental, so don't take the merge as a guarantee that it won't go away at some point.
- Config options may change before too long, if we do the slab thing internally, or end up changing the build system (as we likely will).

davidtgoldblatt:
And I completely forgot that this does actually document the change; I'll put up a PR un-advertising the functionality.

PR #1931, opened 2020-09-10
Mitigate deadlock due to atexit(prof_log_stop_final)

laiwei-rice:
Calling `atexit()` can lead to deadlock in scenarios where there is another call to `atexit()` on the call stack.
Here is a similar note I found in the ChangeLog:
```
  - Disable "opt.prof_final" by default, in order to avoid atexit(3), which can
    internally deadlock on some platforms.
```

The call to `atexit(prof_log_stop_final)` is not necessary unless flag `opt_prof_log` is true.
As a result, we protect the call with the flag to mitigate the deadlock -- the default value of `opt_prof_log` is `false`.

### DIFF
--- a/src/prof_log.c
+++ b/src/prof_log.c
@@ -681,7 +681,7 @@ bool prof_log_init(tsd_t *tsd) {
 		prof_log_start(tsd_tsdn(tsd), NULL);
 	}
 
-	if (atexit(prof_log_stop_final) != 0) {
+	if (opt_prof_log && atexit(prof_log_stop_final) != 0) {
 		malloc_write("<jemalloc>: Error in atexit() "
 			     "for logging\n");
 		if (opt_abort) {


src/prof_log.c
```diff
@@ -681,7 +681,7 @@ bool prof_log_init(tsd_t *tsd) {
 		prof_log_start(tsd_tsdn(tsd), NULL);
 	}
 
-	if (atexit(prof_log_stop_final) != 0) {
```
interwq, line 684: @yinan1048576 : This makes me feel like it was constructed this way on purpose; can we start with prof_log == false, and later start logging with mallctl? In that case we should delay the `atexit` call to the first log start call.

laiwei-rice: Here is what I have seen in the code.
At line 681, `prof_log_start()` is called if flag `opt_prof_log` is true. If `prof_log_start()` is not called, variable `prof_logging_state` will retain its default value as `prof_logging_state_stopped`.

`prof_log_stop_final()` calls `prof_log_stop()`, which does nothing unless `prof_logging_state == prof_logging_state_started`.

Therefore, when `opt_prof_log` is false, `prof_log_start()` is never called and we would have `prof_logging_state == prof_logging_state_stopped`, which means we don't need to call `prof_log_stop_final()`.

yinan1048576: Part of the design was to allow applications to start logging after the program starts, via the `prof.log_start` command in `mallctl`, and then to dump at exit. Therefore, the right way should be to invoke `atexit` within `prof_log_start()`, as @interwq recommended. In case of multiple `atexit` registerings, the OS would likely invoke `prof_log_stop_final()` multiple times, but there should be no problem.

interwq:
Resolved in #2137

Issue #285, opened 2015-09-28
Issues compiling under CYGWIN


thefallentree:
There are several issues compiling under CYGWIN 

```
$ uname -a
CYGWIN_NT-10.0 host 2.2.1(0.289/5/3) 2015-08-20 11:42 x86_64 Cygwin
```

Namely:
1.  there are no `<sys/syscall.h>`  under cygwin.  need wrap it with `#ifndef __CYGWIN__`
2. issues with RTLD_NEXT usage, it is not supported under CYGWIN.
3.  Complain about `error "No madvise(2) flag defined for purging unused dirty pages."`

Full Log:

```
$ ./configure
checking for xsltproc... false
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.exe
checking for suffix of executables... .exe
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether compiler supports -std=gnu99... yes
checking whether compiler supports -Wall... yes
checking whether compiler supports -Werror=declaration-after-statement... yes
checking whether compiler supports -pipe... yes
checking whether compiler supports -g3... yes
checking how to run the C preprocessor... gcc -E
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking whether byte ordering is bigendian... no
checking size of void *... 8
checking size of int... 4
checking size of long... 8
checking size of intmax_t... 8
checking build system type... x86_64-unknown-cygwin
checking host system type... x86_64-unknown-cygwin
checking whether pause instruction is compilable... yes
checking for ar... ar
checking malloc.h usability... yes
checking malloc.h presence... yes
checking for malloc.h... yes
checking whether malloc_usable_size definition can use const argument... no
checking whether __attribute__ syntax is compilable... yes
checking whether compiler supports -Werror... yes
checking whether tls_model attribute is compilable... yes
checking whether compiler supports -Werror... yes
checking whether alloc_size attribute is compilable... yes
checking whether compiler supports -Werror... yes
[...]

PR #1934, opened 2020-09-11
Cygwin fixes

bprager:
I applied the changes recommended in #285. Cygwin now compiles without complaints. Can that be merged into the branch? Thank you.


configure.ac
```diff
@@ -2232,7 +2233,7 @@ fi
 dnl ============================================================================
 dnl Enable background threads if possible.
 
-if test "x${have_pthread}" = "x1" -a "x${je_cv_os_unfair_lock}" != "xyes" ; then
+if test "x${have_pthread}" = "x1" -a "x${je_cv_os_unfair_lock}" != "xyes" -a "x${is_cywgin}" = "x1" ; then
```
davidtgoldblatt, line 2236: I don't think this is the right test -- it will effectively disable background threads on non-cygwin configurations, right?

davidtgoldblatt (commented):
Nit: we keep a linear commit history, and only enable the rebase-and-merge; please squash these down to one commit, and add a more descriptive commit message.

The failures look real; possible fix inline.

configure.ac
bprager: I think you are correct. I will check and correct that. Thanks.

AriFordsham:
Why was this closed?

PR #1937, opened 2020-09-16
Remove --with-slab-maxregs options from INSTALL.md

davidtgoldblatt:
The variable slab sizes feature is still experimental; we don't want people to
start using it willy-nilly, or document its existence as a guarantee.

For some reason, the rebased version triggered some failures that the original PR didn't. Work around a triggered -Wundef issue.


interwq (approved):

interwq (approved):
LGTM

Issue #808, opened 2017-04-28
Dev branch # of CPU detection error when using CPU binding


tanghaodong25:
#### Test platform
ubuntu 16.10
gcc/g++ 6.2.0
numactl 2.0.11
jemalloc commit id: 5c56603e91d8d0021280615f86b95efe0463b6a8

#### Background
We're trying to run benchmark for ceph, a distributed objectstore, and hit some issues with jemalloc if using "numactl" to bind prcocess to specified numa node. After deep analysis, we suspected it was jemalloc problem. So we tried jemalloc "run_test.sh" with prefix "numactl --membind=0 --cpunodebind=0", it ran into error on ubuntu 16.10. We also tried ubuntu 14.04, there was no error.

#### Issue
```shell 
=== test/unit/bitmap ===
test_bitmap_initializer (non-reentrant): pass
test_bitmap_initializer (libc-reentrant): pass
test_bitmap_initializer (arena_new-reentrant): pass
test_bitmap_size (non-reentrant): pass
test_bitmap_size (libc-reentrant): pass
test_bitmap_size (arena_new-reentrant): pass
<jemalloc>: src/jemalloc.c:402: Failed assertion: "ind <= narenas_total_get()"
Aborted
Test harness error: test/unit/bitmap w/ MALLOC_CONF=""
Use prefix to debug, e.g. JEMALLOC_TEST_PREFIX="gdb --args" sh test/test.sh test/unit/bitmap
Makefile:439: recipe for target 'check_unit' failed
make: *** [check_unit] Error 1
test_alignment_and_size (non-reentrant): pass
test_alignment_and_size (libc-reentrant): pass
test_alignment_and_size (arena_new-reentrant): pass
--- pass: 9/9, skip: 0/9, fail: 0/9 ---

=== test/integration/allocated ===
<jemalloc>: src/jemalloc.c:402: Failed assertion: "ind <= narenas_total_get()"
Aborted
```
Thanks,
Haodong

interwq:
@tanghaodong25: yes we are aware of this issue. It's caused by # of CPUs not detected correctly when CPU binding is used (there was some dependency complication to detect it during bootstrapping). 

To workaround this issue for now, you can try one of the following when using binding:
1) add "percpu_arena:disabled" to your jemalloc config string, e.g. add 
const char* malloc_conf = "percpu_arena:disabled" in the source file.
2) manual set number of arenas, e.g. add "narenas:64" where 64 is the total # of cores in your system.

[...]

PR #1939, opened 2020-09-20
[RFC] Fix percpu arena by using all available CPUs (not only eligible)

azat:
Prefer `_SC_NPROCESSORS_ONLN` over `sched_getaffinity()` for number of
arenas for percpu_arena enabled, since `sched_getaffinity()` can be changed in runtime, or it can be
allowed to run only on, say, 8 CPU, but in this case:
- ncpus==1
- but malloc_getcpu() will return 7

And this will trigger the following assertion in debug build:
```
<jemalloc>: ../src/jemalloc.c:321: Failed assertion: "ind <= narenas_total_get()"
```

*Here is a reproducer [1].*

  [1]: https://gist.github.com/azat/9770f751c3e7d2da5422cc1ca26339c4

And AFAICS eligible number of CPUs cannot be used even for
[mutex detection](https://github.com/jemalloc/jemalloc/blob/259c5e3e8f4731f2e32ceac71c66f4bc7d078145/src/mutex.c#L50), since the affinity mask can be changed later.

Refs: #1676
Refs: #808

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -384,4 +384,7 @@ before_script:
 
 script:
   - make check
+  # trigger possible for incorrect detection of ncpus
+  # (assumes getconf if taskset exists)
+  - if command -v taskset >/dev/null; then echo "Running on last CPU" && taskset --cpu-list $(( $(getconf _NPROCESSORS_ONLN)-1 )) make check; fi
 
--- a/scripts/gen_travis.py
+++ b/scripts/gen_travis.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 from itertools import combinations
 
--- a/scripts/gen_travis.py
+++ b/scripts/gen_travis.py
@@ -22,6 +22,9 @@ before_script:
 
 script:
   - make check
+  # trigger possible for incorrect detection of ncpus
+  # (assumes getconf if taskset exists)
+  - if command -v taskset >/dev/null; then echo "Running on last CPU" && taskset --cpu-list $(( $(getconf _NPROCESSORS_ONLN)-1 )) make check; fi
 """
 
 # The 'default' configuration is gcc, on linux, with no compiler or configure
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -790,30 +790,6 @@ malloc_ncpus(void) {
 	SYSTEM_INFO si;
 	GetSystemInfo(&si);
 	result = si.dwNumberOfProcessors;
-#elif defined(CPU_COUNT)
-	/*
-	 * glibc >= 2.6 has the CPU_COUNT macro.
-	 *
-	 * glibc's sysconf() uses isspace().  glibc allocates for the first time
-	 * *before* setting up the isspace tables.  Therefore we need a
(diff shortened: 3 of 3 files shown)


azat:
Any comments on this one?

davidtgoldblatt:
+cc @interwq -- this seems like it's in your wheelhouse.

azat:
@interwq can you please take a look? Thanks in advance!

interwq:
Thanks for sharing the thoughts (also sorry for the delayed response). Indeed the percpu arena feature does not play well with dynamic affinity changes; we didn't really try to optimize for this case -- not only for percpu arena, but for the default number of auto arenas / scalability factor in general.

A workaround for this is to also set `narenas`, see the comments here: https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L1843

for example, if you specify `narenas` to be the actual number of CPUs, percpu_arena will attempt to utilize those arenas.

However, this can cause bad side effects: especially for [decay](http://jemalloc.net/jemalloc.3.html#opt.dirty_decay_ms), which by default relies on arenas being accessed regularly to return unused pages back to OS gradually. In the case of dynamic CPU binding changes + percpu_arena, the cached dirty pages in an arena may become "unaccessible" when the affinity for that CPU / arena is removed, i.e. no threads will be triggering decay for those arenas.

I'd recommend not using percpu_arena if CPU affinity changes over the lifetime of the binary, and manually setting `narenas` based on available resources: higher arena count typically improves scalability, at the cost of increased fragmentation / memory usage.

azat:
>A workaround for this is to also set narenas, see the comments here: 

But right now it is impossible to set narenas dynamically from the binary (w/o wrapping into shell script), am I missing something?
How about adding one more mode for percpu then, say, `percpu_arena:percpu_all` (and maybe add new `percpu_arena:percpu_eligible` and deprecated `percpu_arena:percpu`) that will set narenas to the all available CPUs.

But this will not solve other issues (like the mutex requirement detection), actually I thought about adding two different counters (total and eligible like you wrote in #1947, however later I came to conclusion that there is no point in keeping total CPUs counter)

>However, this can cause bad side effects: especially for decay

But shouldn't be the special case that does not worth supporting? (since sounds pretty tricky especially with dynamic affinity changes)?

interwq:
re: `narenas` setting -- you can set it to the total number of CPUs (e.g. via env var MALLOC_CONF), then the arena slots will be reserved and available for percpu_arena, effectively the percpu_all you asked.

The case with dynamic CPU affinity is tricky to support; to be honest we didn't really try hard to optimize for the case. For the decay issue, enabling background_thread should solve it (since the dedicated jemalloc background threads will be able to purge for the inactive arenas). There might also be other unoptimized cases as well, like the mutex spin hint you linked.

azat:
>re: narenas setting -- you can set it to the total number of CPUs (e.g. via env var MALLOC_CONF), then the arena slots will be reserved and available for percpu_arena, effectively the percpu_all you asked.

Agree, but my point was that you cannot do this from the binary

interwq:
That's true. Unfortunately it won't be easy to make it runtime changeable, and at the moment it's not on our plan.

Closing but feel free to ask if anything.

PR #1942, opened 2020-09-28
Hugepages "glue" code

davidtgoldblatt:
This ties hpa_central and psset into a "complete" implementation of PAI.

I don't regard this as the final version, but canary results have been promising regardless just by moving all slabs onto the hugepage allocator (and using the central allocator solely to back the psset). There are still lots of questions I want to answer, and things to iterate on once we have data from real production runs:
- How do we want stats to be incorporated? Right now we don't do a great job with the central stats at all; they don't make sense as part of merged arena stats, but omitting them is confusing.
- How should the slab-level and bin-level fragmentation avoidance work? Right now, the psset zeros out extent serial numbers, so we pick solely based on address within the bins. Practically this doesn't seem to matter much, but it feels wrong.
- How should purging work?
- How should tuning parameters change? The psset is more memory hungry, but also much faster than the PAC. In canary runs I've been cutting the number of arenas in half, while increasing the number of bin shards.
- This design supports 1G hugepages fairly naturally. We should start using them.
- How do we want to handle the wide variety of kernel tuning knobs across different OSs, and the wildly different behaviors that can happen just on (e.g.) Linux.


src/jemalloc.c
```diff
@@ -1737,6 +1742,24 @@ malloc_init_hard_a0_locked() {
 		return true;
 	}
 	a0 = arena_get(TSDN_NULL, 0, false);
+
+	if (opt_hpa) {
+		/*
+		 * The global HPA uses the edata cache from a0, and so needs to
```
interwq, line 1767: Question: should we limit the HPA to auto arenas only? I see that right now it's guarded under custom hooks, but manual arenas with default hooks may still have it? If we want to make that work, arena_reset would need special care for the HPA.

davidtgoldblatt: I think arena_reset works correctly as is in this implementation -- the slabs / large extents get freed back to PA in the arena-level code, so we'll evict everything from the psset back to the hpa_central before arena destruction.

I'm fairly ambivalent about the broader question of whether or not we ought to (independent of whether or not it's correct to); having manual arenas have drastically different perf characteristics from auto arenas feels wrong, but OTOH most programs currently using manual arenas are probably expecting one set of time/space tradeoffs, and we'd be giving them another.

I think if we're careful with our jemalloc 6 design changes, we can make this question easier to answer.

src/hpa.c
```diff
@@ -0,0 +1,395 @@
...
+
+	edata = edata_cache_get(tsdn, hpa->edata_cache);
+	if (edata == NULL) {
+		malloc_mutex_unlock(tsdn, &hpa->grow_mtx);
+		pages_unmap(addr, alloc_size);
+		return NULL;
+	}
+
+	edata_init(edata, hpa->ind, addr, alloc_size, /* slab */ false,
+	    SC_NSIZES, /* sn */ 0, extent_state_active, /* zeroed */ true,
```
interwq, line 161: nit: comment why sn is 0

src/hpa.c
```diff
@@ -0,0 +1,395 @@
...
+	(void)err;
+
+	edata = edata_cache_get(tsdn, hpa->edata_cache);
+	if (edata == NULL) {
+		malloc_mutex_unlock(tsdn, &hpa->grow_mtx);
+		pages_unmap(addr, alloc_size);
+		return NULL;
+	}
+
+	edata_init(edata, hpa->ind, addr, alloc_size, /* slab */ false,
```
interwq, line 160: why arena ind is hpa->ind? Should we use shard ind instead? (and remove L169?)

src/hpa.c
```diff
@@ -0,0 +1,395 @@
...
+static edata_t *
+hpa_alloc(tsdn_t *tsdn, pai_t *self, size_t size,
+    size_t alignment, bool zero) {
+
+	assert((size & PAGE_MASK) == 0);
+	/* We don't handle alignment or zeroing for now. */
+	if (alignment > PAGE) {
+		return NULL;
+	}
+	if (zero) {
```
interwq, line 239: nit: slightly prefer combining the two branches. feel free to ignore.

src/hpa.c
```diff
@@ -0,0 +1,395 @@
...
+		return NULL;
+	}
+
+	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
+	    WITNESS_RANK_CORE, 0);
+
+	hpa_shard_t *shard = (hpa_shard_t *)self;
```
interwq: nit: might as well add a wrapper function to do the pai <==> shard conversion. Probably can add a few extra assertions in the wrapper.

src/hpa.c
```diff
@@ -0,0 +1,395 @@
...
+
+	edata_t *edata;
+	if (size <= shard->ps_alloc_max) {
+		edata = hpa_alloc_psset(tsdn, shard, size);
+		emap_register_boundary(tsdn, shard->hpa->central.emap, edata,
```
interwq, line 252: check if edata is NULL first?

src/hpa.c
```diff
@@ -0,0 +1,395 @@
...
+	if (edata_ps_get(edata) != NULL) {
+		emap_deregister_boundary(tsdn, shard->hpa->central.emap, edata);
+
+		malloc_mutex_lock(tsdn, &shard->mtx);
+		edata_t *evicted_ps = psset_dalloc(&shard->psset, edata);
+		malloc_mutex_unlock(tsdn, &shard->mtx);
+
+		edata_cache_put(tsdn, shard->edata_cache, edata);
+
+		if (evicted_ps != NULL) {
```
interwq, line 322: The code logic here is pretty smooth and easy to follow. I'd still say to add some high level comments somewhere about the dalloc logic, e.g. what / when goes to shard vs hpa_central.

src/hpa.c
```diff
@@ -0,0 +1,395 @@
...
+	 * everything from the psset, so it shouldn't be able to serve even a
+	 * 1-page allocation.
+	 */
+	if (config_debug) {
+		edata_t edata = {0};
+		malloc_mutex_lock(tsdn, &shard->mtx);
+		bool psset_empty = psset_alloc_reuse(&shard->psset, &edata,
+		    PAGE);
+		malloc_mutex_unlock(tsdn, &shard->mtx);
+		assert(psset_empty);
```
interwq, line 357: nit: maybe also check some stats (if it's easy to do)?

src/jemalloc.c
```diff
@@ -133,6 +133,10 @@ unsigned	ncpus;
 
 /* Protects arenas initialization. */
 malloc_mutex_t arenas_lock;
+
+/* The global hpa, and whether it's on. */
+bool opt_hpa;
```
interwq: nit: let's explicitly set a default value.

src/pa.c
```diff
@@ -55,8 +86,14 @@ pa_alloc(tsdn_t *tsdn, pa_shard_t *shard, size_t size, size_t alignment,
...
 	    WITNESS_RANK_CORE, 0);
 
-	edata_t *edata = pai_alloc(tsdn, &shard->pac.pai, size, alignment,
-	    zero);
+	edata_t *edata = NULL;
+	if (atomic_load_b(&shard->use_hpa, ATOMIC_RELAXED)) {
+		edata = pai_alloc(tsdn, &shard->hpa_shard.pai, size, alignment,
+		    zero);
+	}
+	if (edata == NULL) {
```
interwq, line 95: nit: a quick comment about falling back to pac can help.

src/pa.c
```diff
@@ -79,8 +116,10 @@ pa_expand(tsdn_t *tsdn, pa_shard_t *shard, edata_t *edata, size_t old_size,
 
 	size_t expand_amount = new_size - old_size;
 
-	bool error = pai_expand(tsdn, &shard->pac.pai, edata, old_size,
-	    new_size, zero);
+	pai_t *pai = (edata_pai_get(edata) == EXTENT_PAI_PAC
+	    ? &shard->pac.pai : &shard->hpa_shard.pai);
```
interwq, line 121: nit: add a wrapper to get pai from edata.

interwq (commented):
First wave of comments / questions

interwq:
Have only read the first commit; so some questions might answer themselves in latter commits. Feel free to let me know / ignore these.

include/jemalloc/internal/hpa.h
```diff
@@ -0,0 +1,70 @@
...
+#define JEMALLOC_INTERNAL_HPA_H
+
+#include "jemalloc/internal/geom_grow.h"
+#include "jemalloc/internal/hpa_central.h"
+#include "jemalloc/internal/pai.h"
+#include "jemalloc/internal/psset.h"
+
+typedef struct hpa_s hpa_t;
+struct hpa_s {
+	malloc_mutex_t grow_mtx;
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

include/jemalloc/internal/pa.h
```diff
@@ -66,12 +67,31 @@ struct pa_shard_s {
 	 */
 	atomic_zu_t nactive;
 
+	/*
+	 * Whether or not we should prefer the hugepage allocator.  Atomic since
+	 * it may be concurrently modified by a thread setting extent hooks.
+	 * Note that we still may do HPA operations in this arena; if use_hpa is
+	 * changed from true to false, we'll
```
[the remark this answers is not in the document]
davidtgoldblatt: Yep, fixed.

include/jemalloc/internal/hpa.h
```diff
@@ -0,0 +1,70 @@
...
+	malloc_mutex_t mtx;
+	hpa_central_t central;
+	/* The arena ind we're associated with. */
+	unsigned ind;
+	edata_cache_t *edata_cache;
+	geom_grow_t geom_grow;
```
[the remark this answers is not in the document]
davidtgoldblatt: We could, although I don't know how useful it'd be (maybe as a stat? "Ah, you're not getting any hugepages because all your arenas use custom extent hooks"?). Up to you.

include/jemalloc/internal/hpa.h
```diff
@@ -0,0 +1,70 @@
...
+#include "jemalloc/internal/psset.h"
+
+typedef struct hpa_s hpa_t;
+struct hpa_s {
+	malloc_mutex_t grow_mtx;
+	malloc_mutex_t mtx;
+	hpa_central_t central;
+	/* The arena ind we're associated with. */
+	unsigned ind;
+	edata_cache_t *edata_cache;
```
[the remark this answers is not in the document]
davidtgoldblatt: Added some comments; centralized allocations all use the same edata cache, whereas psset allocations use their "home-base" arena edata cache.

include/jemalloc/internal/hpa.h
```diff
@@ -0,0 +1,70 @@
...
+	malloc_mutex_t mtx;
+	edata_cache_t *edata_cache;
+	hpa_t *hpa;
+	psset_t psset;
+
+	/*
+	 * When we're grabbing a new ps from the central allocator, how big
+	 * would we like it to be?
+	 */
+	size_t ps_goal;
```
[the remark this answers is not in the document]
davidtgoldblatt: Added some comments; ps_goal is mostly about batching, ps_alloc_max is mostly about fragmentation avoidance.

src/hpa.c
```diff
@@ -0,0 +1,395 @@
...
+	 * No such luck. We've dropped mtx, so other allocations can proceed
+	 * while we allocate the new extent.  We know no one else will grow in
+	 * the meantime, though, since we still hold grow_mtx.
+	 */
+	size_t alloc_size;
+	pszind_t skip;
+
+	size_t hugepage_goal_min = HUGEPAGE_CEILING(size_goal);
+
+	err = geom_grow_size_prepare(&hpa->geom_grow, hugepage_goal_min,
```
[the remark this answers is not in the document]
davidtgoldblatt: We don't; it can grow to accomodate however many allocations it gets (same as the ecache).

src/hpa.c
davidtgoldblatt: Done.

src/hpa.c
davidtgoldblatt: Added some comments; the arena ind we use during growth should be the one associated with the global allocator. Once we've allocated a (probably much smaller) extent out of that large extent, it gets the arena index of the arena actually doing the allocation (rather than the index of the centralized allocator).

src/hpa.c
davidtgoldblatt: Agreed, done.

src/hpa.c
davidtgoldblatt: Done.

src/hpa.c
davidtgoldblatt: Good catch, done.

src/hpa.c
davidtgoldblatt: Added some comments.

src/hpa.c
davidtgoldblatt: Done.

src/jemalloc.c

src/pa.c
davidtgoldblatt: Done.

src/pa.c
davidtgoldblatt: Done.

include/jemalloc/internal/hpa.h
```diff
@@ -0,0 +1,92 @@
...
+
+#include "jemalloc/internal/geom_grow.h"
+#include "jemalloc/internal/hpa_central.h"
+#include "jemalloc/internal/pai.h"
+#include "jemalloc/internal/psset.h"
+
+typedef struct hpa_s hpa_t;
+struct hpa_s {
+	/*
+	 * We have to mutexes for the central allocator; mtx protects its state,
```
interwq, line 12: typo: two mutexes

include/jemalloc/internal/hpa.h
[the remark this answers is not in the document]
interwq: No need to do it now. Let's wait and see.

include/jemalloc/internal/hpa.h
```diff
@@ -0,0 +1,92 @@
...
+	 * while grow_mtx protects controls the ability to grow the backing
+	 * store.  This prevents race conditions in which the central allocator
+	 * has exhausted its memory while mutiple threads are trying to
+	 * allocate.  If they all reserved more address space from the OS
+	 * without synchronization, we'd end consuming much more than necessary.
+	 */
+	malloc_mutex_t grow_mtx;
+	malloc_mutex_t mtx;
+	hpa_central_t central;
+	/* The arena ind we're associated with. */
```
interwq, line 22: Does the central allocator also has an associated arena?

include/jemalloc/internal/hpa.h
```diff
@@ -0,0 +1,92 @@
...
+	 * pai must be the first member; we cast from a pointer to it to a
+	 * pointer to the hpa_shard_t.
+	 */
+	pai_t pai;
+	malloc_mutex_t grow_mtx;
+	malloc_mutex_t mtx;
+	/*
+	 * This edata cache is the one we use when allocating a small extent
+	 * from a pageslab.  The pageslab itself comes from the centralized
+	 * allocator, and so will use its edata_cache.
```
interwq, line 51: The comments make sense. Can we make assertions on the correct edata cache being used? My concern is mainly around, if we get one wrong, there won't be any crash but rather poor locality which is hard to track down.

src/hpa.c
```diff
@@ -0,0 +1,444 @@
...
+	assert(edata_szind_get_maybe_invalid(edata) == SC_NSIZES);
+	assert(!edata_slab_get(edata));
+	assert(edata_committed_get(edata));
+	assert(edata_base_get(edata) != NULL);
+
+	/*
+	 * Either the psset field is NULL, in which case this doesn't come from
+	 * the psset and instead comes from the centralized allocator (and
+	 * should be freed there directly).  If it *is* a centralized
+	 * allocation, we return it to the psset instead.
```
interwq, line 344: I'm not following the last sentence?

src/pa.c
```diff
@@ -49,14 +80,30 @@ pa_shard_destroy(tsdn_t *tsdn, pa_shard_t *shard) {
...
 edata_t *
 pa_alloc(tsdn_t *tsdn, pa_shard_t *shard, size_t size, size_t alignment,
     bool slab, szind_t szind, bool zero) {
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, 0);
 
-	edata_t *edata = pai_alloc(tsdn, &shard->pac.pai, size, alignment,
-	    zero);
+	edata_t *edata = NULL;
+	if (atomic_load_b(&shard->use_hpa, ATOMIC_RELAXED)) {
```
interwq, line 105: question: is it true that, with HPA off, this branch is the only runtime overhead?

include/jemalloc/internal/hpa.h
davidtgoldblatt: Fixed.

interwq:
One more question: can we also add an option, to limit the total size of hpa? That may help to balance the CPU / memory tradeoff. 

Also to confirm, since the SEC is tied to HPA shards, so when hpa is off, SEC won't be touched as well, right?

include/jemalloc/internal/hpa.h
davidtgoldblatt: It depends on what you mean; it's edatas have a value in those bits that they get from their edata cache, but practically it serves mostly as a source for edata_ts. The arena index of edata_ts in the central arena isn't particularly meaningful (eventually it might be, if we decide to shard it better).

include/jemalloc/internal/hpa.h
davidtgoldblatt: Added an assert to pa.c's alloc function.

src/hpa.c
davidtgoldblatt: This whole comment doesn't make a ton of sense. I rewrote it.

src/pa.c
davidtgoldblatt: Should be (well, not counting stuff like the size of the arena struct).

davidtgoldblatt:
> One more question: can we also add an option, to limit the total size of hpa? That may help to balance the CPU / memory tradeoff.

Mind deferring out of this PR (slash, potentially until we've got a firmer idea how usage looks)? This is trickier than it sounds because of overcommit; we'd need to start tracking ever-touchedness on a per-extent basis (which, note: need not correspond to any physical hugepage).

> Also to confirm, since the SEC is tied to HPA shards, so when hpa is off, SEC won't be touched as well, right?

Right.

interwq:
re: the total size option, how about we do an oversimplified version first, i.e. not doing the accurate counting at all, but rather serve as an upper limit guarantee. That way it should be simple I think? It might even give us half of the benefit, tradeoff tuning-wise.

davidtgoldblatt:
Added the max functionality, though the more I think about it the less sense I think it makes. We can talk tomorrow.

src/sec.c
```diff
@@ -0,0 +1,263 @@
...
+
+	sec_t *sec = (sec_t *)self;
+
+	if (zero || alignment > PAGE || sec->nshards == 0
+	    || size > sec->alloc_max) {
+		return pai_alloc(tsdn, sec->fallback, size, alignment, zero);
+	}
+	pszind_t pszind = sz_psz2ind(size);
+	sec_shard_t *shard = sec_shard_pick(tsdn, sec);
+	malloc_mutex_lock(tsdn, &shard->mtx);
```
interwq, line 102: maybe do an atomic load on the bytes, and skip the whole lock-alloc-unlock if it's 0?

include/jemalloc/internal/sec.h
```diff
@@ -0,0 +1,107 @@
...
+
+typedef struct sec_shard_s sec_shard_t;
+struct sec_shard_s {
+	/*
+	 * We don't keep per-bin mutexes, even though that would allow more
+	 * sharding; this allows global cache-eviction, which in turn allows for
+	 * better balancing across free lists.
+	 */
+	malloc_mutex_t mtx;
+	/* See the note on the sec 'enabled' field below. */
```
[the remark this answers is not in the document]
davidtgoldblatt: Fixed.

include/jemalloc/internal/sec.h
```diff
@@ -0,0 +1,107 @@
...
+	 * they're larger than alloc_max bytes.
+	 */
+	size_t alloc_max;
+	/*
+	 * Exceeding this amount of cached extents in a shard causes *all* of
+	 * the shards in that bin to be flushed.
+	 */
+	size_t bytes_max;
+
+	/* How many shards are active (which may be less than the max. */
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

src/sec.c
```diff
@@ -0,0 +1,263 @@
...
+
+	return false;
+}
+
+static sec_shard_t *
+sec_shard_pick(tsdn_t *tsdn, sec_t *sec) {
+	/*
+	 * Eventually, we should implement affinity, tracking source shard using
+	 * the edata_t's newly freed up fields.  For now, just randomly
+	 * distribute across all shards.
```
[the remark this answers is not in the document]
davidtgoldblatt: See the next commit -- the SEC is within the arenas; every thread touching a given SEC is already coming from the same arena.

What we really want is *thread* affinity to a particular shard within an arena (the same way bin shards work), with a given thread having a "home-base" shard. But there are some design questions at that point; you could imagine that we should return extents to the allocating shard to handle producer-consumer situations, but you could also imagine that we should instead focus more on batching optimizations to reduce remote-shard lock acquisitions during deallocations. Exploring the design space there is complicated enough (and dependent on other changes, like batch alloc/dalloc in the PAI) that I wanted to defer to a later PR. Randomizing practically works well fairly well it seems.

src/sec.c
davidtgoldblatt: This is probably not compatible with later batching optimizations -- we'll want to hold the lock for a more complicated operation than just examining a linked list. (It also shouldn't fire often -- if it does, it means that caching in general is failing, in which case micro-optimizing the sharded layer is the least of our worries).

src/sec.c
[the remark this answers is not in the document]
interwq: I have no doubt that the randomized sharding scales well in the current workload. However I'd still want us to maintain the stable and predictable behavior (e.g. memory alloc pattern / layout). How about we go with a very simple thread affinity similar to the bin shards, e.g. round robin binding when seeing new threads, and each thread only alloc / dalloc to the current binding shard.

I understand there will be further enhancements, however IMO we should base the next changes on a stable layout. Although it's not something we officially guarantee, I consider it a very useful and necessary feature. IIUC, the randomized approach will have unpredictable layout even with a single thread (which I consider worth more than even noticeable memory / cpu wins).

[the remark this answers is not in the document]
davidtgoldblatt: Done.

include/jemalloc/internal/pa.h
```diff
@@ -85,7 +86,12 @@ struct pa_shard_s {
 	/* Allocates from a PAC. */
 	pac_t pac;
 
-	/* Allocates from a HPA. */
+	/*
+	 * We place a small extent cache in front of the HPA, since we intend
+	 * these configurations to use many fewer arenas, and therefore have a
+	 * higher risk of hot locks.
+	 */
+	sec_t hpa_sec;
```
interwq, line 94: We should also flush the sec, when threads are terminating, e.g. similar to the last thread leaving an arena here: https://github.com/jemalloc/jemalloc/blob/5e41ff9b740258bddebcbd5575e1670a15f8b1ae/src/tcache.c#L721

Also is it true that, threads accessing the same shard all can flush all the sec shards, so we only need to worry about when the last thread is leaving?

include/jemalloc/internal/edata.h
```diff
@@ -193,16 +194,11 @@ struct edata_s {
 	};
 
 	/*
-	 * Reserved for hugepages -- once that allocator is more settled, we
-	 * might be able to claw some of this back.  Until then, don't get any
-	 * funny ideas about using the space we just freed up to keep some other
-	 * bit of metadata around.  That kind of thinking can be hazardous to
```
interwq, line 199: Glad that we remove one hazard for the developers!

interwq (commented):
Looks great!

include/jemalloc/internal/pa.h
davidtgoldblatt: Done. I hooked this into the purge pathway generally, since that's probably what someone wants if they're purging manually, too.

include/jemalloc/internal/edata.h
davidtgoldblatt: Vague threats of bodily harm are an important part of the development process.

interwq (approved):

PR #1943, opened 2020-09-30
Fix compilation with TinyCC compiler

XVilka:
image

As part of the [effort to make radare2 buildable with TinyCC](https://github.com/radareorg/radare2/pull/17295), which is the only option on some systems since easier to port rather than GCC or Clang.

Current [TinyCC compiler](https://bellard.org/tcc/) version is at https://repo.or.cz/tinycc.git/ - `mob` branch.

### DIFF
--- a/include/jemalloc/internal/bit_util.h
+++ b/include/jemalloc/internal/bit_util.h
@@ -75,7 +75,7 @@ fls_u_slow(unsigned x) {
 
 #undef DO_FLS_SLOW
 
-#ifdef JEMALLOC_HAVE_BUILTIN_CLZ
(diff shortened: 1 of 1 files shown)


jasone:
This is better done as a fix to the autoconf feature test which erroneously defines `JEMALLOC_HAVE_BUILTIN_CLZ`. And that leads to the question: are you sure the configure script used tcc? If you configure with gcc/clang, then compile with tcc, these sorts of failures are to be expected.

XVilka:
We use only the parts of the jemalloc, so this is might be the case. I will think then on how to fix this better.

PR #1944, opened 2020-10-01
Add prof_dump_file_overwrite option

davemarchevsky:
Enabling this boolean option results in dumps _NOT_ including seq or
vseq in the filename. Dumps are always written to the same file for a
particular pid and dump type. If a previous dump file exists, it is
unlinked before a new dump is written.

My use-case is as follows:
 * I configure some processes to dump to a known directory via `prof_prefix`, `lg_prof_sample`, `lg_prof_interval` options
    * `prof_prefix` sends the dumps to a small `tmpfs` such that they're never written out to disk
 * Another process ('watcher') periodically checks this directory and does some processing with the _last_ (largest seq num) dump for each dump type
 * When the watcher chooses to process the _last_ files, it deletes the processed files and all earlier dumps

Since my watcher process only cares about the last dump file for each pid, and I have a small `tmpfs` that many processes use, I'd like to limit the number of extraneous files dumped. (Configuring `lg_prof_interval` to closely match the period of 'watcher' such that there aren't any extraneous files is impractical.)

Notes:
* I added an `unlink` before `creat` in `prof_dump_open_file_impl`. This is to deal with the scenario where the watcher is processing an extant dump for a pid while it concurrently decides to do another dump. With `prof_dump_file_overwrite:true` set, this would result in `creat` truncating the existing file. Happy to gate this behind `prof_dump_file_overwrite`.
* I don't know what I'm doing in terms of testing or anything, this is my first contribution. Will look at extant tests and see where I can add tests for this functionality.

### DIFF
--- a/src/prof_sys.c
+++ b/src/prof_sys.c
@@ -373,6 +373,7 @@ prof_dump_check_possible_error(prof_dump_arg_t *arg, bool err_cond,
 
 static int
 prof_dump_open_file_impl(const char *filename, int mode) {
+	unlink(filename);
 	return creat(filename, mode);
 }
 prof_dump_open_file_t *JET_MUTABLE prof_dump_open_file =
@@ -544,16 +545,23 @@ prof_dump_filename(tsd_t *tsd, char *filename, char v, uint64_t vseq) {
 	assert(tsd_reentrancy_level_get(tsd) == 0);
 	const char *prof_prefix = prof_dump_prefix_get(tsd_tsdn(tsd));
 
-	if (vseq != VSEQ_INVALID) {
-	        /* "<prefix>.<pid>.<seq>.v<vseq>.heap" */
-		malloc_snprintf(filename, DUMP_FILENAME_BUFSIZE,
-		    "%s.%d.%"FMTu64".%c%"FMTu64".heap",
-		    prof_prefix, prof_getpid(), prof_dump_seq, v, vseq);
-	} else {
-	        /* "<prefix>.<pid>.<seq>.<v>.heap" */
-		malloc_snprintf(filename, DUMP_FILENAME_BUFSIZE,
-		    "%s.%d.%"FMTu64".%c.heap",
-		    prof_prefix, prof_getpid(), prof_dump_seq, v);
+	if (!opt_prof_dump_overwrite) {
+	    if (vseq != VSEQ_INVALID) {
+	            /* "<prefix>.<pid>.<seq>.v<vseq>.heap" */
+		    malloc_snprintf(filename, DUMP_FILENAME_BUFSIZE,
+		        "%s.%d.%"FMTu64".%c%"FMTu64".heap",
+		        prof_prefix, prof_getpid(), prof_dump_seq, v, vseq);
+	    } else {
+	            /* "<prefix>.<pid>.<seq>.<v>.heap" */
+		    malloc_snprintf(filename, DUMP_FILENAME_BUFSIZE,
+		        "%s.%d.%"FMTu64".%c.heap",
+		        prof_prefix, prof_getpid(), prof_dump_seq, v);
+	    }
+	} else { // opt_prof_dump_overwrite
+	    /* "<prefix>.<pid>.<v>.heap" */
+	    malloc_snprintf(filename, DUMP_FILENAME_BUFSIZE,
+	        "%s.%d.%c.heap",
+	        prof_prefix, prof_getpid(), v);
 	}
 	prof_dump_seq++;
 }
--- a/include/jemalloc/internal/prof_externs.h
+++ b/include/jemalloc/internal/prof_externs.h
@@ -24,6 +24,9 @@ extern bool opt_prof_unbias;
 /* For recording recent allocations */
 extern ssize_t opt_prof_recent_alloc_max;
 
+/* Whether to overwrite seq-less dump filenames */
+extern bool opt_prof_dump_overwrite;
+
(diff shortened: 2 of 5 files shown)


src/prof_sys.c
```diff
@@ -373,6 +373,7 @@ prof_dump_check_possible_error(prof_dump_arg_t *arg, bool err_cond,
 
 static int
 prof_dump_open_file_impl(const char *filename, int mode) {
+	unlink(filename);
```
yinan1048576, line 376: Can `unlink` fail, say if `filename` does not exist?

yinan1048576: I'm not too familiar with the `unlink` function, but we care about reentrancy for system calls. See how we handle other system calls in `include/jemalloc/internal/test_hooks.h`.

For failure handling, if `unlink` always return an error on new file name (when `prof_dump_file_overwrite` is off), then let's gate it.

Further, let's gate it at a higher level, in `prof_dump_open` instead of `prof_dump_open_impl`. The reason is unit testing. We want to test for how well we handle `unlink` failure. See how we define a function pointer for `prof_dump_open_impl` and how we test in `test/unit/prof_mdump.c`.

src/prof_sys.c
```diff
@@ -544,16 +545,23 @@ prof_dump_filename(tsd_t *tsd, char *filename, char v, uint64_t vseq) {
...
+	            /* "<prefix>.<pid>.<seq>.<v>.heap" */
+		    malloc_snprintf(filename, DUMP_FILENAME_BUFSIZE,
+		        "%s.%d.%"FMTu64".%c.heap",
+		        prof_prefix, prof_getpid(), prof_dump_seq, v);
+	    }
+	} else { // opt_prof_dump_overwrite
+	    /* "<prefix>.<pid>.<v>.heap" */
+	    malloc_snprintf(filename, DUMP_FILENAME_BUFSIZE,
+	        "%s.%d.%c.heap",
+	        prof_prefix, prof_getpid(), v);
```
yinan1048576, line 564: Is the new mode intended for every type of dump, or just the interval triggered dump? My guess is interval triggered dump only? If so you might not want to make things different for other types of dumps (e.g. manual dumps).

davemarchevsky: I was planning on having the mode apply to all modes' autogenerated filenames, for consistency's sake and code understandability. Definitely don't feel strongly about this though - my usecase will need to be able to parse both the 'overwrite' filenames and 'normal'. 

Happy to add a check for `v == 'i'` in this code and comment explaining why.

src/prof_sys.c
davemarchevsky: > Can unlink fail, say if filename does not exist?
> For failure handling, if unlink always return an error on new file name (when prof_dump_file_overwrite is off), then let's gate it.

If `unlink` fails, it returns `-1` and sets `errno`. I'm intentionally not checking `errno` because:
 * If file doesn't exist (`ENOENT`), it's not a problem because we `creat` it on next line, and is likely intentional because `prof_dump_file_overwrite` isn't set
* If some other fs issue occurs (fs is readonly, io error, no kernel memory...) it's likely that the `creat` call immediately following will have the same issue, and its error states are handled by `prof_dump_check_possible_error` in `prof_dump_open`. If it's some error that will not affect `creat`, this is also OK - but suboptimal - as `creat` will truncate the overwrite file.

> Further, let's gate it at a higher level, in prof_dump_open instead of prof_dump_open_impl. The reason is unit testing. We want to test for how well we handle unlink failure.

So you're saying `unlink` should occur in `prof_dump_open` before `prof_dump_open_file` is called? Will look into the unit testing aspects

> I'm not too familiar with the unlink function, but we care about reentrancy for system calls

Looks like it's signal-safe (https://man7.org/linux/man-pages/man7/signal-safety.7.html)

yinan1048576: I believe your reasoning is correct, but let's not base our code on external assumptions.

What I was suggesting is basically something like the following (in addition to changing `include/jemalloc/internal/test_hooks.h`):

```c
static int
prof_dump_unlink_file_impl(const char *filename) {
        return unlink(filename);
}
prof_dump_unlink_file_t *JET_MUTABLE prof_dump_unlink_file =
    prof_dump_unlink_file_impl;

static int
prof_dump_open_file_impl(const char *filename, int mode) {
        return creat(filename, mode);
}
prof_dump_open_file_t *JET_MUTABLE prof_dump_open_file =
    prof_dump_open_file_impl;

static void
prof_dump_open(prof_dump_arg_t *arg, const char *filename) {
        if (prof_dump_file_overwrite) {
                int err = prof_dump_unlink_file(filename);
                prof_dump_check_possible_error(...);
        }
        arg->prof_dump_fd = prof_dump_open_file(filename, 0644);
        prof_dump_check_possible_error(arg, arg->prof_dump_fd == -1,
            "<jemalloc>: failed to open \"%s\"\n", filename);
}
```
And in the unit tests, you can point `prof_dump_unlink_file` to a mock function to verify that the error handling logic is working.

Thoughts?

src/prof_sys.c
yinan1048576: By all modes you are basically meaning that there will ever be one dump file of each mode on the file system? It might make sense. I'll leave the decision to you then. The particular mode probably worth more thinking is the peak triggered dumping, a.k.a. `gdump`.

interwq:
Discussed offline. Will use the new prof dump hook to achieve the same goal.

PR #1954, opened 2020-10-13
Tcaches: Fix a subtle race condition.

davidtgoldblatt:
Without a lock held continuously between checking tcaches_past and incrementing
it, it's possible for two threads to go down manual creation path
simultaneously.  If the number of tcaches is one less than the maximum, it's
possible for both to create a tcache and increment tcaches_past, with the second
thread returning a value larger than TCACHES_MAX.

### DIFF
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -767,7 +767,7 @@ static bool
 tcaches_create_prep(tsd_t *tsd, base_t *base) {
 	bool err;
 
-	malloc_mutex_lock(tsd_tsdn(tsd), &tcaches_mtx);
+	malloc_mutex_assert_owner(tsd_tsdn(tsd), &tcaches_mtx);
 
 	if (tcaches == NULL) {
 		tcaches = base_alloc(tsd_tsdn(tsd), base,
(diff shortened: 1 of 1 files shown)


src/tcache.c
```diff
@@ -795,6 +794,8 @@ tcaches_create(tsd_t *tsd, base_t *base, unsigned *r_ind) {
 
 	bool err;
 
+	malloc_mutex_lock(tsd_tsdn(tsd), &tcaches_mtx);
+
 	if (tcaches_create_prep(tsd, base)) {
```
interwq, line 799: It looks there's no simple way to 1) not holding the lock while allocating, and 2) reserving the tcache id and handling OOM properly at the same time?

davidtgoldblatt: Yeah; the other thing we could do is check and the recheck once we reacquire the lock. But that's actually reasonably complex, because it then means that the tcache allocation becomes speculative and we might need to do cleanup for it as well. (Also, tcache-sized object allocation is fast enough that it's not clear that doubling the number of acquisitions/releases is worth it to reduce the hold time).

interwq (approved):

PR #1955, opened 2020-10-13
Add a logo to doc_internal.

davidtgoldblatt:
This is the logo from the jemalloc development team's snazzy windbreakers.  We
don't actually use it in any documentation yet, but there's no reason we
couldn't.  In the meantime, it's probably best if it exists somewhere more
stable than various email inboxes.


davidtgoldblatt:
Discussed offline.

PR #1956, opened 2020-10-14
Add thread name in prof last-N records

yinan1048576:
What I have for now may or may not work: it depends on the integrity of the `tdata` the `tctx` points to. I saw some cautioning comments:
https://github.com/jemalloc/jemalloc/blob/be9548f2bef30b75294fdd0eb6721d1bf6e6a56a/include/jemalloc/internal/prof_structs.h#L47-L52
I think the `tdata` should be living nicely as long as it has a positive number of associated `tctx`s? Does the "teardown" in that comment mean program exit, e.g. mainly for the `prof_final` case? Let me also see if any tests may fail...

### DIFF
--- a/src/prof_recent.c
+++ b/src/prof_recent.c
@@ -484,6 +484,12 @@ prof_recent_alloc_dump_node(emitter_t *emitter, prof_recent_t *node) {
 
 	emitter_json_kv(emitter, "alloc_thread_uid", emitter_type_uint64,
 	    &node->alloc_tctx->thr_uid);
+	prof_tdata_t *alloc_tdata = node->alloc_tctx->tdata;
+	assert(alloc_tdata != NULL);
+	if (alloc_tdata->thread_name != NULL) {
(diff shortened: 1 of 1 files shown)


interwq:
re: tdata liveness; I don't trust my memory right now. @davidtgoldblatt can tdata go away with tctx still alive? The comments seem to indicate its possible.

A related question @yinan1048576, where is the storage of the thread name? Just want to make sure that we copy it to some place we own.

yinan1048576:
> where is the storage of the thread name? Just want to make sure that we copy it to some place we own.

Yes, we copy it, and the `tdata` owns the copy.

yinan1048576:
Discussed offline. Let's ignore the comment for now, and if crashes happen somewhere, we can answer my question from there; otherwise, I'll cleanup the comment as well as the relevant logic at a later time.

interwq (approved):

PR #1957, opened 2020-10-15
Eagerly detect double free and sized dealloc bugs for large sizes.

interwq:
Eagerly detect double free and sized dealloc bugs for large sizes.  The cost is low enough (as edata will be accessed anyway) to be enabled all the time.


include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -203,13 +204,32 @@ arena_vsalloc(tsdn_t *tsdn, const void *ptr) {
 	return sz_index2size(full_alloc_ctx.szind);
 }
 
+JEMALLOC_ALWAYS_INLINE void
+large_dalloc_safety_checks(edata_t *edata, szind_t szind) {
+	/*
```
davidtgoldblatt, line 209: Guard with a `config_opt_safety_checks` check? We've at least thought about trying to turn the boat on code size, and we otherwise strip the safety check stuff during linking.

davidtgoldblatt (commented):
Other comment: can we add a test? I'm thinking something similar to the small object case, where we just leak in case of failure so the test can limp along and return.

include/jemalloc/internal/arena_inlines_b.h
interwq: Makes sense. Added.

interwq:
re: tests -- good point. Added `test/unit/double_free`, also modified `test/unit/size_check` to cover large sizes.

One awkward part though, I had to skip the double-free test for debug builds, otherwise too many assertions will be triggered. Manually tested with debug off, but with `config_opt_safety_check` which would be the production set up this targets anyway.

davidtgoldblatt (approved):

PR #1958, opened 2020-10-15
Add a hard limit on tcache max size class.

interwq:
For locality reasons, tcache bins are integrated in TSD.  Allowing all size
classes to be cached has little benefit, but takes up much thread local storage.
In addition, it complicates the layout which we try hard to optimize.

### DIFF
--- a/include/jemalloc/internal/tcache_structs.h
+++ b/include/jemalloc/internal/tcache_structs.h
@@ -54,7 +54,7 @@ struct tcache_slow_s {
 
 struct tcache_s {
 	tcache_slow_t	*tcache_slow;
-	cache_bin_t	bins[SC_NSIZES];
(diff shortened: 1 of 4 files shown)


include/jemalloc/internal/tcache_structs.h
```diff
@@ -54,7 +54,7 @@ struct tcache_slow_s {
 
 struct tcache_s {
 	tcache_slow_t	*tcache_slow;
-	cache_bin_t	bins[SC_NSIZES];
+	cache_bin_t	bins[TCACHE_NBINS_MAX];
```
interwq, line 57: Verified that this goes up to tcache->bins[72], i.e. the 8M size class. So the default config has 32 bins sitting unused (vs right now >200 unused bins).

davidtgoldblatt (approved):

PR #1961, opened 2020-10-20
Switch from opt.lg_tcache_max to opt.tcache_max

interwq:
Though for convenience, keep parsing lg_tcache_max.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -109,7 +109,7 @@ CTL_PROTO(opt_zero)
 CTL_PROTO(opt_utrace)
 CTL_PROTO(opt_xmalloc)
 CTL_PROTO(opt_tcache)
-CTL_PROTO(opt_lg_tcache_max)
+CTL_PROTO(opt_tcache_max)
 CTL_PROTO(opt_tcache_nslots_small_min)
 CTL_PROTO(opt_tcache_nslots_small_max)
 CTL_PROTO(opt_tcache_nslots_large)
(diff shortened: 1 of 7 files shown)


src/jemalloc.c
```diff
@@ -1379,8 +1382,21 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				CONF_HANDLE_BOOL(opt_xmalloc, "xmalloc")
 			}
 			CONF_HANDLE_BOOL(opt_tcache, "tcache")
-			CONF_HANDLE_SSIZE_T(opt_lg_tcache_max, "lg_tcache_max",
-			    -1, (sizeof(size_t) << 3) - 1)
+			CONF_HANDLE_SIZE_T(opt_tcache_max, "tcache_max",
+			    0, TCACHE_MAXCLASS_LIMIT, CONF_DONT_CHECK_MIN,
+			    CONF_CHECK_MAX, /* clip */ false)
+			if (CONF_MATCH("lg_tcache_max")) {
+				size_t m;
```
davidtgoldblatt, line 1389: Two things:
- I think better is to compare m against TCACHE_LG_MAXCLASS_LIMIT (this will better handle cases where the user erroneously inputs a number larger than the bits in size_t).
- I think we should clip rather than erroring out in case the user exceeds 8MB (i.e. I think "cache as much as you can" is a reasonable thing for the user to request).

src/ctl.c
```diff
@@ -362,7 +362,7 @@ static const ctl_named_node_t opt_node[] = {
 	{NAME("utrace"),	CTL(opt_utrace)},
 	{NAME("xmalloc"),	CTL(opt_xmalloc)},
 	{NAME("tcache"),	CTL(opt_tcache)},
-	{NAME("lg_tcache_max"),	CTL(opt_lg_tcache_max)},
+	{NAME("tcache_max"),	CTL(opt_tcache_max)},
```
davidtgoldblatt, line 365: I think this means we need a doc change as well.

interwq: Yeah. One thing I'm less sure is, do we want to keep `lg_tcache_max` documented? I'm leaning towards not mentioning it, but only slightly.

src/jemalloc.c
interwq: Changed both.

src/ctl.c
davidtgoldblatt: Agreed. A middle ground would be to mention it, but only in the doc for tcache_max (e.g. "as a convenience, this may also be set by specifying lg_tcache_max, which will be taken to be the base-2 logartihm of the setting of tcache_max"). No strong preference either way.

interwq: I like that better. Just updated.

PR #1962, opened 2020-10-20
Route batch allocation of small batch size to tcache

yinan1048576:
For batch allocation requests of small sizes, instead of going to the page allocator for a fresh new slab, follow the ordinary hierarchy i.e. tcache -> arena -> page allocator, to improve performance. The downside is of course memory locality; however, locality was originally a best effort anyway.

A few remarks on the current version:
- I don't yet have a great way to determine the size threshold for deciding which route ("tcache route" / "page allocator route") the request should go through. Right now I just put an arbitrary number as a placeholder. I think there are a few aspects we need to consider here:
  - Improvement on performance. If we plot the difference in performance between the two routes against the batch size, then I imagine there are two big jumps: one is around the point at which the request cannot be fulfilled by the tcache without going to the arena, the other is around the point at which the request cannot be fulfilled by existing slabs in the arena without going to the page allocator. Both these two points can be a wider range instead of a single point (so the jump may be a steady one rather than a sudden one). Meanwhile, not sure about the magnitude of the two jumps: I suspect that the first jump should be quite significant, but how about the second one?
  - Cost on locality: locality is correlated with performance of later memory access by the caller, in particular sequential memory access, so locality also comes with a performance price tag, though a harder-to-measure one. Therefore, in theory, we can also plot the difference in memory access performance between the two routes against the batch size, and try to find a good spot to set the threshold. The order of memory locality, from worst to best, is tcache -> existing slabs in arena -> fresh new slabs; not sure how that'd translate to how the plot would look like.
  - It seems to me that many things I touched above have some dependencies on the caller. So, if we want to be lazy / flexible, we also have the option of asking the caller to choose, and to of course also bear the tuning effort. In such a case, we may have a default (of e.g. always going through one route). For the caller flag, we can use `MALLOCX_TCACHE_AUTO` / `MALLOCX_TCACHE_NONE` for the distinction.
- If we decide to have a size threshold ourselves rather than letting the caller specify, then we may want to allow the caller to specify the mallocx tcache flag. If the batch size is large, we can ignore the specified flag. However, if the batch size is small, then we need to decide what to do when the caller specifies `MALLOCX_TCACHE_NONE`. Obviously, we can ignore it. Meanwhile, it might also make some sense to treat that as "I don't care about performance", so we can go for the "page allocator route", to optimize for locality. If so, then the caller essentially has a "one way" overwrite privilege to our "route decision".
- Previously, I applied special care to the batch allocation API so that, whenever a manual arena is specified, we only serve memory strictly from that arena. Now, if tcache is allowed, then we can no longer guarantee that in the batch allocation API, just as the case for all other APIs we have. Therefore, I decide to simply save the effort and not try even when I can i.e. when the tcache is bypassed.
- When going through the tcache route, I make the tcache bin act as a middleman, and I never explicitly touch the arena level. This approach may not be optimal if the batch size is large enough to require more than one refill from the arena, when compared to the alternative approach where, whenever the tcache is depleted, I directly go to the arena and collect the remaining batch. My current approach is just easier to implement, since it roughly follows the original logic structure I had. Meanwhile, it might not be too bad, because we only decide to route through the tcache when the batch size is small, and it has the advantage of not leaving the tcache empty for the next allocation request. However, I'm also open to the alternative approach, and my guess is that I'd need to do some refactorings first for it.
- For unit tests, I currently short-circuit the locality and stats tests whenever the tcache route is chosen: the locality tests are irrelevant; there might be some opportunity for some stats tests, but it might be complicated to write. Please let me know your thoughts.
- I previously did not optimize for large sizes, but now it's sometimes possible, when (a) the size is not too large to be cached in tcache and (b) the batch size is small. Will do that after the core logic settles.

### DIFF
--- a/include/jemalloc/internal/cache_bin.h
+++ b/include/jemalloc/internal/cache_bin.h
@@ -317,6 +317,18 @@ cache_bin_alloc(cache_bin_t *bin, bool *success) {
 	return cache_bin_alloc_impl(bin, success, true);
 }
 
+JEMALLOC_ALWAYS_INLINE cache_bin_sz_t
+cache_bin_alloc_batch(cache_bin_t *bin, size_t num, void **out) {
+	size_t n = cache_bin_ncached_get_internal(bin);
+	if (n > num) {
+		n = num;
+	}
+	memcpy(out, bin->stack_head, n * sizeof(void *));
+	bin->stack_head += n;
+	cache_bin_low_water_adjust(bin);
+	return n;
+}
+
 /*
  * Free an object into the given bin.  Fails only if the bin is full.
  */
--- a/include/jemalloc/internal/cache_bin.h
+++ b/include/jemalloc/internal/cache_bin.h
@@ -167,16 +167,21 @@ cache_bin_diff(cache_bin_t *bin, uint16_t earlier, uint16_t later) {
 	return later - earlier;
 }
 
-/* Number of items currently cached in the bin. */
+/* Number of items currently cached in the bin, without checking ncached_max. */
 static inline cache_bin_sz_t
-cache_bin_ncached_get(cache_bin_t *bin, cache_bin_info_t *info) {
+cache_bin_ncached_get_internal(cache_bin_t *bin) {
 	cache_bin_sz_t diff = cache_bin_diff(bin,
 	    (uint16_t)(uintptr_t)bin->stack_head, bin->low_bits_empty);
 	cache_bin_sz_t n = diff / sizeof(void *);
-
-	assert(n <= cache_bin_info_ncached_max(info));
 	assert(n == 0 || *(bin->stack_head) != NULL);
+	return n;
+}
 
+/* Number of items currently cached in the bin, with checking ncached_max. */
+static inline cache_bin_sz_t
+cache_bin_ncached_get(cache_bin_t *bin, cache_bin_info_t *info) {
+	cache_bin_sz_t n = cache_bin_ncached_get_internal(bin);
+	assert(n <= cache_bin_info_ncached_max(info));
 	return n;
 }
 
@@ -186,7 +191,7 @@ cache_bin_ncached_get(cache_bin_t *bin, cache_bin_info_t *info) {
  * A pointer to the position one past the end of the backing array.
  */
 static inline void **
-cache_bin_empty_position_get(cache_bin_t *bin, cache_bin_info_t *info) {
+cache_bin_empty_position_get(cache_bin_t *bin) {
 	cache_bin_sz_t diff = cache_bin_diff(bin,
 	    (uint16_t)(uintptr_t)bin->stack_head, bin->low_bits_empty);
 	uintptr_t empty_bits = (uintptr_t)bin->stack_head + diff;
@@ -204,7 +209,7 @@ cache_bin_empty_position_get(cache_bin_t *bin, cache_bin_info_t *info) {
 static inline void
 cache_bin_assert_empty(cache_bin_t *bin, cache_bin_info_t *info) {
 	assert(cache_bin_ncached_get(bin, info) == 0);
-	assert(cache_bin_empty_position_get(bin, info) == bin->stack_head);
+	assert(cache_bin_empty_position_get(bin) == bin->stack_head);
 }
 
 /*
@@ -213,7 +218,7 @@ cache_bin_assert_empty(cache_bin_t *bin, cache_bin_info_t *info) {
  * ncached >= low_water during flush).
  */
 static inline cache_bin_sz_t
-cache_bin_low_water_get_internal(cache_bin_t *bin, cache_bin_info_t *info) {
+cache_bin_low_water_get_internal(cache_bin_t *bin) {
 	return cache_bin_diff(bin, bin->low_bits_low_water,
 	    bin->low_bits_empty) / sizeof(void *);
 }
@@ -221,7 +226,7 @@ cache_bin_low_water_get_internal(cache_bin_t *bin, cache_bin_info_t *info) {
 /* Returns the numeric value of low water in [0, ncached]. */
 static inline cache_bin_sz_t
 cache_bin_low_water_get(cache_bin_t *bin, cache_bin_info_t *info) {
-	cache_bin_sz_t low_water = cache_bin_low_water_get_internal(bin, info);
+	cache_bin_sz_t low_water = cache_bin_low_water_get_internal(bin);
 	assert(low_water <= cache_bin_info_ncached_max(info));
 	assert(low_water <= cache_bin_ncached_get(bin, info));
 
@@ -238,6 +243,14 @@ cache_bin_low_water_get(cache_bin_t *bin, cache_bin_info_t *info) {
 static inline void
 cache_bin_low_water_set(cache_bin_t *bin) {
 	bin->low_bits_low_water = (uint16_t)(uintptr_t)bin->stack_head;
+}
+
+static inline void
+cache_bin_low_water_adjust(cache_bin_t *bin) {
+	if (cache_bin_ncached_get_internal(bin)
+	    < cache_bin_low_water_get_internal(bin)) {
+		cache_bin_low_water_set(bin);
+	}
 }
 
 JEMALLOC_ALWAYS_INLINE void *
@@ -365,8 +378,8 @@ struct cache_bin_ptr_array_s {
 static inline void
 cache_bin_init_ptr_array_for_fill(cache_bin_t *bin, cache_bin_info_t *info,
     cache_bin_ptr_array_t *arr, cache_bin_sz_t nfill) {
-	assert(cache_bin_ncached_get(bin, info) == 0);
-	arr->ptr = cache_bin_empty_position_get(bin, info) - nfill;
+	cache_bin_assert_empty(bin, info);
+	arr->ptr = cache_bin_empty_position_get(bin) - nfill;
 }
 
 /*
@@ -377,8 +390,8 @@ cache_bin_init_ptr_array_for_fill(cache_bin_t *bin, cache_bin_info_t *info,
 static inline void
 cache_bin_finish_fill(cache_bin_t *bin, cache_bin_info_t *info,
     cache_bin_ptr_array_t *arr, cache_bin_sz_t nfilled) {
-	assert(cache_bin_ncached_get(bin, info) == 0);
-	void **empty_position = cache_bin_empty_position_get(bin, info);
+	cache_bin_assert_empty(bin, info);
+	void **empty_position = cache_bin_empty_position_get(bin);
 	if (nfilled < arr->n) {
 		memmove(empty_position - nfilled, empty_position - arr->n,
 		    nfilled * sizeof(void *));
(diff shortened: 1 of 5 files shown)


yinan1048576:
> Previously, I applied special care to the batch allocation API so that, whenever a manual arena is specified, we only serve memory strictly from that arena. Now, if tcache is allowed, then we can no longer guarantee that in the batch allocation API, just as the case for all other APIs we have. Therefore, I decide to simply save the effort and not try even when I can i.e. when the tcache is bypassed.

Oops - forgot to revise the tests. Revised.

include/jemalloc/internal/cache_bin.h
```diff
@@ -294,6 +298,21 @@ cache_bin_alloc(cache_bin_t *bin, bool *success) {
...
+JEMALLOC_ALWAYS_INLINE cache_bin_sz_t
+cache_bin_alloc_batch(cache_bin_t *bin, size_t num, void **out) {
+	size_t n = cache_bin_ncached_get_internal(bin);
+	if (n > num) {
+		n = num;
+	}
+	memcpy(out, bin->stack_head, n * sizeof(void *));
+	bin->stack_head += n;
+	uint16_t head_low_bits = (uint16_t)(uintptr_t)bin->stack_head;
+	if (head_low_bits > bin->low_bits_low_water) {
```
davidtgoldblatt, line 310: This check is not necessarily correct in the case where a bin spans a 64k boundary. I think, given that we don't need to save every cycle on the batch paths, it's fine to compute the low_water_mark position explicitly.

include/jemalloc/internal/cache_bin.h
```diff
@@ -294,6 +298,21 @@ cache_bin_alloc(cache_bin_t *bin, bool *success) {
 	return cache_bin_alloc_impl(bin, success, true);
 }
 
+JEMALLOC_ALWAYS_INLINE cache_bin_sz_t
+cache_bin_alloc_batch(cache_bin_t *bin, size_t num, void **out) {
```
davidtgoldblatt, line 302: We should add a test for this.

davidtgoldblatt (commented):
There's a middle-ground on locality vs. tcache reuse that makes intuitive sense to me:
- Fill up as many of the allocation requests as we can with full slabs
- Fill up any remainder (i.e. nrequested % objects_per_slab_of_that_size) out of the tcache pathways.

This is not necessarily objectively correct (i.e. there's also memory benefits of batch allocating), but it does have the nice property of aligning with perf expectations of the regular alloc pathways.

davidtgoldblatt:
> i.e. there's also memory benefits of batch allocating

Rereading, this is unclear -- what I mean is that we lose the ability to fill in "gaps" in existing slabs, as opposed to allocating new ones (full as though they may be).

include/jemalloc/internal/cache_bin.h
interwq: It feels we gotta to better make sure the low_bits fields have only equality check against them. Maybe similar to the TSD, i.e. change the member names to something tedious to type, and then provide wrappers that we must always go through.

yinan1048576 (commented):
This round of revision is mainly for the following two things:
- cache bin: slightly restructured the functions, added some wrappers. Also added tests for the batch alloc function.
> I previously did not optimize for large sizes, but now it's sometimes possible, when (a) the size is not too large to be cached in tcache and (b) the batch size is small. Will do that after the core logic settles.
- I thought a bit more about it. First, what I said earlier is not necessarily correct, because the sizes in tcache may or may not be a super set of small sizes, and I have to handle all cases. In this revision, I don't assume anything about whether the maximum size in tcache is larger or smaller than the maximum small size, and make the logic general. The logic is now also general enough to deal with the following: (a) for large sizes in the tcache, we don't really fill the tcache when it's empty, and (b) for large sizes not in the tcache, the fallback logic no longer needs to be extracted out.

Regarding the route determination question I raised earlier, I slightly augmented my placeholder logic, but it still remains a placeholder for now. I think the "integral-fraction separation" mechanism being suggested can also make sense. I don't see a scientific way to make a decision here. If we decide to go for the "integral-fraction separation", then I have a few questions in mind:
- Unless the batch size is an exact multiple of `nregs`, the batch will for sure contain some "ill-locality" memory, though the count is always less than `nregs`. I think this may or may not be desired by callers: the caller is paying roughly the same latency, but is not getting the best locality. Also, down to the implementation level, the chunk of "ill-locality" memory may be concentrated at the front of the batch, which may or may not be desired say if the front elements are to be most frequently traversed through; we may at least want to scatter the "ill-locality" memory throughout the batch?
- If the cache bin for the requested size is small but the fraction is large, then we may need to refill the cache bin multiple times, which may or may not be more efficient than simply getting one slab through the page allocator? We may want some additional criteria here?

davidtgoldblatt:
> Unless the batch size is an exact multiple of nregs, the batch will for sure contain some "ill-locality" memory, though the count is always less than nregs. I think this may or may not be desired by callers: the caller is paying roughly the same latency, but is not getting the best locality. Also, down to the implementation level, the chunk of "ill-locality" memory may be concentrated at the front of the batch, which may or may not be desired say if the front elements are to be most frequently traversed through; we may at least want to scatter the "ill-locality" memory throughout the batch?

I think the way to think about this is that we're at least doing-no-harm. The lead/trail allocations may not be locality-optimizing, but at the same time, they wouldn't be if the user were doing them manually. I.e. batch allocation is more or less unconditionally a win relative to the user's best alternative. We try to give them both locality and speed improvements; if we fail, we give them only speed improvements.

Scattering the tcache allocations throughout seems like unnecessary complexity.

> If the cache bin for the requested size is small but the fraction is large, then we may need to refill the cache bin multiple times, which may or may not be more efficient than simply getting one slab through the page allocator? We may want some additional criteria here?

I think going directly to the arena bins for the excess allocations should work here; they already know how to fill N allocations into an array for arbitrarily large N.

davidtgoldblatt:
BTW first four commits look fine -- I'd be happy merging them if you wanted to split the last into a separate PR.

yinan1048576:
Rebase, and changed the design based on offline discussions.

The main principles underlying the new design are the following:
- We don't want to waste any memory on any freshly allocated slabs; otherwise, memory waste can grow quickly on certain memory allocation patterns.
- Conditioned on the above, we want to maximize locality, even at some potential cost of performance (non-optimal use of thread cache + higher page allocator level lock contention) and overall memory compactness (non-optimal use of non-full arena bins), as compared with the alternative of naively batching a series of `malloc()` calls. We believe the tradeoff we're making is sensible, but we can always revisit at a later time.

On the implementation level, a few things I think are worth mentioning:
- I tried not to place "ill-locality" memory at the front of the batch as I was mentioning before. It should benefit some usage cases e.g. where the batch is to be served as a batch of nodes arranged in BFS order for a binary search tree.
- The arena and the thread cache bin are now lazily initialized, since they may or may not be needed at all, and it's hard to tell in advance (due to profiling). This unfortunately also results in a huge `while` loop - I can try to simplify that later.
- I previously had an early termination whenever the fresh slab allocation fails. Now given the new logic flow, it's slightly trickier to have that, so I decided to just get rid of it, and mainly rely on the OOM termination mechanism at the end of the `while` loop.
- I was previously thinking "if the cache bin for the requested size is small but the fraction is large, then we may need to refill the cache bin multiple times". I realized that it shouldn't be a big deal - the thread cache is large relative to the slab size, and the fraction is strictly smaller than the slab size. I'm still not bothering going to the arena to fetch excess allocations when the thread cache is not enough, which I think has the downside of slightly complicating the logic as well as likely leaving the tcache empty for the next allocation request. I can change it if needed.
- I changed a lot of the unit test logic, mainly getting rid of things. First, the tests around bin stats becomes more tricky even when profiling is off, because it now also has a dependency on the thread cache, so I decided to get rid of it. Second, I'm not testing locality for profiling-on case, because it'd just be complicated. Let me know if there are any thoughts on the unit test part.

src/jemalloc.c
```diff
@@ -4132,9 +4115,63 @@ batch_alloc(void **ptrs, size_t num, size_t size, int flags) {
...
+				    && unlikely(ind >= SC_NBINS)) {
+					for (size_t i = 0; i < n; ++i) {
+						prof_tctx_reset_sampled(tsd,
+						    ptrs[filled + i]);
+					}
+				}
+				progress += n;
+				filled += n;
+			}
+		}
```
davidtgoldblatt, line 4175: It seems like this will never allocate out of the arena bins -- is that intentional?

yinan1048576: No, it first allocates from the arena bins, but only in exact multiples of `nregs`. See the `if (likely(ind < SC_NBINS) && batch >= nregs)` branch. You can also see in the unit tests that I'm testing locality for the segment `[0, batch / nregs * nregs)`. I understand that the logic is a bit confusing here though; the `while` loop gets overly long in this revision - I'll try to do some cleanup.

yinan1048576: Sorry - forget about what I just said - I thought you meant fresh slabs...

yinan1048576: The allocation out of arena bins is done implicitly. When the cache bin is not enough, the `if (progress < batch || prof_sample_event)` branch at the end of the `while` loop serves to fill the cache bin, and then in the next round of the `while` loop, the batch gets allocations from the cache bin. There might be some unnecessary steps here, compared to directly filling in from arena bins whenever the cache bin is not enough, but I believe the overhead should not be a big deal: we at most try to fill `batch % nregs` allocations from the cache, and the cache has a size of twice of `nregs` (with the condition that it's between 20 and 200), so I'd imagine there's once or twice of filling in the worst case, and perhaps close to no filling at all in the average case. I didn't do very careful computation though, so may be wrong. I can change the logic if we figure that the overhead is not ideal.

davidtgoldblatt: Ahhh I see, that's tricky. Could you add a comment explaining?

yinan1048576:
Added comments in several places to better clarify the intention behind the logic.

src/jemalloc.c
```diff
@@ -4148,8 +4170,41 @@ batch_alloc(void **ptrs, size_t num, size_t size, int flags) {
 					bin = &tcache->bins[ind];
 				}
 			}
+			/*
+			 * If we don't have a tcache bin, we don't want to
+			 * immediately give up, because there's the possibility
```
davidtgoldblatt, line 4175: This (i.e. that the user specified MALLOCX_TCACHE_NONE) is the only possibility, right?

src/jemalloc.c
```diff
@@ -4148,8 +4170,41 @@ batch_alloc(void **ptrs, size_t num, size_t size, int flags) {
...
+				 * In such cases, we rely on the slow path,
+				 * i.e. the mallocx() call at the end of the
+				 * while loop, to fill in the cache, and in the
+				 * next iteration of the while loop, the tcache
+				 * will contain a lot of memory, and we can
+				 * harvest them here.  Compared to the
+				 * alternative approach where we directly go to
+				 * the arena bins here, the overhead of our
+				 * current approach should usually be minimal,
+				 * since the tcache is usually larger than a
```
davidtgoldblatt, line 4201: Is this true? I think lg_fill_div is often positive, so we fill less than a slab amount into tcaches.

davidtgoldblatt (commented):
I think the last comment is really the key one -- i.e. noting that mallocx will cause a tcache fill, and so we don't really call it in a loop.

The others seem a lot like they emphasize the wrong thing -- i.e. the reason for the various fallback choices isn't really like, OOM (i.e. it's perfectly fine for our answer to all OOM questions to be "who cares? let's bail out early and return an error."). It's more about time/space/locality tradeoffs.

src/jemalloc.c
yinan1048576: As far as I understand, that should be the main one. Some other possibilities include:
- `tcache` has not been initialized, which should be rare.
- `tcache` is disabled by the caller explicitly.

src/jemalloc.c
yinan1048576: You're right.

yinan1048576 (commented):
Simplified / revised the comments.

davidtgoldblatt (approved):

PR #1967, opened 2020-10-29
Linux with musl does not implement pthread_getname_np but possible to…

devnexen:
… get

 from the related file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1745,6 +1745,37 @@ dnl Check if we have dlsym support.
   if test "x${je_cv_pthread_setname_np}" = "xyes" ; then
     AC_DEFINE([JEMALLOC_HAVE_PTHREAD_SETNAME_NP], [ ])
   fi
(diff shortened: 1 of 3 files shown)


davidtgoldblatt:
I think this sort of macro detection is probably not quite right (it also fires for non-musl non-glibc implementations, for instance, and will fail if musl gains pthread_getname_np, as it some mailing list messages imply it may). It seems like better would be to change the configure check; i.e. that pthread_getname_np and pthread_setname_np both work together as expected.

davidtgoldblatt:
OK, this looks good. Could you rebase down to a single commit with a cleaned up title/description? Something like

```
Detect pthread_getname_np explicitly.

At least one libc (musl) defines pthread_setname_np without defining
pthread_getname_np.  Detect the presence of each individually, rather than
inferring both must be defined if set is.
```

davidtgoldblatt:
Thanks!

PR #1969, opened 2020-10-30
Cache edatas in the HPA

davidtgoldblatt:
The intent with the HPA is to run with fewer arenas, to the point that contention on the edata_cache_ts becomes observable. This PR rewrites the edata_cache_small_t in response to the design evolution that's happened since it was introduced, sprinkles usages into the HPA.

### DIFF
--- a/src/edata_cache.c
+++ b/src/edata_cache.c
@@ -27,8 +27,7 @@ edata_cache_get(tsdn_t *tsdn, edata_cache_t *edata_cache) {
 		return base_alloc_edata(tsdn, edata_cache->base);
 	}
 	edata_avail_remove(&edata_cache->avail, edata);
-	size_t count = atomic_load_zu(&edata_cache->count, ATOMIC_RELAXED);
-	atomic_store_zu(&edata_cache->count, count - 1, ATOMIC_RELAXED);
+	atomic_load_sub_store_zu(&edata_cache->count, 1);
 	malloc_mutex_unlock(tsdn, &edata_cache->mtx);
 	return edata;
 }
@@ -37,8 +36,7 @@ void
 edata_cache_put(tsdn_t *tsdn, edata_cache_t *edata_cache, edata_t *edata) {
 	malloc_mutex_lock(tsdn, &edata_cache->mtx);
 	edata_avail_insert(&edata_cache->avail, edata);
-	size_t count = atomic_load_zu(&edata_cache->count, ATOMIC_RELAXED);
-	atomic_store_zu(&edata_cache->count, count + 1, ATOMIC_RELAXED);
+	atomic_load_add_store_zu(&edata_cache->count, 1);
 	malloc_mutex_unlock(tsdn, &edata_cache->mtx);
 }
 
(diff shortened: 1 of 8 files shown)


davidtgoldblatt:
Of note is that this PR has some bug that shows up in canaries but not the tests here. Poking at it now.

davidtgoldblatt:
(Bugs are fixed -- forgot to include some of the edata_cache_small accesses under the right locks).

src/edata_cache.c
```diff
@@ -62,48 +60,97 @@ edata_cache_small_init(edata_cache_small_t *ecs, edata_cache_t *fallback) {
...
+		atomic_load_sub_store_zu(&ecs->fallback->count, 1);
 	}
-	return false;
+	malloc_mutex_unlock(tsdn, &ecs->fallback->mtx);
+	edata = edata_list_inactive_first(&ecs->list);
+	if (edata != NULL) {
+		edata_list_inactive_remove(&ecs->list, edata);
+		ecs->count--;
+	} else {
+		edata = base_alloc_edata(tsdn, ecs->fallback->base);
```
interwq, line 102: Question: this implies the fall back never really allocates, and the ecs always handles that part? It's a bit unintuitive as I'd expect the "fallback" to deal with this. Maybe move the batch alloc loop above to a separate fallback alloc function?

src/edata_cache.c
```diff
@@ -62,48 +60,97 @@ edata_cache_small_init(edata_cache_small_t *ecs, edata_cache_t *fallback) {
...
+		edata_list_inactive_remove(&ecs->list, edata);
+		edata_avail_insert(&ecs->fallback->avail, edata);
+		nflushed++;
+	}
+	atomic_load_add_store_zu(&ecs->fallback->count, ecs->count);
+	malloc_mutex_unlock(tsdn, &ecs->fallback->mtx);
+	assert(nflushed == ecs->count);
+	ecs->count = 0;
+}
+void
```
interwq, line 129: nit: need a blank line

src/edata_cache.c
```diff
@@ -62,48 +60,97 @@ edata_cache_small_init(edata_cache_small_t *ecs, edata_cache_t *fallback) {
...
+	return edata;
 }
 
-void edata_cache_small_finish(tsdn_t *tsdn, edata_cache_small_t *ecs,
-    size_t num) {
-	while (ecs->count > num) {
-		/* Same deal here -- we should be batching. */
-		edata_t *edata = edata_cache_small_get(ecs);
+static void
+edata_cache_small_flush(tsdn_t *tsdn, edata_cache_small_t *ecs) {
```
interwq, line 108: nit: maybe _flush_all()

src/edata_cache.c
```diff
@@ -62,48 +60,97 @@ edata_cache_small_init(edata_cache_small_t *ecs, edata_cache_t *fallback) {
...
+		return;
 	}
+
+	/*
+	 * Prepend rather than append, to do LIFO ordering in the hopes of some
+	 * cache locality.
+	 */
+	edata_list_inactive_prepend(&ecs->list, edata);
+	ecs->count++;
+	if (ecs->count > EDATA_CACHE_SMALL_MAX) {
```
interwq, line 147: can `>` really happen?

src/edata_cache.c
davidtgoldblatt: Done.

src/edata_cache.c
davidtgoldblatt: Done.

src/edata_cache.c
davidtgoldblatt: Done.

src/edata_cache.c
davidtgoldblatt: The preceding count increment can push the count to 1 past the max. Added an assert that that holds.

interwq (approved):
LGTM

PR #1970, opened 2020-10-31
Add experimental.thread.activity_callback.

davidtgoldblatt:
This (experimental, undocumented) functionality can be used by users to track various statistics of interest at a finer level of granularity than the thread.

The intended use case is things like peak tracking, but on a per-request basis.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -264,6 +264,7 @@ CTL_PROTO(stats_retained)
 CTL_PROTO(stats_zero_reallocs)
 CTL_PROTO(experimental_hooks_install)
 CTL_PROTO(experimental_hooks_remove)
+CTL_PROTO(experimental_thread_activity_callback)
 CTL_PROTO(experimental_utilization_query)
 CTL_PROTO(experimental_utilization_batch_query)
 CTL_PROTO(experimental_arenas_i_pactivep)
@@ -712,6 +713,11 @@ static const ctl_named_node_t experimental_hooks_node[] = {
 	{NAME("remove"),	CTL(experimental_hooks_remove)}
 };
 
+static const ctl_named_node_t experimental_thread_node[] = {
+	{NAME("activity_callback"),
+		CTL(experimental_thread_activity_callback)}
+};
+
 static const ctl_named_node_t experimental_utilization_node[] = {
 	{NAME("query"),		CTL(experimental_utilization_query)},
 	{NAME("batch_query"),	CTL(experimental_utilization_batch_query)}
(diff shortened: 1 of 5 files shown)


src/peak_event.c
```diff
@@ -19,6 +21,12 @@ peak_event_update(tsd_t *tsd) {
 	uint64_t dalloc = tsd_thread_deallocated_get(tsd);
 	peak_t *peak = tsd_peakp_get(tsd);
 	peak_update(peak, alloc, dalloc);
+
+	activity_callback_thunk_t *thunk = tsd_activity_callback_thunkp_get(
+	    tsd);
+	if (thunk->callback != NULL) {
+		thunk->callback(thunk->uctx, alloc, dalloc);
+	}
```
interwq, line 29: Move these to "peak_alloc_event_handler"? It feels more natural to me there. Or do we explicitly call _event_update somewhere that also requires triggering callback?

davidtgoldblatt: Do you mean this whole function? The peak_update call and the thunk call sort of logically go together.

I think the structure here might have been copied from the other event handlers? No strong opinions, other than that either both should move or both should stay here.

interwq: Oh I meant just the new lines. IIRC, the event handler is only called when the event fires, while the event_update()'s job is only to update the threshold for the next event (in some rare cases like reentrancy, event_update may be called w/o triggering the event).

interwq: Wait I think I was confused.

interwq: Ah so, the `peak_event_update` is not part of the thread event machinery? I thought that it's part of the thread_event module functions like the event handler.

interwq: Okay so now the question is, do we want to trigger the callback with the mallctl here: https://github.com/jemalloc/jemalloc/blob/ea32060f9ca5e14077cda7fa2401a1f91f55ad82/src/ctl.c#L2084

Seems not?

davidtgoldblatt: Ahhh I think it's coming back to me. I think we do want to trigger it there. This is code-sharing between CTL and the thread_event triggered activity updates. Doing the update at CTL time makes it so that calls are always exact if we're at a peak during a thread.peak.read call, which is sometimes handy.

interwq: Yep makes sense.

interwq (approved):

src/ctl.c
```diff
@@ -337,7 +338,7 @@ static const ctl_named_node_t	thread_node[] = {
 	{NAME("tcache"),	CHILD(named, thread_tcache)},
 	{NAME("peak"),		CHILD(named, thread_peak)},
 	{NAME("prof"),		CHILD(named, thread_prof)},
-	{NAME("idle"),		CTL(thread_idle)}
+	{NAME("idle"),		CTL(thread_idle)},
```
interwq, line 341: Is this intentional?

src/peak_event.c
davidtgoldblatt: Oh wait I meant that as a reason why the activity callback ought to be moved (no reason to invoke it at the point of a mallctl call after all). I moved it in a race; want me to move it back?

src/ctl.c
davidtgoldblatt: Nope. Reverted.

src/peak_event.c
interwq: Nope I think we are on the same page: updating the counters on mallctl, but not invoking the callbacks.

interwq (approved):

PR #1973, opened 2020-11-06
Add NULL check when coalescing backwards.

semihalf-gorecki-dawid:
When testing ASLR implementation on FreeBSD we have noticed that every time we tried to build FreeBSD world with ASLR enabled on 32-bit target machines, the build failed with following message:
```
<jemalloc>: /usr/home/mindal/git/freebsd/contrib/jemalloc/include/jemalloc/internal/rtree.h:329: Failed assertion: "key != 0"
Abort trap (core dumped)
```
Looking at ktrace logs from multiple test runs, it seems that this assertion always fails when `mmap()` returns 0x1000. When trying to get previous extent, `extent_before_get()` subtracts page size from current address, which resolves to zero and this triggers the assertion in `rtree_leaf_elm_looup()`.

I looked at `dev` branch and the same fix should apply, however I had some problems when trying to import it into FreeBSD source tree and I wasn't able to test it further, so I'm leaving this as PR to master for now.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -1641,8 +1641,11 @@ extent_try_coalesce_impl(tsdn_t *tsdn, arena_t *arena,
 		}
 
 		/* Try to coalesce backward. */
-		extent_t *prev = extent_lock_from_addr(tsdn, rtree_ctx,
-		    extent_before_get(extent), inactive_only);
+		extent_t *prev = NULL;
+		if (extent_before_get(extent) != NULL) {
+			prev = extent_lock_from_addr(tsdn, rtree_ctx,
+			    extent_before_get(extent), inactive_only);
+		}
 		if (prev != NULL) {
 			bool can_coalesce = extent_can_coalesce(arena, extents,
 			    extent, prev);


davidtgoldblatt:
Good catch.

I'm not sure about the dev vs. master thing -- I think we ordinarily try to only do merge commits to master (and we may be on the verge of cutting a release anyways). @jasone, you do the FreeBSD integrations -- any opinions?

jasone:
Yeah, only releases should be merged with the master branch. A backport of the fix to any of the release branches is fine though.

Re: me maintaining jemalloc in FreeBSD, I don't plan to do so anymore. I haven't even booted a FreeBSD machine in nearly a year, and the most recent update, to 5.2.1, was a re-commit (by someone else) of my attempt that I had to back out after hitting temporary toolchain failures for secondary platforms. Hopefully, someone in the FreeBSD developer community will continue to periodically merge stable jemalloc releases into libc.

davidtgoldblatt:
@semihalf-gorecki-dawid, that given, what do you think about applying this straight to FreeBSD's copy, and in parallel we can apply it to the dev branch so that it makes it into the next release (which we intend to be soon, although it's hit some delays)?

semihalf-gorecki-dawid:
Yeah, I think that this is the best option. I already opened a diff on FreeBSD phabricator (https://reviews.freebsd.org/D27025) with the same patch. From what I've seen the only difference between dev and master branches in this part of the file is caused by renaming functions and data types, the logic is exactly the same.

davidtgoldblatt:
As discussed, closing in favor of https://github.com/jemalloc/jemalloc/pull/2003, which applies cleanly to dev HEAD -- this should catch the next release.

PR #1976, opened 2020-11-09
Update config.{guess,sub} to 2020-11-07@77632d9

concatime:
https://git.savannah.gnu.org/cgit/config.git/commit/?id=77632d92f25e26b95c64afd3700d51bd03587613


davidtgoldblatt:
Thanks! Looks like we've fallen a ways behind.

PR #1981, opened 2020-11-12
prof_log: delay calling atexit

BusyJay:
atexit call can allocate on some platforms, which can lead to deadlock.
This PR delays the function call after really starting logging. So it won't
be called during initialization unless it's required.

### DIFF
--- a/src/prof_log.c
+++ b/src/prof_log.c
@@ -27,6 +27,8 @@ enum prof_logging_state_e {
  */
 prof_logging_state_t prof_logging_state = prof_logging_state_stopped;
 
+bool prof_logging_final_hook_registerred = false;
+
 /* Used in unit tests. */
 static bool prof_log_dummy = false;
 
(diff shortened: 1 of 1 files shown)


davidtgoldblatt (commented):
Sorry for the long delay on this; it tickled some scheduling issues (because the end-goal is for the prof_log stuff to get entirely replaced with prof_recent) which made it easy to drop off the radar.

I think actually, probably something simpler is called for here; prof_log_init can instead take a bool parameter, `install_atexit` (or some more descriptive name). Then, in prof_init (here: https://github.com/jemalloc/jemalloc/blob/025d8c37c93a69ec0aa5d8a55e3793cb480a5ac8/src/prof.c#L615 ), we can pass in `opt_prof_final` as that parameter. This makes the prof atexit calls handled similarly throughout the module.

I'm not sure this is strictly correct as is, because in a world where atexit allocates, it's possible for that allocation to get sampled, triggering reentrancy, and I don't think prof_log was written with reentrancy in mind (although that might get handled with existing reentrancy-handling facilities; it's been a while).

interwq:
Resolved in #2137

PR #1986, opened 2020-11-23
utrace support with label based signature.

devnexen:


davidtgoldblatt:
Is this for OpenBSD (just from Googling, that's the only OS I can find that takes a utrace label). Could you confirm that running tests when configure with both:
```
./configure --enable-debug --enable-utrace
```
and
```
./configure --enable-utrace
```

Pass? We don't have CI set up for those cases, and I don't have an OpenBSD system readily available to test on.

devnexen:
Yes they pass for both.

davidtgoldblatt:
Thanks!

PR #1988, opened 2020-11-29
Fix gcc mem barriers for arm.

devnexen:
membar is a sparc instruction, using DMB here.


davidtgoldblatt:
I *think* this code is right as is -- note that we're checking `__arch64__` (single a) rather than `__aarch64__` (two a's); i.e. 64-bit SPARC (I don't totally remember why this was the decision; perhaps that's the only static clue we get that the barrier is definitely present?).

The ARM case doesn't match any of specific preprocessor defines, so it'll go down the default path and call `__sync_synchronize()`, which should be right.

devnexen:
yes you re right.

PR #1990, opened 2020-12-01
Allow fractional settings for narenas_ratio.

davidtgoldblatt:
There's no fundamental reason this ratio ought to be an integer. With the HPA allowing (and benefitting from) much lower arena counts, this has become a day-to-day pain. Setting the number of arenas to (say) half the number of CPUs is hard to do without knowing the environment you'll run on.

### DIFF
--- a/include/jemalloc/internal/fxp.h
+++ b/include/jemalloc/internal/fxp.h
@@ -1,0 +1,100 @@
+#ifndef JEMALLOC_INTERNAL_FXP_H
+#define JEMALLOC_INTERNAL_FXP_H
+
+/*
+ * A simple fixed-point math implementation, supporting only unsigned values
+ * (with overflow being an error).
+ *
+ * It's not in general safe to use floating point in core code, because various
+ * libc implementations we get linked against can assume that malloc won't touch
+ * floating point state and call it with an unusual calling convention.
+ */
+
+/*
+ * High 16 bits are the integer part, low 16 are the fractional part.  Or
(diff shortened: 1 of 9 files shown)


include/jemalloc/internal/fxp.h
```diff
@@ -0,0 +1,100 @@
...
+	uint32_t fractional_part = (a  & ((1U << 16) - 1));
+	uint32_t increment = (uint32_t)(fractional_part >= (1U << 15));
+	return (a >> 16) + increment;
+}
+
+/*
+ * Returns true on error.  Otherwise, returns false and updates *ptr to point to
+ * the first character not parsed (because it wasn't a digit).
+ */
+bool fxp_parse(fxp_t *a, const char *ptr, char **end);
```
interwq, line 97: Would it be better if we pass in a length (instead of returning the *end back), and return error if not fully consumed? With the current approach it looks we will always check two things on return.

interwq (commented):
Looks great. Just one question.

include/jemalloc/internal/fxp.h
davidtgoldblatt: The honest answer is that it's this way because it's sort of morally the same as strtoumax, so I just made the API the same. If we want to have a variant of malloc_conf_multi_sizes_next that allows for non-integer values (say, for tcache sizing), we'll need this style API.

No strong opinions, up to you.

interwq: Thanks for explaining. Keeping it this way makes sense.

interwq (approved):

PR #1991, opened 2020-12-01
HPA: More directly assign hugepage ownership to arenas

davidtgoldblatt:
Each of these commits is individually small, but put together they are a fairly big design shift in the HPA. We now explicitly track a whole hugepage's state in an hpa_shard-local context. This lets us make smarter decisions about when to hugeify or not, and also allows purging hugepages in a reasonable way.

For now, we only purge whole hugepages, and only when they become completely empty, but this structure will allow finer-grained / smarter strategies as we continue to evolve.

### DIFF
--- a/include/jemalloc/internal/bit_util.h
+++ b/include/jemalloc/internal/bit_util.h
@@ -179,12 +179,97 @@ fls_u(unsigned x) {
 }
 #endif
 
-#ifdef JEMALLOC_INTERNAL_POPCOUNTL
+#if LG_SIZEOF_LONG_LONG > 3
+#  error "Haven't implemented popcount for 16-byte ints."
+#endif
+
+#define DO_POPCOUNT(x, type) do {					\
+	/*								\
+	 * Algorithm from an old AMD optimization reference manual.	\
+	 * We're putting a little bit more work than you might expect	\
+	 * into the no-instrinsic case, since we only support the	\
+	 * GCC intrinsics spelling of popcount (for now).  Detecting	\
+	 * whether or not the popcount builtin is actually useable in	\
+	 * MSVC is nontrivial.						\
+	 */								\
+									\
+	type bmul = (type)0x0101010101010101ULL;			\
+									\
+	/*								\
+	 * Replace each 2 bits with the sideways sum of the original	\
+	 * values.  0x5 = 0b0101.					\
+	 *								\
+	 * You might expect this to be:					\
+	 *   x = (x & 0x55...) + ((x >> 1) & 0x55...).			\
+	 * That costs an extra mask relative to this, though.		\
+	 */								\
+	x = x - ((x >> 1) & (0x55U * bmul));				\
+	/* Replace each 4 bits with their sideays sum.  0x3 = 0b0011. */\
+	x = (x & (bmul * 0x33U)) + ((x >> 2) & (bmul * 0x33U));		\
+	/*								\
+	 * Replace each 8 bits with their sideways sum.  Note that we	\
+	 * can't overflow within each 4-bit sum here, so we can skip	\
+	 * the initial mask.						\
+	 */								\
+	x = (x + (x >> 4)) & (bmul * 0x0FU);				\
+	/*								\
+	 * None of the partial sums in this multiplication (viewed in	\
+	 * base-256) can overflow into the next digit.  So the least	\
+	 * significant byte of the product will be the least		\
+	 * significant byte of the original value, the second least	\
+	 * significant byte will be the sum of the two least		\
+	 * significant bytes of the original value, and so on.		\
+	 * Importantly, the high byte will be the byte-wise sum of all	\
+	 * the bytes of the original value.				\
+	 */								\
+	x = x * bmul;							\
+	x >>= ((sizeof(x) - 1) * 8);					\
+	return (unsigned)x;						\
+} while(0)
+
+static inline unsigned
+popcount_u_slow(unsigned bitmap) {
+	DO_POPCOUNT(bitmap, unsigned);
+}
+
+static inline unsigned
+popcount_lu_slow(unsigned long bitmap) {
+	DO_POPCOUNT(bitmap, unsigned long);
+}
+
+static inline unsigned
+popcount_llu_slow(unsigned long long bitmap) {
+	DO_POPCOUNT(bitmap, unsigned long long);
+}
+
+#undef DO_POPCOUNT
+
+static inline unsigned
+popcount_u(unsigned bitmap) {
+#ifdef JEMALLOC_INTERNAL_POPCOUNT
+	return JEMALLOC_INTERNAL_POPCOUNT(bitmap);
+#else
+	return popcount_u_slow(bitmap);
+#endif
+}
+
 static inline unsigned
 popcount_lu(unsigned long bitmap) {
-  return JEMALLOC_INTERNAL_POPCOUNTL(bitmap);
-}
+#ifdef JEMALLOC_INTERNAL_POPCOUNTL
+	return JEMALLOC_INTERNAL_POPCOUNTL(bitmap);
+#else
+	return popcount_lu_slow(bitmap);
 #endif
+}
+
+static inline unsigned
+popcount_llu(unsigned long long bitmap) {
+#ifdef JEMALLOC_INTERNAL_POPCOUNTLL
+	return JEMALLOC_INTERNAL_POPCOUNTLL(bitmap);
+#else
+	return popcount_llu_slow(bitmap);
+#endif
+}
 
 /*
  * Clears first unset bit in bitmap, and returns
--- a/include/jemalloc/internal/hpa.h
+++ b/include/jemalloc/internal/hpa.h
@@ -35,8 +35,7 @@ struct hpa_s {
 /* Used only by CTL; not actually stored here (i.e., all derived). */
 typedef struct hpa_shard_stats_s hpa_shard_stats_t;
 struct hpa_shard_stats_s {
-	psset_bin_stats_t psset_full_slab_stats;
-	psset_bin_stats_t psset_slab_stats[PSSET_NPSIZES];
+	psset_stats_t psset_stats;
 };
 
 typedef struct hpa_shard_s hpa_shard_t;
@@ -89,6 +88,9 @@ bool hpa_init(hpa_t *hpa, base_t *base, emap_t *emap,
 bool hpa_shard_init(hpa_shard_t *shard, hpa_t *hpa,
     edata_cache_t *edata_cache, unsigned ind, size_t ps_goal,
     size_t ps_alloc_max, size_t small_max, size_t large_min);
+
+void hpa_stats_accum(hpa_shard_stats_t *dst, hpa_shard_stats_t *src);
+void hpa_stats_merge(tsdn_t *tsdn, hpa_shard_t *shard, hpa_shard_stats_t *dst);
 /*
  * Notify the shard that we won't use it for allocations much longer.  Due to
  * the possibility of races, we don't actually prevent allocations; just flush
--- a/include/jemalloc/internal/hpa.h
+++ b/include/jemalloc/internal/hpa.h
@@ -21,6 +21,8 @@ struct hpa_shard_s {
 	pai_t pai;
 	malloc_mutex_t grow_mtx;
 	malloc_mutex_t mtx;
+	/* The base metadata allocator. */
+	base_t *base;
 	/*
 	 * This edata cache is the one we use when allocating a small extent
 	 * from a pageslab.  The pageslab itself comes from the centralized
@@ -45,7 +47,14 @@ struct hpa_shard_s {
 	 *
 	 * Guarded by grow_mtx.
 	 */
-	edata_list_inactive_t unused_slabs;
+	hpdata_list_t unused_slabs;
+
+	/*
+	 * How many grow operations have occurred.
+	 *
+	 * Guarded by grow_mtx.
+	 */
+	uint64_t age_counter;
 
 	/*
 	 * Either NULL (if empty), or some integer multiple of a
@@ -54,7 +63,8 @@ struct hpa_shard_s {
 	 *
 	 * Guarded by grow_mtx.
 	 */
-	edata_t *eden;
+	void *eden;
+	size_t eden_len;
 
 	/* The arena ind we're associated with. */
 	unsigned ind;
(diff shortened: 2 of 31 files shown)


src/psset.c
```diff
@@ -65,6 +65,47 @@ psset_assert_ps_consistent(edata_t *ps) {
...
 
+void
+psset_insert(psset_t *psset, edata_t *ps) {
+	psset_assert_ps_consistent(ps);
+	size_t longest_free_range = edata_longest_free_range_get(ps);
+
+	if (longest_free_range == 0) {
+		psset_bin_stats_adjust(&psset->full_slab_stats, ps,
+		    /* inc */ true);
+		return;
```
interwq, line 99: Let's add some comments regarding that the full slabs don't need to be tracked (maybe similar to in `struct psset_stats_s`).

src/hpa.c
```diff
@@ -89,6 +89,18 @@ hpa_shard_init(hpa_shard_t *shard, hpa_t *hpa, edata_cache_t *edata_cache,
 	return false;
 }
 
+void
+hpa_stats_accum(hpa_shard_stats_t *dst, hpa_shard_stats_t *src) {
+	psset_stats_accum(&dst->psset_stats, &src->psset_stats);
```
interwq, line 91: nit: explain why this needs no locking.

include/jemalloc/internal/psset.h
```diff
@@ -62,6 +65,8 @@ void psset_stats_accum(psset_stats_t *dst, psset_stats_t *src);
 void psset_insert(psset_t *psset, edata_t *ps);
 void psset_remove(psset_t *psset, edata_t *ps);
 
+void psset_hugeify(psset_t *psset, edata_t *ps);
```
interwq: `hugify` might be the better term?

include/jemalloc/internal/hpa.h
```diff
@@ -53,44 +27,46 @@ struct hpa_shard_s {
 	 * allocator, and so will use its edata_cache.
 	 */
 	edata_cache_small_t ecs;
-	hpa_t *hpa;
```
interwq, line 57: Is it true that, this diff removed the central huge page allocator completely? If so what's the high level motivation behind it? My impression was that the hpa level contention didn't hurt much, and it potentially can enable flexible sharing? I also had hope to deal with "global" purging at the hpa level.

src/psset.c
davidtgoldblatt: Will do.

src/hpa.c
davidtgoldblatt: Will do.

include/jemalloc/internal/psset.h
davidtgoldblatt: Something about that feels very wrong to me, but some googling indicates that it's by far a more popular term.

include/jemalloc/internal/hpa.h
davidtgoldblatt: Right; hpa_central could be deleted after this stack. A couple things pushed in this direction:
- hpa_central was actually the source of quite a bit of contention; it was by far the hottest of our locks on test binaries.
- Splitting a single hugepage across multiple suballocators means that there's no global view of it for the purposes of making purging / hugification decisions, which turns out to be important both for fragmentation avoidance and purging choices.

Practically this seems to be well-borne out; this design is performing far better than the previous one in canaries.

I think eventually we'll want the central allocator back, but only once we start trying to support large allocations (ones too big to serve out of the hpa_shard_t, or in case of hpa_shard_t exhaustion) and as the source of individual hpdata_ts for the hpa_shard_t.

Perhaps more intuitively, if thinking of it in terms of the earlier design: if we make the slab goal a whole hugepage, we can implement more efficient purging / hugification policies for free. But once we've done that (and, given that we don't actually try to serve large allocations yet), the hpa_central_t doesn't really add much (but still costs some complexity and lock contention).

davidtgoldblatt:
Newest pushed version has all the promised changes.

BTW, the purging changes stacked on top of this are [here](https://github.com/davidtgoldblatt/jemalloc/tree/huge_lazy-dalloc), if you want to see what things are building towards. I figured I'd batch up PRs, though, to avoid dumping a big long stack all at once.

include/jemalloc/internal/hpa.h
interwq: Sounds good. I think we are on the same page here -- 

> I think eventually we'll want the central allocator back, but only once we start trying to support large allocations

include/jemalloc/internal/hpdata.h
```diff
@@ -0,0 +1,123 @@
...
+#include "jemalloc/internal/flat_bitmap.h"
+#include "jemalloc/internal/ph.h"
+#include "jemalloc/internal/ql.h"
+#include "jemalloc/internal/typed_list.h"
+
+/*
+ * The metadata representation we use for extents in hugepages.  While the PAC
+ * uses the edata_t to represent both active and inactive extents, the HP only
+ * uses the edata_t for active ones; instead, inactive extents are tracked by
+ * the hugepage they're in.
```
interwq, line 13: ` tracked by the hugepage they're in` is a bit unclear to me. That means they don't need metadata storage, other than their vaddr?

src/hpa.c
```diff
@@ -96,176 +89,319 @@ hpa_shard_init(hpa_shard_t *shard, hpa_t *hpa, edata_cache_t *edata_cache,
...
+	 * NULL, we have to allocate it too.  Otherwise, we just have to
+	 * allocate an edata_t for the new psset.
 	 */
-	bool commit = true;
-	void *addr = pages_map(NULL, alloc_size, HUGEPAGE, &commit);
-	if (addr == NULL) {
-		malloc_mutex_unlock(tsdn, &hpa->grow_mtx);
-		return NULL;
+	if (shard->eden == NULL) {
+		/*
```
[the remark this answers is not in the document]
interwq: Ignore -- just saw that it got simplified in later commits.

src/hpa.c
```diff
@@ -247,12 +247,20 @@ hpa_try_alloc_no_grow(tsdn_t *tsdn, hpa_shard_t *shard, size_t size, bool *oom)
 	}
 	assert(edata_arena_ind_get(edata) == shard->ind);
 
-	err = psset_alloc_reuse(&shard->psset, edata, size);
-	if (err) {
+	hpdata_t *ps = psset_fit(&shard->psset, size);
```
interwq, line 256: Shall we call out _oldest_fit?

include/jemalloc/internal/hpa.h
```diff
@@ -67,6 +69,12 @@ struct hpa_shard_s {
 	/* The arena ind we're associated with. */
 	unsigned ind;
 	emap_t *emap;
+
+	/*
+	 * The number of times we've purged a hugepage.  Guarded by the grow
```
interwq, line 74: nit: mention that each eviction purges a single hugepage.

include/jemalloc/internal/bit_util.h
```diff
@@ -179,11 +179,105 @@ fls_u(unsigned x) {
...
+}
+
+static inline unsigned
+popcount_llu_slow(unsigned long long bitmap) {
+	DO_POPCOUNT(bitmap, unsigned long long);
+}
+
+#undef DO_POPCOUNT
+
+#ifdef JEMALLOC_INTERNAL_POPCOUNT
```
interwq, line 247: nit: have the `#ifdef` covers only the return statement

interwq:
I just realized that our arm CI is gone; Scaleway retired their arm servers.

include/jemalloc/internal/hpa.h
```diff
@@ -53,44 +27,46 @@ struct hpa_shard_s {
...
+	 * fall back to the more general (for now) PAC, which can always handle
+	 * any allocation request.
 	 */
-	size_t ps_alloc_max;
+	size_t alloc_max;
+
 	/*
-	 * What's the maximum size we'll try to allocate out of the shard at
-	 * all?
+	 * Slabs currently purged away.  They are non-hugepages.
```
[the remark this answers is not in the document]
davidtgoldblatt: Yes; clarified the comment a little.

src/hpa.c
```diff
@@ -15,43 +17,37 @@ static bool hpa_shrink(tsdn_t *tsdn, pai_t *self, edata_t *edata,
 static void hpa_dalloc(tsdn_t *tsdn, pai_t *self, edata_t *edata);
 
 bool
-hpa_init(hpa_t *hpa, base_t *base, emap_t *emap, edata_cache_t *edata_cache) {
-	bool err;
-
+hpa_supported() {
```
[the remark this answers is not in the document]
davidtgoldblatt: Added a check. I think eventually we'll have to think a little harder about what precisely we want in what circumstances. For now I'm fine with saying "HPA is supported if your environment looks David's test machine" as we're quickly iterating on the design and implementation.

src/hpa.c
```diff
@@ -96,176 +89,319 @@ hpa_shard_init(hpa_shard_t *shard, hpa_t *hpa, edata_cache_t *edata_cache,
...
+		 * with overcommit.  Eventually, we should be more careful here.
+		 */
+		bool commit = true;
+		/* Allocate address space, bailing if we fail. */
+		void *new_eden = pages_map(NULL, HPA_EDEN_SIZE, HUGEPAGE,
+		    &commit);
+		if (new_eden == NULL) {
+			return NULL;
+		}
+		malloc_mutex_lock(tsdn, &shard->mtx);
```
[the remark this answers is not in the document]
davidtgoldblatt: Added a comment. We drop the lock so that other allocations or frees can succeed while growing (though, the allocations must be smaller than the size of this alloc, except in the case of races in which an intervening dalloc frees up a longer free range). By analogy with the PAC: you don't hold the dirty ecache lock during an entire grow_retained call.

src/hpa.c
```diff
@@ -96,176 +89,319 @@ hpa_shard_init(hpa_shard_t *shard, hpa_t *hpa, edata_cache_t *edata_cache,
...
+	edata_hugeified_set(ps, false);
+	edata_addr_set(shard->eden, edata_past_get(ps));
+	edata_size_set(shard->eden,
+	    edata_size_get(shard->eden) - HUGEPAGE);
+
+	return ps;
+}
 
+static void
+hpa_handle_ps_eviction(tsdn_t *tsdn, hpa_shard_t *shard, edata_t *ps) {
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

src/hpa.c
```diff
@@ -96,176 +89,319 @@ hpa_shard_init(hpa_shard_t *shard, hpa_t *hpa, edata_cache_t *edata_cache,
...
+		 * Do the metadata modification while holding the lock; we'll
+		 * actually change state with the lock dropped.
+		 */
+		psset_hugify(&shard->psset, ps);
+	}
 	malloc_mutex_unlock(tsdn, &shard->mtx);
-	if (!err) {
+	if (hugify) {
+		/*
+		 * Hugeifying with the lock dropped is safe, even with
```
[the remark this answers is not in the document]
davidtgoldblatt: Expanded the comment. Dropping the lock doesn't just allow other operations on the given pageslab; while we hold the lock, no other thread can do any other operation on the whole shard (even on unrelated hugepages).

src/hpa.c
```diff
@@ -427,17 +541,15 @@ hpa_shard_destroy(tsdn_t *tsdn, hpa_shard_t *shard) {
...
+		hpa_assert_empty(tsdn, shard, &shard->psset);
 		malloc_mutex_unlock(tsdn, &shard->mtx);
-		assert(psset_empty);
-		hpa_shard_assert_stats_empty(&shard->psset.stats.full_slabs);
-		for (pszind_t i = 0; i < PSSET_NPSIZES; i++) {
-			hpa_shard_assert_stats_empty(
-			    &shard->psset.stats.nonfull_slabs[i]);
-		}
+	}
+	edata_t *ps = NULL;
```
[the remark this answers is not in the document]
davidtgoldblatt: Fixed.

src/jemalloc.c
```diff
@@ -1797,31 +1797,19 @@ malloc_init_hard_a0_locked() {
 	}
 	a0 = arena_get(TSDN_NULL, 0, false);
 
-	if (opt_hpa && LG_SIZEOF_PTR == 2) {
+	if (opt_hpa && !hpa_supported()) {
+		malloc_printf("<jemalloc>: HPA not currently supported "
+		    "in the current configuration; %s.",
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

src/psset.c
```diff
@@ -45,29 +49,62 @@ psset_stats_accum(psset_stats_t *dst, psset_stats_t *src) {
...
+	    ? &binstats->ninactive_huge : &binstats->ninactive_nonhuge;
 
 	size_t npages = edata_size_get(ps) >> LG_PAGE;
 	size_t ninactive = edata_nfree_get(ps);
 	size_t nactive = npages - ninactive;
-	binstats->npageslabs += mul * 1;
-	binstats->nactive += mul * nactive;
-	binstats->ninactive += mul * ninactive;
+
+	*npageslabs_dst += mul * 1;
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

include/jemalloc/internal/hpdata.h
davidtgoldblatt: Right; active vs. inactive state is tracked in the bitmap in the hpdata (and, in the purging stack commits, so is dirty state). I beefed up the comment some.

src/hpa.c
davidtgoldblatt: My inclination is no:
- This matches the eset naming
- The fragmentation avoidance policy is an internal implementation choice of the psset (and in fact, it isn't quite oldest fit -- it's the oldest fit of the best fits. And even there, I think we'll start doing some degree of hugepage-preference).
- Eventually, we'll have a psset function for picking hpdatas to hugify or purge, and we'll want some consistent naming scheme for the various ways to pick things we have.

include/jemalloc/internal/hpa.h
davidtgoldblatt: Done. (Actually, this was also done in the purging stack, so it's doubly done now since I liked that phrasing better).

include/jemalloc/internal/bit_util.h
davidtgoldblatt: Done.

davidtgoldblatt:
Updates pushed.

I filed https://github.com/jemalloc/jemalloc/issues/1993 for the ARM issues.

src/hpa.c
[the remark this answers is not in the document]
interwq: Makes sense. I misunderstood and forgot it was for the whole shard.

interwq (approved):
LGTM!

Issue #1334, opened 2018-10-02
Coverity scan results for jemalloc-5.1.0


vashirov:
We use jemalloc in our project and run Coverity on a regular basis.
It reported some issues in jemalloc code, I'm not sure if they are false positives or not, so please take a look.

jemalloc-5.1.0/src/arena.c:

```
288 static void
289 arena_large_malloc_stats_update(tsdn_t *tsdn, arena_t *arena, size_t usize) {
290         szind_t index, hindex;
291 
>>     	1. Condition 0L /* !!!config_stats */, taking false branch.
292         cassert(config_stats);
293 
>>     	2. Condition usize < (16384UL /* (size_t)1 << 14 */), taking false branch.
294         if (usize < LARGE_MINCLASS) {
295                 usize = LARGE_MINCLASS;
296         }
>>     	3. return_constant: Function call sz_size2index(usize) returns 232.
>>     	4. assignment: Assigning: index = sz_size2index(usize). The value of index is now 232.
297         index = sz_size2index(usize);
>>     	5. Condition index >= 36, taking true branch.
>>     	6. assignment: Assigning: hindex = (index >= 36U) ? index - 36U : 0U. The value of hindex is now 196.
298         hindex = (index >= NBINS) ? index - NBINS : 0;
299 
>>    	
>> CID 17477 (#1 of 1): Out-of-bounds read (OVERRUN)
>> 7. overrun-local: Overrunning array arena->stats.lstats of 196 32-byte elements at element index 196 (byte offset 6272) using index hindex (which evaluates to 196).
300         arena_stats_add_u64(tsdn, &arena->stats,
301             &arena->stats.lstats[hindex].nmalloc, 1);
302 }
```

```
304 static void
305 arena_large_dalloc_stats_update(tsdn_t *tsdn, arena_t *arena, size_t usize) {
306         szind_t index, hindex;
307 
>>    	1. Condition 0L /* !!!config_stats */, taking false branch.
308         cassert(config_stats);
309 
>>    	2. Condition usize < (16384UL /* (size_t)1 << 14 */), taking false branch.
310         if (usize < LARGE_MINCLASS) {
311                 usize = LARGE_MINCLASS;
312         }
>>    	3. return_constant: Function call sz_size2index(usize) returns 232.
>>    	4. assignment: Assigning: index = sz_size2index(usize). The value of index is now 232.
313         index = sz_size2index(usize);
>>    	5. Condition index >= 36, taking true branch.
[...]

PR #1994, opened 2020-12-07
Move n_search outside of assert as reported by static analyzer

aditya7fb:
Reported in:
https://github.com/jemalloc/jemalloc/issues/1334

>> CID 17475 (#1 of 1): Side effect in assertion (ASSERT_SIDE_EFFECT)
>> assert_side_effect: Argument n_search++ of assert() has a side effect. The containing function might work differently in a non-debug build.
191                assert(n_search++ < lg_floor(SMOOTHSTEP_NSTEPS) + 1);

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -185,7 +185,8 @@ arena_decay_compute_purge_interval_impl(tsdn_t *tsdn, decay_t *decay,
 			lb = target;
 			npurge_lb = npurge;
 		}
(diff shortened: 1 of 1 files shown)


davidtgoldblatt (approved):
Looks reasonable. (As you noted on the other issue, this couldn't *really* trigger, but it's also pointlessly suspicious-looking).

PR #1999, opened 2020-12-17
Binary space optimizations

davidtgoldblatt:
This change implements a few binary size reductions:
- Even when stats and profiling are disabled, some of those functions still get included. Adding a cassert tells the compiler they're unreachable, and lets them be optimized into no-ops.
- Marking cold functions as cold treats them as such, including some space-saving optimizations. For now, the definition of cold is fairly conservative (basically, things that are manually-triggered, or can only happen once; things like prof_lookup aren't counted as cold, even though they arguably are).
- Allowing code reuse between realloc and rallocx (by removing the force inline) saves nontrivial space overhead. This does cost slightly more CPU (since we have to parse a constant 0 flag), but that's quite cheap in practice, especially given that the realloc pathways are cold generally.

Binary size before this change, under common configurations:
(no stats, no profiling) 456k
(stats but no profiling) 624k
(stats and profiling): 720k

And size after:
(no stats, no profiling) 372k (18.4% savings)
(stats but no profiling) 512k (17.9% savings)
(stats and profiling) 620k (13.9% savings)

The first of these is the one I care about the most, since it's the configuration being shipped in the space-sensitive applications we know about.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -909,6 +909,18 @@ if test "x${je_cv_fallthrough}" = "xyes" ; then
   JE_CXXFLAGS_ADD([-Wimplicit-fallthrough])
 fi
 
+dnl Check for cold attribute support.
+JE_CFLAGS_SAVE()
+JE_CFLAGS_ADD([-Werror])
+JE_CFLAGS_ADD([-herror_on_warning])
+JE_COMPILABLE([cold attribute], [],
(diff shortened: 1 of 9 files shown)


aditya7fb (approved):

interwq (approved):
Sounds great!

Issue #1844, opened 2020-05-12
Why would je_calloc not return zero-initialized memory?


goffioul:
I'm working on a ARM-to-x86 translation layer for android-x86 [1], and I'm facing a really odd behavior whereby je_calloc in the emulated layer returns memory that is not zero-initialized.

The code is based on AOSP/Android-Q, which uses jemalloc from here [2]. The emulation layer is based on QEMU, which operates in user mode, with a minimal ARM-compiled runtime environment and a set of interface libraries that bridge the emulated environment and the native Android OS.

Maybe of importance is that there are 2 versions of jemalloc operating at the same time:
- the regular Android native jemalloc (x86)
- the jemalloc version in the emulated runtime environment (arm)

I've been tracking a series of crashes that all point to the same root cause: at some point during the execution, the emulated code performs a `calloc` call, but the returned memory is not fully zero-initialized. I've patched libc.so (in the emulated environment) with the following snippet to verify it.

```
--- a/libc/bionic/malloc_common.cpp
+++ b/libc/bionic/malloc_common.cpp
@@ -67,10 +67,19 @@ extern "C" void* calloc(size_t n_elements, size_t elem_size) {
   if (__predict_false(result == nullptr)) {
     warning_log("calloc(%zu, %zu) failed: returning null pointer", n_elements, elem_size);
   }
+  for (size_t i = 0; i < (n_elements * elem_size); i++) {
+    if (((char*)result)[i] != 0) {
+      warning_log("calloc(%zu, %zu): %p byte %zu is not zero!!! [calloc=%p]", n_elements, elem_size, result, i, Malloc(calloc));
+      break;
+    }
+  }
   return result;
 }
```

I would appreciate if you could help me understand in what circumstances such behavior could ever happen?

[1] https://github.com/goffioul/ax86-nb-qemu
[2] https://android.googlesource.com/platform/external/jemalloc_new/+/refs/tags/android-10.0.0_r2

davidtgoldblatt:
It's hard to say without a little bit more info. Could you print the sizes of the allocations that are failing? Small and large allocations get zeroed at different places.

My best guess right now is that something about the unusual environment might be making some of the configuration go a little screwy. We assume that certain system calls (e.g. madvise with the right flags) will zero memory. One way you could check this would be to instrument ehooks_zero (in ehooks.h) to do the zero-checking as well.

goffioul:
This is the `calloc` call that is failing (print using the patch snippet posted previously):
[...]

PR #2005, opened 2020-12-18
[RFC] Add runtime detection for MADV_DONTNEED zeroes pages (mostly for qemu)

azat:
qemu does not support this, yet [1], and you can get very tricky assert
if you will run program with jemalloc in use under qemu:

    <jemalloc>: ../contrib/jemalloc/src/extent.c:1195: Failed assertion: "p[i] == 0"

  [1]: https://patchwork.kernel.org/patch/10576637/

Here is a simple example that shows the problem [2]:

    // Gist to check possible issues with MADV_DONTNEED
    // For example it does not supported by qemu user
    // There is a patch for this [1], but it hasn't been applied.
    //   [1]: https://lists.gnu.org/archive/html/qemu-devel/2018-08/msg05422.html

    #include <sys/mman.h>
    #include <stdio.h>
    #include <stddef.h>
    #include <assert.h>
    #include <string.h>

    int main(int argc, char **argv)
    {
        void *addr = mmap(NULL, 1<<16, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
        if (addr == MAP_FAILED) {
            perror("mmap");
            return 1;
        }
        memset(addr, 'A', 1<<16);

        if (!madvise(addr, 1<<16, MADV_DONTNEED)) {
            puts("MADV_DONTNEED does not return error. Check memory.");
            for (int i = 0; i < 1<<16; ++i) {
                assert(((unsigned char *)addr)[i] == 0);
            }
        } else {
            perror("madvise");
        }

        if (munmap(addr, 1<<16)) {
            perror("munmap");
            return 1;
        }

        return 0;
    }

  ### unpatched qemu

      $ qemu-x86_64-static /tmp/test-MADV_DONTNEED
      MADV_DONTNEED does not return error. Check memory.
      test-MADV_DONTNEED: /tmp/test-MADV_DONTNEED.c:19: main: Assertion `((unsigned char *)addr)[i] == 0' failed.
      qemu: uncaught target signal 6 (Aborted) - core dumped
      Aborted (core dumped)

  ### patched qemu (by returning ENOSYS error)

      $ qemu-x86_64 /tmp/test-MADV_DONTNEED
      madvise: Success

  ### patch for qemu to return ENOSYS

      diff --git a/linux-user/syscall.c b/linux-user/syscall.c
      index 897d20c076..5540792e0e 100644
      --- a/linux-user/syscall.c
      +++ b/linux-user/syscall.c
      @@ -11775,7 +11775,7 @@ static abi_long do_syscall1(void *cpu_env, int num, abi_long arg1,
                  turns private file-backed mappings into anonymous mappings.
                  This will break MADV_DONTNEED.
                  This is a hint, so ignoring and returning success is ok.  */
      -        return 0;
      +        return ENOSYS;
       #endif
       #ifdef TARGET_NR_fcntl64
           case TARGET_NR_fcntl64:

  [2]: https://gist.github.com/azat/12ba2c825b710653ece34dba7f926ece

Refs: https://github.com/jemalloc/jemalloc/issues/1844#issuecomment-627496992

*This change is needed for [ClickHouse](https://github.com/ClickHouse/ClickHouse/) to enable tests of [AArch64 builds under QEMU](https://github.com/ClickHouse/ClickHouse/issues/15174)*

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -94,6 +94,13 @@ bool	opt_junk_free =
     false
 #endif
     ;
+bool	opt_trust_madvise =
+#ifdef JEMALLOC_PURGE_MADVISE_DONTNEED_ZEROS
+    false
+#else
+    true
+#endif
+    ;
 
 zero_realloc_action_t opt_zero_realloc_action =
     zero_realloc_action_strict;
@@ -1256,6 +1263,7 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 
 			CONF_HANDLE_BOOL(opt_abort, "abort")
 			CONF_HANDLE_BOOL(opt_abort_conf, "abort_conf")
+			CONF_HANDLE_BOOL(opt_trust_madvise, "trust_madvise")
 			if (strncmp("metadata_thp", k, klen) == 0) {
 				int i;
 				bool match = false;
--- a/src/pages.c
+++ b/src/pages.c
@@ -42,6 +42,57 @@ thp_mode_t init_system_thp_mode;
 /* Runtime support for lazy purge. Irrelevant when !pages_can_purge_lazy. */
 static bool pages_can_purge_lazy_runtime = true;
 
+#ifdef JEMALLOC_PURGE_MADVISE_DONTNEED_ZEROS
+static int madvise_dont_need_zeros_is_faulty = -1;
+/**
+ * Check that MADV_DONTNEED will actually zero pages on subsequent access.
+ *
+ * Since qemu does not support this, yet [1], and you can get very tricky
+ * assert if you will run program with jemalloc in use under qemu:
+ *
+ *     <jemalloc>: ../contrib/jemalloc/src/extent.c:1195: Failed assertion: "p[i] == 0"
+ *
+ *   [1]: https://patchwork.kernel.org/patch/10576637/
+ */
+static int madvise_MADV_DONTNEED_zeroes_pages()
+{
+	int works = -1;
+	size_t size = PAGE;
+
+	void * addr = mmap(NULL, size, PROT_READ|PROT_WRITE,
+	    MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
+
+	if (addr == MAP_FAILED) {
+		malloc_write("<jemalloc>: Cannot allocate memory for "
+		    "MADV_DONTNEED check\n");
+		if (opt_abort) {
+			abort();
+		}
+	}
+
+	memset(addr, 'A', size);
+	if (madvise(addr, size, MADV_DONTNEED) == 0) {
+		works = memchr(addr, 'A', size) == NULL;
+	} else {
+		/*
+		 * If madvise() does not support MADV_DONTNEED, then we can
+		 * call it anyway, and use it's return code.
+		 */
+		works = 1;
+	}
+
+	if (munmap(addr, size) != 0) {
+		malloc_write("<jemalloc>: Cannot deallocate memory for "
+		    "MADV_DONTNEED check\n");
+		if (opt_abort) {
+			abort();
+		}
+	}
+
+	return works;
+}
+#endif
+
 /******************************************************************************/
 /*
  * Function prototypes for static functions that are referenced prior to
@@ -351,10 +402,12 @@ pages_purge_forced(void *addr, size_t size) {
 
 #if defined(JEMALLOC_PURGE_MADVISE_DONTNEED) && \
     defined(JEMALLOC_PURGE_MADVISE_DONTNEED_ZEROS)
-	return (madvise(addr, size, MADV_DONTNEED) != 0);
+	return (unlikely(madvise_dont_need_zeros_is_faulty) ||
+	    madvise(addr, size, MADV_DONTNEED) != 0);
 #elif defined(JEMALLOC_PURGE_POSIX_MADVISE_DONTNEED) && \
     defined(JEMALLOC_PURGE_POSIX_MADVISE_DONTNEED_ZEROS)
-	return (posix_madvise(addr, size, POSIX_MADV_DONTNEED) != 0);
+	return (unlikely(madvise_dont_need_zeros_is_faulty) ||
+	    posix_madvise(addr, size, POSIX_MADV_DONTNEED) != 0);
 #elif defined(JEMALLOC_MAPS_COALESCE)
 	/* Try to overlay a new demand-zeroed mapping. */
 	return pages_commit(addr, size);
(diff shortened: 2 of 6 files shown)


alexey-milovidov:
This change is needed for [ClickHouse](https://github.com/ClickHouse/ClickHouse/) to enable tests of [AArch64 builds under QEMU](https://github.com/ClickHouse/ClickHouse/issues/15174).

AArch64 becomes more popular as a server platform and we see increasing demand of running ClickHouse on AArch64 servers. But there are difficulties in integrating it in CI environment that only has x86_64. That's why we are trying to run it under QEMU.

davidtgoldblatt:
Everyone's on vacation now, so it'll probably be early next year before anyone looks at this closely. That being said, my first thought is that I don't love adding an extra syscall to every malloc-using process; this is something that the FreeBSD people have tried to get down quite a bit recently (out of concern for minimizing fixed-cost overhead of process creation). I wonder if we can get by with something in configuration. Something like one off:
- Building with `./configure --dont_trust_madvise`
- Trying out a purge during configuration time, and only turning on the runtime check if that fails

azat:
@davidtgoldblatt thank you for the feedback.

>That being said, my first thought is that I don't love adding an extra syscall to every malloc-using process

I'm kind of understand, but this quirk is pretty simple and has almost zero overhead, and to me it looks like better to have the check instead of silently works incorrectly.

>Building with ./configure --dont_trust_madvise

But this will require building separate binaries to run in qemu

>Trying out a purge during configuration time, and only turning on the runtime check if that fails

Same here

azat:
@davidtgoldblatt holidays are passed, maybe someone can take a look is suitable for upstream or not? (Feel free to close if not)

davidtgoldblatt:
Sorry, we celebrate the new year by immediately doing our performance reviews for the previous 6 months, which ends up sucking up lots of time.

I think I'm sold on this generally, so long as it's guarded by a config option (which can at least be off by default outside of Linux). I've got a couple other nits (e.g. using atomics for the testing, some constant choices, etc.). Up to you if we iterate on this, or I can just put up a parallel PR if you'd rather not deal with our configure scripts.

azat:
Sure no problem, I can finish this to make it ready for upstream:
- config option - ok
- as for atomic, is it a problem? the flag is set only in pages_boot, like other flags and they don't atomic
- you can post what constant naming you will prefer, or if you prefer to finish this by yourself, I'm completely fine with this.

src/pages.c
```diff
@@ -42,6 +42,53 @@ thp_mode_t init_system_thp_mode;
 /* Runtime support for lazy purge. Irrelevant when !pages_can_purge_lazy. */
 static bool pages_can_purge_lazy_runtime = true;
 
+#ifdef JEMALLOC_PURGE_MADVISE_DONTNEED_ZEROS
+static int madvise_MADV_DONTNEED_zeros_pages = -1;
```
davidtgoldblatt, line 46: Rather than having an ifdef and a runtime flag with the same name but different semantics, it would be better to have (e.g.) `static bool madvise_dont_need_is_faulty` or something.

src/pages.c
```diff
@@ -42,6 +42,53 @@ thp_mode_t init_system_thp_mode;
...
+ * assert if you will run program with jemalloc in use under qemu:
+ *
+ *     <jemalloc>: ../contrib/jemalloc/src/extent.c:1195: Failed assertion: "p[i] == 0"
+ *
+ *   [1]: https://patchwork.kernel.org/patch/10576637/
+ */
+static int madvise_MADV_DONTNEED_zeroes_pages()
+{
+	int works = -1;
+	size_t size = 1 << 16;
```
davidtgoldblatt, line 60: Should be `PAGE` rather than `1 << 16`.

src/pages.c
```diff
@@ -42,6 +42,53 @@ thp_mode_t init_system_thp_mode;
...
+
+	void * addr1 = mmap(NULL, size, PROT_READ|PROT_WRITE,
+	    MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
+	void * addr2 = mmap(NULL, size, PROT_READ|PROT_WRITE,
+	    MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
+
+	if (addr1 == MAP_FAILED || addr2 == MAP_FAILED) {
```
davidtgoldblatt, line 67: This should also be guarded by `opt_abort`.

src/pages.c
```diff
@@ -42,6 +42,53 @@ thp_mode_t init_system_thp_mode;
...
+	memset(addr1, 'A', size);
+	if (madvise(addr1, size, MADV_DONTNEED) == 0) {
+		works = memcmp(addr1, addr2, size) == 0;
+	} else {
+		/* If madvise() does not support MADV_DONTNEED, then we can call it
+		 * anyway, and use it's return code. */
+		works = 1;
+	}
+
+	if (munmap(addr1, size) != 0 || munmap(addr2, size) != 0) {
```
davidtgoldblatt, line 82: Same as above.

src/pages.c
```diff
@@ -42,6 +42,53 @@ thp_mode_t init_system_thp_mode;
...
+ *     <jemalloc>: ../contrib/jemalloc/src/extent.c:1195: Failed assertion: "p[i] == 0"
+ *
+ *   [1]: https://patchwork.kernel.org/patch/10576637/
+ */
+static int madvise_MADV_DONTNEED_zeroes_pages()
+{
+	int works = -1;
+	size_t size = 1 << 16;
+
+	void * addr1 = mmap(NULL, size, PROT_READ|PROT_WRITE,
```
davidtgoldblatt, line 62: I think we just need one mapping, right?

davidtgoldblatt (commented):
One other thing I forgot to mention -- the configure option should be called out in the installation documentation.

davidtgoldblatt:
Oh, forgot to mention -- good point about the atomic, didn't realize that this wasn't an initialize-on-first-use thing.

src/pages.c
azat: That was for simplicity, to compare just two regions, but yes the second can be removed

src/pages.c
azat: BTW this is not new macros, it is used in code to decide "does madvise zero page?", so I guess it is fine to keep it

davidtgoldblatt: Yeah, the macro is fine as is I think (I guess arguably it should be renamed to JEMALLOC_PURGE_MADVISE_DONTNEED_SHOULD_ZERO or something); it's the pseudo-shadowing I think is confusing. Really, the macro describes the documented behavior of the system, and the bool describes whether or not the actual behavior conforms.

azat: > and the bool describes whether or not the actual behavior conforms.

Yep, I was talking just about macro, option added - `opt_dont_trust_madvise`

davidtgoldblatt (commented):
Looks good -- one substantive naming change and a few style nits (full rules here: https://github.com/jemalloc/jemalloc/wiki/Coding-Style).

src/jemalloc.c
```diff
@@ -1256,6 +1263,7 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 
 			CONF_HANDLE_BOOL(opt_abort, "abort")
 			CONF_HANDLE_BOOL(opt_abort_conf, "abort_conf")
+			CONF_HANDLE_BOOL(opt_dont_trust_madvise, "dont_trust_madvise")
```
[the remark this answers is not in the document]
azat: Added (with a smoke test in `test/unit/mallctl.c`)

davidtgoldblatt (approved):
Thanks!

PR #2012, opened 2021-01-10
Update config.{sub,guess} to support support-aarch64-apple-darwin as a target

### LINKED ISSUE #2010: Re-release 5.2.1 with updated config.guess/config.sub
The current `config.guess/config.sub` files don't support (cross-)compiling to the Apple M1 process. The build fails with 

```
Invalid configuration `arm64-apple-darwin20.0.0': machine `arm64-apple' not recognized
configure: error: /bin/sh build-aux/config.sub arm64-apple-darwin20.0.0 failed
```

This is easily fixed by updating the `config.guess/config.sub` files. Manually doing this and running the tests shows that `jemalloc` runs fine on Apple Silicon

But in Apache Arrow we built `jemalloc` during Arrow C++'s build from the latest release tarball. We would neither like to have a `autoconf` dependency nor can we ship those files with the Arrow source tarball due to licensing restrictions inside the Apache Software Foundation. Working with a jemalloc release tarball that contains them is though ok for us/the ASF. Happy to do volunteer some work into getting this minor release, not sure if I can add valuable things here for such a minor release.

cc @pitrou @nealrichardson @kou

xhochy:
Fixes #2010 

This is intentionally against master to allow for a 5.2.2 release to simply add support for building there without patching. I can do the same PR against `dev` to also update the files there to the same revision.

### DIFF
--- a/build-aux/config.guess
+++ b/build-aux/config.guess
@@ -1,8 +1,8 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
-#   Copyright 1992-2020 Free Software Foundation, Inc.
+#   Copyright 1992-2021 Free Software Foundation, Inc.
 
-timestamp='2020-11-07'
+timestamp='2021-01-01'
 
 # This file is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


xhochy:
Build failures look like CI configuration issues and are unrelated to this PR.

davidtgoldblatt:
This should be against `dev` directly -- (broadly, everything new should go into dev, and changes to `master` should happen only at the time of actual release). I think we'll probably try to roll out a 5.2.2 regardless, just to wrap together a few similar sorts of fixes we've accumulated since 5.2.1.

xhochy:
Rebased and changed PR base to `dev`.

davidtgoldblatt:
Thanks!

PR #2016, opened 2021-01-25
Fix duration in prof log

yinan1048576:


davidtgoldblatt (approved):
This makes sense on its own. Does it explain the huge duration prod bug we saw, though? The high-res an monotonic timers should at least be consistent with one another on modern Linuxes, right?

yinan1048576:
Yes, I thought about that, and did some checking.

It seems that the high resolution timestamp is a correct epoch timestamp. For example, I got one `1611605368747782988` earlier today, which is actually representing a valid epoch timestamp today. However, the default resolution timestamp doesn't seem correct. I just got one `25243412105673041`, which points to year 1970. If we do a subtraction, then we get an underflow.

davidtgoldblatt:
Huh, TIL.

PR #2021, opened 2021-02-01
Optimize for memory locality on tcache flush pathways

davidtgoldblatt:
This is a sequence of commits to minimize cache misses and missed compiler-optimization opportunities down the tcache flush pathways. Each commit has a relatively descriptive message, so I won't bother repeating it here. I think probably the trickiest thing is ticker_geom_t; through some careful randomization, we can avoid need to keep per-arena state in the TSD without changing the average time until decay triggers.

Stacked together, these commits result in something like 0.4% end-to-end reduction in cycles spent in malloc in various large binaries. They also dramatically reduce wait times spent on arena bin locks (something like a 40% reduction in wait times on the hotter mutexes), but I'm having trouble figuring out the exact CPU reduction attributable to these; I'm going to think a bit more on it.


include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -441,4 +442,89 @@ arena_cache_oblivious_randomize(tsdn_t *tsdn, arena_t *arena, edata_t *edata,
...
+arena_dalloc_bin_locked_begin(arena_dalloc_bin_locked_info_t *info,
+    szind_t binind) {
+	info->div_info = arena_binind_div_info[binind];
+	info->nregs = bin_infos[binind].nregs;
+	info->ndalloc = 0;
+}
+
+/* Returns true if arena_slab_dalloc must be called on slab */
+JEMALLOC_ALWAYS_INLINE bool
+arena_dalloc_bin_locked_step(tsdn_t *tsdn, arena_t *arena, bin_t *bin,
```
interwq, line 485: nit: document that the stats is updated outside of this function, maybe also why it's called _step.

include/jemalloc/internal/cache_bin.h
```diff
@@ -441,29 +441,18 @@ cache_bin_finish_fill(cache_bin_t *bin, cache_bin_info_t *info,
...
 cache_bin_init_ptr_array_for_flush(cache_bin_t *bin, cache_bin_info_t *info,
     cache_bin_ptr_array_t *arr, cache_bin_sz_t nflush) {
-	arr->ptr = cache_bin_empty_position_get(bin) - 1;
+	arr->ptr = cache_bin_empty_position_get(bin) - nflush;
 	assert(cache_bin_ncached_get_local(bin, info) == 0
 	    || *arr->ptr != NULL);
 }
 
-/*
- * These accessors are used by the flush pathways -- they reverse ordinary array
```
interwq, line 454: The change on the reserver order made this part better in every way. Which made me wonder, was there any particular reason this was reverse flushed? Or was it really just out of "convenience"?

interwq (commented):
Looking great! Just a few nits. Also discussed offline about the new ticker.

Another question, any insights regarding which parts of the changes made the biggest dent? (not asking to testing them individually, but in case you had intermediate numbers along the way).

include/jemalloc/internal/emap.h
```diff
@@ -234,15 +234,34 @@ emap_edata_lookup_batch(tsd_t *tsd, emap_t *emap, size_t nptrs,
...
+	rtree_ctx_t *rtree_ctx = tsd_rtree_ctxp_get(tsd);
 
 	for (size_t i = 0; i < nptrs; i++) {
-		emap_full_alloc_ctx_t full_alloc_ctx;
 		const void *ptr = ptr_getter(ptr_getter_ctx, i);
+		/*
+		 * Reuse the edatas array as a temp buffer, lying a little about
+		 * the types.
+		 */
+		r_edatas[i] = (edata_t *)rtree_leaf_elm_lookup(tsd_tsdn(tsd),
```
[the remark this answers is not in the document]
davidtgoldblatt: I ended up making it more complex :(. I had convinced myself that this sort of cheating was allowed, but it's not really without a union. Given that we do actually care about the ordering of the reuse, and this could be broken by TBAA, made it UB-free.

include/jemalloc/internal/emap.h
```diff
@@ -234,15 +234,34 @@ emap_edata_lookup_batch(tsd_t *tsd, emap_t *emap, size_t nptrs,
...
+	for (size_t i = 0; i < nptrs; i++) {
+		rtree_leaf_elm_t *elm = (rtree_leaf_elm_t *)r_edatas[i];
+		rtree_contents_t contents = rtree_leaf_elm_read(tsd_tsdn(tsd),
+		    &emap->rtree, elm, /* dependent */ true);
+		r_edatas[i] = contents.edata;
+		emap_full_alloc_ctx_t alloc_ctx;
+		/*
+		 * Of these, only szind actually gets read by the metadata
+		 * visitor in practice.  But these get optimized away anyways,
+		 * no sense in being lazy about it.
```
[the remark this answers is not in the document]
davidtgoldblatt: Rephrased. The compiler optimizing away (e.g.) the setting of "slab" (which doesn't get used) is what I mean.

include/jemalloc/internal/arena_inlines_b.h
davidtgoldblatt: Done.

include/jemalloc/internal/cache_bin.h
davidtgoldblatt: I think it's mostly a historical accident. Originally, there was just a for loop in the arena code accessing tcache internals directly, where it was just as easy to write `for (void **ptr = bin_end; ptr != bin_end - nflush; ptr--)`.

When extracting cache_bin out of the tcache/arena interactions, I accidentally inverted the flush ordering during development and it caused a hard-to-track-down regression. So after that, I just kept every access the exact same, even at the cost of extra syntactic overhead. I don't think anyone ever consciously decided that the syntactic overhead was necessary to accomplish some other goal (except for accommodating my laziness).

davidtgoldblatt (commented):
The batch edata lookup and prefetching the edata contents are the two big ones; something like 2/3rds of the improvement is those two (it's hard to break them down more finely, because you can't really do the edata prefetching without also doing the batching). It's fairly evenly split between the rest (which are two some extent just following the pattern of "find a cache miss and make it go away").

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -115,18 +116,22 @@ arena_prof_info_set(tsd_t *tsd, edata_t *edata, prof_tctx_t *tctx,
...
-	if (unlikely(ticker_ticks(decay_ticker, nticks))) {
+	tsd_t *tsd = tsdn_tsd(tsdn);
+	/*
+	 * We use the ticker_geom_t to avoid having per-arena state in the tsd.
+	 * Instead of having a countdown-until-decay timer running for every
+	 * arena in every thread, we flip a coin once per tick, whose
+	 * probability of coming up heads is 1/nticks; this is effectively the
+	 * operation of the ticker_geom_t.  Each arena has the same chance of a
+	 * coinflip coming up heads (1/ARENA_DECAY_NTICKS_PER_UPDATE), so we can
+	 * use a single ticker for all of them.
```
interwq, line 130: Looks great. Thanks for the clarification!

test/unit/ticker.c
```diff
@@ -64,10 +64,32 @@ TEST_BEGIN(test_ticker_copy) {
...
+	ticker_geom_init(&ticker, ticks);
+	uint64_t total_ticks = 0;
+	/* Just some random constant. */
+	uint64_t prng_state = 0x343219f93496db9fULL;
+	for (uint64_t i = 0; i < niters; i++) {
+		while(!ticker_geom_tick(&ticker, &prng_state)) {
+			total_ticks++;
+		}
+	}
+	expect_u64_ge(total_ticks / niters, ticks * 9 / 10,
```
interwq, line 81: nit: add comment to include a typical result; usually the mean should be very close?

interwq (approved):
LGTM!

test/unit/ticker.c
davidtgoldblatt: Done.

Really, we quantize coarsely enough that there is some bias in the resulting output, but shrinking that by making the ticker table bigger didn't seem worth it to me (we don't really care if the average tick time is exactly 1000 or 1050 or whatever).

PR #2023, opened 2021-02-04
Hugepage scalability improvements

davidtgoldblatt:
This stack of commits reduces SEC shard lock hold times (by dropping the lock while flushing), increases SEC hit rates (by only flushing individual bins in the SEC until we get down below some threshold), and adds batch allocation/deallocation facilities to the hpa shard.


davidtgoldblatt:
Closed in favor of https://github.com/jemalloc/jemalloc/pull/2029.

PR #2027, opened 2021-02-10
Implement opt.cache_oblivious.

interwq:
Add `opt.cache_oblivious` to make the option runtime configurable. Keep `config.cache_oblivious` for now to remain backward-compatible.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -102,6 +102,14 @@ bool	opt_trust_madvise =
 #endif
     ;
 
+bool opt_cache_oblivious =
+#ifdef JEMALLOC_CACHE_OBLIVIOUS
+    true
+#else
+    false
+#endif
+    ;
+
 zero_realloc_action_t opt_zero_realloc_action =
     zero_realloc_action_strict;
 
@@ -1697,7 +1705,7 @@ malloc_init_hard_a0_locked() {
 		prof_boot0();
 	}
 	malloc_conf_init(&sc_data, bin_shard_sizes);
-	sz_boot(&sc_data);
+	sz_boot(&sc_data, opt_cache_oblivious);
 	bin_info_boot(&sc_data, bin_shard_sizes);
 
 	if (opt_stats_print) {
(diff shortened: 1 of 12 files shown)


src/jemalloc.c
```diff
@@ -2790,12 +2798,7 @@ isfree(tsd_t *tsd, void *ptr, size_t usize, tcache_t *tcache, bool slow_path) {
 			 * usize can be trusted to determine szind and slab.
 			 */
 			alloc_ctx.szind = sz_size2index(usize);
-			if (config_cache_oblivious) {
-				alloc_ctx.slab = (alloc_ctx.szind < SC_NBINS);
-			} else {
-				/* Non page aligned must be slab allocated. */
```
interwq, line 2796: @davidtgoldblatt this is something I didn't realize. It's not a huge deal, but still adding a branch on the fastpath (of a common case -- config_prof:true, opt_prof:false; ptr not page aligned) isn't something we enjoy.

I'm thinking of this: first add opt.cache_oblivious, then benchmark on the binaries we care about. Kill the entire feature if none measures any benefit (plus verifying that newer CPUs have non-direct mapping for cache index, although I'm less sure on the ARM front). WDYT?

davidtgoldblatt: That makes sense to me, at least as a first step.

If there's a regression, we should still be able to restructure things to eliminate it, though. The `slab` field doesn't actually get used in the cases where the allocation is going into the tcache, so we should be able to defer the computation until much later (basically, until we're in the arena_sdalloc slow path). This is probably the right direction for the fast path code in general (i.e. tcache as a strict user of the arena, rather than bouncing around as we do now), but it's a little tricky. I wouldn't hold it against you if you didn't want to do it all right now.

I think the compiler may even be able to reason all this out on its own -- it wouldn't surprise me if the check is moved outside the fast path in the generated assembly.

src/sz.c
```diff
@@ -106,6 +107,7 @@ sz_boot_size2index_tab(const sc_data_t *sc_data) {
 
 void
 sz_boot(const sc_data_t *sc_data) {
+	sz_large_pad = opt_cache_oblivious ? PAGE : 0;
```
davidtgoldblatt, line 110: Should take cache_oblivious as an argument rather than looking at the global -- this is new-style code where where we're trying to minimize dependencies on global state.

src/jemalloc.c
jasone: Re: benchmarking `cache_oblivious`, it won't make a difference for many applications, but when it matters, it *really* matters. I originally added the feature due to terrible HPC benchmark results reported by a FreeBSD user (IIRC large matrix allocations/operations, but other use cases can also trip on cache-associativity limits). You may decide that the feature doesn't justify the complexity (I'm okay with that decision if you come to it), but benchmarking will only help if you're trying to verify there's no performance impact of the feature when it's disabled.

davidtgoldblatt: Oh yeah, I think I misread -- I don't think we should kill off cache_oblivious entirely, but rather the runtime-enablement feature if it's too much work to make it dynamic.

(I think we should probably also flip the default for the non-libc case -- the median "apt install jemalloc" user is probably doing so to save memory, so we would ideally help that case in the ways we can).

interwq: Thanks a lot for sharing the inputs Jason and David! I'm convinced to keep it; also I just realized that the `isfree` wasn't the fast-fastpath that I thought, since `free_fastpath` now handles the most common cases for both sized and unsized deallocations. Let's not change the default until we got more concrete numbers to support it.

davidtgoldblatt (approved):

PR #2028, opened 2021-02-10
Fix stats for "tcache_max" (was "lg_tcache_max")

jordalgo:
Seems like this option was changed here: c8209150f9d219a137412b06431c9d52839c7272
And maybe this got missed.

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -1498,7 +1498,7 @@ stats_general_print(emitter_t *emitter) {
 	OPT_WRITE_BOOL("utrace")
 	OPT_WRITE_BOOL("xmalloc")
 	OPT_WRITE_BOOL("tcache")
-	OPT_WRITE_SSIZE_T("lg_tcache_max")
+	OPT_WRITE_SIZE_T("tcache_max")
 	OPT_WRITE_UNSIGNED("tcache_nslots_small_min")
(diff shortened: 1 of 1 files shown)


jordalgo:
No idea how to test this though.

davidtgoldblatt (commented):
Looks good; mind cleaning up the commit message a bit (e.g. capitalization, and mention `lg_tcache_max` as the old name)?

There's not really a great way to test, unfortunately (other than manually inspecting the stats output). Ideally, these would be hard errors in debug mode; but there are some config/opt interactions that make this not exactly trivial.

src/stats.c
```diff
@@ -1498,7 +1498,7 @@ stats_general_print(emitter_t *emitter) {
 	OPT_WRITE_BOOL("utrace")
 	OPT_WRITE_BOOL("xmalloc")
 	OPT_WRITE_BOOL("tcache")
-	OPT_WRITE_SSIZE_T("lg_tcache_max")
+	OPT_WRITE_SSIZE_T("tcache_max")
```
interwq, line 1501: nit: it should be `OPT_WRITE_SIZE_T`, the option is now unsigned.

interwq (approved):
Thanks for the fix!

PR #2029, opened 2021-02-10
Hpa purging and scalability improvements

davidtgoldblatt:
This is a few logical pieces of functionality, stacked together per reviewer preference.

The initial (SEC-centered commits) stack of commits reduces SEC shard lock hold times (by dropping the lock while flushing), increases SEC hit rates (by only flushing individual bins in the SEC until we get down below some threshold), and adds batch allocation/deallocation facilities to the hpa shard.

We then have a few miscellaneous cleanups and performance improvements (edata change through the psset bitmap renaming).

The last two commits rework purging; first by doing a "so-so but principled" approach (in which every huge extent with dirty pages goes on a FIFO purging list, segregated by hugification status so that we purge hugified huge extents last). This is mostly to set up the framework with something simple.

In the last commit, we switch to a dirtiest-first strategy selection strategy. We still purge non-hugified extents before hugified ones, but only within a given size class bucket. This seems to work much better in practice. (I think there's lots more tuning here to do, though -- maybe linear dirtiness bucketing is better than exponential, for example).

### DIFF
--- a/include/jemalloc/internal/hpdata.h
+++ b/include/jemalloc/internal/hpdata.h
@@ -7,6 +7,42 @@
 #include "jemalloc/internal/typed_list.h"
 
 /*
+ * How badly we want to purge some region of memory.  This is a temporary
+ * definition; it gets deleted in the next commit (where we adopt a more
+ * explicit dirtiest-first policy that only considers hugification status).
+ */
+enum hpdata_purge_level_e {
+	/*
+	 * The level number is important -- we use it as indices into an array
+	 * of size 2 (one for each purge level).
+	 */
+
+	/* "Regular" candidates for purging. */
+	hpdata_purge_level_default = 0,
+
+	/*
+	 * Candidates for purging, but as a last resort.  Practically,
+	 * nonpreferred corresponds to hugified regions that are below the
+	 * hugification threshold but have not yet reached the dehugification
+	 * threshold, while strongly nonpreferred candidates are those which are
+	 * above the hugification threshold.
+	 */
+	hpdata_purge_level_nonpreferred = 1,
+	hpdata_purge_level_strongly_nonpreferred = 2,
+
+	/* Don't purge, no matter what. */
+	hpdata_purge_level_never = 2,
+
+	/*
+	 * How big an array has to be to accomodate all purge levels.  This
+	 * relies on the fact that we don't actually keep unpurgable hpdatas in
+	 * a container.
+	 */
+	hpdata_purge_level_count = hpdata_purge_level_never
+};
+typedef enum hpdata_purge_level_e hpdata_purge_level_t;
+
+/*
  * The metadata representation we use for extents in hugepages.  While the PAC
  * uses the edata_t to represent both active and inactive extents, the HP only
  * uses the edata_t for active ones; instead, inactive extent state is tracked
@@ -52,8 +88,8 @@ struct hpdata_s {
 	bool h_in_psset_alloc_container;
 
 	/* The same, but with purging. */
-	bool h_purge_allowed;
-	bool h_in_psset_purge_container;
+	uint8_t h_purge_level;
+	uint8_t h_purge_container_level;
 
 	/* And with hugifying. */
 	bool h_hugify_allowed;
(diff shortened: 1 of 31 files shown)


src/psset.c
```diff
@@ -195,6 +196,51 @@ psset_alloc_container_remove(psset_t *psset, hpdata_t *ps) {
...
+
+static void
+psset_maybe_remove_purge_list(psset_t *psset, hpdata_t *ps) {
+	/*
+	 * Remove the hpdata from its purge list (if it's in one).  Even if it's
+	 * going to stay in the same one, by appending it during
+	 * psset_update_end, we move it to the end of its queue, so that we
+	 * purge LRU within a given dirtiness bucket.
+	 */
+	if (hpdata_purge_allowed_get(ps)) {
```
interwq, line 221: Question: in what case do we go down this path while purge is not allowed? (same question for the insert below). Does that mean setting purge allowed and the list operation need to be ordered?

src/sec.c
```diff
@@ -107,6 +141,49 @@ sec_shard_alloc_locked(tsdn_t *tsdn, sec_t *sec, sec_shard_t *shard,
...
+	if (ret != NULL) {
+		edata_list_active_remove(&result, ret);
+	}
+
+	malloc_mutex_lock(tsdn, &shard->mtx);
+	bin->being_batch_filled = false;
+	/*
+	 * Handle the easy case first: nothing to cache.  Note that this can
+	 * only happen in case of OOM, since sec_alloc checks the expected
+	 * number of allocs, and doesn't bother going down the batch_fill
```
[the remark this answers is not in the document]
davidtgoldblatt: It's possible that pai_alloc_batch was *asked* to allocate multiple items, but OOM'd after the first one. Beefed up the comment a little.

src/hpa.c
```diff
@@ -487,42 +486,62 @@ hpa_try_alloc_no_grow(tsdn_t *tsdn, hpa_shard_t *shard, size_t size, bool *oom)
...
+	bool oom = false;
+
+	size_t nsuccess = hpa_try_alloc_batch_no_grow(tsdn, shard, size, &oom,
+	    nallocs, results);
 
-	edata = hpa_try_alloc_no_grow(tsdn, shard, size, &oom);
-	if (edata != NULL) {
-		return edata;
+	if (nsuccess == nallocs || oom) {
+		assert(!oom);
```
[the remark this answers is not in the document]
davidtgoldblatt: Left in accidentally while debugging, removed now. (This probably says something about our testing of OOMs, but for now I think that's inevitable).

src/hpa.c
```diff
@@ -487,42 +486,62 @@ hpa_try_alloc_no_grow(tsdn_t *tsdn, hpa_shard_t *shard, size_t size, bool *oom)
...
+		return nsuccess;
 	}
 
-	/* Nothing in the psset works; we have to grow it. */
+	/*
+	 * We didn't OOM, but weren't able to fill everything requested of us;
+	 * try to grow.
```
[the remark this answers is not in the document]
davidtgoldblatt: I think of this as more like the tcache pathway; we affirmatively want everything we ask for (and in fact, I think that ideally, tcache fills and flushes of page-sized allocations would be able to use the batch alloc/dalloc functionality eventually).

> However no strong feeling either way; IIUC, similar to PAC, the grow operation should only reserve / consume the vaddr, as newly allocated regions are unlikely to get hugified?

Right.

include/jemalloc/internal/hpdata.h
```diff
@@ -6,6 +6,42 @@
 #include "jemalloc/internal/ql.h"
 #include "jemalloc/internal/typed_list.h"
 
+/*
```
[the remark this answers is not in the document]
davidtgoldblatt: Done. Fixed another typo, too ("more worth candidate" -> "more worthy candidate").

src/psset.c
davidtgoldblatt: I'm not totally sure I follow; a pageslab is never in the purge list in between psset_update_begin and psset_update_end. So in that sense, the operations need to be ordered, but not any more so than how every pageslab metadata update happens between begin and end.

interwq: If we want to disallow a pageslab (which is currently in the purge list) from being purged, the order should be, remove it from the purge list first, and then update the metadata for `purge_allowed`? Because otherwise if we update the metadata first, this function would simply be skipped entirely?

interwq: > a pageslab is never in the purge list in between psset_update_begin and psset_update_end

Is this ensured by the API or the logic flow like we always take a pageslab off the lists before, or it's disallowed since it doesn't make sense? I might be missing something here; is it right that a pageslab on the purge list can be partially in use, and keeps serving requests, and be a part of the psset that way? (which is different than PAC "purge list").

interwq: Discussed offline; I forgot that the LRU-nature of the purgelist means that every time a pageslab gets updated it will also needs to be adjusted for the LRU.

Do we have assertions to check for that case, i.e. pageslab not on purge list during update (I seem to recall we do)?

davidtgoldblatt: Not for purging (I don't think there's in general an easy way to check if a pageslab is in a list in our linked list implementation). So the "can purge" bool tells us if something is in the list when it's not being updated.

interwq: Sounds good. Thanks!

interwq (approved):

PR #2031, opened 2021-02-16
Update INSTALL.md to mention 'autoconf'

jordalgo:
'autoconf' needs to be installed for './autogen.sh' to work.


davidtgoldblatt (commented):
Good thing to note, thanks!

PR #2034, opened 2021-02-17
Darwin malloc_size override support proposal.

devnexen:
Darwin has similar api than Linux/FreeBSD's malloc_usable_size.


davidtgoldblatt:
I think this makes sense overall -- probably though, we should keep malloc_usable_size (since it's a jemalloc extension on OS X at this point, even if it wasn't really meant to be), and just have both mallooc_usable_size and malloc_size (probably both just being thin wrappers around some shared implementation function).

interwq:
Merged in cf9724531af2864b243668d82aa63114e9737bfd. Thanks for the patch @devnexen!

PR #2035, opened 2021-02-25
Add m4 to .cirrus.yml.

davidtgoldblatt:
This is a precondition for building.

I'm not actually sure this will fix the CI errors, but we'll see.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ env:
 task:
   freebsd_instance:
     matrix:
-      image: freebsd-12-1-release-amd64
+      image: freebsd-12-2-release-amd64
   install_script:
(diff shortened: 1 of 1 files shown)


interwq:
Doesn't seem to work even though m4 is installed according to the log
```
[3/5] Fetching m4-1.4.18_1,1.txz: .......... done
[4/5] Fetching perl5-5.32.1_1.txz: .......... done
[5/5] Fetching autoconf-wrapper-20131203.txz: . done
Checking integrity... done (0 conflicting)
[1/5] Installing m4-1.4.18_1,1...
[1/5] Extracting m4-1.4.18_1,1: .......... done
```
Somehow the path is not expected or set properly?

interwq (approved):
Looks good!

Would it be possible to do multiple builds / configs with Cirrus? Right now it's only doing the default config.

davidtgoldblatt:
I think it should be; I think cirrus supports a travis-style configuration matrix. It should "just" be a question of syntax and names and whatnot.

PR #2037, opened 2021-02-26
Extent-state synchronization improvements

interwq:
This PR adds tracking of extent state in rtree leaf, so that we can stop relying on the address based mutex pool.

This avoids the addr-based mutexes (i.e. the mutex_pool), and instead relies on
the metadata tracked in rtree leaf: the head state and extent_state.  Before
trying to access the neighbor edata (e.g. for coalescing), the states will be
verified first -- only neighbor edatas from the same arena and with the same
state will be accessed.

Production benchmark shows ~20% of mutex-related cycles reduction within jemalloc.


src/emap.c
```diff
@@ -35,82 +26,160 @@ emap_init(emap_t *emap, base_t *base, bool zeroed) {
...
+		return false;
+	}
+	if (opt_retain) {
+		assert(edata_arena_ind_get(edata) ==
+		    edata_arena_ind_get(neighbor));
+	} else {
+		/*
+		 * This isn't entirely safe with the presence of arena_reset /
+		 * destroy, in which case the neighbor edata can be destoryed if
+		 * it belongs to a manual arena.  More on that later.
```
interwq, line 97: Note that the current PR doesn't deal with that yet. We discussed a few solutions (for the !opt_retain + arena_reset case). I'll be working on that next.

src/emap.c
```diff
@@ -35,82 +26,184 @@ emap_init(emap_t *emap, base_t *base, bool zeroed) {
...
 }
 
-void
-emap_unlock_edata2(tsdn_t *tsdn, emap_t *emap, edata_t *edata1,
-    edata_t *edata2) {
-	assert(edata1 != NULL && edata2 != NULL);
-	mutex_pool_unlock2(tsdn, &emap->mtx_pool, (uintptr_t)edata1,
-	    (uintptr_t)edata2);
+static inline bool
+edata_neighbor_head_state_mergeable(bool edata_is_head,
```
davidtgoldblatt: This and the related functions belong in extent.c; the emap should be about keeping edatas and the rtree in sync; making policy decisions like when things should be merged belongs there. The HPA doesn't care about any of this functionality; it has its own policy decisions about when extents are mergeable; those decisions should go with the allocator.

For a related reason, we should probably have the PAI bit in the rtree metadata; it buys us flexibility in the edata representation.

src/hpa_central.c
```diff
@@ -189,21 +169,24 @@ hpa_central_dalloc(tsdn_t *tsdn, hpa_central_t *central, edata_t *edata) {
...
-	}
-	edata_t *trail = hpa_central_dalloc_get_merge_candidate(tsdn, central,
-	    edata_past_get(edata));
-	if (trail != NULL && !edata_is_head_get(trail)) {
+	/*
+	 *  Merge forward first, so that the original *edata stays active state
+	 *  for the second acquire (only necessary for sanity checking).
+	 */
+	edata_t *trail = emap_try_acquire_edata_neighbor(tsdn, central->emap,
+	    edata, EXTENT_PAI_HPA, extent_state_dirty, true /* forward */);
```
davidtgoldblatt, line 177: I'm not religious about this sort of thing, but everywhere else in the codebase we put the bool name comment before the value rather than after.

src/pac.c
```diff
@@ -158,6 +158,7 @@ pac_expand_impl(tsdn_t *tsdn, pai_t *self, edata_t *edata, size_t old_size,
 	if (trail == NULL) {
 		return true;
 	}
+	/////
```
davidtgoldblatt: Accident?

src/emap.c
davidtgoldblatt: We settled on have arena destroy remove all extents from the rtree first, then acquire the existing mutex locks (and releasing them immediately), right? Can we rebase that commit in front of this one, just so we don't have an intermediate period where we introduce and then fix an arena destroy race (just so history stays clean / bisectable)?

include/jemalloc/internal/ehooks.h
```diff
@@ -338,21 +337,10 @@ ehooks_split(tsdn_t *tsdn, ehooks_t *ehooks, void *addr, size_t size,
 
 static inline bool
 ehooks_merge(tsdn_t *tsdn, ehooks_t *ehooks, void *addr_a, size_t size_a,
-    bool head_a, void *addr_b, size_t size_b, bool head_b, bool committed) {
```
davidtgoldblatt, line 341: Love it.

src/emap.c
interwq: Yes I'll push another commit soon, implementing the empty critical section trick we discussed. I'll make sure to rebase that commit to before this one.

include/jemalloc/internal/edata.h
```diff
@@ -13,6 +13,12 @@
 #include "jemalloc/internal/sz.h"
 #include "jemalloc/internal/typed_list.h"
 
+/*
+ * sizeof(edata_t) is 128 bytes on 64-bit architectures.  Ensure the alignment
+ * to free up the low bits in the rtree leaf.
+ */
+#define EDATA_ALIGNMENT 64
```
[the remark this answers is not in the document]
interwq: Will do. Let me also merge the other commit (the one changing it to 128) with this one.

include/jemalloc/internal/emap.h
```diff
@@ -136,6 +136,35 @@ emap_assert_not_mapped(tsdn_t *tsdn, emap_t *emap, edata_t *edata) {
 	}
 }
 
+static inline void
+rtree_edata_state_update(tsdn_t *tsdn, rtree_t *rtree, edata_t *edata,
```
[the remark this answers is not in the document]
interwq: Makes sense. Will update.

include/jemalloc/internal/edata.h
```diff
@@ -13,6 +13,12 @@
 #include "jemalloc/internal/sz.h"
 #include "jemalloc/internal/typed_list.h"
```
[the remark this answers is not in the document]
interwq: > one of which is disallowed by anyone but the emap

That is true. I'm with you that we should have one copy of metadata instead of cached / duplicated. The only / main advantage of that is probably convenience.

How about we remove the duplicated states in the next PR? There will be quite some code motion for that change. This one is getting pretty large, and I already have a few things want to do but not included (e.g. the metadata_brief we also discussed).

test/unit/rtree.c
```diff
@@ -32,8 +32,10 @@ TEST_END
 #undef NITERS
 #undef SEED
 
+JEMALLOC_ATTR(aligned(128)) edata_t edata_a = {0}, edata_b = {0},
```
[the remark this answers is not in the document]
interwq: Will change them to heap allocated.

include/jemalloc/internal/edata.h
```diff
@@ -17,7 +17,7 @@
  * sizeof(edata_t) is 128 bytes on 64-bit architectures.  Ensure the alignment
```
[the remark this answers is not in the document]
interwq: Yep will do.

include/jemalloc/internal/edata.h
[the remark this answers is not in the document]
davidtgoldblatt: Yep, totally fine.

src/emap.c
interwq: Makes sense. Will move. (Maybe later on the PAI bit part though).

src/pac.c
interwq: Yeah will fix.

src/hpa_central.c
interwq: I'm with you. Will fix.

include/jemalloc/internal/edata.h
```diff
@@ -143,7 +149,7 @@ struct edata_s {
 #define EDATA_BITS_ZEROED_SHIFT  (EDATA_BITS_PAI_WIDTH + EDATA_BITS_PAI_SHIFT)
 #define EDATA_BITS_ZEROED_MASK  MASK(EDATA_BITS_ZEROED_WIDTH, EDATA_BITS_ZEROED_SHIFT)
 
-#define EDATA_BITS_STATE_WIDTH  2
+#define EDATA_BITS_STATE_WIDTH  4
```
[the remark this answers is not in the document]
interwq: Yeah had more stuff planned earlier but will not happen in this PR. Will update.

include/jemalloc/internal/witness.h
```diff
@@ -286,6 +287,20 @@ witness_assert_lockless(witness_tsdn_t *witness_tsdn) {
 	witness_assert_depth(witness_tsdn, 0);
 }
 
+static inline void
+witness_assert_not_lockless(witness_tsdn_t *witness_tsdn) {
```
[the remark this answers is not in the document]
interwq: Makes sense. Will update.

src/arena.c
```diff
@@ -605,6 +605,88 @@ arena_reset(tsd_t *tsd, arena_t *arena) {
...
+	if (opt_retain) {
+		return;
+	}
+	unsigned destroy_ind = base_ind_get(base_to_destroy);
+	assert(destroy_ind >= manual_arena_base);
+
+	tsdn_t *tsdn = tsd_tsdn(tsd);
+	malloc_mutex_t *delayed_mtx[ARENA_DESTROY_MAX_DELAYED_MTX];
+	unsigned n_delayed = 0;
+	for (unsigned i = 0; i < narenas_total_get(); i++) {
```
interwq, line 671: @davidtgoldblatt what do you think of this approach? Slightly tweaked based on what we discussed: instead of tracking the is_manual_arena / is_destroyable state in rtree leaf, here we ask the destroy operation to go through all arenas' locks, versus only the manual arenas'. Since this can make the # of locks involved a lot higher, I added a simple trylock + batching mechanism to compensate.

(only the commit adding this part is new (and rebased to the earliest), other commits not updated yet.

davidtgoldblatt: I like just acquiring all the locks.

The trylock/batching thing feels like premature optimization to me; do we care that much about arena destroy latency / do we even have a benchmark implying this would help[1]? Not really opposed per se, but it feels like an edge-casey thing to take on the code complexity for.

[1] You could imagine a world where if an ecache lock is hot, we're always gonna have to do a contended spin to acquire it. So the initial trylock pointlessly slows down other threads without the benefit. I'm not saying we live in this world, but without a use case it's hard to even be sure if batching helps/hurts.

interwq: Yeah it's not a straight win. My thoughts were: in the case where it ends up hurting, it should be just another small dent in an already terrible situation, and won't really gets blamed for the whole issue; but in the case where it helps (e.g. a busy workload resulting semi-hot ecache locks), the out of order locking could make the overhead / latency tolerable.

Though the truth is I don't really have a preference; I admit the initial motivation was like, oh I made the this path slower, let's see if there's any remedy for that. The final call is your @davidtgoldblatt :)

davidtgoldblatt: Eh fair enough. I feel like status quo wins in a situation where no one has a strong preference, so I'm fine leaving it the way it is now.

interwq: Haha okay. I'll update the PR and let you know when it's ready.

interwq:
@davidtgoldblatt all the feedback were addressed. The part of moving edata merge-ability related functions to `extent.h` was done as a new commit (tried to update the previous commits but the conflicts looked risky).

I think we made the right call regarding the edata liveness synchronization -- the changes corresponding to that is well contained near `arena_destroy`; it's fairly close to the "edata live forever" world already.

include/jemalloc/internal/witness.h
```diff
@@ -286,6 +287,21 @@ witness_assert_lockless(witness_tsdn_t *witness_tsdn) {
 	witness_assert_depth(witness_tsdn, 0);
 }
 
+static inline void
+witness_assert_min_depth_to_rank(witness_tsdn_t *witness_tsdn,
```
davidtgoldblatt, line 290: Nit: should probably rename or actually take a depth parameter.

davidtgoldblatt (approved):

include/jemalloc/internal/witness.h
interwq: Good point. How about `witness_assert_min_rank()`? I'm afraid the relative depth isn't very intuitive to the users; it may even change if we remove / reorder locks.

davidtgoldblatt: `witness_assert_positive_depth_to_rank` maybe?

interwq: Makes sense. Will change.

PR #2038, opened 2021-03-12
Set is_head in extent_alloc_wrapper w/ retain.

interwq:
When retain is on, when extent_grow_retained failed (e.g. due to split hook
failures), we'll try extent_alloc_wrapper as the last resort.  Set the is_head
bit in that case to be consistent.  The allocated extent in that case will be
retained properly, but not merged with other extents.


interwq:
This is the only place where alloc_wrapper is called: https://github.com/jemalloc/jemalloc/blob/dev/src/extent.c#L114

On that path, if retain is on, it implies that grow_retained failed (e.g. because of custom split hooks in `test/integration/extent`). We should still mark the is_head bit for the newly mmaped edata. It will disallow merging. However if split fails w/ retain, no merging is kind of the right thing to do -- otherwise we may keep merging retained extents to be larger and larger, and never be able to reuse them.

davidtgoldblatt (approved):

PR #2042, opened 2021-03-17
Red black tree summarize/filter functionality

davidtgoldblatt:
This expands the red-black tree to support various sorts of augmented tree data structures (e.g. trees supporting rank/select queries, interval trees, etc.). Insert/remove now have the ability to call a user hook on tree rotation operations, to inform parents of updates in their children. Users can thereby summarize information about the subtree rooted at some particular node, and use those summarize to guide later "filtered" searches. A more extended example is included in the comments.

Also are a couple other cleanups "flat_bitmap.h -> fb.h" (to match the function names, and the spelling that's now in my muscle memory), and macro naming standardization.


davidtgoldblatt:
Fixed an incoherent comment and rebased past the cirrus fix.

davidtgoldblatt:
OK, last version is *really* the last version I think. I had forgotten to add some of the extended API functions to rb_summarized_proto, and I expanded a little more on the comment.

include/jemalloc/internal/rb.h
```diff
@@ -1001,6 +1350,465 @@ a_prefix##destroy(a_rbt_type *rbtree, void (*cb)(a_type *, void *),	\
...
+  void *filter_ctx) {							\
+    a_type *node = rbtree->rbt_root;					\
+    if (node == NULL || !filter_subtree(filter_ctx, node)) {		\
+	return NULL;							\
+    }									\
+    return a_prefix##last_filtered_from_node(node, filter_node,		\
+	filter_subtree, filter_ctx);					\
+}									\
+static inline a_type *							\
+a_prefix##search_with_filter_bounds(a_rbt_type *rbtree,			\
```
interwq, line 1456: nit: explain the bounds and the usage a bit. They can't both be true at the same time, is that right?

test/unit/rb.c
```diff
@@ -341,15 +473,546 @@ TEST_BEGIN(test_rb_random) {
...
+			check_consistency(&tree, nodes, nnodes);
+		}
+	}
+}
+
+TEST_BEGIN(test_rb_update_search) {
+	summarize_always_returns_true = false;
+	do_update_search_test(2, 100, 3, 50);
+	do_update_search_test(5, 100, 3, 50);
+	do_update_search_test(12, 100, 5, 1000);
```
interwq, line 980: The synthetic tests look great; any reason why not testing every possible nnodes (i.e. upto UPDATE_TEST_MAX)?

include/jemalloc/internal/rb.h
```diff
@@ -338,8 +411,186 @@ a_prefix##destroy(a_rbt_type *rbtree, void (*cb)(a_type *, void *),	\
...
+ *         const a_type *a_child2);
+ *
+ *     This function should update a_node with the summary of the subtree rooted
+ *     there, using the data contained in it and the summaries in a_child1 and
+ *     a_child2.  One or both of them may be NULL.  When the tree changes due to
+ *     an insertion or removal, it updates the summaries of all nodes whose
+ *     subtrees have changed (always updating the summaries of children before
+ *     their parents).  If the user alters a node in the tree in a way that
+ *     may change its summary, they can call the generated update_summaries
+ *     function to bubble up the summary changes to the root.  It should return
```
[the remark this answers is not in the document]
davidtgoldblatt: Right.

include/jemalloc/internal/rb.h
```diff
@@ -338,8 +411,186 @@ a_prefix##destroy(a_rbt_type *rbtree, void (*cb)(a_type *, void *),	\
...
+ * larger than 12345 (with ex_nsearch_filtered), iterate over just those
+ * multiples of 64 that are in the set (with ex_iter_filtered), or determine if
+ * the set contains any multiples of 1024 (with ex_empty_filtered).
+ *
+ * Some possibly subtle API notes:
+ * - The node argument to ex_next_filtered and ex_prev_filtered need not pass
+ *   the filter; it will find the next/prev node that passes the filter.
+ * - ex_search_filtered will fail even for a node in the tree, if that node does
+ *   not pass the filter.  ex_psearch_filtered and ex_nsearch_filtered behave
+ *   similarly; they may return a node larger/smallerthan the key, even if a
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

include/jemalloc/internal/rb.h
```diff
@@ -338,8 +411,186 @@ a_prefix##destroy(a_rbt_type *rbtree, void (*cb)(a_type *, void *),	\
...
+ * find_div_128(ex_tree_t *tree) {
+ *     unsigned min_ffs = 7;
+ *     return ex_first_filtered(tree, &ex_filter_node, &ex_filter_subtree,
+ *         &min_ffs);
+ * }
+ *
+ * We could with similar ease find the next multiple of 128 in the set that's
+ * larger than 12345 (with ex_nsearch_filtered), iterate over just those
+ * multiples of 64 that are in the set (with ex_iter_filtered), or determine if
+ * the set contains any multiples of 1024 (with ex_empty_filtered).
```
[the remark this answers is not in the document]
davidtgoldblatt: Done.

include/jemalloc/internal/rb.h
davidtgoldblatt: Right. (I mean, it would be logically coherent to have them both true at once, but we don't have a `get_bounds` API that would use it[1]). Added some comments.

[1] Maybe we should? Not in this PR of course, but it's a not-uncommon red-black tree library feature.

test/unit/rb.c
davidtgoldblatt: This test is fairly slow in debug mode; doing much more than this causes a noticeable slowdown.

(The special cases 2 and 5 are handy when debugging, since you can easily draw out those trees by hand).

interwq: Makes sense. One more request though -- can you do a one-time manual stress test with a lot of nodes / updates etc?

interwq (approved):

test/unit/rb.c
davidtgoldblatt: Added ` do_update_search_test(100, 1, 50, 500);`, I think much more than that and we should spend some time trying to optimize the test itself (e.g. qsort pointers to the tree nodes or something).

interwq: Oh I was thinking try a high # of nodes like this on your devserver once to see it pass, and not including it in the commit. Totally up to you though.

PR #2047, opened 2021-03-30
Fix a prof_recent/prof_sys_thread_name interaction

davidtgoldblatt:
When both of these are enabled, the output format changes slightly.  Teach the
unit test about the interaction.


interwq (approved):

davidtgoldblatt:
Note added commit -- that's OK too I take it?

interwq:
Yep looks good too!

PR #2048, opened 2021-03-30
Avoid the release & re-acquire of the ecache locks around the split and merge hooks.

interwq:


src/extent.c
```diff
@@ -502,14 +489,16 @@ static edata_t *
 extent_recycle_split(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
     ecache_t *ecache, edata_t *expand_edata, size_t size, size_t alignment,
     edata_t *edata, bool growing_retained) {
+	malloc_mutex_assert_owner(tsdn, &ecache->mtx);
```
LeeHowes, line 492: Presumably this verifies that the lock is held, and implies that we should use extent_deactivate_locked below?

src/extent.c
```diff
@@ -541,8 +530,11 @@ extent_recycle_split(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 		}
 		if (to_leak != NULL) {
 			extent_deregister_no_gdump_sub(tsdn, pac, to_leak);
+			/* May go down the purge path. */
+			malloc_mutex_unlock(tsdn, &ecache->mtx);
```
LeeHowes, line 534: Is this comment obvious enough for future understanding? Why does this imply we should unlock and relock around this operation?

LeeHowes: Has something changed that makes extents_abandon_vm not callable under the lock, or has something changed to make this function hold the lock?

src/extent.c
```diff
@@ -559,23 +551,26 @@ extent_recycle(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, ecache_t *ecache,
     bool *commit, bool growing_retained) {
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, growing_retained ? 1 : 0);
+
+	malloc_mutex_lock(tsdn, &ecache->mtx);
```
LeeHowes, line 555: I see, you moved the lock here so extent_recycle_split now assumes it is locked. Should the function name make that obvious in addition to the assert?

src/extent.c
LeeHowes: So one big change in this diff is moving the locking outside this function and locking once rather than repeatedly.

src/extent.c
```diff
@@ -1116,10 +1107,13 @@ extent_purge_forced_wrapper(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata,
  */
 static edata_t *
 extent_split_impl(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
-    edata_t *edata, size_t size_a, size_t size_b, bool growing_retained) {
+    edata_t *edata, size_t size_a, size_t size_b, bool holding_core_locks) {
 	assert(edata_size_get(edata) == size_a + size_b);
-	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
-	    WITNESS_RANK_CORE, growing_retained ? 1 : 0);
+	/* Only the shrink path may split w/o holding core locks. */
+	if (holding_core_locks) {
```
LeeHowes, line 1113: The second major change is changing when this happens so it only happens under the lock rather than when retained was growing.

LeeHowes (commented):
I think I understand the logic here. It wasn't obvious why the change throughout is to move from growing_retained being the deciding factor, to the more specific locking question.

interwq (commented):
re: move from growing_retained to the more specific locking question -- replied inline in `extent_split_impl()`.

LeeHowes (approved):
Looks good, the comments help thanks.

PR #2066, opened 2021-05-11
mpsc_queue: Add module.

davidtgoldblatt:
This is a simple multi-producer, single-consumer queue.  The intended use case
is in the HPA, as we begin supporting hpdatas that move between hpa_shards.  We
take just a single CAS as the cost to send a message (or a batch of messages) in
the low-contention case, and lock-freedom lets us avoid some lock-ordering
issues.

### DIFF
--- a/include/jemalloc/internal/mpsc_queue.h
+++ b/include/jemalloc/internal/mpsc_queue.h
@@ -1,0 +1,134 @@
+#ifndef JEMALLOC_INTERNAL_MPSC_QUEUE_H
+#define JEMALLOC_INTERNAL_MPSC_QUEUE_H
+
+#include "jemalloc/internal/atomic.h"
+
+/*
+ * A concurrent implementation of a multi-producer, single-consumer queue.  It
+ * supports three concurrent operations:
+ * - Push
+ * - Push batch
+ * - Pop batch
+ *
+ * These operations are all lock-free.
+ *
+ * The implementation is the simple two-stack queue built on a Treiber stack.
+ * It's not terribly efficient, but this isn't expected to go into anywhere with
+ * hot code.  In fact, we don't really even need queue semantics in any
+ * anticipated use cases; we could get away with just the stack.  But this way
+ * lets us frame the API in terms of the existing list types, which is a nice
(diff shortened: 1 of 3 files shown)


include/jemalloc/internal/mpsc_queue.h
```diff
@@ -0,0 +1,134 @@
...
+ * get faster.  Since we're currently providing queue semantics though, we use
+ * the prev field in the link rather than the next field for Treiber-stack
+ * linkage, so that we can preserve order for bash-pushed lists (recall that the
+ * two-stack tricks reverses orders in the lock-free first stack).
+ */
+
+#define mpsc_queue(a_type)						\
+struct {								\
+	atomic_p_t tail;						\
+}
```
interwq, line 31: nit: maybe in dbg build, track the consumer and verifies its truly single consumer? If it's too much hassle then let's not bother.

interwq (commented):
LGTM. What's the use cases you have in mind, for which part of the HPA?

interwq:
re: use cases; never mind just saw the commit msg

include/jemalloc/internal/mpsc_queue.h
```diff
@@ -0,0 +1,134 @@
...
+ * - Push batch
+ * - Pop batch
+ *
+ * These operations are all lock-free.
+ *
+ * The implementation is the simple two-stack queue built on a Treiber stack.
+ * It's not terribly efficient, but this isn't expected to go into anywhere with
+ * hot code.  In fact, we don't really even need queue semantics in any
+ * anticipated use cases; we could get away with just the stack.  But this way
+ * us frame the API in terms of the existing list types, which is a nice
```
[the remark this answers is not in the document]
davidtgoldblatt: Fixed.

include/jemalloc/internal/mpsc_queue.h
davidtgoldblatt: Yeah, it's a little tricky. The consumer here isn't any particular *thread* necessarily, it's just whoever is in charge of an hpdata (i.e. it's single consumer *at a time*, not single consumer in the sense of having a unique tsd_t).

PR #2068, opened 2021-05-12
Fix an interaction between the oversize_threshold test and bgthds.

interwq:
Also added the shared utility to check if background_thread is enabled.

### DIFF
--- a/test/include/test/bgthd.h
+++ b/test/include/test/bgthd.h
@@ -1,0 +1,17 @@
+/*
+ * Shared utility for checking if background_thread is enabled, which affects
+ * the purging behavior and assumptions in some tests.
+ */
+
+static inline bool
+is_background_thread_enabled(void) {
+	bool enabled;
+	size_t sz = sizeof(bool);
+	int ret = mallctl("background_thread", (void *)&enabled, &sz, NULL,0);
+	if (ret == ENOENT) {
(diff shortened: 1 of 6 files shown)


test/include/test/bgthd.h
```diff
@@ -0,0 +1,17 @@
...
+
+static inline bool
+check_background_thread_enabled(void) {
+	bool enabled;
+	size_t sz = sizeof(bool);
+	int ret = mallctl("background_thread", (void *)&enabled, &sz, NULL,0);
+	if (ret == ENOENT) {
+		return false;
+	}
+	expect_d_eq(ret, 0, "Unexpected mallctl error");
```
davidtgoldblatt, line 14: Should probably be `assert_d_eq` (I mean, I know it's just copied and pasted, but might as well fix it too). This isn't a test failure, it's a core jemalloc assumption failure.

Maybe `is_background_thread_enabled()`? `check` to me implies it actually asserts that background threads are enabled.

davidtgoldblatt (approved):
Looks good. Some suggestions, but feel free to disregard if you disagree.

test/include/test/bgthd.h
interwq: Makes sense. Updated.

PR #2075, opened 2021-06-02
support 'make uninstall' to make hacking life easy.

sangshuduo:


davidtgoldblatt:
I think that these are improperly escaped, no? I.e. if the install directory contains spaces, Bad Things will happen. Is that right?

Nit: Clean up commit message; something like "Add support for 'make uninstall'".

sangshuduo:
Unfortunately, GNU make has a bug [https://savannah.gnu.org/bugs/?712](https://savannah.gnu.org/bugs/?712) for handling filename with spaces. Doesn't that mean it's impossible  using --prefix to specify a directory name with space?

davidtgoldblatt:
Thinking on this more, the spaces thing spooks me a little bit. Is it reasonable to detect if the install prefix has a space in it, and error out if so? (Or, omit the uninstall target?)

I guess this applies for install, too, but that one feels safer, since we're not deleting things that we don't "own".

sangshuduo:
I agree with you it's dangerous to uninstall files if the install directory has space. But my point is it seems would not happen since gnu make will exit with error if a directory with space be specified. make uninstall only reuse the name from install phase.

Lapenkov:
Thank you for the effort. Close in favor of https://github.com/jemalloc/jemalloc/pull/2212, but I kept your authorship in the commit.

sangshuduo:
Happy to hear that from you. Thanks.

PR #2076, opened 2021-06-04
HPA: Fix typos with big performance implications.

davidtgoldblatt:
This fixes two simple but significant typos in the HPA batch_alloc_extra settings:
- The conf string parsing accidentally set a min value of PAGE for
  hpa_sec_batch_fill_extra; i.e. allocating 4096 extra pages every time we
  attempted to allocate a single page.  This puts us over the SEC flush limit,
  so we then immediately flush all but one of them (probably triggering
  purging).
- The HPA was using the default PAI batch alloc implementation, which meant it
  did not actually get any locking advantages.

This snuck by because I did all the performance testing without using the PAI
interface or config settings.  When I cleaned it up and put everything behind
nice interfaces, I only did correctness checks, and didn't try any performance
ones.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1482,8 +1482,8 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			    "hpa_sec_bytes_after_flush", PAGE, 0,
 			    CONF_CHECK_MIN, CONF_DONT_CHECK_MAX, true);
 			CONF_HANDLE_SIZE_T(opt_hpa_sec_opts.batch_fill_extra,
-			    "hpa_sec_batch_fill_extra", PAGE, 0, CONF_CHECK_MIN,
-			    CONF_DONT_CHECK_MAX, true);
+			    "hpa_sec_batch_fill_extra", 0, HUGEPAGE_PAGES,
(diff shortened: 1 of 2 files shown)


src/jemalloc.c
```diff
@@ -1482,8 +1482,8 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			    "hpa_sec_bytes_after_flush", PAGE, 0,
 			    CONF_CHECK_MIN, CONF_DONT_CHECK_MAX, true);
 			CONF_HANDLE_SIZE_T(opt_hpa_sec_opts.batch_fill_extra,
-			    "hpa_sec_batch_fill_extra", PAGE, 0, CONF_CHECK_MIN,
```
interwq, line 1485: This was unfortunate. Did we have flush stats for the hpa?

interwq (approved):

src/jemalloc.c
davidtgoldblatt: Yes, but I wasn't looking at them with custom malloc_confs after the interface changes.

This only triggers if you try non-default settings, which I didn't in perf-sensitive situations until recently (at which point, the programs failed to boot in a reasonable amount of time).

PR #2082, opened 2021-06-21
Add opt.experimental_infallible_new.

davidtgoldblatt:
This allows a guarantee that operator new never throws, without the heavyweight option of turning on xmalloc.

Fix the .gitignore rules to include test/integration/cpp while we're here.

### DIFF
--- a/test/integration/cpp/infallible_new_true.cpp
+++ b/test/integration/cpp/infallible_new_true.cpp
@@ -1,0 +1,61 @@
+#include <stdio.h>
+
+/*
+ * We can't test C++ in unit tests, and we can't change the safety check failure
+ * hook in integration tests.  So we check that we *actually* abort on failure,
+ * by forking and checking the child process exit code.
+ */
+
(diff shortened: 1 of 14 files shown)


test/integration/cpp/infallible_new_true.cpp
```diff
@@ -0,0 +1,61 @@
...
+		try {
+			/* Too big of an allocation to succeed. */
+			void *volatile ptr = ::operator new((size_t)-1);
+			(void)ptr;
+		} catch (...) {
+			/*
+			 * Swallow the exception; remember, we expect this to
+			 * fail via an abort within new, not because an
+			 * exception didn't get caught.
+			 */
```
interwq, line 43: Add a unreachable of sort here?

interwq: Actually, never mind since the parent checks the abort anyway.

interwq (approved):

PR #2089, opened 2021-07-09
Clean up background thread sleep computation

Lapenkov:
Isolate the computation of purge interval from background thread logic and
move into more suitable file.

### DIFF
--- a/include/jemalloc/internal/decay.h
+++ b/include/jemalloc/internal/decay.h
@@ -3,6 +3,8 @@
 
 #include "jemalloc/internal/smoothstep.h"
 
+#define DECAY_UNBOUNDED_TIME_TO_PURGE ((uint64_t)-1)
+
 /*
  * The decay_t computes the number of pages we should purge at any given time.
  * Page allocators inform a decay object when pages enter a decay-able state
@@ -145,5 +147,14 @@ void decay_reinit(decay_t *decay, nstime_t *cur_time, ssize_t decay_ms);
 /* Returns true if the epoch advanced and there are pages to purge. */
 bool decay_maybe_advance_epoch(decay_t *decay, nstime_t *new_time,
     size_t current_npages);
+
+/*
+ * Calculates wait time until at least npages_threshold pages should be purged.
+ *
+ * Returns number of nanoseconds or DECAY_UNBOUNDED_TIME_TO_PURGE in case of
+ * indefinite wait.
+ */
+uint64_t decay_ns_until_purge(decay_t *decay, size_t npages_current,
+    uint64_t npages_threshold);
 
 #endif /* JEMALLOC_INTERNAL_DECAY_H */
--- a/include/jemalloc/internal/decay.h
+++ b/include/jemalloc/internal/decay.h
@@ -149,7 +149,8 @@ bool decay_maybe_advance_epoch(decay_t *decay, nstime_t *new_time,
     size_t current_npages);
 
 /*
- * Calculates wait time until at least npages_threshold pages should be purged.
+ * Calculates wait time until a number of pages in the interval
+ * [0.5 * npages_threshold .. 1.5 * npages_threshold] should be purged.
  *
  * Returns number of nanoseconds or DECAY_UNBOUNDED_TIME_TO_PURGE in case of
  * indefinite wait.
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -104,134 +104,6 @@ set_current_thread_affinity(int cpu) {
 /* Minimal sleep interval 100 ms. */
 #define BACKGROUND_THREAD_MIN_INTERVAL_NS (BILLION / 10)
 
-static inline size_t
-decay_npurge_after_interval(decay_t *decay, size_t interval) {
-	size_t i;
-	uint64_t sum = 0;
-	for (i = 0; i < interval; i++) {
-		sum += decay->backlog[i] * h_steps[i];
-	}
-	for (; i < SMOOTHSTEP_NSTEPS; i++) {
-		sum += decay->backlog[i] * (h_steps[i] - h_steps[i - interval]);
-	}
-
-	return (size_t)(sum >> SMOOTHSTEP_BFP);
-}
-
-static uint64_t
-arena_decay_compute_purge_interval_impl(tsdn_t *tsdn, decay_t *decay,
-    ecache_t *ecache) {
-	if (malloc_mutex_trylock(tsdn, &decay->mtx)) {
-		/* Use minimal interval if decay is contended. */
-		return BACKGROUND_THREAD_MIN_INTERVAL_NS;
-	}
-
-	uint64_t interval;
-	ssize_t decay_time = decay_ms_read(decay);
-	if (decay_time <= 0) {
-		/* Purging is eagerly done or disabled currently. */
-		interval = BACKGROUND_THREAD_INDEFINITE_SLEEP;
-		goto label_done;
-	}
-
-	uint64_t decay_interval_ns = decay_epoch_duration_ns(decay);
-	assert(decay_interval_ns > 0);
-	size_t npages = ecache_npages_get(ecache);
-	if (npages == 0) {
-		unsigned i;
-		for (i = 0; i < SMOOTHSTEP_NSTEPS; i++) {
-			if (decay->backlog[i] > 0) {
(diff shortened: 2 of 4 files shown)


include/jemalloc/internal/decay.h
```diff
@@ -146,4 +146,14 @@ void decay_reinit(decay_t *decay, nstime_t *cur_time, ssize_t decay_ms);
 bool decay_maybe_advance_epoch(decay_t *decay, nstime_t *new_time,
     size_t current_npages);
 
+typedef struct purge_interval_opts_s purge_interval_opts_t;
```
davidtgoldblatt, line 149: This is the way we'd ordinarily write an options struct, although in this case I think we can pare down the details enough that we don't need one here.

(We also generally tend to reserve options structs for things like configuration settings rather than function arguments, although that's more of a case-by-case sort of thing).

include/jemalloc/internal/decay.h
```diff
@@ -146,4 +146,14 @@ void decay_reinit(decay_t *decay, nstime_t *cur_time, ssize_t decay_ms);
...
+struct purge_interval_opts_s {
+	uint64_t min_interval_ns;
```
davidtgoldblatt, line 151: I think that `min_interval_ns` probably belongs in the background thread code, since it's not really fundamental to the decay logic, but rather is just to avoid short background thread sleeps. A little cleaner would be to have the decay module *just* return the number of nanos there will be before there are pages that should be purged.

include/jemalloc/internal/decay.h
```diff
@@ -146,4 +146,14 @@ void decay_reinit(decay_t *decay, nstime_t *cur_time, ssize_t decay_ms);
...
+	uint64_t npages_threshold;
+	uint64_t indefinite_sleep;
```
davidtgoldblatt, line 153: Ditto here; we should just return `(uint64_t)-1` if there are no pages left to purge for the foreseeable future. (This is a theme in jemalloc; we often use -1 to stand in for infinity). It would also be reasonable to pull this into a macro in the header; e.g.

`#define DECAY_UNBOUNDED_TIME_TO_PURGE ((uint64_t)-1)`. The background thread code can see if it gets that value back, and if so convert it to whatever internal representation it likes (I don't think it will need to, but it code).

src/background_thread.c
```diff
@@ -285,6 +171,10 @@ static inline void
 background_work_sleep_once(tsdn_t *tsdn, background_thread_info_t *info, unsigned ind) {
 	uint64_t min_interval = BACKGROUND_THREAD_INDEFINITE_SLEEP;
 	unsigned narenas = narenas_total_get();
+	const purge_interval_opts_t purge_interval_opts = {
```
davidtgoldblatt, line 174: Doesn't matter if we're making these just raw arguments instead of an options struct (as suggested above), but just as a note: we can't use designated initializers in jemalloc, because of limited MSVC support.

include/jemalloc/internal/decay.h
```diff
@@ -146,4 +146,14 @@ void decay_reinit(decay_t *decay, nstime_t *cur_time, ssize_t decay_ms);
...
+};
+
+uint64_t arena_decay_compute_purge_interval(
```
davidtgoldblatt, line 156: Shouldn't have `arena` in the name any more; it's a property of "just" the decay object. A name like `decay_ns_until_purge` or something might be clearer as well.

src/decay.c
```diff
@@ -175,3 +175,123 @@ decay_maybe_advance_epoch(decay_t *decay, nstime_t *new_time,
...
+		    (h_steps[i] - h_steps[i - interval]);
+	}
+
+	return (size_t)(sum >> SMOOTHSTEP_BFP);
+}
+
+static uint64_t
+arena_decay_compute_purge_interval_impl(tsdn_t *tsdn, decay_t *decay,
+    ecache_t *ecache, const purge_interval_opts_t *opts) {
+	if (malloc_mutex_trylock(tsdn, &decay->mtx)) {
```
davidtgoldblatt, line 197: Belongs in the background thread code; the `decay_t` morally doesn't know about its mutex, and this is a backgronud thread locking optimization to begin with.

src/decay.c
```diff
@@ -175,3 +175,123 @@ decay_maybe_advance_epoch(decay_t *decay, nstime_t *new_time,
...
+label_done:
+	interval = (interval < opts->min_interval_ns) ? opts->min_interval_ns :
+	    interval;
+	malloc_mutex_unlock(tsdn, &decay->mtx);
+
+	return interval;
+}
+
+uint64_t
+arena_decay_compute_purge_interval(tsdn_t *tsdn, arena_t *arena,
```
davidtgoldblatt, line 281: This still belongs in the background thread code.

src/decay.c
```diff
@@ -175,3 +175,123 @@ decay_maybe_advance_epoch(decay_t *decay, nstime_t *new_time,
...
+	ssize_t decay_time = decay_ms_read(decay);
+	if (decay_time <= 0) {
+		/* Purging is eagerly done or disabled currently. */
+		interval = opts->indefinite_sleep;
+		goto label_done;
+	}
+
+	uint64_t decay_interval_ns = decay_epoch_duration_ns(decay);
+	assert(decay_interval_ns > 0);
+	size_t npages = ecache_npages_get(ecache);
```
davidtgoldblatt, line 212: This belongs in the background thread code; the background thread can get the number of pages in the ecache on its own, and then pass it to the decay function.

With that gone, we don't need the ecache parameter to these functions. And with the locking removed above, we don't need the tsdn_t parameter. So this becomes just a "regular" object that has no knowledge of its existence as part of a larger system; it's a small component that can be tested in isolation.

include/jemalloc/internal/decay.h
Lapenkov: Maybe it logically belongs in the background thread, but its removal affects the binary search in `arena_decay_compute_purge_interval_impl`. Lower bound of the binary search is calculated based on `min_interval_ns`. We could always assign lower bound to 2 though, but at the cost of additional binary search iterations. Is it worth it?

davidtgoldblatt: Hmm, that's a good point. My inclination is that the extra computation is fine if it gets us a slightly nicer API (this code isn't hot at all), but I don't feel so strongly about it that I'll argue about it if you disagree.

davidtgoldblatt:
Looks great.

One request: now that we've got a tamer API on this (and someone who gets the logic a little bit): can we improve the testing a little? Nothing big / time-intensive, but something to catch silly regressions would be nice.

Lapenkov:
> Looks great.
> 
> One request: now that we've got a tamer API on this (and someone who gets the logic a little bit): can we improve the testing a little? Nothing big / time-intensive, but something to catch silly regressions would be nice.

Yes, totally. I've been thinking about this as the next step. I'll add the tests and make this draft ready for review.

test/unit/decay.c
```diff
@@ -2,10 +2,63 @@
...
+	nstime_t curtime;
+	nstime_init(&curtime, 0);
+
+	ssize_t decay_ms = 1000;
+	assert_true(decay_ms_valid(decay_ms), "");
+
+	expect_false(decay_init(&decay, &curtime, decay_ms),
+	    "Should be able to initialize");
+	expect_zd_eq(decay_ms_read(&decay), decay_ms,
+	    "Correctly initializes decay ms");
```
davidtgoldblatt, line 18: Very nitty, but since this is also in part for codebase acclimation: A little more in keeping with our other tests is to say what went wrong rather than what the expectation is testing. I.e. here it might be "decay ms initialized incorrectly".

davidtgoldblatt (approved):
One nit, otherwise looks great.

interwq:
Test failures look real:
(AppVeyor)
test/unit/decay.c(82): error C2146: syntax error: missing ')' before identifier 'PRIu64'
test/unit/decay.c(82): error C2059: syntax error: ')'

(Travis, -m32 builds)
test/unit/decay.c:29:57: error: implicit conversion from 'unsigned long long' to
      'ssize_t' (aka 'int')

test/unit/decay.c
Lapenkov: Got it, will make sure all `expect`s follow this rule.

PR #2092, opened 2021-07-19
Hpa central redux

davidtgoldblatt:
This reintroduces the `hpa_central_t`, as the source of virtual address space for hpa_shards. For now, this flow is one-directional; eventually, we'll start allowing huge extents to flow between shards.


include/jemalloc/internal/hpa.h
```diff
@@ -7,6 +7,42 @@
...
+	 */
+	malloc_mutex_t grow_mtx;
+	/*
+	 * Either NULL (if empty), or some integer multiple of a
+	 * hugepage-aligned number of hugepages.  We carve them off one at a
+	 * time to satisfy new pageslab requests.
+	 *
+	 * Guarded by grow_mtx.
+	 */
+	void *eden;
```
interwq, line 30: This points to at most 1 region, right? What happens if the request size is greater than it, or the central handles requests up to hugepage size?

include/jemalloc/internal/hpa.h
```diff
@@ -7,6 +7,42 @@
...
+	size_t eden_len;
+	/* Source for metadata. */
+	base_t *base;
+	/* Number of grow operations done on this hpa_central_t. */
+	uint64_t age_counter;
+
+	/*
+	 * The HPA hooks for this shard.  Eventually, once we have the
+	 * hpa_central_t back, these should live there (since it doesn't make
```
interwq, line 39: nit: update the comments.

src/hpa.c
```diff
@@ -51,9 +51,125 @@ hpa_supported() {
...
+hpa_alloc_ps(tsdn_t *tsdn, hpa_central_t *central) {
+	return (hpdata_t *)base_alloc(tsdn, central->base, sizeof(hpdata_t),
+	    CACHELINE);
+}
+
+hpdata_t *
+hpa_central_extract(tsdn_t *tsdn, hpa_central_t *central, size_t size,
+    bool *oom) {
+	/* Don't yet support big allocations; these should get filtered out. */
+	assert(size < HUGEPAGE);
```
interwq, line 86: <= HUGEPAGE?

interwq (approved):

include/jemalloc/internal/hpa.h
interwq: Never mind it got answered.

davidtgoldblatt: Central will eventually handle > hugepage-sized requests; for now, it doesn't permit them.

include/jemalloc/internal/hpa.h
davidtgoldblatt: Done.

src/hpa.c
davidtgoldblatt: Done.

(Really, doesn't matter since the shard should reject these; but I think this is morally right).

PR #2096, opened 2021-07-23
Remove unless function prof_dump_ctl

Clcanny:
`prof_dump_ctl` is a static function, it seems that no one in src/ctl.c use it.

And it seems that`prof_dump_ctl` is the only caller of `prof_mdump`, should we delete it too?


Clcanny:
Sorry for that.

PR #2097, opened 2021-07-23
edata_cache: Allow unbounded fast caching.

davidtgoldblatt:
The edata_cache_small had a fill/flush heuristic.  In retrospect, this was a
premature optimization; more testing indicates that an unbounded cache is
effectively fine here, and moreover we spend a nontrivial amount of time doing
unnecessary filling/flushing.

As the HPA takes on a larger and larger fraction of all allocations, any
theoretical differences in allocation patterns should shrink.  The HPA is more
efficient with its metadata in general, so it still comes out ahead on metadata
usage anyways.

### DIFF
--- a/include/jemalloc/internal/edata_cache.h
+++ b/include/jemalloc/internal/edata_cache.h
@@ -3,15 +3,8 @@
 
 #include "jemalloc/internal/base.h"
 
-/*
- * Public for tests.  When we go to the fallback when the small cache is empty,
- * we grab up to 8 items (grabbing less only if the fallback is exhausted).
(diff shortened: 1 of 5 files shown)


davidtgoldblatt:
(Only last commit is relevant, obviously)

interwq (approved):

interwq:
Just noticed there's a Travis failure: https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/526973551

davidtgoldblatt:
Travis failure is spurious (well, flaky test on CI environment, which isn't quite "spurious" but also not caused by this PR).

PR #2101, opened 2021-08-04
Mutex: Make spin count configurable.

davidtgoldblatt:
Don't document it since we don't want to support this as a "real" setting, but
it's handy for testing.


Lapenkov (approved):
LGTM

interwq (approved):

PR #2103, opened 2021-08-05
Verify that HPA is used before calling its functions

Lapenkov:
This change eliminates the possibility of PA calling functions of uninitialized
HPA.


davidtgoldblatt (approved):
Good catch!

PR #2104, opened 2021-08-07
Add Cirrus CI testing matrix

interwq:
Contains 16 testing configs -- a mix of debug, prof, -m32 plus a few uncommon options.

This helps our FreeBSD coverage.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,6 +3,30 @@ env:
   ARCH: amd64
 
 task:
+  matrix:
+      env:
+        DEBUG_CONFIG: --enable-debug
+      env:
+        DEBUG_CONFIG: --disable-debug
+  matrix:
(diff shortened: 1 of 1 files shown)


interwq:
The 16 tests should finish within 40 mins -- each run has 2 CPUs and takes <10 minutes, and Cirrus allows max 8 concurrent CPUs for FreeBSD builds, so 4 builds can run in parallel.

interwq:
Added a one-liner to upgrade gcc -- 

The default compiler on Cirrus FreeBSD doesn't support -Werror or
-herror_on_warning, which makes our configure script fail to detect supported
attributes, and then results in thousands of warnings produced.  Upgrading gcc
fixed that.

davidtgoldblatt (approved):
woooo

PR #2106, opened 2021-08-10
[CI testing only] Cirrus CI: upgrade compiler

interwq:


interwq:
The default compiler on FreeBSD 12 is llvm10; tried llvm12 and gcc both worked fine. I'm gonna close this since it doesn't feel worth adding to our CI matrix, unless @davidtgoldblatt or @Lapenkov feels differently.

PR #2107, opened 2021-08-11
Allow PAI to calculate time until deferred work

Lapenkov:
Previously the calculation of sleep time between wakeups was implemented within
background_thread. This resulted in some parts of decay and hpa specific
logic mixing with background thread implementation. In this change, background
thread delegates this calculation to arena and it, in turn, delegates it to PAI.
The next step is to implement the actual calculation of time until deferred work
in HPA.


src/arena.c
```diff
@@ -462,6 +484,112 @@ arena_decay(tsdn_t *tsdn, arena_t *arena, bool is_background_thread, bool all) {
...
+
+/*
+ * Get time until next deferred work ought to happen. If there are multiple
+ * things that have been deferred, this function calculates the time until
+ * the soonest of those things.
+ */
+uint64_t
+arena_time_until_deferred(tsdn_t *tsdn, arena_t *arena) {
+	uint64_t soonest;
+	soonest = arena_ns_until_purge(tsdn,
```
davidtgoldblatt, line 508: I think this logic can be moved into the pa_shard_time_until_differed_word call.

Really, given how often the background thread reaches into PA internals anyways, I'm not sure this function needs to exist at all.

davidtgoldblatt (commented):
I think I like the overall structure of this modulo some nits, but @interwq, would you mind taking a look at the background thread parts?

include/jemalloc/internal/background_thread_structs.h
```diff
@@ -19,6 +19,9 @@
 #define BACKGROUND_THREAD_HPA_INTERVAL_MAX_UNINITIALIZED (-2)
 #define BACKGROUND_THREAD_HPA_INTERVAL_MAX_DEFAULT_WHEN_ENABLED 5000
 
+#define DEFERRED_ASAP UINT64_C(0)
```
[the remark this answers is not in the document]
Lapenkov: Sure, thanks for pointing out!

src/arena.c
Lapenkov: Makes sense, will do.

include/jemalloc/internal/background_thread_structs.h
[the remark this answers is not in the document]
interwq: Maybe `DEFERRED_MIN`?

src/arena.c
```diff
@@ -189,6 +198,20 @@ arena_stats_merge(tsdn_t *tsdn, arena_t *arena, unsigned *nthreads,
 	}
 }
 
+void
```
interwq, line 201: nit: static

src/pa.c
```diff
@@ -96,6 +96,11 @@ pa_shard_reset(tsdn_t *tsdn, pa_shard_t *shard) {
 	}
 }
 
+bool
```
interwq, line 99: Make this static? Or plan to use it in other places as well?

interwq (commented):
The structure makes sense to me too.

include/jemalloc/internal/background_thread_structs.h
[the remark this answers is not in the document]
Lapenkov: I wanted these constants to answer the question `when should deferred work happen?`. But you're right, in this case they should contain a time rather than an interval.

If I rename it to `BACKGROUND_THREAD_DEFERRED_MIN`, then I should also rename `DEFERRED_NEVER` to `BACKGROUND_THREAD_DEFERRED_MAX`.

src/pa.c
Lapenkov: No such plans yet, hence making it static.

src/arena.c
Lapenkov: Although, while doing this, I found that PA currently doesn't do anything decay-wise. The current implementation puts the responsibility of handling deferred decay on arena, hence having anything related to decay in PA seems a bit off without other changes in deferring decay.

For instance, arena has functions like `arena_dirty_decay_ms_default_get`, `arena_decay`, etc. In my diff, arena is supposed to know `ARENA_DEFERRED_PURGE_NPAGES_THRESHOLD`, and PA is oblivious to this.

Also, can you please elaborate on the frequency of background thread reaching into PA internals? I can see only calls to `pa_shard_set_deferral_allowed`, doesn't seem that often.

How I see it working within previously set boundaries: decay is deferred on arena level, but there can be possibly other stuff deferred by PAC/HPA which we want to add support for. Consequently, backgound_thread would ask each arena for time until deferred work, arena would check itself via `arena_ns_until_purge` and its pa_shard via `pa_shard_time_until_deferred_work`

Lapenkov:
Addressed some feedback and CI errors.

src/arena.c
davidtgoldblatt: The broad structure we're heading towards is that an arena is a set of bins, some state tracking for large allocations, and a handle to a page allocator shard (which is in turn tied to a pa_central_t). So that the arena's job becomes translating malloc requests to page-level ones. Because historically these were all sort of tied together (the arena was in charge of everything), we've got a lot of places where responsibility for some aspect of the page allocator doesn't really live in the PA module (like with decay, as you note). So e.g. when the background threads / arena access decay_t objects directly (rather than some PA-level wrapper), that's what I mean by "reaching into PA internals" (and similarly, when touching the PA shard ecaches directly).

So when we're adding new functionality that deals with things logically scoped to page management, we should break ties by putting that logic into pa.c or one of the PAI implementations instead of arena.c.

What's called in this diff arena_time_until_deferred should really be pac_time_until_deferred_work (with a little bit of reworking probably; stuff like BACKGROUND_THREAD_DEFERRED_MIN doesn't seem fundamental to the PAC decay logic, it's something that belongs in the caller rather than the callee). The PA shard should have a function that calls pai_time_until_deferred_work on each of its PAI implementations and returns the min (possibly with some adjustment).

Does that make sense? Sorry, I know it's not necessarily super easy to glean that direction from the current state of the codebase.

Lapenkov: Yes, this direction is clear.

Should I try to move other decay-related stuff from arena into PA?

davidtgoldblatt: I think that would be a valuable code-cleanup thing, but I wouldn't spend a *lot* of time on it. So if it seems easy to do I'd so go for it, but I wouldn't delay the functional improvements if you start getting bogged down for some unforeseen reason.

Lapenkov:
I think it needs more tests, I'll add them soon.

src/arena.c
```diff
@@ -462,6 +476,73 @@ arena_decay(tsdn_t *tsdn, arena_t *arena, bool is_background_thread, bool all) {
...
+/*
+ * Check if deferred work needs to be done sooner than planned.
+ * For decay we might want to wake up earlier because of an influx of dirty
+ * pages. Rather than waiting for previously estimated time, we proactively
+ * purge those pages.
+ * If background thread sleeps indefinitely, always wake up because some
+ * deferred work has been generated.
+ */
+static void
+arena_maybe_do_deferred_work(tsdn_t *tsdn, arena_t *arena, decay_t *decay,
```
Lapenkov, line 517: The current implementation has a drawback: imagine we deferred hugification for 5 seconds and the thread went to sleep for 5 seconds. Then after 1 second we realize, that we want to hugify much sooner, say, in 2 seconds since sleep began. Then the background thread won't be woken up until after the initial 5-second sleep.
Or if we defer hugification for 5 seconds and realize that we want to purge now, we'll still have to wait for 5 seconds.
The only scenario for HPA to wake up early is to wake up from indefinite sleep.

I'm not sure if this drawback is that critical, but I do plan to further change the mechanism to accommodate this scenario and push early wake up logic down to PA.

src/decay.c
```diff
@@ -3,6 +3,13 @@
 
 #include "jemalloc/internal/decay.h"
 
+const uint64_t h_steps[SMOOTHSTEP_NSTEPS] = {
```
davidtgoldblatt, line 6: Love it. I think this can be static now, right?

src/ctl.c
```diff
@@ -97,6 +97,7 @@ CTL_PROTO(opt_hpa)
 CTL_PROTO(opt_hpa_slab_max_alloc)
 CTL_PROTO(opt_hpa_hugification_threshold)
 CTL_PROTO(opt_hpa_hugify_delay_ms)
+CTL_PROTO(opt_hpa_min_purge_interval_ms)
```
davidtgoldblatt, line 100: Nit: We should add this to stats output, too. (Broadly, this is true for any new config option; it should go in jemalloc.c opt parsing, in the ctl opt namespace, and in the stats opt output).

src/arena.c
```diff
@@ -803,11 +816,15 @@ arena_destroy(tsd_t *tsd, arena_t *arena) {
...
+	bool deferred_work_generated;
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, 0);
 
 	edata_t *slab = pa_alloc(tsdn, &arena->pa_shard, bin_info->slab_size,
-	    PAGE, /* slab */ true, /* szind */ binind, /* zero */ false);
+	    PAGE, /* slab */ true, /* szind */ binind, /* zero */ false,
+	    &deferred_work_generated);
+
+	assert(deferred_work_generated == false);
```
davidtgoldblatt, line 827: Is this assert right? It seems like a slab allocation could trigger a (delayed) hugification, so that we would in fact generate deferred work. No?

(Also, just stylistically, for booleans we'd usually assert the bool computation directly rather than its value; i.e. `assert(!deferred_work_generated)` instead of `assert(deferred_work_generated == false)`).

src/background_thread.c
```diff
@@ -177,13 +176,21 @@ background_work_sleep_once(tsdn_t *tsdn, background_thread_info_t *info,
...
 	unsigned narenas = narenas_total_get();
+	bool slept_indefinitely = background_thread_indefinite_sleep(info);
 
 	for (unsigned i = ind; i < narenas; i += max_background_threads) {
 		arena_t *arena = arena_get(tsdn, i, false);
 		if (!arena) {
 			continue;
 		}
-		arena_do_deferred_work(tsdn, arena);
+		/*
```
davidtgoldblatt, line 186: I don't totally follow the reasoning here; why should we skip work depending on the conditions of the wakeup?

src/hpa.c
```diff
@@ -718,6 +726,11 @@ hpa_alloc_batch(tsdn_t *tsdn, pai_t *self, size_t size, size_t nallocs,
 	size_t nsuccess = hpa_alloc_batch_psset(tsdn, shard, size, nallocs,
 	    results);
 
+	malloc_mutex_lock(tsdn, &shard->mtx);
```
davidtgoldblatt, line 729: This ends up acquiring and releasing the shard mutex twice; instead, we should fold the deferred work computation into the other lock acquistion, by passing `deferred_work_generated` as a parameter to `hpa_alloc_batch_psset`.

src/hpa.c
```diff
@@ -840,21 +857,73 @@ hpa_dalloc_batch(tsdn_t *tsdn, pai_t *self, edata_list_active_t *list) {
...
-	hpa_dalloc_batch(tsdn, self, &dalloc_list);
+	hpa_dalloc_batch(tsdn, self, &dalloc_list, deferred_work_generated);
 }
 
+/*
+ * Calculate time until either purging or hugification ought to happen.
+ * Called by background threads.
+ */
 static uint64_t
 hpa_time_until_deferred_work(tsdn_t *tsdn, pai_t *self) {
```
davidtgoldblatt, line 881: Hmmm, this gets called after most operations, right? I would really rather avoid acquiring the lock a second time again if so.

Instead, we can update the "time until next deferred operation is ready" on every function call, and have a lock-free way of reading it in this function. (Probably using a seqlock, as in seq.h, since we don't have a guarantee that 64-bit atomics are available on every platform).

src/hpa.c
```diff
@@ -840,21 +857,73 @@ hpa_dalloc_batch(tsdn_t *tsdn, pai_t *self, edata_list_active_t *list) {
...
+
+	hpdata_t *to_hugify = psset_pick_hugify(&shard->psset);
+	if (to_hugify != NULL) {
+		nstime_t time_hugify_allowed =
+		    hpdata_time_hugify_allowed(to_hugify);
+		nstime_t nstime;
+		shard->central->hooks.curtime(&nstime);
+		nstime_subtract(&nstime, &time_hugify_allowed);
+		uint64_t since_last_hugify_ms = nstime_msec(&nstime);
+		/* If not enough time has passed since hugification was allowed,
```
davidtgoldblatt, line 895: Nit: comment style should have leading and trailing empty lines for multi-line comments.

src/large.c
```diff
@@ -88,8 +89,13 @@ large_ralloc_no_move_expand(tsdn_t *tsdn, edata_t *edata, size_t usize,
 	size_t new_size = usize + sz_large_pad;
 
 	szind_t szind = sz_size2index(usize);
+
+	bool deferred_work_generated;
 	bool err = pa_expand(tsdn, &arena->pa_shard, edata, old_size, new_size,
-	    szind, zero);
+	    szind, zero, &deferred_work_generated);
+
+	assert(deferred_work_generated == false);
```
davidtgoldblatt, line 97: Same here as in the slab alloc case; I'm not sure this will be true once we have "real" realloc support in the HPA.

davidtgoldblatt (commented):
The overall structure here looks good; some functional changes though.

src/arena.c
Lapenkov: Nice catch. Aside from removing the assertion, I need to call `arena_handle_deferred_work` to wake up from possible indefinite sleep.

src/background_thread.c
Lapenkov: We wake up a thread from indefinite sleep when we generate deferred work. At this point we've just generated the work and it doesn't need to be done instantly. We wake up the thread from the indefinite sleep and put it into another sleep, this time for the period of deferral.

Here is a hypothetical example:
1. Thread sleeps indefinitely
2. HPA generated deferred work in 5 sec
3. We wake up the thread, it checks that the work should be done in 5 sec
4. Thread sleeps for another 5 sec
5. In 5 sec the thread wakes up and does the work
6. Thread checks if there's still deferred work, then it sleeps for `time_until_deferred_work`. Otherwise, it sleeps indefinitely.

If the thread did all the work instantly, in case of HPA if would keep flipping between indefinite sleep and immediate work execution.

src/ctl.c
Lapenkov: Thanks for letting me know, will add it there.

src/decay.c
Lapenkov: Correct!

src/large.c
Lapenkov: Right. For some reason I thought I'd leave the assert here and after we have the support we'll change that. But really, why not implement it right now.

src/hpa.c
Lapenkov: It's called only by background thread after it's done with some background work or woke up from indefinite sleep. Do you think it will significantly contribute to the contention?

src/hpa.c
Lapenkov: I can do that, but then I'll have to propagate `deferred_work_generated` all the way to `hpa_try_alloc_batch_no_grow` and assign it after `hpa_shard_maybe_do_deferred_work`. It's because there's no single point in `hpa_try_alloc_batch_no_grow` that holds `shard->mtx`, gets called every time and doesn't potentially modify the outcome of `deferred_work_generated` later.

src/jemalloc.c
```diff
@@ -1447,10 +1447,15 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
...
 
-			CONF_HANDLE_SIZE_T(
+			CONF_HANDLE_UINT64_T(
 			    opt_hpa_opts.hugify_delay_ms, "hpa_hugify_delay_ms",
 			    0, 0, CONF_CHECK_MIN, CONF_DONT_CHECK_MAX, true);
 
+			CONF_HANDLE_UINT64_T(
+			    opt_hpa_opts.min_purge_interval_ms,
+			    "hpa_min_purge_interval_ms",
+			    0, 0, CONF_CHECK_MIN, CONF_DONT_CHECK_MAX, true);
```
interwq, line 1457: With the uint64 type, `CONF_CHECK_MIN` is not needed. Also the `clip` (last arg) should be false since it doesn't really have a range.

src/arena.c
```diff
@@ -471,6 +476,43 @@ arena_decay(tsdn_t *tsdn, arena_t *arena, bool is_background_thread, bool all) {
 	arena_decay_muzzy(tsdn, arena, is_background_thread, all);
 }
 
+static bool
+arena_should_decay_early(tsdn_t *tsdn, arena_t *arena, decay_t *decay,
+    background_thread_info_t *info, nstime_t *remaining_sleep,
+    size_t npages_new) {
+	if (malloc_mutex_trylock(tsdn, &decay->mtx)) {
```
interwq, line 483: Let's assert we hold `info->mtx` upon entry here.

test/unit/hpa_background_thread.c
```diff
@@ -65,6 +65,19 @@ set_background_thread_enabled(bool enabled) {
...
+wait_until_thread_is_enabled(unsigned arena_id) {
+	bool sleeping = false;
+	tsd_t* tsd = tsd_fetch();
+	do {
+		background_thread_info_t *info =
+		    background_thread_info_get(arena_id);
+		malloc_mutex_lock(tsd_tsdn(tsd), &info->mtx);
+		malloc_mutex_unlock(tsd_tsdn(tsd), &info->mtx);
+		sleeping = background_thread_indefinite_sleep(info);
+	} while (!sleeping);
```
interwq, line 78: nit: add a loop count and assert the count is <= 1M or so, to make sure the test doesn't get stuck.

src/background_thread.c
```diff
@@ -609,88 +573,23 @@ background_threads_disable(tsd_t *tsd) {
 	return false;
 }
 
-/* Check if we need to signal the background thread early. */
+bool
+background_thread_running(background_thread_info_t *info) {
```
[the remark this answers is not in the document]
Lapenkov: Agree, using `running` is incorrect with current implementation.
It's a bit confusing that thread state is set to `started` even before it's created, but I'm not sure it can be simplified without bloating the code with unnecessary states.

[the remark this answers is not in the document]
Lapenkov: I wish I could do `background_thread_started`, but it's already an existing symbol from `enum background_thread_state_t`. What about `background_thread_is_started`? Or I can remove that whatsoever and revert to `==`, but my gut tells me that thread state should be internal to `background_thread`.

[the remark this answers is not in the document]
interwq: `_is_started` sounds good to me.

re: why it's set to `started` before it's created -- the context is we want to avoid holding any locks when calling pthread_create, while we also want make sure only one thread is created for the specific index. Therefore marking it as started while holding the index lock. Like you said it can use another "creating" state, but the value is a bit limited given that it's a one-time transition (and usually very quick).

src/jemalloc.c
Lapenkov: Makes sense. I basically reused these parameters from `hpa_hugify_delay_ms` above, which should also be corrected accordingly.

src/background_thread.c
davidtgoldblatt: Ahh, so just to confirm my understanding: when a background thread's sleep time gets shortened, it wakes up twice: the first time to sleep with a new parameter, and the second time to actually do the work?

src/hpa.c
davidtgoldblatt: I think that's worthwhile; this is a hot mutex, we should be willing to do some gymnastics to avoid redundant acquisitions.

src/hpa.c
davidtgoldblatt: OK, I think that makes sense; that case should be rare.

src/background_thread.c
Lapenkov: Almost. It wakes up twice only when sleep time gets shortened and it is indefinite. Otherwise it doesn't wake up twice, which is a limitation of current implementation.

src/hpa.c
Lapenkov: Fixed - I put it into `hpa_try_alloc_batch_no_grow`

include/jemalloc/internal/background_thread_externs.h
```diff
@@ -13,6 +12,9 @@ extern background_thread_info_t *background_thread_info;
 bool background_thread_create(tsd_t *tsd, unsigned arena_ind);
 bool background_threads_enable(tsd_t *tsd);
 bool background_threads_disable(tsd_t *tsd);
+bool background_thread_is_started(background_thread_info_t* info);
+void background_thread_wakeup_early(background_thread_info_t *info,
+    nstime_t *remaining_sleep);
 void background_thread_interval_check(tsdn_t *tsdn, arena_t *arena,
```
interwq, line 18: Let's remove this one too.

src/background_thread.c
```diff
@@ -609,88 +573,24 @@ background_threads_disable(tsd_t *tsd) {
...
-		 * future epoch.
-		 */
+background_thread_wakeup_early(background_thread_info_t *info,
+    nstime_t *remaining_sleep) {
+	/*
+	 * This is an optimization to increase batching. At this point
+	 * we know that background thread wakes up soon, so the time to cache
+	 * the just freed memory is bounded and low.
+	 */
+	if (remaining_sleep && nstime_ns(remaining_sleep) <
```
interwq, line 589: nit: `if (remaining_sleep != NULL ...`. We try always making the type conversion explicit.

src/pa.c
```diff
@@ -226,15 +232,71 @@ pa_decay_ms_get(pa_shard_t *shard, extent_state_t state) {
...
+	    &shard->pac.decay_muzzy,
+	    ecache_npages_get(&shard->pac.ecache_muzzy));
+	if (muzzy < time) {
+		time = muzzy;
+		if (time == BACKGROUND_THREAD_DEFERRED_MIN) {
+			return time;
+		}
+	}
+
+	uint64_t pac = pai_time_until_deferred_work(tsdn, &shard->pac.pai);
```
interwq, line 286: Is this simply returning MAX right now? How about moving the above two into it here?

test/unit/hpa_background_thread.c
```diff
@@ -65,6 +65,23 @@ set_background_thread_enabled(bool enabled) {
...
+
+	bool sleeping = false;
+	int iterations = 0;
+	do {
+		background_thread_info_t *info =
+		    background_thread_info_get(arena_id);
+		malloc_mutex_lock(tsd_tsdn(tsd), &info->mtx);
+		malloc_mutex_unlock(tsd_tsdn(tsd), &info->mtx);
+		sleeping = background_thread_indefinite_sleep(info);
+		assert_d_lt(iterations, (int)1e6,
```
interwq, line 80: It looks the `1e` syntax is part of the floating point literals? To be safe I'd say stick with `UINT64_C(1000000)` or so.

interwq (approved):
LGTM just a few nits.

Let's wait a day before landing though, in case @davidtgoldblatt wants to take another look.

src/pa.c
Lapenkov: Makes sense, moved decay-based deferral into PAC.

PR #2108, opened 2021-08-11
Makefile.in: make sure doc generated before install

Mingli-Yu:
There is a race between the doc generation and the doc installation,
so make the install depend on the build for doc to fix the error occurs
sometimes as below:
 | TOPDIR/tmp-glibc/hosttools/install: cannot stat 'doc/jemalloc.3': No such file or directory
 | make: *** [Makefile:513: install_doc_man] Error 1

Signed-off-by: Mingli Yu <mingli.yu@windriver.com>

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -582,14 +582,14 @@ install_lib: install_lib_static
 endif
 install_lib: install_lib_pc
 
-install_doc_html:
+install_doc_html: build_doc_html
 	$(INSTALL) -d $(DATADIR)/doc/jemalloc$(install_suffix)
 	@for d in $(DOCS_HTML); do \
 	echo "$(INSTALL) -m 644 $$d $(DATADIR)/doc/jemalloc$(install_suffix)"; \
 	$(INSTALL) -m 644 $$d $(DATADIR)/doc/jemalloc$(install_suffix); \
 done
 
-install_doc_man:
+install_doc_man: build_doc_man
 	$(INSTALL) -d $(MANDIR)/man3
 	@for d in $(DOCS_MAN3); do \
 	echo "$(INSTALL) -m 644 $$d $(MANDIR)/man3"; \


interwq:
I'm not sure I follow where the issue is: the `build_doc` part is done by `install_doc`

> install_doc: build_doc install_doc_html install_doc_man

Where `build_doc` does both `build_doc_html` and `build_doc_man`. @Mingli-Yu do you mean you see the `install_doc_*` parts above started before the `build_doc` step finishes?

Mingli-Yu:
@interwq 
There is indeed a race as below.
install_doc: build_doc install_doc_html install_doc_man
The build_doc, install_doc_html and install_doc_man have no dependencies and can run concurrently. And sometimes the install_doc_* parts will execute before build_doc finishes and  you can also reproduce it you run the build many times.

BTW, it proves that the above patch solves my problem.

@davidtgoldblatt 
For the xsltproc issue,  it can always generate the doc even without xsltproc installed and doesn't introduce any install error. The difference is using xsltproc to generate the real doc when xlstproc installed and echo "Missing xsltproc.  Doc not built."  >  $(objroot)doc/jemalloc.3 to generate a fake manual as below logic.
ifneq ($(XSLROOT),)
        $(XSLTPROC) -o $@ $(objroot)doc/manpages.xsl $<
else
ifeq ($(wildcard $(DOCS_MAN3)),)
        @echo "Missing xsltproc.  Doc not built." > $@
endif
        @echo "Missing xsltproc.  "$@" not (re)built."
endif

interwq:
Thanks for the confirmation @Mingli-Yu 

In this case, can you please update the commit a tiny bit: remove the `build_doc` from `install_doc:` since the build step is now redundant (updating the existing commit is fine; since we don't allow squash at merge time).

interwq:
Made the tweak and committed in e5062e9f under your name @Mingli-Yu. Thanks for the patch!

PR #2111, opened 2021-08-13
Fix warnings when compiled with clang

Lapenkov:
When clang sees an unknown warning option, unlike gcc it doesn't fail the build
with error. It issues a warning. Hence JE_CFLAGS_ADD with warning options that
didnt't exist in clang would still mark those options as available. This led to
several warnings when built with clang or "gcc" on OSX. This change fixes those
warnings by simply making clang fail builds with non-existent warning options.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -244,6 +244,7 @@ if test "x$GCC" = "xyes" ; then
       AC_DEFINE_UNQUOTED([JEMALLOC_HAS_RESTRICT])
     fi
   fi
+  JE_CFLAGS_ADD([-Werror=unknown-warning-option])
   JE_CFLAGS_ADD([-Wall])
   JE_CFLAGS_ADD([-Wextra])
   JE_CFLAGS_ADD([-Wshorten-64-to-32])


Lapenkov:
You can see the warnings this change is supposed to fix in this build:
https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/531139568#L518

davidtgoldblatt:
I think I'm misreading something; if previously we had warnings because of unknown warning flags, and we now treat an unknown warning flag as an error, why doesn't that imply that we start seeing build errors?

Lapenkov:
Yes, there's some misunderstanding.

First, let's see how `JE_CFLAGS_ADD(flag)` works. If adding `flag` to `CFLAGS` doesn't result in build errors, we add it. Otherwise, `flag` is ignored.

Now, the problem was clang **not failing** build with `-Wunknown-flag`. It just warns you that the warning flag is unknown. On the contrary, gcc fails the build.

How it worked before the change:
We try to add compilation flag `-Wno-missing-attributes`. Clang doesn't support this, but still successfully builds a trivial program with this flag (throwing a warning), so `JE_CFLAGS_ADD` appends it to `CFLAGS`. Later, when the project is built, this warning is repeated many times.

How it works after the change:
We first successfully add compilation flag `-Werror=unknown-warning-option`. After that, when we do `JE_CFLAGS_ADD([-Wno-missing-attributes])`, clang fails to build a trivial program (due to `-Werror=unknown-warning-option`), hence `JE_CFLAGS_ADD` doesn't add `-Wno-missing-attributes` to `CFLAGS`.

In other words, this change fixes the mechanism that detects which compilation flags are supported, rather than affecting the actual warnings we get after building the project.

davidtgoldblatt:
Ahhh, I see. Makes sense!

davidtgoldblatt (approved):

interwq (approved):
Thanks to @Lapenkov who had to explain it to me offline as well. It took me a while to realize this is fixing the option detection itself.

PR #2115, opened 2021-08-20
HPA: Correctly calculate retained pages

Lapenkov:
Retained pages are those which haven't been touched and are unbacked from OS
perspective. For a pageslab their number should equal "total pages in slab"
minus "touched pages".


davidtgoldblatt (approved):
LGTM assuming tests come back green.

Good catch.

PR #2116, opened 2021-08-23
Not assume initialized arena in the default alloc hook.

interwq:
A valid use case, is to invoke the default alloc hook in a customized hook
arena.  Specifically, this allows the default alloc hook to used during
arenas.create.


test/unit/arena_reset.c
```diff
@@ -255,6 +255,21 @@ TEST_BEGIN(test_arena_destroy_hooks_default) {
...
+	extent_hooks_t *default_hooks;
+	expect_d_eq(mallctl("arena.0.extent_hooks", (void *)&default_hooks,
+	    &sz, NULL, 0), 0, "Unexpected mallctlnametomib() failure");
+
+	/* Default impl; but wrapped as "customized". */
+	extent_hooks_t new_hooks = *default_hooks;
+	extent_hooks_t *hook = &new_hooks;
+	sz = sizeof(unsigned);
+	expect_d_eq(mallctl("arenas.create", (void *)&arena_ind, &sz,
+	    (void *)&hook, sizeof(void *)), 0,
```
interwq, line 270: This test case is what's being fixed.

davidtgoldblatt (approved):
Nit: clean up the commit message a little maybe? E.g. "Not" -> "Don't" in the title, and maybe clarify the "read hooks out of a default arena, then create using those hooks" use case I.e. this intends to be a temporary expansion of API surface.

PR #2118, opened 2021-08-30
add experimental.arenas_create_ext mallctl

### LINKED ISSUE #2095: Question: Extent_hooks, retain memory and abandoned_vm
Hi,

I got a question regarding abandoned_vm metrics extent hooks and retain memory.

I am using custom extent_hooks (https://github.com/jemalloc/jemalloc/issues/1329#issuecomment-427748503) - my goal is to achieve a custom limited heap that is based on allocation to the file. I would like to retain memory (opt-out from deallocation) to reuse the once mapped memory:

```cpp
static void *custom_extent_alloc(extent_hooks_t *extent_hooks,
                        void *new_addr,
                        size_t size,
                        size_t alignment,
                        bool *zero,
                        bool *commit,
                        unsigned arena_ind)
{
    void *addr = NULL;
    if (new_addr != NULL) {
        /* not supported */
        goto exit;
    }

    if (g_current_file_size + size > SIZE_LIMIT) {
        goto exit;
    }

    if ((errno = posix_fallocate(g_fd, (off_t)g_current_file_size, (off_t)size)) != 0) {
        goto exit;
    }

    if ((addr = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, g_fd,
                      (off_t)g_current_file_size)) != MAP_FAILED) {
        g_current_file_size += size;
    }

    if (addr != MAP_FAILED) {
        *zero = true;
        *commit = true;
        if ((uintptr_t)addr & (alignment - 1)) {
            std::cerr << "Not alligned address " << size <<" alignment " <<alignment << std::endl;
            abort();
        }
    } else {
        addr = NULL;
    }
exit:
    return addr;
}

static bool custom_extent_dalloc(extent_hooks_t *extent_hooks,
                        void *addr,
                        size_t size,
                        bool committed,
                        unsigned arena_ind)
{
    return true;
}

static bool custom_extent_commit(extent_hooks_t *extent_hooks,
                        void *addr,
                        size_t size,
                        size_t offset,
                        size_t length,
                        unsigned arena_ind)
{
    /* do nothing - report success */
    return false;
}

static bool custom_extent_decommit(extent_hooks_t *extent_hooks,
                          void *addr,
                          size_t size,
                          size_t offset,
                          size_t length,
                          unsigned arena_ind)
{
    /* do nothing - report failure (opt-out) */
    return true;
}

static bool custom_extent_purge(extent_hooks_t *extent_hooks,
                       void *addr,
                       size_t size,
                       size_t offset,
                       size_t length,
                       unsigned arena_ind)
{
    /* do nothing - report failure (opt-out) */
    return true;
}

static bool custom_extent_split(extent_hooks_t *extent_hooks,
                       void *addr,
                       size_t size,
                       size_t size_a,
                       size_t size_b,
                       bool committed,
                       unsigned arena_ind)
{
    /* do nothing - report success */
    return false;
}

static bool custom_extent_merge(extent_hooks_t *extent_hooks,
                       void *addr_a,
                       size_t size_a,
                       void *addr_b,
                       size_t size_b,
                       bool committed,
                       unsigned arena_ind)
{
    /* do nothing - report success */
    return false;
}

static void custom_extent_destroy(extent_hooks_t *extent_hooks,
                         void *addr,
                         size_t size,
                         bool committed,
                         unsigned arena_ind)
{
    if (munmap(addr, size) == -1) {
        std::cerr << "munmap failed!";
    }
}

static extent_hooks_t custom_extent_hooks = {
    .alloc = custom_extent_alloc,
    .dalloc = custom_extent_dalloc,
    .destroy = custom_extent_destroy,
    .commit = custom_extent_commit,
    .decommit = custom_extent_decommit,
    .purge_lazy = custom_extent_purge,
    .purge_forced = NULL,
    .split = custom_extent_split,
    .merge = custom_extent_merge,
};

static void create_hooks(void) {
    int err;
    size_t unsigned_size = sizeof(unsigned);
    //create new arena
    err = je_mallctl("arenas.create", (void *)&g_arena_index, &unsigned_size, NULL, 0);
    if(err) {
        std::cerr << "Could not create arena";
        exit(1);
    }
    char cmd[64];
    const size_t retainGrowLimit = 2 * 1024 * 1024; // 2 MB
    snprintf(cmd, sizeof(cmd), "arena.%u.retain_grow_limit", g_arena_index);
    err = je_mallctl(cmd, NULL, NULL, (void *)&retainGrowLimit,
                        sizeof(retainGrowLimit));
    if (err) {
        std::cerr << "retain_grow_limit failed";
        exit(1);
    }

    //setup extent_hooks for newly created arena
    extent_hooks_t *new_hooks = &custom_extent_hooks;
    snprintf(cmd, sizeof(cmd), "arena.%u.extent_hooks", g_arena_index);
    err = je_mallctl(cmd, NULL, NULL, (void *)&new_hooks, sizeof(extent_hooks_t *));
    if(err) {
        std::cerr << "Could not setup extent_hooks";
        exit(1);
    }
}

static void create_temporary_file(const char *dir) {

    const size_t size = sysconf(_SC_PAGESIZE);
    static char template_name[] = "/test_file.XXXXXX";
    int oerrno;
    int dir_len = strlen(dir);

    if (dir_len > PATH_MAX) {
        std::cerr << "Could not create temporary file: too long path.";
        exit(1);
    }

    char fullname[dir_len + sizeof(template_name)];
    (void)strcpy(fullname, dir);
    (void)strcat(fullname, template_name);

    sigset_t set, oldset;
    sigfillset(&set);
    (void)sigprocmask(SIG_BLOCK, &set, &oldset);

    if ((g_fd = mkstemp(fullname)) < 0) {
        std::cerr << "Could not create temporary file: errno=" <<errno;
        exit(1);
    }

    (void)unlink(fullname);
    (void)sigprocmask(SIG_SETMASK, &oldset, NULL);
}

```

Regarding the abandoned_vm:

I perform the following scenario:
- create the file designated to be the heap
- create one arena and adding extent hooks
- I limit the maximum mapped memory on the extent_alloc_t hook
- I perform allocation to the file based heap with a specified range of sizes
- When mallocx to dedicated arena return NULL - I free random allocation and try to allocate once more
- Finally, I observe that abandoned_vm metrics increase - see the jemalloc stats below

I added the attachment with a full example and reproduction and jemalloc stats output.
[main.cpp](https://github.com/jemalloc/jemalloc/files/6861698/main.txt)
[jemalloc_stats.txt](https://github.com/jemalloc/jemalloc/files/6861697/jemalloc_stats.txt)

As you can see abandoned_vm metrics are different from zero which points to a "virtual memory leak", can I somehow prevent this leak?

pbalcer:
This mallctl accepts an arena_config_t structure which
can be used to customize the behavior of the arena.
Right now it contains extent_hooks and a new option,
metadata_use_hooks, which controls whether the extent
hooks are also used for metadata allocation.

The medata_use_hooks option has two main use cases:

1. In heterogeneous memory systems, to avoid metadata
being placed on potentially slower memory.

2. Avoiding virtual memory from being leaked as a result
of metadata allocation failure originating in an extent hook.

Closes #2095 

---
This is more of an RFC than anything else. I'm not entirely sure if adding this option is the right thing to do to solve the underlying problem. We want to use jemalloc extent hooks to allocate from a file-based DAX* mapping, which initially appeared to be working just fine. Still, we found physically deallocating pages to be very expensive. This is expected because deallocation has to punch a hole in the file and deallocate the underlying file system extent. So we've tried to avoid deallocating pages in the hooks implementation, but this led to runaway problems with virtual address space leaks in near OOM conditions (abandoned_vm) - which we then traced back to jemalloc's inability to allocate additional metadata when splitting extents. These address space leaks (from `extents_abandon_vm` function) are especially problematic for this type of file-based mapping because they only call the purge hooks (to avoid leaking physical space), which are difficult to implement differently than through `madvise(..., MADV_REMOVE)` which again has to punch a hole in the file...
Avoiding the use of hooks for internal metadata made it so that address space isn't leaked unless the OS runs out of main memory. And, as a nice bonus, using main memory (which is very likely the lowest-latency memory in a system) for metadata is ought to have a positive impact on overall system performance. 

I'm also not confident about the API - I considered multiple different solutions like adding a "use_for_metadata" to the extent hooks structure or adding a flag indicating metadata allocation to alloc/dealloc hook functions, but this would not be backward-compatible. I also briefly considered adding a global option, but configuring this on a per-arena basis is useful.
Instead, I added a single independent arena option that users can set after configuring the extent hooks. This is fully backward-compatible and unintrusive. The downside is that some base allocations might happen using the hooks when an arena is first created with the extent hooks defined - but the reverse is already the case when *not* providing the extent hooks during arena creation (from `arena.<i>.extent_hooks ` docs: `The functions must be capable of operating on all extant extents associated with arena <i>, usually by passing unknown extents to the replaced functions`), so I'm not sure if this is a problem.

\* DAX - direct access, this is used to "directly access" heterogeneous memory on Linux

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -55,6 +55,11 @@ static unsigned nbins_total;
 
 static unsigned huge_arena_ind;
 
+const arena_config_t arena_config_default = {
+	/* .extent_hooks = */ (extent_hooks_t *)&ehooks_default_extent_hooks,
+	/* .metadata_use_hooks = */ true,
+};
+
 /******************************************************************************/
 /*
  * Function prototypes for static functions that are referenced prior to
@@ -1427,7 +1432,6 @@ arena_set_extent_hooks(tsd_t *tsd, arena_t *arena,
 	return ret;
 }
 
-
 dss_prec_t
 arena_dss_prec_get(arena_t *arena) {
 	return (dss_prec_t)atomic_load_u(&arena->dss_prec, ATOMIC_ACQUIRE);
@@ -1494,7 +1498,7 @@ arena_nthreads_dec(arena_t *arena, bool internal) {
 }
 
 arena_t *
-arena_new(tsdn_t *tsdn, unsigned ind, extent_hooks_t *extent_hooks) {
+arena_new(tsdn_t *tsdn, unsigned ind, const arena_config_t *config) {
 	arena_t *arena;
 	base_t *base;
 	unsigned i;
@@ -1502,7 +1506,8 @@ arena_new(tsdn_t *tsdn, unsigned ind, extent_hooks_t *extent_hooks) {
 	if (ind == 0) {
 		base = b0get();
 	} else {
-		base = base_new(tsdn, ind, extent_hooks);
+		base = base_new(tsdn, ind, config->extent_hooks,
+		    config->metadata_use_hooks);
 		if (base == NULL) {
 			return NULL;
 		}
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -313,6 +313,7 @@ INDEX_PROTO(experimental_arenas_i)
 CTL_PROTO(experimental_prof_recent_alloc_max)
 CTL_PROTO(experimental_prof_recent_alloc_dump)
 CTL_PROTO(experimental_batch_alloc)
+CTL_PROTO(experimental_arenas_create_ext)
 
 #define MUTEX_STATS_CTL_PROTO_GEN(n)					\
 CTL_PROTO(stats_##n##_num_ops)						\
@@ -867,6 +868,7 @@ static const ctl_named_node_t experimental_node[] = {
 	{NAME("hooks"),		CHILD(named, experimental_hooks)},
 	{NAME("utilization"),	CHILD(named, experimental_utilization)},
 	{NAME("arenas"),	CHILD(indexed, experimental_arenas)},
+	{NAME("arenas_create_ext"),	CTL(experimental_arenas_create_ext)},
 	{NAME("prof_recent"),	CHILD(named, experimental_prof_recent)},
 	{NAME("batch_alloc"),	CTL(experimental_batch_alloc)},
 	{NAME("thread"),	CHILD(named, experimental_thread)}
@@ -1239,7 +1241,7 @@ ctl_arena_refresh(tsdn_t *tsdn, arena_t *arena, ctl_arena_t *ctl_sdarena,
 }
 
 static unsigned
-ctl_arena_init(tsd_t *tsd, extent_hooks_t *extent_hooks) {
+ctl_arena_init(tsd_t *tsd, const arena_config_t *config) {
 	unsigned arena_ind;
 	ctl_arena_t *ctl_arena;
 
@@ -1257,7 +1259,7 @@ ctl_arena_init(tsd_t *tsd, extent_hooks_t *extent_hooks) {
 	}
 
 	/* Initialize new arena. */
-	if (arena_init(tsd_tsdn(tsd), arena_ind, extent_hooks) == NULL) {
+	if (arena_init(tsd_tsdn(tsd), arena_ind, config) == NULL) {
 		return UINT_MAX;
 	}
 
@@ -2874,8 +2876,11 @@ arena_i_extent_hooks_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
 				extent_hooks_t *new_extent_hooks
 				    JEMALLOC_CC_SILENCE_INIT(NULL);
 				WRITE(new_extent_hooks, extent_hooks_t *);
+				arena_config_t config = arena_config_default;
+				config.extent_hooks = new_extent_hooks;
+
 				arena = arena_init(tsd_tsdn(tsd), arena_ind,
-				    new_extent_hooks);
+				    &config);
 				if (arena == NULL) {
 					ret = EFAULT;
 					goto label_return;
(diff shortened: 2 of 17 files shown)


davidtgoldblatt:
We've got a lot of sympathy for this style of use case (it's the longest-standing thing on our list of stuff to change the next time we do a breaking release).

The thing that makes us nervous about this as a temporary fix is that first extent, which results in hooked-extents getting used partially for metadata. The alternative we liked better was having a new arena creation mallctl, which could take some creation parameters. Since we know we want to change this in the next release anyways, it should live in the experimental mallctl namespace (and not get top-level documentation in the man page).

So I think we like the direction here, but with some tweaks:
- A new mallctl, say, `experimental.arenas.create2` or `experimental.arenas.create_with_args` or something
- An arguments struct with a hidden definition (i.e. not exposed in a public header; you need to reach into the jemalloc internals to use this functionality). That struct should include both the hooks pointer as well as a `metadata_use_hooks` bool.

Does that make sense? Is it still useful for you if we're not willing to make a firm API commitment yet?

pbalcer:
Thanks for your reply - it makes perfect sense. I agree that creating a new arena creation mallctl would be the best approach for now. And yes, it's still useful for our purpose.

src/ctl.c
```diff
@@ -867,6 +868,7 @@ static const ctl_named_node_t experimental_node[] = {
 	{NAME("hooks"),		CHILD(named, experimental_hooks)},
 	{NAME("utilization"),	CHILD(named, experimental_utilization)},
 	{NAME("arenas"),	CHILD(indexed, experimental_arenas)},
+	{NAME("arenas_create_ext"),	CTL(experimental_arenas_create_ext)},
```
pbalcer, line 871: I've decided to create a new mid-level `arenas_create_ext` leaf CTL because the `arenas` is an indexed node - this is different than the `arenas.`. Looks like an oversight, but I'm not sure - I'm open to suggestions.

src/arena.c
```diff
@@ -55,6 +55,11 @@ static unsigned nbins_total;
 
 static unsigned huge_arena_ind;
 
+const arena_config_t arena_config_default = {
+	.extent_hooks = (extent_hooks_t *)&ehooks_default_extent_hooks,
```
davidtgoldblatt, line 59: Nit: we can't use designated initializers because they don't compile on some older versions of MSVC that we still try to support. (Unfortunately, this isn't well tested in CI).

src/ctl.c
davidtgoldblatt: No, I think that makes sense. Ideally we would have named the existing `experimental.arenas` `experimental.arena` instead, but no need to fix that here (it would be difficult for some of the other experimental functionality experiments).

test/unit/mallctl.c
```diff
@@ -1223,6 +1223,27 @@ TEST_BEGIN(test_thread_activity_callback) {
...
+TEST_BEGIN(test_arenas_create_ext) {
+	unsigned arena, arena1;
+	void *ptr;
+	size_t sz = sizeof(unsigned);
+
+	arena_config_t config;
+	config.extent_hooks = NULL;
+	config.metadata_use_hooks = false;
+
+	expect_d_eq(mallctl("experimental.arenas_create_ext",
```
davidtgoldblatt, line 1235: Sorry for the extra pain, but could you also add an extra test that passes in non-default hooks as well (and, perhaps, asserts that they don't get called, or at least don't get called until a real user alloc happens)? I don't think you missed anywhere in this diff, but I can imagine it being easy to accidentally use the extent hooks for metadata in future refactors.

davidtgoldblatt (commented):
Basically looks good, just a few minor nits.

Thanks for the work, this solves a number of annoying use cases others have hit as well.

test/unit/mallctl.c
pbalcer: No worries, you are right. I was just being lazy :) I've added a couple of new tests to verify that there are no base allocations performed during arena create ctl. I also moved those tests from `unit/mallctl.c` to `integration/extent.c` - it seemed like a more appropriate place, but I'm not sure.

src/arena.c
```diff
@@ -1494,15 +1498,20 @@ arena_nthreads_dec(arena_t *arena, bool internal) {
...
+arena_new(tsdn_t *tsdn, unsigned ind, const arena_config_t *config) {
 	arena_t *arena;
 	base_t *base;
 	unsigned i;
 
 	if (ind == 0) {
 		base = b0get();
 	} else {
-		base = base_new(tsdn, ind, extent_hooks);
+		extent_hooks_t *extent_hooks = config->extent_hooks ?
```
interwq, line 1509: This has one side effect: if the custom hooks passed in by the user is really NULL, then it's going to be using the default hook with this change, instead of crashing like previously.

For the existing `arenas.create`, it's a behavior change that might cover application bugs. How about we remove this check, and simply use config->extent_hooks here? I'm guessing you are trying to support the use case of setting `metadata_use_hook` w/o setting a custom hook (by using a NULL hook value in the config)? I wouldn't worry about that since it's practically the same as the all default.

pbalcer: That's right - I was trying to cover the case where the config is set with NULL hooks. I've now removed this check and the corresponding test.

Thanks for the review.

interwq:
LGTM. Thanks for the patch @pbalcer!

PR #2121, opened 2021-09-01
Reject --enable-prof-libunwind without --enable-prof

Lapenkov:
Prior to the change you could specify --enable-prof-libunwind without
--enable-prof which would do effectively nothing. This was confusing as I
expected --enable-prof-libunwind to act like --enable-prof, but use libunwind.


configure.ac
```diff
@@ -1282,6 +1282,10 @@ AC_ARG_ENABLE([prof-libunwind],
 else
   enable_prof_libunwind="1"
 fi
+
+if test "x$enable_prof" = "x0" ; then
```
interwq, line 1286: This should go into the `enable_prof_libunwind="1"` branch? Right now it applies to both cases.

Lapenkov: You're right, I forgot about the `--enable-prof-libnwind=no` case. Thanks!

interwq (approved):

PR #2124, opened 2021-09-08
Guard prof related mallctl with opt_prof.

interwq:
The prof initialization is done only when opt_prof is true.  This change makes
sure the prof_* mallctls only have limited read access (i.e. no access to prof
internals) when opt_prof is false.

In addition, initialize the global prof mutexes even if opt_prof is false.  This
makes sure the mutex stats are set properly.

### DIFF
--- a/test/unit/mallctl.c
+++ b/test/unit/mallctl.c
@@ -914,7 +914,7 @@ TEST_BEGIN(test_prof_active) {
 	old = true;
 	expect_d_eq(mallctl("prof.active", &old, &len, &active, len), ENOENT,
 	    "Setting prof_active to true should fail when opt_prof is off");
-	expect_true(old, "old valud should not be touched when mallctl fails");
+	expect_true(old, "old value should not be touched when mallctl fails");
 	active = false;
 	expect_d_eq(mallctl("prof.active", NULL, NULL, &active, len), 0,
 	    "Setting prof_active to false should succeed when opt_prof is off");
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -2338,7 +2338,7 @@ thread_prof_name_ctl(tsd_t *tsd, const size_t *mib,
     size_t newlen) {
 	int ret;
 
-	if (!config_prof) {
+	if (!config_prof || !opt_prof) {
 		return ENOENT;
 	}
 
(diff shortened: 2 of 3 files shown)


interwq:
This has a side effect: the prof related clt such as prof_active becomes unreadable when opt_prof is false. The change is necessary because, even reading the values, we still access the (uninitialized) mutexes w/o opt_prof.

Alternatively, if we want to avoid the behavior change, we may do a "partial init" on prof_boot when opt_prof==false. It feels hacky, but I'm indeed open to that route as well.

test/unit/mallctl.c
```diff
@@ -916,11 +916,11 @@ TEST_BEGIN(test_prof_active) {
...
-	expect_d_eq(mallctl("prof.active", NULL, NULL, &active, len), 0,
-	    "Setting prof_active to false should succeed when opt_prof is off");
-	expect_d_eq(mallctl("prof.active", &old, &len, &active, len), 0,
-	    "Setting prof_active to false should succeed when opt_prof is off");
-	expect_false(old, "prof_active should be false when opt_prof is off");
+	expect_d_eq(mallctl("prof.active", NULL, NULL, &active, len), ENOENT,
+	    "Setting prof_active to false should fail when opt_prof is off");
+	expect_d_eq(mallctl("prof.active", &old, &len, &active, len), ENOENT,
+	    "Setting prof_active to false should fail when opt_prof is off");
+	expect_true(old, "old valud should not be touched when mallctl fails");
```
davidtgoldblatt, line 923: Nit: valud -> value

davidtgoldblatt (approved):
One nit

test/unit/mallctl.c
interwq: Good catch! Will fix.

interwq:
Discussed offline about this. The current commit should have no notable behavior change -- the unit tests around the prof_* mallctls don't need to be changed anymore.

When opt_prof==false, a few prof_* mallctls are allowed read access (specifically those with a boolean type, such as prof.active). Also makes sure that the profiling section in malloc_stats is still included when opt_prof==false.

@davidtgoldblatt could use your review again.

Lapenkov (approved):

PR #2129, opened 2021-09-22
Redefine functions with test hooks only for tests

Lapenkov:
Android build has issues with these defines, this will allow the build to
succeed if it doesn't need to build the tests.

### DIFF
--- a/include/jemalloc/internal/test_hooks.h
+++ b/include/jemalloc/internal/test_hooks.h
@@ -4,16 +4,21 @@
 extern JEMALLOC_EXPORT void (*test_hooks_arena_new_hook)();
 extern JEMALLOC_EXPORT void (*test_hooks_libc_hook)();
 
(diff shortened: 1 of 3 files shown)


include/jemalloc/internal/test_hooks.h
```diff
@@ -6,14 +6,16 @@ extern JEMALLOC_EXPORT void (*test_hooks_libc_hook)();
 
 #define JEMALLOC_HOOK(fn, hook) ((void)(hook != NULL && (hook(), 0)), fn)
```
interwq, line 7: How about we define this to be simply `fn` in the non testing case, and also rename the macro to `JEMALLOC_TEST_HOOK`? Might be a bit easier to read.

Lapenkov: Agree, these are good suggestions. Addressed.

PR #2140, opened 2021-10-06
Initialize deferred_work_generated

Lapenkov:
As the code evolves, some code paths that have previously assigned
`deferred_work_generated` may cease being reached. This would leave the value
uninitialized. This change initializes the value for safety.


interwq (changes requested):
> One thing I'd suggest is clearing out the redundant initializations elsewhere

That's a good point. Let's do that @Lapenkov. It's reasonable for the pa / pai to assume that `generated` starts with false, and only set to true when necessary.

PR #2143, opened 2021-10-14
Separate retained region for guarded extents

Lapenkov:
To avoid interference with unguarded extent coalescing, guarded extents are
allocated out of a separate region. Also, now instead of two guard pages, only
one is needed per guarded extent.


test/unit/san.c
```diff
@@ -72,20 +79,15 @@ TEST_BEGIN(test_guarded_large) {
 	}
 
 	/* Verify the pages are not continuous, i.e. separated by guards. */
-	uintptr_t min_diff = (uintptr_t)-1;
```
Lapenkov, line 75: I figured this is unnecessary, since if all diffs are less than expected value, then min must be less too.

src/extent.c
```diff
@@ -1024,6 +1119,12 @@ extent_dalloc_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 			return;
 		}
 		extent_reregister(tsdn, pac, edata);
+		/* If failed to dalloc, restore the guard pages */
+		if (edata_guarded_get(edata)) {
+			san_guard_pages(tsdn, ehooks, edata, pac->emap,
+			    /* left */ true, /* right */ true,
```
Lapenkov, line 1125: I'm not a big fan of having many bool args, but I don't know if creating helper functions for all used combinations (like `san_guard_pages_right`, etc.) is better.

src/extent.c
```diff
@@ -91,9 +91,49 @@ ecache_alloc(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, ecache_t *ecache,
...
 }
 
+
+/*
+ * Allocates an extent as a separate mapping. On the other hand,
+ * ecacle_alloc_retained can allocate multiple extents out of a single retained
+ * mapping if opt_retain is enabled.
+ */
+static edata_t*
+extent_alloc_as_mapping(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
```
Lapenkov, line 105: Not sure about the name here.

interwq (commented):
Discussed offline. Some of the comments might not be applicable anymore.

PR #2148, opened 2021-10-20
Optimize away the tsd_fast() check on free fastpath.

interwq:


src/jemalloc.c
```diff
@@ -2925,15 +2932,6 @@ bool free_fastpath(void *ptr, size_t size, bool size_hint) {
 		}
 		assert(alloc_ctx.szind != SC_NSIZES);
 	} else {
-		/*
-		 * The size hinted fastpath does not involve rtree lookup, thus
-		 * can tolerate an uninitialized tsd.  This allows the tsd_fast
```
interwq, line 2930: In other words, this PR makes the free fastpath tolerate unintialized / nonfast tsd, so that it can avoid the tsd_fast() check the same way, using the next_event branch.

interwq:
Confirmed that these two instructions at the beginning of `free` are gone with this PR:

    │     if (unlikely(tsd == NULL || !tsd_fast(tsd))) {
      cmpb $0x0,%fs:0x330(%rcx)
    . jne  c0

interwq:
Just realized there's an issue on freeing null pointers. The uninitialized tsd case isn't right. Will update and reopen.

PR #2151, opened 2021-10-26
Implement bump allocator for sanitized allocations

Lapenkov:
The new simple allocator is designed to allocate slabs for small guarded allocations.

This PR doesn't integrate this allocator into PAC, only introduces the allocator itself and its interface to address the high-level changes before integrating.


src/pac.c
```diff
@@ -105,47 +108,65 @@ pac_may_have_muzzy(pac_t *pac) {
...
+		    NULL, size, alignment, zero, true);
+	}
+	if (edata == NULL) {
+		edata = ecache_alloc(tsdn, pac, ehooks, &pac->ecache_retained,
+		    NULL, size, alignment, zero, true);
+
+		if (edata != NULL && config_prof) {
+			extent_gdump_add(tsdn, edata);
+		}
+	}
```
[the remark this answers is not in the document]
Lapenkov: Discussed offline. Totally agree, that `dirty->muzzy->retained` duplication sucks and must be avoided. I just went to redo it and remembered the original reasoning behind these proposed design changes.

The original approach dragged `guarded` flag deep into `extent.c` with some inevitable assumptions about guarded control flow upstream. For instance, `ecache_alloc_grow` won't actually grow for guarded, because it knows it was called solely to perform retained lookup, without growing intentions. This, while involves less duplication, makes it very implicit for the uninitiated: "Why doesn't `ecache_alloc_grow` grow retained for guarded allocations?". To answer this question, readers should go up the stack to PAC and find the answer there.

My approach, on the other hand, tired to localize most changes in the control flow part of this - PAC. But yes, that came at the cost of duplication, which I too don't want to pay.

src/pac.c
```diff
@@ -105,47 +108,65 @@ pac_may_have_muzzy(pac_t *pac) {
...
+	if (size >= SC_LARGE_MINCLASS || !opt_retain) {
+		edata = pac_alloc_unguarded(tsdn, pac,
```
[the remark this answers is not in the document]
Lapenkov: Discussed offline - `bool slab` flag will be added to PAI to pass this data down to `PAC`.

src/pac.c
```diff
@@ -233,27 +250,6 @@ pac_dalloc_impl(tsdn_t *tsdn, pai_t *self, edata_t *edata,
 	pac_t *pac = (pac_t *)self;
 	ehooks_t *ehooks = pac_ehooks_get(pac);
 
-	if (edata_guarded_get(edata)) {
```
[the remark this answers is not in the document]
Lapenkov: Discussed offline. That was removed to potentially reuse large guarded allocs, but that's an extremely rare scenario which is not worth it. Well, basically the removed comment mentions it already :)

src/extent.c
```diff
@@ -89,43 +89,50 @@ ecache_alloc(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, ecache_t *ecache,
...
 }
 
+
+/*
+ * Allocates an extent as a separate mapping. On the other hand,
+ * ecacle_alloc_retained can allocate multiple extents out of a single retained
+ * mapping if opt_retain is enabled.
+ */
+static edata_t*
+extent_alloc_as_mapping(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
```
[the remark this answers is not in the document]
Lapenkov: Returning it back to `alloc_grow`

src/pac.c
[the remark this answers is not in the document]
davidtgoldblatt: Can you go into some more detail? It was a lot of work for the page-level allocator not to know about the purpose for which its pages is used, and it's fairly abstraction breaking to do so. We should try to do something more specific if at all possible.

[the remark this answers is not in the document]
interwq: Agreed that we shouldn't pass down another param really mean "is_slab". What we want is actually a hint, more like `bool frequent_reuse_expected`, which doesn't have to be exactly slab (but we may simply use the slab as the criteria). The PA will then use it to determine if a guarded extent should be cached (prevents coalescing but speeds up reuse), or getting rid of the guards (mprotect needed).

The other thought is, I think it's also worth trying out some hints with the HPA later, e.g. to group slabs into the same hugepage.

An alternative to the additional hint param, is to have a size threshold within the PA, to indicate that, e.g. sizes <= MAX_SLAB_SIZE are frequently reused. It may get a bit awkward when the dynamic slab sizes are used though (then a large range of sizes may get covered).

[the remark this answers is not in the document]
davidtgoldblatt: Feels compelling to me, thanks.

include/jemalloc/internal/extent.h
```diff
@@ -125,6 +125,9 @@ extent_can_acquire_neighbor(edata_t *edata, rtree_contents_t contents,
 			return false;
 		}
 	}
+	if (edata_guarded_get(edata) != edata_guarded_get(neighbor)) {
+		return false;
+	}
```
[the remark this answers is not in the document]
Lapenkov: I was thinking we would potentially try to acquire the head of the exponential mapping from san_bump, but that one is marked as `EXTENT_IS_HEAD`, so it would be filtered out by of the previous checks.

Will change this to an assertion

src/san.c
```diff
@@ -10,16 +10,58 @@
...
+	*addr = (uintptr_t)edata_base_get(edata);
+	if (right) {
+		*guard2 = *addr + size;
+	} else {
+		*guard2 = 0;
+	}
+
+	if (left) {
+		*guard1 = *addr - SAN_PAGE_GUARD;
+		*addr = *guard1;
```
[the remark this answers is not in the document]
Lapenkov: Could you please elaborate on what this protects from?

src/pages.c
```diff
@@ -365,33 +365,54 @@ pages_decommit(void *addr, size_t size) {
...
+	assert(head != NULL || tail != NULL);
+	assert(head == NULL || tail == NULL ||
+	    (uintptr_t)head < (uintptr_t)tail);
 #ifdef JEMALLOC_HAVE_MPROTECT
+	if (head != NULL && tail != NULL) {
+
+	}
 	size_t range = (uintptr_t)tail - (uintptr_t)head + PAGE;
-	if (range <= SC_LARGE_MINCLASS) {
+	if (head != NULL && tail != NULL && range <= SC_LARGE_MINCLASS) {
```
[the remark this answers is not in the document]
Lapenkov: Could you elaborate please?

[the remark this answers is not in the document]
interwq: I'm thinking like:

    bool head_and_tail = (head != NULL) && (tail != NULL);
    size_t range = head_and_tail ? (uintptr_t)tail - (uintptr_t)head + PAGE : -1;
    bool ranged_mprotect = head_and_tail && range <= SC_LARGE_MINCLASS;

    if (ranged_mprotect) {
        mprotect(head, range, PROT_READ | PROT_WRITE);
    } else ...

src/san.c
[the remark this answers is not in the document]
interwq: This is basically saying that, *addr != PAGE, i.e. it's a left-guarded extent, the min addr would needs to be PAGE * 2.

Also because the guard1 = 0 was an acceptable returning result to the caller, and the caller doesn't have to check since NULL is a valid input to ehooks_unguard, which happens right after. So it's asserting if left is true, *guard1 is non zero.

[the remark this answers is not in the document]
Lapenkov: Clever, I'll add it.

src/san_bump.c
```diff
@@ -0,0 +1,121 @@
...
+		if (extent_commit_wrapper(tsdn, ehooks, edata, 0,
+		    edata_size_get(edata), true)) {
+			extent_record(tsdn, pac, ehooks, &pac->ecache_retained,
+			    edata);
+			goto label_err;
+		}
+		edata_committed_set(edata, true);
+	}
+
+	if (zero && !edata_zeroed_get(edata)) {
```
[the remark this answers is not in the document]
interwq: Just realized, don't we also need a `exp_grow_size_commit` for san_bump_alloc_grow (maybe rename it to san_bump_grow since it doesn't do the actual alloc part)?

For the commit error handling, I think we can leave the grow_retained part unchanged (i.e. not using the new utility created for this). That function can handle an arbitrarily large allocation, so the chance of commit failure is real. Avoiding exp_grow_size_commit in such failure case also has its point.

While the bump alloc here is for small alloc only, it's probably fine that we don't even use the exp_grow.

src/extent.c
```diff
@@ -124,8 +125,8 @@ ecache_alloc_grow(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, ecache_t *ecache,
 		}
 		void *new_addr = (expand_edata == NULL) ? NULL :
 		    edata_past_get(expand_edata);
-		edata = extent_alloc_wrapper(tsdn, pac, ehooks, new_addr,
-		    size, alignment, zero, &commit);
+		return extent_alloc_wrapper(tsdn, pac, ehooks,
+		    new_addr, size, alignment, zero, &commit);
```
interwq, line 129: Is this intentional?

src/extent.c
```diff
@@ -416,7 +412,18 @@ extent_recycle_extract(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
 		unsigned lg_max_fit = ecache->delay_coalesce
 		    ? (unsigned)opt_lg_extent_max_active_fit : SC_PTR_BITS;
-		edata = eset_fit(eset, size, alignment, exact_only, lg_max_fit);
+
+		/*
+		 * If split and merge are not allowed (Windows w/o retain), try
+		 * exact fit only.
+		 *
+		 * For simplicity purposes, splitting guarded extents is not
+		 * supported. Hence, we do only exact fit for guarded
```
interwq, line 421: nit: double space after periods in comment

src/extent.c
```diff
@@ -1008,11 +1021,11 @@ extent_dalloc_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
 	/* Avoid calling the default extent_dalloc unless have to. */
 	if (!ehooks_dalloc_will_fail(ehooks)) {
-		/* Restore guard pages for dalloc / unmap. */
+		/* Remove guard pages for dalloc / unmap. */
 		if (edata_guarded_get(edata)) {
 			assert(ehooks_are_default(ehooks));
-			san_unguard_pages_two_sided(tsdn, ehooks, edata,
-			    pac->emap);
+			san_unguard_pages(tsdn, ehooks, edata, pac->emap,
+			    /* left */ true, /* right */ true);
```
interwq, line 1028: You probably intend to keep the _two_sided()?

src/extent.c
```diff
@@ -1171,6 +1184,7 @@ extent_split_impl(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 	    /* slab */ false, SC_NSIZES, edata_sn_get(edata),
 	    edata_state_get(edata), edata_zeroed_get(edata),
 	    edata_committed_get(edata), EXTENT_PAI_PAC, EXTENT_NOT_HEAD);
+	edata_guarded_set(trail, edata_guarded_get(edata));
```
interwq, line 1187: nit: edata_init sets guarded to false 
https://github.com/jemalloc/jemalloc/blob/deb8e62a837b6dd303128a544501a7dc9677e47a/include/jemalloc/internal/edata.h#L611

so this isn't needed.

src/pac.c
```diff
@@ -189,8 +201,8 @@ pac_expand_impl(tsdn_t *tsdn, pai_t *self, edata_t *edata, size_t old_size,
 	}
 	if (trail == NULL) {
 		trail = ecache_alloc_grow(tsdn, pac, ehooks,
-		    &pac->ecache_retained, edata, expand_amount, PAGE,
-		    zero, /* guarded */ false);
+		    &pac->ecache_retained, edata, expand_amount, PAGE, zero,
+		    false);
```
interwq, line 205: nit: keep the `/* guarded */`

src/pac.c
```diff
@@ -158,12 +167,15 @@ pac_alloc_impl(tsdn_t *tsdn, pai_t *self, size_t size, size_t alignment,
 	pac_t *pac = (pac_t *)self;
 	ehooks_t *ehooks = pac_ehooks_get(pac);
 
-	edata_t *edata = pac_alloc_real(tsdn, pac, ehooks, size, alignment,
-	    zero, guarded);
+	edata_t *edata = NULL;
+	if (!guarded || frequent_reuse) {
```
interwq, line 171: Is it true that the `frequent_reuse` condition is an optimization? i.e. if it's a guarded && large request, going down pac_alloc_real will return NULL gracefully, because we never put large back into the guarded ecache (and won't attempt to grow retained)?

If so, some comments here can certainly help.

interwq (commented):
It's coming together pretty nicely!

src/extent.c
Lapenkov: No, changing back.

src/pac.c
interwq: Sounds good. Maybe it's slightly easier to read with

    bool try_cached_guarded = guarded && frequent_reuse;
    if (!try_cached_guarded) { ...

interwq: Well, maybe we simply don't do this optimization? By default, slab sizes can be 5 and 7 pages (which are large sizes), and that mean even if frequent_reuse == false, we may still find cached & guarded extents. We can also say that we don't really want those cached extents to be used that way, but it doesn't hurt much.

Lapenkov:
Pushed mostly for CI, will be available for review on Monday.

What has changed:
- Addressed nits
- No more exponential growth for san_bump. Now it's constant 4MB mappings.
- Reuse `extent_alloc_wrapper` in san_bump
- Unmap trailing `curr_reg` via `extent_destroy_wrapper`
- Committing and zeroing is done outside the locked section.

What's left:
- Create a separate function for committing + zeroing a new extent.
- Perform self-review to make sure no obvious mistakes/unintended changes are remaining.

src/pac.c
Lapenkov: I've added a comment. Let me know if it makes it clear what's going on.

Lapenkov:
If CI succeeds, then ready for review.

src/san_bump.c
```diff
@@ -0,0 +1,115 @@
...
+		if (err) {
+			goto label_err;
+		}
+	}
+	assert(guarded_size <= edata_size_get(sba->curr_reg));
+	size_t trail_size = edata_size_get(sba->curr_reg) - guarded_size;
+
+	edata_t* edata;
+	if (trail_size != 0) {
+		edata_t* curr_reg_trail = extent_split_wrapper(tsdn, pac,
```
interwq, line 57: We need err handling here; split can fail, e.g. from allocating new edata.

src/san_bump.c
```diff
@@ -0,0 +1,115 @@
...
+}
+
+static bool
+san_bump_grow_locked(tsdn_t *tsdn, san_bump_alloc_t *sba, pac_t *pac,
+    ehooks_t *ehooks) {
+	malloc_mutex_assert_owner(tsdn, &sba->mtx);
+
+	bool committed = false, zeroed = false;
+	sba->curr_reg = extent_alloc_wrapper(tsdn, pac, ehooks, NULL,
+	    SBA_RETAINED_ALLOC_SIZE, PAGE, zeroed, &committed,
```
interwq, line 100: re: ALLOC_SIZE, we still need to pass the input requested size, and use the larger of the two. Slab sizes can be dynamic and >4MB

src/extent.c
```diff
@@ -1001,6 +969,37 @@ extent_dalloc_wrapper_try(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
+	    &zero, commit);
+	if (addr == NULL) {
+		edata_cache_put(tsdn, pac->edata_cache, edata);
+		return NULL;
+	}
+	edata_init(edata, ecache_ind_get(&pac->ecache_dirty), addr,
+	    size, /* slab */ false, SC_NSIZES, extent_sn_next(pac),
+	    extent_state_active, zero, *commit, EXTENT_PAI_PAC,
+	    opt_retain ? EXTENT_IS_HEAD : EXTENT_NOT_HEAD);
+	bool gdump_add = !growing_retained;
```
interwq, line 1000: Is it true that only this part is changed of `extent_alloc_wrapper()`? Let's add some comments to explain why no gdump_add when growing_retained.

include/jemalloc/internal/extent.h
```diff
@@ -30,22 +30,29 @@ void ecache_dalloc(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
 bool extent_decommit_wrapper(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata,
     size_t offset, size_t length);
 bool extent_purge_lazy_wrapper(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata,
     size_t offset, size_t length);
 bool extent_purge_forced_wrapper(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata,
     size_t offset, size_t length);
 edata_t *extent_split_wrapper(tsdn_t *tsdn, pac_t *pac,
-    ehooks_t *ehooks, edata_t *edata, size_t size_a, size_t size_b);
+    ehooks_t *ehooks, edata_t *edata, size_t size_a, size_t size_b,
+    bool holding_core_locks);
```
interwq, line 55: One thing makes worry is the newly exposed details in the interfaces here, i.e. growing_retained, holding_core_locks and gdump_add which were previously all internal to extent.c.  I do like that the wrapper functions did not require any of these internal details, and it did take quite some refactoring effort to achieve that.

In a way it says that the san_bump.c is very related to extent.c. I was hoping adding the new `_commit_zero()` facility would help avoid some of the exposure here. But it looks that wasn't enough. Perhaps we should consider merging them? WDYT @Lapenkov?

Lapenkov: On the one hand, I don't like it exposed either, on the other hand the role of this argument is to simply ensure correct locking. I don't think it's a deal breaker to expose this sanity checking mechanism to the outside. I can make `locked` and `unlocked` versions of most of these functions, which would essentially achieve the same goal, without exposing `growing_retained` or `holding_core_locks`.

The only exception is `extent_alloc_wrapper`, where `gdump_add` depens on `growing_retained`. This relation is pretty simple if you think about it - when you allocate a retained extent, you don't add retained memory to gdump. When you allocate an extent as a mapping solely dedicated to it, you instantly add that mapping's size to gdump.

I don't like merging `extent` and `san_bump_alloc`, because `san_bump_alloc` is a stateful entity that uses `extent.h` functionality to manage extents. `extent` is a collection of functions to operate extents. I see clear distinction here, and with `extent.c` already being 1300 lines, feels like merging them will complicate things. I would rather put extra effort into exposing a more user-friendly interface through `extent.h` than merging it with `san_bump`.

src/extent.c
Lapenkov: This is not the only change, also `witness_assert_depth_to_rank` now depends on `growing_retained` (ternary operator added).

Added the comment.

src/san_bump.c
```diff
@@ -0,0 +1,109 @@
...
+}
+
+static bool
+san_bump_grow_locked(tsdn_t *tsdn, san_bump_alloc_t *sba, pac_t *pac,
+    ehooks_t *ehooks, size_t size) {
+	malloc_mutex_assert_owner(tsdn, &sba->mtx);
+
+	bool committed = false, zeroed = false;
+	size_t alloc_size = size > SBA_RETAINED_ALLOC_SIZE ? size :
+	    SBA_RETAINED_ALLOC_SIZE;
```
interwq, line 101: nit: assert alloc_size & PAGE_MASK == 0

include/jemalloc/internal/extent.h
interwq: I'm convinced we should keep it separate. But still feel we should try tightening the interface changes. The biggest issue I have is the gdump_add(); usually that's done when registering the extent. It looks we register the whole extent w/o gdump_add when calling alloc_wrapper -- where do we register the new extent after the split?

A couple of other thoughts:
1) for alloc_wrapper(), will it work better if we pass in an additional `size_t over_reserve`, instead of the `bool growing_retained`? In other words, it's like having a new `alloc_grow_wrapper()` which handles the alloc + reserving together.
2) it looks `extent_commit_wrapper` doesn't have to be changed? Probably because we have the new commit_zero() now.
3) the new use of extent_record() if always in the failure case of commit_zero(), should we integrate extent_record into the failure path of commit_zero()?

Appreciate the effort and patience on tidying everything. We are getting very close now.

Lapenkov: Correct, we first register the whole extent allocated by `alloc_wrapper`, but don't add it to `gdump`. Later we cut a piece of that extent and, if no errors have occurred, add it to gdump (san_bump.c:88). `extent_grow_retained` works similarly.

1. Can you elaborate? By alloc + reserving you mean in should allocate a retained extent and cut a smaller piece of it?
2. Yes, reverting it back. Thanks.
3. I don't think so, if we add error handling there, it will be easy to miss it when reading the call site, unless we somehow reflect it in the function's name. And this function's name isn't ideal already :)

Lapenkov: Discussed offline and decided to evaluate if gdump management could be moved into `commit_zero` (and renaming the function).

I've checked, and it doesn't really fit in the picture. `commit_zero` is called in `extent_recycle`, which can recycle an extent from any ecache, not just retained. Hence, gdump management is done separately. So, either I do not reuse `commit_zero`, and add gdump management there, or I keep the code as is. To me, keeping as is seems better.

test/unit/san_bump.c
```diff
@@ -0,0 +1,111 @@
...
+	san_bump_alloc_t sba;
+	san_bump_alloc_init(&sba);
+
+	unsigned arena_ind = do_arena_create(0, 0);
+	assert_u_ne(arena_ind, UINT_MAX, "Failed to create an arena");
+
+	arena_t *arena = arena_get(tsdn, arena_ind, false);
+	pac_t *pac = &arena->pa_shard.pac;
+
+	size_t alloc_size = 1024 * 1024 * 12; /* 12 MB */
```
interwq, line 86: nit: maybe `SBA_RETAINED_ALLOC_SIZE * 2`, in case we later increase that threshold

include/jemalloc/internal/extent.h
```diff
@@ -55,6 +55,8 @@ edata_t *extent_split_wrapper(tsdn_t *tsdn, pac_t *pac,
     bool holding_core_locks);
 bool extent_merge_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
     edata_t *a, edata_t *b);
+bool extent_commit_zero(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata,
+    bool growing_retained, bool commit, bool zero);
```
interwq, line 59: nit: make growing_retained the last arg

interwq (commented):
Looks great! Only a few nits left.

@Lapenkov as discussed, given that the scope of change is fairly large now, let's try production runs before landing?

src/pac.c
```diff
@@ -132,18 +135,24 @@ pac_alloc_real(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, size_t size,
...
-	/* Alloc a non-guarded extent first.*/
-	edata_t *edata = pac_alloc_real(tsdn, pac, ehooks, size_with_guards,
-	    /* alignment */ PAGE, zero, /* guarded */ false);
-	if (edata != NULL) {
-		/* Add guards around it. */
-		assert(edata_size_get(edata) == size_with_guards);
-		san_guard_pages(tsdn, ehooks, edata, pac->emap, true, true,
-		    true);
+	edata_t *edata;
+	if (maps_coalesce && opt_retain && frequent_reuse) {
```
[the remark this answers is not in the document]
Lapenkov: It's destroy plus deallocating a trailing piece of retained guarded region. I'll follow your suggestion, thanks.

interwq (approved):
@Lapenkov as mentioned, the only left is updating the msvc source files.

PR #2152, opened 2021-11-02
San: Unguard guarded slabs during arena destruction

Lapenkov:
When opt_retain is on, slab extents remain guarded in all states, even
retained. This works well if arena is never destroyed, because we
anticipate those slabs will be eventually reused. But if the arena is
destroyed, the slabs must be unguarded to prevent leaking guard pages.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -1057,12 +1057,14 @@ extent_destroy_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
     edata_t *edata) {
 	assert(edata_base_get(edata) != NULL);
 	assert(edata_size_get(edata) != 0);
+	assert(edata_state_get(edata) == extent_state_retained);
+	assert(emap_edata_is_acquired(tsdn, pac->emap, edata));
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, 0);
 
-	/* Deregister first to avoid a race with other allocating threads. */
-	extent_deregister(tsdn, pac, edata);
-
+	if (edata_guarded_get(edata)) {
+		unguard_pages_pre_destroy(tsdn, ehooks, edata, pac->emap);
+	}
 	edata_addr_set(edata, edata_base_get(edata));
 
 	/* Try to destroy; silently fail otherwise. */
--- a/src/guard.c
+++ b/src/guard.c
@@ -32,10 +32,16 @@ guard_pages(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata, emap_t *emap) {
 	/* The new boundary will be registered on the pa_alloc path. */
 }
 
-void
-unguard_pages(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata, emap_t *emap) {
+static void
+unguard_pages_impl(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata, emap_t *emap,
+    bool reg_emap) {
 	/* Remove the inner boundary which no longer exists. */
-	emap_deregister_boundary(tsdn, emap, edata);
+	if (reg_emap) {
+		assert(edata_state_get(edata) == extent_state_active);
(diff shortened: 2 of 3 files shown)


src/extent.c
```diff
@@ -1057,11 +1057,19 @@ extent_destroy_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
-	extent_deregister(tsdn, pac, edata);
+	/*
+	 * Deregister to avoid a race with other allocating threads. If guarded,
+	 * then unguard without re-registering in the emap.
+	 */
+	if (edata_guarded_get(edata)) {
+		unguard_pages(tsdn, ehooks, edata, pac->emap, false);
+	} else {
+		extent_deregister(tsdn, pac, edata);
+	}
```
interwq, line 1072: If we always do deregister on this path, can we avoid the additional flag for the unguard_pages? It might seem to do redundant operations that way, but since the destroy very rarely happens (if at all), it should be a net win (both simpler API, and one less branch).

Lapenkov: We can't do it that easily, because `unguard_pages` does `emap_register_boundary`, which has `assert(edata_state_get(edata) == extent_state_active);`. I would prefer to keep the assertion to serve its purpose and catch bugs, because `emap_register_boundary` is used in two more places.

interwq: I see. In that case, let's do `emap_update_edata_state(tsdn, pac->emap, edata, extent_state_active);` before calling unguard? That way also feels safer since `active` is an acquired state (meaning other threads won't try accessing it), and it can remain only unguard active extents. Then we can still contain the logic to the destroy side instead of tweaking the unguard side.

interwq: Good point regarding gdump. I'll take a look again tomorrow; a bit out of energy now.

Talking about gdump, don't we have a double `deregsiter`, i.e. double gdump subtraction along the evict => destroy path?

Lapenkov: We have double `deregister`, but in `extent_gdump_sub` there's a check that subtracts only if the extent is `active`. Tricky.

interwq: Indeed. Thanks for the discussion -- let's continue and see if this will work:

Context: the deregister happens under ecache_evict makes sense, because we are holding the ecache lock. A general rule we follow is, an edata should either be covered by a state lock (i.e. mutex for the ecache it's part of), or it's in a acquired state.  See this [sanity check function for details](https://github.com/jemalloc/jemalloc/blob/4d56aaeca5883ae5f4b5550c528503fb51fdf479/include/jemalloc/internal/emap.h#L169). Therefore, the deregister before releasing the lock in evict is doing the right thing.
1) let's get rid of the dup deregister in `extent_destroy_wrapper`? We can use the above assert to verify it's acquired in place of the deregister. To be safe, I'd say add a separate commit to make it clear it's intentional.
2) then maybe a new `unguard_pages_pre_destroy()`, which simply calls the `static unguard_pages_impl()` with the `reg_emap == false` (the flag should cover both the deregister and register), while the original `unguard_pages()` calls it with `reg_emap==true`. This way we makes sure only the destroy path gets the special treatment of bypassing emap re-register.

Lapenkov: Sounds good. This is one of the approaches I've considered, let's implement it!

src/guard.c
```diff
@@ -44,16 +47,35 @@ unguard_pages(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata, emap_t *emap) {
 	uintptr_t guard1 = addr - PAGE;
 	uintptr_t guard2 = addr + size;
 
-	assert(edata_state_get(edata) == extent_state_active);
+	assert(edata_state_get(edata) == extent_state_active
+	    || edata_state_get(edata) == extent_state_retained);
```
interwq, line 51: nit: Split into 2 assertions in the first `if (reg_emap)` branch? I assume the retained only holds if !reg_emap

Lapenkov: Right, let's do that.

interwq (approved):

PR #2154, opened 2021-11-04
CI: Refactor gen_travis.py

Lapenkov:
The CI consolidation project adds more operating systems to Travis. This
refactoring is aimed to decouple the configuration of each individual OS
from the actual job matrix generation and formatting. Otherwise,
format_job function would turn into a huge collection of ad-hoc
conditions.

The job matrix stays intact, all changes to `.travis.yml` should not affect the jobs being run, despite some syntactic changes like job order or comments.


.travis.yml
```diff
@@ -1,17 +1,15 @@
...
 language: generic
 dist: focal
 
-matrix:
+jobs:
   include:
     - os: linux
       arch: amd64
       env: CC=gcc CXX=g++ COMPILER_FLAGS="" CONFIGURE_FLAGS="" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
-    - os: osx
```
interwq, line 9: Question: is it true that only the order of the jobs changed, not the actual combination?

interwq (approved):

.travis.yml
interwq: Never mind; just saw the PR summary.

PR #2155, opened 2021-11-04
Add ctl interface for experimental_infallible_new.

interwq:
W/o the ctl interface, the stats isn't output the option -- the below mallctl is failing:
https://github.com/jemalloc/jemalloc/blob/dev/src/stats.c#L1509


davidtgoldblatt (approved):
Sadly, computer science has not yet invented a way to repeat the same string in three different places. My bad.

interwq:
No worries :) and thanks for the quick review!

PR #2160, opened 2021-11-16
Fix uninitialized nstime reading / updating on the stack in hpa.

interwq:
In order for nstime_update to handle non-monotonic clocks, it requires the input
nstime to be initialized -- when reading for the first time, zero init has to be
done.  Otherwise random stack value may be seen as clocks and returned.


Lapenkov:
Maybe we can enhance non-monotonic clock check by adding a reasonable threshold? If a new reading is 1 year back, something must be wrong.

Lapenkov:
Initialized flag is way better, thanks for working on it.

PR #2161, opened 2021-11-17
Track the initialized state of nstime_t on debug build.

interwq:
Some nstime_t operations require and assume the input nstime is initialized
(e.g. nstime_update) -- uninitialized input may cause silent failures which is
difficult to reproduce / debug.  Add an explicit flag to track the state
(limited to debug build only).

Also fixed an use case in hpa (time of last_purge).

### DIFF
--- a/include/jemalloc/internal/nstime.h
+++ b/include/jemalloc/internal/nstime.h
@@ -3,10 +3,19 @@
 
 /* Maximum supported number of seconds (~584 years). */
 #define NSTIME_SEC_MAX KQU(18446744072)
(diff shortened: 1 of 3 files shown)


interwq:
As discussed in #2160, the uninitialized nstime_t use is hard to track down. Adding the explicit flag (which also helped uncover another similar issue).

Manually verified that it catches the issue being fixed in #2160.

interwq:
Good point. Let me update.

Lapenkov:
Thanks, now it's next to impossible to forget to initialize it!

Lapenkov (approved):

PR #2164, opened 2021-11-23
Correct opt.prof_leak documentation

Lapenkov:
The option has been misleading, because it stays disabled unless
prof_final is also specified. In practice, it's impossible to detect that
the option is silently disabled, because it just doesn't provide any output
as if there are no memory leaks detected.


interwq:
nit: typo in the commit msg `disnabled`

interwq (approved):

PR #2168, opened 2021-12-01
Mark slab as true on sized dealloc fast path.

interwq:
For sized dealloc, fastpath only handles lookup-able sizes, which must be slabs.

### DIFF
--- a/include/jemalloc/internal/sc.h
+++ b/include/jemalloc/internal/sc.h
@@ -248,15 +248,20 @@
 
 /* The largest size class in the lookup table, and its binary log. */
 #define SC_LG_MAX_LOOKUP 12
-#define SC_LOOKUP_MAXCLASS ((size_t)1 << SC_LG_MAX_LOOKUP)
+#define SC_LOOKUP_MAXCLASS (1 << SC_LG_MAX_LOOKUP)
 
 /* Internal, only used for the definition of SC_SMALL_MAXCLASS. */
-#define SC_SMALL_MAX_BASE ((size_t)1 << (LG_PAGE + SC_LG_NGROUP - 1))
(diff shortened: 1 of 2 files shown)


davidtgoldblatt (commented):
Am I right in thinking that this doesn't affect fast-path assembly? Slab vs. not-slab only matters if the object isn't headed to the tcache, in which case we fall off the fast path regardless. Hence the current dead-store comment; nobody ever reads alloc_ctx.slab, so it should get optimized away.

interwq:
Right. The dead store comment is correct and it's not affecting the opt assembly. I'd say this is more for avoiding confusion. I was looking at this for a bit and thought I missed something. Also changing it to an assert which should be more useful.

davidtgoldblatt (approved):
I think it's not necessarily true, right? I.e. lookup-able sizes *could* be large; it's mostly an accident of our choice of `SC_LOOKUP_MAXCLASS` (small relative to the page size) that they're not.

Not a strong opinion, don't want to hold up velocity over arguing about it, so stamping if you disagree. Two nits though:
- Could you add a comment here, explaining why the assert's right?
- Maybe another assert during sc booting or something, that SC_LOOKUP_MAXCLASS <= SC_SMALL_MAXCLASS

davidtgoldblatt:
Actually, we can make it a compile-time failure, both constants are available to the preprocessor:
```
#if SC_SMALL_MAXCLASS <= SC_LOOKUP_MAXCLASS
#  error "whatever"
#endif
```

interwq:
Will update with the compile time check. Yeah I wanted to make it clear that the fastpath only handles small sizes, and if anything like the lookup-able threshold changes it better fail some assertion. For example, the tcache bins accessed on the path don't cover all size classes (used to be the case), but only small plus ~a dozen of large ones now.

PR #2169, opened 2021-12-02
freebsd 14 build fix proposal.

devnexen:
seems to have introduced finally more linux api cpu affinity (sched_* family)
compatibility detected at configure time thus adjusting accordingly.


interwq:
Thanks for the patch!

PR #2171, opened 2021-12-03
Enforce abort_conf:true when malloc_conf is not fully recognized.

interwq:
Ensures the malloc_conf "ends with key", "ends with comma" and "malform conf
string" cases abort under abort_conf:true.


interwq:
Previously, there cases wouldn't abort --

MALLOC_CONF=abort_conf:true,
MALLOC_CONF=abort_conf:true,,
MALLOC_CONF=abort_conf:true,asdf
MALLOC_CONF=abort_conf:true,asdf,

Verified that w/ this PR they all abort as expected.

Lapenkov (approved):

PR #2175, opened 2021-12-13
Use visibility("default") attribute for libc overrides

neobrain:
This fixed problems we observed with [FEX's jemalloc fork](https://github.com/FEX-Emu/jemalloc) overriding libc functions, where the libc's `malloc` implementation got called instead of jemalloc's. I'm guessing `attribute((used))` is purely a compiler directive, whereas the linker only considers visibility (which is set via `-fvisibility=hidden` in our build configuration).

I haven't reproduced this with upstream jemalloc (since I'm only using our fork), but it's worth noting both tcmalloc and mimalloc have similar overrides that used `attribute((used))` in the past and have later been changed to use the `visibility` attribute instead, so this seems like a reasonable way to follow. (Compare [old tcmalloc](https://github.com/chromium/chromium/blob/84ce36326b9e2985278ae1ac8095b4b85a0653a7/third_party/tcmalloc/chromium/src/libc_override_gcc_and_weak.h#L58) to [new tcmalloc](https://github.com/google/tcmalloc/blob/98b28bc6861036aabd1bd8d3b17800012c4d8f23/tcmalloc/libc_override_gcc_and_weak.h#L35-L36), and see [mimalloc](https://github.com/microsoft/mimalloc/blob/0560fc27c08d28d523b7f741a42deb26cd01c0c6/src/alloc-override.c#L33-L38))

I'm not sure if upstream needs to keep `attribute((used))` around, though. If the user builds without `-fvisibility=hidden`, it might still have some effect. Opinions?

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -3088,7 +3088,7 @@ JEMALLOC_EXPORT void *(*__memalign_hook)(size_t alignment, size_t size) =
  * be implemented also, so none of glibc's malloc.o functions are added to the
  * link.
  */
-#    define ALIAS(je_fn)	__attribute__((alias (#je_fn), used))
+#    define ALIAS(je_fn)	__attribute__((alias (#je_fn), visibility("default")))
 /* To force macro expansion of je_ prefix before stringification. */
 #    define PREALIAS(je_fn)	ALIAS(je_fn)
 #    ifdef JEMALLOC_OVERRIDE___LIBC_CALLOC


Lapenkov:
Hello. To better understand the motivation to make these visible, I'd like to know more about the nature of the issues you had. Do you have a link to a github issue with more information or something?

neobrain:
We don't have a formal issue report, but let me summarize the problem in more detail:

FEX uses its jemalloc fork as a drop-in replacement for the stdlib's malloc/free allocators. However, we observed that in some workloads, glibc's malloc would still be called (usually in some external library rather than FEX itself), causing crashes later down the road due to passing libc-managed pointers to jemalloc allocator functions (`je_free`, `je_malloc_usable_size`). Notably our minimalistic FEXConfig application was affected early during startup, where glibc's malloc was called deep inside a `SDL_Init > libsystemd > libudev` stack trace.

These crashes are reliably fixed by changing the attribute from `used` to `visibility(default)` to counteract `-fvisibility=hidden` that we build jemalloc with.

Note that upstream jemalloc only overrides the `__libc_` family of functions, whereas FEX also overrides `malloc`/`free`/... themselves. Also, FEX is generally built with `clang`. I'm not sure to what extent either of these play a role in the problem though.

Lapenkov:
```
Note that upstream jemalloc only overrides the __libc_ family of functions, whereas FEX also overrides malloc/free/... themselves.
```

Upstream jemalloc also overrides `malloc` and `free`: https://github.com/FEX-Emu/jemalloc/blob/dea850b30a00354f08b72f4df94e70f3c093f9f9/include/jemalloc/jemalloc_protos.h.in
And they have `visibility("default")`.

Did you try removing your declarations? I think that might be the cause.

Lapenkov:
You might say that those declarations have `je_` prefix, but when they're added to `jemalloc.h`, there's `#define je_malloc malloc` placed before those declarations. See `include/jemalloc/jemalloc.sh` that generates `jemalloc.h`. I admit - that's very confusing.

neobrain:
Thanks for the quick reply, @Lapenkov, and for the note about `jemalloc.h` overriding these functions already - easy to miss indeed :)

> Did you try removing your declarations? I think that might be the cause.

I gave this some testing and it seems to work fine without my visibility change indeed. Not sure why we had added the overrides in the first place - I'll discuss internally and let you know if something significant comes up!

PR #2180, opened 2021-12-17
Fix `base_ehooks_get_for_metadata`

mweisgut:
Since #2118, jemalloc has the experimental mallctl `experimental.arenas_create_ext`, which allows passing an `arena_config_t` structure, which can be used to customize the behavior of the arena. With this addition, `base_ehooks_get_for_metadata` was also added, which is intended to return the corresponding extend hook functions:
- (1) the user-specified extend hook functions when `metadata_use_hooks` is `true`
- (2) the default extend hook functions when `metadata_use_hooks` is `false`.

However, in case (2), accessing the return value leads to a segmentation fault due to the static cast `(struct ehooks_s *)&ehooks_default_extent_hooks`. Note that `&ehooks_default_extent_hooks` is of type `const extent_hooks_t *`. I'm new to the jemalloc code base, but from my current understanding, this is not a correct way to wrap `&ehooks_default_extent_hooks` into an `ehooks_t` (or `ehooks_s`) object. Instead, `ehooks_init(...)` can be used to do so.

This PR adds two test cases for `base_ehooks_get_for_metadata` and suggests a fix for the described segmentation fault. This fix, however, also changes the interface. I am glad to receive feedback about the changes.

To reproduce the segmentation fault, a1e8bf54ff1f01ac5392e4c8387f57ad97f192dd only contains two additional test cases. Test case `test_base_ehooks_get_for_metadata_default_hook` reveals the failure.

### DIFF
--- a/src/base.c
+++ b/src/base.c
@@ -295,12 +295,6 @@ base_block_alloc(tsdn_t *tsdn, base_t *base, ehooks_t *ehooks, unsigned ind,
 	return block;
 }
 
-static ehooks_t *
-base_ehooks_get_for_metadata(base_t *base) {
-	return base->metadata_use_hooks ? &base->ehooks :
-	    (ehooks_t *)&ehooks_default_extent_hooks;
-}
-
 /*
  * Allocate an extent that is at least as large as specified size, with
  * specified alignment.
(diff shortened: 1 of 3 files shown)


src/base.c
```diff
@@ -295,10 +295,14 @@ base_block_alloc(tsdn_t *tsdn, base_t *base, ehooks_t *ehooks, unsigned ind,
 	return block;
 }
 
-static ehooks_t *
-base_ehooks_get_for_metadata(base_t *base) {
-	return base->metadata_use_hooks ? &base->ehooks :
-	    (ehooks_t *)&ehooks_default_extent_hooks;
```
interwq, line 301: Good catch! This is indeed an invalid conversion.

I think we can make the fix simpler though; the current PR attempts to workaround the difference between ehooks_t and extent_hooks_t by adding an ehooks_t on the stack. However we generally avoid this pattern, unless its forced by bootstrapping (e.g. the `fake_ehooks` in base_new).

Instead we can simply add another ehooks_t in base_s, `ehooks_base` to replace `bool metadata_use_hooks`. Then the added ehooks_base can be initialized to either the same as `ehooks`, or pointing to the default hooks, based on the `metadata_use_hooks` pass into `base_new`.

The other request is, please squash everything into a single commit, including the test cases. Then you can `git push -f` to update the PR.

interwq:
Looks great. Thank again @mweisgut!

PR #2181, opened 2021-12-17
[RFC] Disable percpu arena in case of non deterministic CPU count

azat:
Determinitic number of CPUs is important for percpu arena to work
correctly, since it uses cpu index - sched_getcpu(), and if it will
greater then number of CPUs bad thing will happen, or assertion will be
failed in debug build:

    <jemalloc>: ../contrib/jemalloc/src/jemalloc.c:321: Failed assertion: "ind <= narenas_total_get()"
    Aborted (core dumped)

Number of CPUs can be obtained from the following places:
- sched_getaffinity()
- sysconf(_SC_NPROCESSORS_ONLN)
- sysconf(_SC_NPROCESSORS_CONF)

For the sched_getaffinity() you may simply use taskset(1) to run program
on a different cpu, and in case it will be not first, percpu will work
incorrectly, i.e.:

    $ taskset --cpu-list $(( $(getconf _NPROCESSORS_ONLN)-1 )) <your_program>

_SC_NPROCESSORS_ONLN uses /sys/devices/system/cpu/online, LXD/LXC
virtualize /sys/devices/system/cpu/online file [1], and so when you run
container with limited limits.cpus it will bind randomly selected CPU to
it

  [1]: https://github.com/lxc/lxcfs/issues/301

_SC_NPROCESSORS_CONF uses /sys/devices/system/cpu/cpu*, and AFAIK nobody
playing with dentries there.

So if all three of these are equal, percpu arenas should work correctly.

And a small note regardless _SC_NPROCESSORS_ONLN/_SC_NPROCESSORS_CONF,
musl uses sched_getaffinity() for both. So this will also increase the
entropy.

Also note, that you can check is percpu arena really applied using
abort_conf:true.

Refs: https://github.com/jemalloc/jemalloc/pull/1939
Refs: https://github.com/ClickHouse/ClickHouse/issues/32806

v2: move malloc_cpu_count_is_deterministic() into
malloc_init_hard_recursible() since _SC_NPROCESSORS_CONF does
allocations for readdir()

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -148,6 +148,8 @@ unsigned	opt_narenas = 0;
 fxp_t		opt_narenas_ratio = FXP_INIT_INT(4);
 
 unsigned	ncpus;
+/* ncpus is determinstinc, see malloc_cpu_count_is_deterministic() */
+static int	cpu_count_is_deterministic = -1;
 
 /* Protects arenas initialization. */
 malloc_mutex_t arenas_lock;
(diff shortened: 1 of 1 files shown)


src/jemalloc.c
```diff
@@ -1721,6 +1759,15 @@ malloc_init_hard_recursible(void) {
 	malloc_init_state = malloc_init_recursible;
 
 	ncpus = malloc_ncpus();
+	cpu_count_is_deterministic = malloc_cpu_count_is_deterministic();
+
+	if (!cpu_count_is_deterministic) {
```
Lapenkov, line 1764: This condition is weak enough to instantly abort. If percpu arena wasn't enabled by the user, we shouldn't worry about CPU count. In fact, this caused one of the tests to abort unnecessarily:
https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/552738103

Another thing is, if the user supplied `narenas` and enabled `percpu_arena`, they probably know what they're doing and should've set `nareanas` to a number large enough to accommodate the largest possible CPU ID. In this case, it might be worth it to just print a message saying `Warning: if narenas is less than the maximum CPU ID, you'll get undefined behavior`. This scenario is described here in the comment:
https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L1925-L1938

Only if the user asked for the percpu arena and didn't provide `narenas`, and this happened on a system that has a non-deterministic cpu count, we should either abort or change the setting (but also notify about the change in the setting).

Lapenkov:
Thanks for the PR, it's a step in the right direction to make the allocator more robust! I added some feedback, but if you don't have the capacity to address it, let me know, and I'll finish it myself.

src/jemalloc.c
```diff
@@ -128,6 +128,8 @@ bool	opt_zero = false;
 unsigned	opt_narenas = 0;
 
 unsigned	ncpus;
+/* ncpus is determinstinc, see malloc_cpu_count_is_deterministic() */
+int		cpu_count_is_deterministic = -1;
```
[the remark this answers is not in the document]
azat: Indeed, fixed

src/jemalloc.c
azat: >If percpu arena wasn't enabled by the user, we shouldn't worry about CPU count

Indeed, nice catch, thanks, fixed!

>Another thing is, if the user supplied narenas and enabled percpu_arena, they probably know what they're doing and should've set nareanas to a number large enough to accommodate the largest possible CPU ID

Agree, added a message, though it is different, but I'm OK with using your message as-is, just let me know.

Lapenkov:
Can you please rebase on dev for CI?

azat:
>Can you please rebase on dev for CI?

Done.

PR #2182, opened 2021-12-20
Fix the time-since computation in HPA.

interwq:
nstime module guarantees monotonic clock update within a single nstime_t.  This
means, if two separate nstime_t variables are read and updated separately,
nstime_subtract between them may result in underflow.  Fixed by switching to the
time since utility provided by nstime.

### DIFF
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -479,10 +479,7 @@ hpa_try_hugify(tsdn_t *tsdn, hpa_shard_t *shard) {
 
 	/* Make sure that it's been hugifiable for long enough. */
 	nstime_t time_hugify_allowed = hpdata_time_hugify_allowed(to_hugify);
-	nstime_t nstime;
-	shard->central->hooks.curtime(&nstime, /* first_reading */ true);
-	nstime_subtract(&nstime, &time_hugify_allowed);
-	uint64_t millis = nstime_msec(&nstime);
+	uint64_t millis = shard->central->hooks.ms_since(&time_hugify_allowed);
 	if (millis < shard->opts.hugify_delay_ms) {
 		return false;
 	}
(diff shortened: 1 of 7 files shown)


src/hpa.c
```diff
@@ -479,10 +479,7 @@ hpa_try_hugify(tsdn_t *tsdn, hpa_shard_t *shard) {
 
 	/* Make sure that it's been hugifiable for long enough. */
 	nstime_t time_hugify_allowed = hpdata_time_hugify_allowed(to_hugify);
-	nstime_t nstime;
-	shard->central->hooks.curtime(&nstime, /* first_reading */ true);
-	nstime_subtract(&nstime, &time_hugify_allowed);
```
interwq, line 484: As observed by @Lapenkov, this may trigger 
`
<jemalloc>: src/nstime.c:118: Failed assertion: "nstime_compare(time, subtrahend) >= 0"
`
such as in https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/552738061

src/hpa.c
```diff
@@ -924,11 +919,8 @@ hpa_time_until_deferred_work(tsdn_t *tsdn, pai_t *self) {
 			malloc_mutex_unlock(tsdn, &shard->mtx);
 			return BACKGROUND_THREAD_DEFERRED_MIN;
 		}
-		nstime_t nstime;
-		shard->central->hooks.curtime(&nstime,
-		    /* first_reading */ true);
```
Lapenkov, line 929: I'm a bit worried about this now - if you need to mock time for the whole HPA, you now have to employ two mechanisms: this hook and `nstime_update`. Maybe we need to make another `nstime_ns_since_with_hook` that has a function pointer argument?

interwq: The `nstime_update` pointer switch does feel like cheating, given that we already have the time related hooks. Let me update with another hook then.

interwq:
@Lapenkov updated with the new time since hook.

Lapenkov (approved):

PR #2184, opened 2021-12-21
Only invoke malloc_cpu_count_is_deterministic() when necessary.

interwq:
Also refactor the handling of the non-deterministic case.  Notably allow the
case with narenas set to proceed w/o warnings, to not affect existing valid use
cases.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -148,8 +148,6 @@ unsigned	opt_narenas = 0;
 fxp_t		opt_narenas_ratio = FXP_INIT_INT(4);
 
 unsigned	ncpus;
-/* ncpus is determinstinc, see malloc_cpu_count_is_deterministic() */
-static int	cpu_count_is_deterministic = -1;
 
 /* Protects arenas initialization. */
 malloc_mutex_t arenas_lock;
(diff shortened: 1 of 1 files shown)


src/jemalloc.c
```diff
@@ -1931,22 +1930,24 @@ malloc_init_narenas(void) {
...
+		bool cpu_count_is_deterministic =
+		    malloc_cpu_count_is_deterministic();
 		if (!cpu_count_is_deterministic) {
-			if (opt_narenas) {
-				malloc_write("<jemalloc>: Number of CPUs is not deterministic, "
-					"but narenas is set. Hope you not what you are doing and "
-					"you have set narenas to largest possible CPU ID.\n");
-				if (opt_abort) {
-					abort();
-				}
```
Lapenkov, line 1941: I liked the warning when somebody uses `opt_narenas` + `opt_percpu_arena`. Maybe this could help someone detect an issue early.

interwq: I considered that part too. But then I feel that `narenas` and `percpu_arena` are really conflicting options in principle -- if they are both set, it should be intentional.

The other thing is, I can't seem to find what to say in that case, especially since the setting might be a valid case. Then the warning msg will never go away, no matter how narenas is set. So I went with the approach to preserve the existing behavior (i.e. not causing any fuss for that case).

Lapenkov: Ok, that's a good way to think about it, too. I'm fine with it.

Lapenkov (approved):

PR #2187, opened 2021-12-23
Fix shadowed variable usage.

### LINKED ISSUE #1863: error: declaration (prof_active) shadows a variable in the global scope [-Werror,-Wshadow]
```
error: declaration shadows a variable in the global scope [-Werror,-Wshadow]
    bool prof_active, bool updated, const void *old_ptr, size_t old_usize,

jemalloc/include/jemalloc/internal/prof_externs.h:31:13: note: previous declaration is here
extern bool     prof_active;
                ^
```

interwq:
Verified with EXTRA_CFLAGS=-Wshadow.

Resolves #1863

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -43,7 +43,7 @@ static counter_accum_t prof_idump_accumulated;
  * Initialized as opt_prof_active, and accessed via
  * prof_active_[gs]et{_unlocked,}().
  */
-bool prof_active;
+bool prof_active_state;
 static malloc_mutex_t prof_active_mtx;
 
 /*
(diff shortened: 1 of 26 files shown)


src/prof.c
```diff
@@ -43,7 +43,7 @@ static counter_accum_t prof_idump_accumulated;
  * Initialized as opt_prof_active, and accessed via
  * prof_active_[gs]et{_unlocked,}().
  */
-bool prof_active;
+bool prof_active_state;
```
Lapenkov, line 46: This variable must be shadowed by something rather than shadowing something itself?

interwq: Yep. Many functions have an input arg named prof_active. Since the global one usually shouldn’t be accessed anyway (it’s supposed to go through the getter), I renamed this one instead.

interwq:
> Maybe let's enable `-Wshadow` by default? Otherwise, these will slowly accumulate again and is seems to be error-prone to fix them in bulk.

Agreed. Let’s do that part separately, after the release maybe? I’d say not tweak the flags right before a release.

interwq:
A side note is, although most changes here are to avoid scope confusion, removing the redundant `tcache_get` in jemalloc_internal_inline was a 2KB reduction (~0.2%) on the code size of libjemalloc.so.

Lapenkov (approved):

PR #2189, opened 2021-12-26
linux MAP_FIXED_NOREPLACE support proposal.

devnexen:


src/pages.c
```diff
@@ -278,6 +278,30 @@ pages_map(void *addr, size_t size, size_t alignment, bool *commit) {
...
+	{
+		if (os_overcommits) {
+			*commit = true;
+		}
+
+		int prot = *commit ? PAGES_PROT_COMMIT : PAGES_PROT_DECOMMIT;
+		int flags = mmap_flags;
+
+		if (addr != NULL) {
+			flags |= MAP_FIXED_NOREPLACE;
```
interwq, line 296: It looks there was a bad corruption bug of this feature when introduced in 4.17:
https://marc.info/?l=linux-mm&m=153918527100638
https://mobile.twitter.com/danielmicay/status/1050233089794834434

which makes me a bit hesitant to add it. The only case triggering this path w/ `addr != NULL` is the in place expansion of large allocations: https://github.com/jemalloc/jemalloc/blob/d038160f3b76ac1e5203e11008169366629c81cd/src/pac.c#L209

which should be rather uncommon, plus the chance of the expansion being successful is slim (if we did a good job packing the address space tight) so IMO that pattern is unlikely to get very popular in the future. So I'm inclined to say the benefit isn't enough, given that the kernel bug could be hard to detect and investigate.

cc: @Lapenkov and @davidtgoldblatt in case I missed something.

devnexen: ah true was unaware of the 4.17 particular bug, fair enough we can close this PR if wished.

interwq: Thanks for the PR @devnexen! Closing now but we can revisit if things change.

PR #2190, opened 2021-12-27
Fix symbol conflict with musl libc

georgthegreat:
`__libc` prefixed functions are used by musl libc as non-replaceable malloc stubs.

Fix this conflict by checking if we are linking against glibc.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -3175,7 +3175,7 @@ JEMALLOC_EXPORT void *(*__memalign_hook)(size_t alignment, size_t size) =
     je_memalign;
 #  endif
 
-#  ifdef CPU_COUNT
+#  ifdef __GLIBC__
 /*
  * To enable static linking with glibc, the libc specific malloc interface must
  * be implemented also, so none of glibc's malloc.o functions are added to the


Lapenkov:
`__GLIBC__` is defined in `features.h`.  Currently this header is transitively included by something else, but this might not hold true in the future.

Could you please include `features.h` inside `#if defined(JEMALLOC_IS_MALLOC) && defined(JEMALLOC_GLIBC_MALLOC_HOOK)` ?

Lapenkov:
Also, do you have a reference to any issues caused by jemalloc replacing `__libc` functions with musl? Not that I'm doubting it should be fixed, but for future testing.

georgthegreat:
@Lapenkov, I have added `#include <features.h>`, as requested.

As for the second question — no, I don't have any public reference with corresponding issue, and musl does not have any design doc regarding this decision.

Some clues can be found in [musl soyrce code](https://git.musl-libc.org/cgit/musl/tree/src/thread/sem_open.c?h=v1.2.2#n17).

I have the following comment in Yandex internal codebase:

```
A couple of musl functions call its internal malloc directly because
they anticipate external mallocs to call musl and deadlock. This does
not happen with Yandex allocators, and since currently the musl
allocator does not support static linking with an external allocator,
we route the internal malloc calls to the external allocator.
```

PR #2195, opened 2021-12-30
Fix test config of lg_san_uaf_align.

interwq:
The option may be configure-disabled, which resulted in the invalid options
output from the tests.

### DIFF
--- a/test/unit/tcache_max.c
+++ b/test/unit/tcache_max.c
@@ -1,4 +1,7 @@
 #include "test/jemalloc_test.h"
+#include "test/san.h"
+
+const char *malloc_conf = TEST_SAN_UAF_ALIGN_DISABLE;
 
 enum {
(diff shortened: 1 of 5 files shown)


test/unit/tcache_max.c
```diff
@@ -1,4 +1,7 @@
 #include "test/jemalloc_test.h"
+#include "test/san.h"
+
+const char *malloc_conf = TEST_SAN_UAF_ALIGN_DISABLE;
```
interwq, line 4: This is the only test not working with `san_uaf_align` -- it assumes and checks the exact caching behavior which is a bit hard to work around. Explicitly disables the uaf check option, as I envision it will become default on in the future (otherwise the test will be always skipped when the default changes).

Lapenkov (approved):

PR #2196, opened 2021-12-30
Fix unnecessary returns in san_(un)guard_pages_two_sided.

interwq:


src/tcache.c
```diff
@@ -134,7 +134,11 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
 
 	size_t nflush = low_water - (low_water >> 2);
 	if (nflush < tcache_slow->bin_flush_delay_items[szind]) {
-		tcache_slow->bin_flush_delay_items[szind] -= nflush;
+		/* Workaround for a conversion warning. */
+		uint8_t nflush_uint8 = (uint8_t)nflush;
+		assert(sizeof(tcache_slow->bin_flush_delay_items[0]) ==
+		    sizeof(nflush_uint8));
```
Lapenkov, line 140: Did you put it here to remember to change this workaround if `bin_flush_delay_items` type is ever changed?

Lapenkov (approved):

src/tcache.c
interwq: Exactly. Unfortunately this feels like paying a cost to get rid of the warning.

PR #2198, opened 2022-01-05
FreeBSD also needs the OS-X "don't declare system functions as nothrow" fix

leres:
... since it also has jemalloc in the base system


interwq:
@davidtgoldblatt this looks to be the right thing to do. Should we broaden the coverage further?

PR #2200, opened 2022-01-07
Fix the background thread sleep in `test/unit/hpa_background_thread`

interwq:
Under high concurrency / heavy test load (e.g. using run_tests.sh), the
background thread may not get scheduled for a longer period of time.  Retry 100
times max before bailing out.

Verified that the fix works with `run_tests.sh` -- it was failing reliably every ~100 runs.

### DIFF
--- a/test/unit/hpa_background_thread.c
+++ b/test/unit/hpa_background_thread.c
@@ -4,8 +4,8 @@
 static void
 sleep_for_background_thread_interval() {
 	/*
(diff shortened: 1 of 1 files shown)


Lapenkov:
Thanks for fixing this one!  I've also seen it fail, but didn't know it was due to thread not getting scheduled.

Lapenkov (approved):

PR #2204, opened 2022-01-10
fix: exit 1 when memory leak is found

yunxuo:
In actual use, it is often necessary to use jemalloc to check for memory leaks, and jemalloc will exit 0 when checking for memory leaks, which is very unfriendly to automated detection. If it exit 1 like tcmalloc when exiting, you can use the automated process to check memory leak.

### DIFF
--- a/src/prof.c
+++ b/src/prof.c
@@ -1790,6 +1790,7 @@ prof_leakcheck(const prof_cnt_t *cnt_all, size_t leak_ngctx,
 		malloc_printf(
 		    "<jemalloc>: Run jeprof on \"%s\" for leak detail\n",
 		    filename);
+		_exit(1);    // we don't want to call atexit() routines!
 	}
 #endif
 }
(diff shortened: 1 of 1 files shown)


src/prof.c
```diff
@@ -1790,6 +1790,7 @@ prof_leakcheck(const prof_cnt_t *cnt_all, size_t leak_ngctx,
 		malloc_printf(
 		    "<jemalloc>: Run jeprof on \"%s\" for leak detail\n",
 		    filename);
+		_exit(1);    // we don't want to call atexit() routines!
```
interwq, line 1793: This function is invoked through `atexit` already, via `prof_fdump`: https://github.com/jemalloc/jemalloc/blob/dev/src/prof.c#L646

Since this is changing the behavior / not backward compatible, we will need to guard it with a runtime option, e.g. maybe `opt.prof_leak_error`?

The other thing is, you need to work on the dev branch, and submit PR for dev only. The master branch is only for releases. Closing but feel free to ask anything.

yunxuo: nice suggestion, thank u

interwq: In case you want an example of adding runtime options, here is one: https://github.com/jemalloc/jemalloc/pull/2027/files

PR #2205, opened 2022-01-10
use isb instead of yield for spin locks on arm

AWSjswinney:
In my tests using a micro benchmark, I found that in some cases using the `isb` (instruction synchronization barrier) instead of `yield` can improve allocation performance by 1.3% and free performance by 1.1%. Other projects have seen a similar change, such as [mysql](https://bugs.mysql.com/bug.php?id=100664), [rust](https://github.com/rust-lang/rust/pull/84725), and [wiredtiger](https://github.com/wiredtiger/wiredtiger/pull/6080) have also seen improvement with similar changes.

The idea is that the the `isb` instruction takes longer to execute than a `yield` instruction (which is usually implemented as a no-op) and therefore allows changes to a spin lock flag to propagate, eliminating the need for a slower path such as a futex or other call which results in a context switch.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -427,10 +427,10 @@ case "${host_cpu}" in
 	HAVE_CPU_SPINWAIT=1
     AC_CACHE_VAL([je_cv_yield],
       [JE_COMPILABLE([yield instruction], [],
-                    [[__asm__ volatile("yield"); return 0;]],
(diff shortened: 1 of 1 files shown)


interwq:
Thanks for the PR! Merged as 89fe8ee6bf7a23556350d883a310c0224a171879 with small tweaks.

PR #2207, opened 2022-01-11
Add FreeBSD and enable Windows in TravisCI

Lapenkov:
This PR finalizes the consolidation of all OS we want to run CI on within TravisCI. With the usage based plan, it now takes as long as the slowest Windows job (26 min) to complete the CI. As discussed, the other systems we use (AppVeyor and CirrusCI) will be kept for redundancy purposes.


Lapenkov:
This is now ready for review.

interwq (approved):
Looks great!

PR #2211, opened 2022-01-18
Avoid calculating size of size class twice & delete sc_data_global.

Clcanny:


interwq:
Thanks for the PR! Tweaked a tiny bit and merged in eb196815d670f0937d2117ff0f2b885bd23c80de.

PR #2214, opened 2022-01-22
Initialize prof_leak during prof init.

interwq:
Otherwise, prof_leak may get set after prof_leak_error, and disagree with each
other.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1578,26 +1578,8 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				CONF_HANDLE_BOOL(opt_prof_gdump, "prof_gdump")
 				CONF_HANDLE_BOOL(opt_prof_final, "prof_final")
 				CONF_HANDLE_BOOL(opt_prof_leak, "prof_leak")
-				if (CONF_MATCH("prof_leak_error")) {
-					if (CONF_MATCH_VALUE("true")) {
-						if (!opt_prof_final) {
-							CONF_ERROR(
-							    "prof_leak_error is"
-							    " not allowed"
-							    " without"
-							    " prof_final",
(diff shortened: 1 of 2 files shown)


src/jemalloc.c
```diff
@@ -1585,7 +1585,7 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 							    "prof_leak_error is"
```
interwq, line 1585: The other thing I realized, is that the `if (!opt_prof_final)` branch creates an order dependency with prof_final, i.e. it will warn / abort if you specify "prof_leak_error:true,prof_final:true" (or prof_final specified in a later stage conf source). I'm also thinking, should we simply enable prof_final as well, since this is a new option?

Lapenkov: Riiiight, the fact that order in which they are specified matters is a problem. I missed that.

I think you suggestion is OK if we specify it in the docs. But then again, someone can explicitly turn off `prof_final` after they set `prof_leak_error`. This will be undetected and `prof_leak_error` won't work. Even if we specify that `prof_leak_error` enables `prof_final`, it's not obvious that the code only works if they both are set to `true`. Rejecting on options validation phase seems robust and correct here.

I think we should separate parsing options from their validation and malloc initialization. Currently it's unclear where to validate options. The `malloc_conf_init_helper` just proved to be pretty inconvenient for that, because you don't know all the values at the moment of validation.

Lapenkov:
What do you think about adding a "stage" to `malloc_conf_init` that would check such that after all option values are set, they adhere to a list of invariants?

src/jemalloc.c
interwq: Yep I added a `malloc_conf_init_check_deps` step at the end. Kept it simple since we want to include it in the release.

Lapenkov (approved):

PR #2217, opened 2022-01-27
Avoid overflow warnings in test/unit/safety_check.

interwq:
Warning by `-Wstringop-overflow`.

### DIFF
--- a/test/unit/safety_check.c
+++ b/test/unit/safety_check.c
@@ -13,6 +13,13 @@ void fake_abort(const char *message) {
 	fake_abort_called = true;
 }
 
+static void
+buffer_overflow_write(char *ptr, size_t size) {
+	/* Avoid overflow warnings. */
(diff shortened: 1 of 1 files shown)


test/unit/safety_check.c
```diff
@@ -13,14 +13,21 @@ void fake_abort(const char *message) {
 	fake_abort_called = true;
 }
 
+static void
+buffer_overflow_write(char *ptr, size_t size) {
+	/* Avoid overflow warnings. */
+	volatile size_t idx = 128;
```
Lapenkov, line 19: Why is the size argument needed if it's never used?

interwq: Good catch. Just fixed.

davidtgoldblatt:
Logging in from leave to say: LMAO.

Lapenkov (approved):

PR #2225, opened 2022-02-15
Properly detect background thread support on Darwin.

interwq:
When cross-compile, the host type / abi should be checked to determine
background thread compatibility.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2394,7 +2394,8 @@ fi
 dnl ============================================================================
 dnl Enable background threads if possible.
 
-if test "x${have_pthread}" = "x1" -a "x${je_cv_os_unfair_lock}" != "xyes" ; then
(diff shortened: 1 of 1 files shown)


interwq:
@jeremiahpslewis: could you help to verify this PR fixes the issue you observed?

interwq:
On a related note, we may consider supporting background thread on Darwin. We decided against it as there was OSSpinLock, which was removed in #1367. Now if we assume pthread, it looks the missing part is cpuset support, which isn't critical.

Lapenkov (approved):

jeremiahpslewis:
This works great for me! Thanks!! 

See: https://github.com/JuliaPackaging/Yggdrasil/pull/4381/checks?check_run_id=5201637375

PR #2226, opened 2022-02-15
Fix failed assertion due to racy memory access

Lapenkov:
While calculating the number of stashed pointers, multiple variables
potentially modified by a concurrent thread were used for the
calculation.  This led to some inconsistencies, correctly detected by
the assertions.  The change eliminates some possible inconsistencies by
using unmodified variables and only once a concurrently modified one.
The assertions are omitted for the cases where we acknowledge potential
inconsistencies too.

### DIFF
--- a/include/jemalloc/internal/cache_bin.h
+++ b/include/jemalloc/internal/cache_bin.h
@@ -194,8 +194,15 @@ cache_bin_assert_earlier(cache_bin_t *bin, uint16_t earlier, uint16_t later) {
  * be associated with the position earlier in memory.
  */
 static inline uint16_t
-cache_bin_diff(cache_bin_t *bin, uint16_t earlier, uint16_t later) {
-	cache_bin_assert_earlier(bin, earlier, later);
+cache_bin_diff(cache_bin_t *bin, uint16_t earlier, uint16_t later, bool racy) {
+	/*
+	 * When it's racy, bin->low_bits_full can be modified concurrently. It
+	 * can cross the uint16_t max value and become less than
+	 * bin->low_bits_empty at the time of the check.
+	 */
+	if (!racy) {
+		cache_bin_assert_earlier(bin, earlier, later);
+	}
 	return later - earlier;
 }
 
@@ -207,7 +214,7 @@ cache_bin_diff(cache_bin_t *bin, uint16_t earlier, uint16_t later) {
 static inline cache_bin_sz_t
 cache_bin_ncached_get_internal(cache_bin_t *bin, bool racy) {
 	cache_bin_sz_t diff = cache_bin_diff(bin,
-	    (uint16_t)(uintptr_t)bin->stack_head, bin->low_bits_empty);
+	    (uint16_t)(uintptr_t)bin->stack_head, bin->low_bits_empty, racy);
 	cache_bin_sz_t n = diff / sizeof(void *);
 	/*
 	 * We have undefined behavior here; if this function is called from the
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/cache_bin.h
```diff
@@ -194,8 +194,15 @@ cache_bin_assert_earlier(cache_bin_t *bin, uint16_t earlier, uint16_t later) {
  * be associated with the position earlier in memory.
  */
 static inline uint16_t
-cache_bin_diff(cache_bin_t *bin, uint16_t earlier, uint16_t later) {
-	cache_bin_assert_earlier(bin, earlier, later);
+cache_bin_diff(cache_bin_t *bin, uint16_t earlier, uint16_t later, bool racy) {
```
interwq, line 197: It looks to me that, we can't have the bool racy here? Isn't it already fixed, w/o the bool racy here? 

The other aspect is, if the assertion fails, it means we have an underflow bug, i.e. the diff calculated isn't correct. IIUC, the bug was that `cache_bin_low_bound_get` was calling `cache_bin_empty_position_get`, while `cache_bin_empty_position_get` itself isn't safe to call remotely (since the stack_head might load twice which results in a wrong empty position), does that match what you saw @Lapenkov ?

Lapenkov: This is indeed unrelated to the issue reported, but it's for `cache_bin_ncached_get_internal(bin, racy = true)`.

There, we calculate the difference between `(uint16_t)(uintptr_t)stack_head` and `low_bits_empty`. It can happen that at the moment we calculate the difference, `(uint16_t)(uintptr_t)stack_head > low_bits_empty`, but by the time we reach `assert(bin->low_bits_full > bin->low_bits_empty);`, we could've stashed some items and changed `low_bits_full` so that the assertion fails too.

interwq: Thanks for explaining. I see the point now.

Maybe we should update the assertion in `cache_bin_assert_earlier`? There the `bin->low_bits_full` really was referring to `cache_bin_low_bits_low_bound`. Am I understanding correctly that, if we use the low_bound, we wouldn't have to bypass it for the racy case? (not asking to change to that way, but for sanity checking).

Lapenkov: Actually, yes, you're right. Instead of skipping the assertion, I can calculate the low bits low bound with read-only variables and use it for comparison. This will make the code easier. Will update the PR tomorrow.

interwq: Discussed offline. Let's still keep the current way, i.e. skipping the assertion in the remote reading case.

include/jemalloc/internal/cache_bin.h
```diff
@@ -239,11 +246,15 @@ cache_bin_ncached_get_local(cache_bin_t *bin, cache_bin_info_t *info) {
  * Internal.
  *
  * A pointer to the position one past the end of the backing array.
+ *
+ * Do not call if racy, because both 'bin->stack_head' and 'bin->low_bits_empty'
```
interwq, line 250: `low_bits_empty` should be `low_bits_full` instead?

interwq (approved):

include/jemalloc/internal/cache_bin.h
```diff
@@ -252,6 +263,21 @@ cache_bin_empty_position_get(cache_bin_t *bin) {
...
 }
 
+/*
+ * Internal.
+ *
+ * Calculates low bits of the lower bound of the usable cache bin's range (see
+ * cache_bin_t visual representation above).
+ *
+ * No values are concurrently modified, so should be safe to read in a
+ * multithreaded environment.
```
interwq, line 273: nit: also mention that the multithreaded case is only for stats reading

include/jemalloc/internal/cache_bin.h
Lapenkov: Correct, thank you!

PR #2228, opened 2022-02-17
Test on more FreeBSD versions

lwhsu:


mchoo7:
When `clock_gettime(MONOTONIC_COARSE, *struct timespec` is used on 64bit machine but compiled in 32bit, it sets tv_sec (time_t) as -1. On 32bit FreeBSD, clock_gettime works as expected.

mchoo7:
A [commit](https://cgit.freebsd.org/src/commit/?id=93626d54370292b09cd0ca604b144737109e9071) that fixes this bug is now on `main`. I tested on -CURRENT and it works. I'm not sure when this will be bring into 12-RELEASE or 13-RELEASE, so this pull request should stay until next releases.

**Edit**
12.0-RELEASE and 13.0-RELEASE are unaffected by this problem, so can you please re-run the tests when next 14-CURRENT snapshots comes out? (I guess it will be around August 18-19)

interwq:
Thank you @lwhsu and @fel1x-developer . Merged as #2574

PR #2232, opened 2022-02-23
Fix FreeBSD system jemalloc TSD cleanup

Lapenkov:
Before this commit, in case FreeBSD libc jemalloc was overridden by another
jemalloc, proper thread shutdown callback was involved only for the overriding
jemalloc. A call to _malloc_thread_cleanup from libthr would be redirected to
user jemalloc, leaving data about dead threads hanging in system jemalloc. This
change tackles the issue in two ways. First, for current and old system
jemallocs, which we can not modify, the overriding jemalloc would locate and
invoke system cleanup routine. For upcoming jemalloc integrations, the cleanup
registering function will also be redirected to user jemalloc, which means that
system jemalloc's cleanup routine will be registered in user's jemalloc and a
single call to _malloc_thread_cleanup will be sufficient to invoke both
callbacks.

### DIFF
--- a/src/tsd.c
+++ b/src/tsd.c
@@ -9,9 +9,6 @@
 /******************************************************************************/
 /* Data. */
 
-static unsigned ncleanups;
-static malloc_tsd_cleanup_t cleanups[MALLOC_TSD_CLEANUPS_MAX];
-
 /* TSD_INITIALIZER triggers "-Wmissing-field-initializer" */
 JEMALLOC_DIAGNOSTIC_PUSH
 JEMALLOC_DIAGNOSTIC_IGNORE_MISSING_STRUCT_FIELD_INITIALIZERS
@@ -337,6 +334,9 @@ malloc_tsd_dalloc(void *wrapper) {
 }
 
 #if defined(JEMALLOC_MALLOC_THREAD_CLEANUP) || defined(_WIN32)
+static unsigned ncleanups;
+static malloc_tsd_cleanup_t cleanups[MALLOC_TSD_CLEANUPS_MAX];
+
 #ifndef _WIN32
 JEMALLOC_EXPORT
 #endif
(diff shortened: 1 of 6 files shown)


src/tsd.c
```diff
@@ -360,11 +365,21 @@ _malloc_thread_cleanup(void) {
 			}
 		}
 	} while (again);
+#if defined(JEMALLOC_HAVE_DLSYM) && !defined(_WIN32)
+	void (*libc_thread_cleanup)() = dlsym(RTLD_NEXT,
+	    "_malloc_thread_cleanup");
+	if (libc_thread_cleanup != NULL) {
+		libc_thread_cleanup();
+	}
+#endif
```
interwq, line 374: Discussed offline. IIUC, we are making sure that there is only one `_malloc_tsd_cleanup_register`, just like `_malloc_thread_cleanup`. So if the user jemalloc wins these definitions, the system jemalloc can just register its cleanup callbacks to be handled together by the user jemalloc, is that right?

Also, if we don't want to workaround for older versions, we can avoid the `dlsym` (which would be my preference)?

Lapenkov: Yes, this is correct understanding. And I've removed `dlsym`.

src/tsd.c
```diff
@@ -363,8 +363,11 @@ _malloc_thread_cleanup(void) {
 }
 #endif
```
interwq, line 364: Shall we also include the register() below into this #if (since we are now exporting the symbol)?

interwq (approved):

src/tsd.c
Lapenkov: Yes, I think so, but it has some other stuff that depends on it and needs to be changed. It's the right thing to do, though, so I'll change it and update the PR.

interwq (approved):

PR #2236, opened 2022-03-15
Fix size class calculation for sec

Lapenkov:
Due to a bug in sec initialization, the number of cached size classes
was equal to 198. The bug caused the creation of more than a hundred of
unused bins, although it didn't affect the caching logic.

### DIFF
--- a/src/sec.c
+++ b/src/sec.c
@@ -25,7 +25,7 @@ sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, pai_t *fallback,
     const sec_opts_t *opts) {
 	assert(opts->max_alloc > 0);
 
-	size_t max_alloc = opts->max_alloc & ~PAGE_MASK;
+	size_t max_alloc = PAGE_FLOOR(opts->max_alloc);
 	pszind_t npsizes = sz_psz2ind(max_alloc) + 1;
 
 	size_t sz_shards = opts->nshards * sizeof(sec_shard_t);
--- a/src/sec.c
+++ b/src/sec.c
@@ -23,11 +23,11 @@ sec_bin_init(sec_bin_t *bin) {
 bool
 sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, pai_t *fallback,
     const sec_opts_t *opts) {
-	size_t max_alloc = opts->max_alloc & PAGE_MASK;
-	pszind_t npsizes = sz_psz2ind(max_alloc);
-	if (sz_pind2sz(npsizes) > opts->max_alloc) {
-		npsizes--;
-	}
+	assert(opts->max_alloc > 0);
+
+	size_t max_alloc = opts->max_alloc & ~PAGE_MASK;
+	pszind_t npsizes = sz_psz2ind(max_alloc) + 1;
+
 	size_t sz_shards = opts->nshards * sizeof(sec_shard_t);
 	size_t sz_bins = opts->nshards * (size_t)npsizes * sizeof(sec_bin_t);
 	size_t sz_alloc = sz_shards + sz_bins;
@@ -232,6 +232,8 @@ sec_alloc(tsdn_t *tsdn, pai_t *self, size_t size, size_t alignment, bool zero,
 		    deferred_work_generated);
 	}
 	pszind_t pszind = sz_psz2ind(size);
+	assert(pszind < sec->npsizes);
+
 	sec_shard_t *shard = sec_shard_pick(tsdn, sec);
 	sec_bin_t *bin = &shard->bins[pszind];
 	bool do_batch_fill = false;
@@ -305,6 +307,7 @@ sec_shard_dalloc_and_unlock(tsdn_t *tsdn, sec_t *sec, sec_shard_t *shard,
 	assert(shard->bytes_cur <= sec->opts.max_bytes);
 	size_t size = edata_size_get(edata);
 	pszind_t pszind = sz_psz2ind(size);
+	assert(pszind < sec->npsizes);
 	/*
 	 * Prepending here results in LIFO allocation per bin, which seems
 	 * reasonable.
--- a/include/jemalloc/internal/pages.h
+++ b/include/jemalloc/internal/pages.h
@@ -13,6 +13,9 @@
 /* Return the smallest pagesize multiple that is >= s. */
 #define PAGE_CEILING(s)							\
 	(((s) + PAGE_MASK) & ~PAGE_MASK)
+/* Return the largest pagesize multiple that is <=s. */
+#define PAGE_FLOOR(s) 							\
+	((s) & ~PAGE_MASK)
 
 /* Huge page size.  LG_HUGEPAGE is determined by the configure script. */
 #define HUGEPAGE	((size_t)(1U << LG_HUGEPAGE))
--- a/test/unit/sec.c
+++ b/test/unit/sec.c
@@ -46,6 +46,7 @@ test_sec_init(sec_t *sec, pai_t *fallback, size_t nshards, size_t max_alloc,
 
 	bool err = sec_init(TSDN_NULL, sec, base, fallback, &opts);
 	assert_false(err, "Unexpected initialization failure");
+	assert_u_ge(sec->npsizes, 0, "Zero size classes allowed for caching");
 }
 
 static inline edata_t *


src/sec.c
```diff
@@ -23,11 +23,10 @@ sec_bin_init(sec_bin_t *bin) {
...
     const sec_opts_t *opts) {
-	size_t max_alloc = opts->max_alloc & PAGE_MASK;
-	pszind_t npsizes = sz_psz2ind(max_alloc);
-	if (sz_pind2sz(npsizes) > opts->max_alloc) {
-		npsizes--;
-	}
+	assert((opts->max_alloc & PAGE_MASK) == 0);
+	assert(opts->max_alloc > 0);
+
+	pszind_t npsizes = sz_psz2ind(opts->max_alloc) + 1;
```
davidtgoldblatt, line 29: Maybe another variant of the same question -- I think we should decrement this again if the (possibly-not-a-page-multiple) max_alloc is smaller than its quantization?

Lapenkov:
> Very nice catch, the intent was `& ~PAGE_MASK` on the first line I believe.
> 
> I _think_ that this doesn't actually affect operation, except for scanning some useless empty bins during shard flushing. Can you double-check my thinking and make sure you agree? (In particular, I think that the last line of the commit message isn't quite true).

Right, I lied - the rate of flushes is unchanged, but the scanning for non-empty bins will take longer in some cases. Thanks for the sanity check.

src/sec.c
```diff
@@ -23,11 +23,10 @@ sec_bin_init(sec_bin_t *bin) {
 bool
 sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, pai_t *fallback,
     const sec_opts_t *opts) {
-	size_t max_alloc = opts->max_alloc & PAGE_MASK;
-	pszind_t npsizes = sz_psz2ind(max_alloc);
-	if (sz_pind2sz(npsizes) > opts->max_alloc) {
-		npsizes--;
-	}
+	assert((opts->max_alloc & PAGE_MASK) == 0);
```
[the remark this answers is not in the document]
Lapenkov: I was thinking that if we deal with extents here, it would be easier to force `max_alloc` to be a PAGE multiple, but thinking more about it, you're right. `max_alloc` could still be a PAGE multiple, but fall between two size classes, so this restriction doesn't help.

src/sec.c
Lapenkov: Right, since all `size` inputs of `sec_alloc` will match size classes, if `max_alloc` falls in between two size classes, we should 'truncate' it to the previous size class.

include/jemalloc/internal/sec_opts.h
```diff
@@ -46,7 +46,7 @@ struct sec_opts_s {
 	/* nshards */							\
 	4,								\
 	/* max_alloc */							\
-	32 * 1024,							\
+	8 * PAGE,							\
```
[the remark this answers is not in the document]
Lapenkov: I'll revert this change, because restricting max_alloc to PAGE multiples doesn't make sense.

src/sec.c
```diff
@@ -23,10 +23,15 @@ sec_bin_init(sec_bin_t *bin) {
 bool
 sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, pai_t *fallback,
     const sec_opts_t *opts) {
-	size_t max_alloc = opts->max_alloc & PAGE_MASK;
-	pszind_t npsizes = sz_psz2ind(max_alloc);
-	if (sz_pind2sz(npsizes) > opts->max_alloc) {
-		npsizes--;
+	assert(opts->max_alloc > 0);
+
+	pszind_t npsizes = sz_psz2ind(opts->max_alloc);
```
interwq, line 28: `sz_psz2ind` kinda reads like `page size to ind` to me, so maybe let's only pass page sizes to it? Plus this is an init function so the extra `& ~PAGE_MASK` doesn't really hurt.

src/sec.c
```diff
@@ -23,10 +23,15 @@ sec_bin_init(sec_bin_t *bin) {
...
+	/*
+	 * If max_alloc falls exactly on the size class boundary, we need one
+	 * more bin.
```
interwq, line 31: I'm a bit lost here. Isn't the size class inclusive?

src/sec.c
Lapenkov: I agree, will change.

src/sec.c
Lapenkov: For some reason, I was doing it under the assumption the all extents we would cache had sizes exactly matching the size class boundaries (obviously incorrect). IDK why I thought so.

Let's reiterate - when we do `sz_psz2ind`, it returns the index of the smallest page size class, which covers `max_alloc`. The number of bins should be equal to this number + 1, because the indices are zero-based.

Before the change, we were subtracting one if the size class boundary is larger than `opts->max_alloc`. I don't quite understand why, though. I think I need David's help with that, because I can't figure it out. To me, it seems that we should always add 1.

Some examples (if we always do sz_psz2ind of truncated opts->max_alloc and add one):
* opts->max_alloc = 4096 => max_alloc = 4096, sz_psz2ind = 0, npsizes = 1
* opts->max_alloc = 4097 => max_alloc = 4096, sz_psz2ind = 0, npsizes = 1
* opts->max_alloc = 36900 => max_alloc = 36864, sz_psz2ind = 8, npsizes = 9. Here, I should clarify: the 7th bin is 32768, the 8th bin is 40960. Even though `sz_pind2sz(npsizes) > opts->max_alloc` is true, I don't see why we need to do `npsizes--`, because extent with size `36864` (exactly between two size classes) should also be cached, and it falls into the 8th bin.

davidtgoldblatt: > all extents we would cache had sizes exactly matching the size class boundaries (obviously incorrect)

Is it? I think this was right, and that the original subtraction of 1 is just a bug that never got noticed because the other bug covered it up.

Are we all agreed that the actual logic should be:
1. Round opts->max_alloc down to a page size to obtain the effective max size.
2. Make that the maximum pszind_t we can access (so that the number of bins is that + 1)
?

I think just reimplementing that from scratch will be easier than trying to make sense of whatever nonsense I initially wrote.

Lapenkov: > all extents we would cache had sizes exactly matching the size class boundaries (obviously incorrect)
Yes, we can cache extents with size falling in between the boundaries. For instance, this happens when you do a large allocation and add sz_large_pad. This only happens if you tweak the default `hpa_sec_max_alloc`.

> Are we all agreed that the actual logic should be
Yes, I think that's the correct approach here.

src/sec.c
```diff
@@ -23,11 +23,11 @@ sec_bin_init(sec_bin_t *bin) {
...
 sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, pai_t *fallback,
     const sec_opts_t *opts) {
-	size_t max_alloc = opts->max_alloc & PAGE_MASK;
-	pszind_t npsizes = sz_psz2ind(max_alloc);
-	if (sz_pind2sz(npsizes) > opts->max_alloc) {
-		npsizes--;
-	}
+	assert(opts->max_alloc > 0);
+
+	size_t max_alloc = opts->max_alloc & ~PAGE_MASK;
```
interwq, line 28: Maybe add a wrapper / macro for the `& ~` operation? around here: https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/pages.h#L14

interwq (approved):
LGTM. @davidtgoldblatt in case you want to take a look as well.

src/sec.c
```diff
@@ -23,11 +23,11 @@ sec_bin_init(sec_bin_t *bin) {
 bool
 sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, pai_t *fallback,
     const sec_opts_t *opts) {
-	size_t max_alloc = opts->max_alloc & PAGE_MASK;
-	pszind_t npsizes = sz_psz2ind(max_alloc);
-	if (sz_pind2sz(npsizes) > opts->max_alloc) {
-		npsizes--;
-	}
+	assert(opts->max_alloc > 0);
```
davidtgoldblatt, line 26: Hmm, not sure, but I think max_alloc == 0 should be a valid way of saying "disable caching"? I don't think we handle that at a higher level of the stack.

src/sec.c
```diff
@@ -23,11 +23,11 @@ sec_bin_init(sec_bin_t *bin) {
...
+
+	size_t max_alloc = opts->max_alloc & ~PAGE_MASK;
+	pszind_t npsizes = sz_psz2ind(max_alloc) + 1;
```
davidtgoldblatt, line 29: OK, maybe I'm about to propose the exact same bug I already introduced the first time around, but: shouldn't we have
```
if (sz_pind2psz(sz_psz2ind(max_alloc)) > max_alloc) {
  npsizes--;
}
```
after this?

Otherwise, if  max_alloc is between two size classes, I think we'll allow sz_psz2ind(max_alloc)-sized extents to be cached, violating the max guarantee.

Lapenkov: No, because we decide whether to cache an extent based on `size <= sec->opts.max_alloc`, not `npsizes`. `npsizes` only describes how many bins we have. If `max_alloc` is between bin 9 and bin 10 (zero based), we want to have 11 bins, because otherwise we won't have bin 10, but may need to cache an extent that falls in that bin.

davidtgoldblatt (approved):
Thanks, all makes sense to me.

src/sec.c
Lapenkov: Good suggestion, was thinking about it myself. Applied.

Issue #2213, opened 2022-01-19
5.3 Release Candidate


interwq:
This issue is for tracking progress and communication related to the upcoming 5.3 release.

Current release candidate: ed5fc14 which is going through our production testing pipeline.

A [previous commit](c9ebff0fd6ab90d5eed0d11f48dfedcc21222ab0
) was production deployed in November and no issues were observed.

CC: @jasone

zct:
Is any description or docs about HPA , I want to test this feature in our workloads

interwq:
@zct: thanks for offering help testing. jemalloc's internal huge page allocator (HPA) is still in an early stage and will not be officially introduced in the 5.3 release. We have several binaries using it and showing great results, but also have been tuning and fixing things around. When it reaches a semi-stable state we will share more details in a separate thread (I'll make sure to tag you).

josh-m-sharpe:
Hi - sorry if this is the wrong place.  Where are releases pushed these days?  Latest I see is 5.2.1 from a few years back: https://github.com/jemalloc/jemalloc/releases - thank you!

Lapenkov:
@josh-m-sharpe You're looking at the right place - it's been a while since the previous release.

interwq:
Update: the current [5.3 release candidate](https://github.com/jemalloc/jemalloc/commit/20f9802e4f25922884448d9581c66d76cc905c0c). CC folks who might be interested in testing it before the official release: @gnzlbg, @rkmisra, @hatvik, @rustyx,  @davidlt, @paravoid, @marxin, @strejda, @sharkcz, @ingvarha, @cferris1000, @maksqwe, @trasz, @tamird, @xhochy, @rlibby, @zonyitoo, @AWSjswinney, @yunxuo, @leres, @georgthegreat, @devnexen, @azat, @mweisgut, @wangjl-uos, @pbalcer, @Mingli-Yu, @sangshuduo, @vineetgarc, @DeannaGelbart, @everschen, @AndroidInternal, @jqian-aurora, @ezeeyahoo, @hliu202, @brooksdavis, @tyroguru, @krytarowski, @zoulasc, @ethercflow, @lwhsu, @nickdesaulniers, @santagada, @RingsC, @Ignition, @giridharprasath, @j0t, @ffontaine, @mgrice, @daverigby

marxin:
[...]

PR #2242, opened 2022-03-21
Add a default page size when cross-compile for Apple M1.

interwq:
When cross-compile for M1 and no page size specified, use the default 16K and
skip detecting the page size (which is likely incorrect).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1677,6 +1677,15 @@ fi
 AC_ARG_WITH([lg_page],
   [AS_HELP_STRING([--with-lg-page=<lg-page>], [Base 2 log of system page size])],
   [LG_PAGE="$with_lg_page"], [LG_PAGE="detect"])
+case "${host}" in
+  aarch64-apple-darwin*)
+      dnl When cross-compile for Apple M1 and no page size specified, use the
+      dnl default and skip detecting the page size (which is likely incorrect).
+      if test "x${host}" != "x${build}" -a "x$LG_PAGE" = "xdetect"; then
+        LG_PAGE=14
+      fi
+      ;;
+esac
 if test "x$LG_PAGE" = "xdetect"; then
   AC_CACHE_CHECK([LG_PAGE],
                [je_cv_lg_page],


interwq:
@jeremiahpslewis this PR adds the default page size for cross compiling M1, as mentioned in #2213. To be clear I still suggest specifying page size explicitly when cross compile, since that is the safest way (i.e. obtaining the page size from the target host, and tell that to the configure script). But I agree that this would be more convenient for many people.

Lapenkov (approved):

GuillaumeDIDIER:
This fix things when compiling specifically for M1, but generic aarch64 will keep crashing there, and in practice most software aren't going to make specific builds for M1 / Apple Silicon. Should I open a separate issue ?

interwq:
@GuillaumeDIDIER : we can use this thread to discuss. To confirm, are you thinking about binaries built for generic aarch64, and run them on M1? For the page size setting, you can specific `--with-lg-page=14` when building the aarch64 binary -- it will use the larger page size, but will be compatible with smaller system page sizes. The larger page size may cause a bit extra fragmentation, but at the same time can also speed up some common paths.

jeremiahpslewis:
@interwq Would you consider tagging a new release with this fix?

interwq:
@jeremiahpslewis : it will be included in the upcoming 5.3 release #2213 (which we have been testing for a couple of months now)

GuillaumeDIDIER:
@interwq I'm not the main person behind the Apple Silicon Linux port, but they are trying to minimise differences with stock linux and have as much as possible be upstreamed. (Right now their Alpha distribution is arch + a few custom packages mostly dealing with kernel, driver and boot loader).

As such they definitely are interested in making sure that the generic arch linux arm package will run. I guess the two alternatives are a. get the distro to recompile with `--with-lg-page=14` or to ship a jemalloc library that selects at runtime the correct implementation.

How does jemalloc deal with macOS nowadays ? (given that macOS on ARM uses both 16k and 4k pages)

Chainfire:
@interwq 

Just to clarify, this patch does not fix the problem for Linux/ARM64 on M1, only for macOS. Either run-time page-size detection or making the default 16k for all Linux/ARM64 would.

Odds are most ARM64 packages will not be compiled on M1 and thus use a 4K page-size, and the binaries will crash on M1. Using 16K as default fixes this and allow packages from generic distros to run.

Note that 64K ARM64's also exist, and problems for those would persist.

interwq:
Right. The intention of this PR was to only cover the specific macOS case (since 16K is the common case there).

> Either run-time page-size detection or making the default 16k for all Linux/ARM64 would.

While I agree either can fix the page size issue, they do come at a cost though. The runtime page size support will add runtime overhead on many common paths. At the moment that doesn't look like a good tradeoff to me.

It looks page sizes 4K, 16K and 64K are common for aarch64 on different distros. That means enforcing a larger page size at jemalloc level for all aarch64 would penalize the 4K page case -- they will get a higher fragmentation and it won't be obvious why. While on the other hand, if the page size isn't configured properly, it crashes w/ message right away.

To sum up, for binaries intended to cover multiple platforms, I don't think jemalloc is the right place to solve the page size issue. The `--with-lg-page` configure option is intended for cross compile usage like this (i.e. setting the highest page size needs to be covered, in order to gain the compatibility).

GuillaumeDIDIER:
I think with a little dynamic tricks it could be possible to move the page size check out of the hot path, but it would make such binaries a bit fatter (multiply jemalloc size by 3) :
- Compile jemalloc for the three size,
- When jemmaloc initialise itself, check the page size, make the various function point to the right malloc (function pointer tricks)
- Pretty quickly the CPU will have its predictor trained which should negate the cost of the extra indirect call

It does require implementation of the subtle implementation selection and has a cost in term of size on disk (the non-selected implem will quickly get booted out of the caches if unused).

Any opinion on this compared to the `--with-lg-page` option (which doesn't need much implementation effort, has no code size penalty but may be less optimal at run-time) ?

interwq:
@GuillaumeDIDIER : it's a fun idea but I'm afraid the complexity is too high. Going that direction, it's probably easier to compile different page sizes into separate .so, and have a wrapper to detect page size and then re-launch and LD_PRELOAD the preferred one. Even that might have gone too far IMO.

The penalty of configuring with a larger `--with-lg-page` is mostly on memory fragmentation. In my experience a larger page size like 16K may cause a few % of extra RSS (comparing to 4K pages) for the entire process, however that's also workload dependent. It may also improve CPU usage a bit (larger slabs, plus the extra fragmented pages serve as cache).

Whether the fragmentation and CPU changes matter also depend on the application, if it's a CPU heavy application w/ enough memory to spare, higher page size likely is preferred and can improve throughput. However if it's a small daemon-like binary, or there are many instances running simultaneously, memory usage may become the key metric. Tradeoffs like these make it hard to pick the page size for all of ARM at the jemalloc level.

davidtgoldblatt:
The only place where we *need* a static page size (in the sense of "being on a hot path") is for "maybe-sampled" detection; that's special, in that we're OK with an under-approximation. So being fully dynamic is another possibility, along with a little bit of branching in places like the rtree and the edata_t slab info.

(Still a lot of work, so definitely not on the 5.3 timeline, but doable).

GuillaumeDIDIER:
Would it be reasonable for 5.3 to default to 16k (or larger) --with-lg-page on ARM64 to ensure portability as an interim solution ? (And so any bistro that doesn't do anything special will get the portable version by default) ?

interwq:
I want to clarify that David meant it’s possible to make full dynamic page size work. But some slowdowns here and there are unavoidable, even if we can workaround it on the hottest paths. So please don’t consider it a guaranteed feature. Given the tradeoffs I’d consider it unlikely to happen, and certainly not anytime soon.

Bumping the page size for all arm64 at jemalloc level still doesn’t sound right. The with-lg-page option is an easy way for application maintainers to make this choice for their use cases. But if it’s set at the jemalloc level, we would be penalizing 4K page users —many of them will see it as a fragmentation / memory regression (even worse is, it would be hard to diagnose).

WhyNotHugo:
Is there a separate issue tracking this for Linux? There's a few programs that won't run due to this issue.

I understand that detecting page size at runtime adds some overhead. Would it maybe be possible to do this _only_ on aarch64, given the diversity of page sizes?

PR #2245, opened 2022-03-25
Do not disable SEC by default for 64k pages platforms

Lapenkov:
Default SEC max_alloc option value was 32k, disabling SEC for platforms with
lg-page=16. This change enables SEC for all platforms, making minimum max_alloc
value equal to PAGE.

### DIFF
--- a/src/sec.c
+++ b/src/sec.c
@@ -23,7 +23,7 @@ sec_bin_init(sec_bin_t *bin) {
 bool
 sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, pai_t *fallback,
     const sec_opts_t *opts) {
-	assert(opts->max_alloc > 0);
+	assert(opts->max_alloc >= PAGE);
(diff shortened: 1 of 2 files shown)


src/sec.c
```diff
@@ -23,7 +23,7 @@ sec_bin_init(sec_bin_t *bin) {
 bool
 sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, pai_t *fallback,
     const sec_opts_t *opts) {
-	assert(opts->max_alloc > 0);
+	assert(opts->max_alloc >= PAGE);
```
Lapenkov, line 26: I decided to make this requirement more stringent, because if `opts->max_alloc` is less than PAGE, there's something fishy going on. SEC can be disabled by setting number of shards to zero, otherwise `max_alloc` should have value >= PAGE.

interwq (approved):

PR #2257, opened 2022-04-20
Implement pvalloc replacement

Lapenkov:
Despite being an obsolete function, pvalloc is still present in GLIBC and should
work correctly when jemalloc replaces libc allocator.


Lapenkov:
One thing I'm not sure is how it's supposed to work when size is zero.

src/jemalloc.c
```diff
@@ -3245,6 +3245,45 @@ je_valloc(size_t size) {
...
+	sopts.null_out_result_on_error = true;
+	sopts.min_alignment = PAGE;
+	sopts.oom_string =
+	    "<jemalloc>: Error allocating aligned memory: out of memory\n";
+	sopts.invalid_alignment_string =
+	    "<jemalloc>: Error allocating aligned memory: invalid alignment\n";
+
+	dopts.result = &ret;
+	dopts.num_items = 1;
+	dopts.item_size = PAGE_CEILING(size);
```
interwq, line 3276: This is the only logical difference from valloc, right? If so let's add a short comment.

interwq (approved):

PR #2264, opened 2022-04-28
PerCPU cache

Lapenkov:
An overview of the implementation:
Ccache by itself is very similar to the tcache. It serves size classes on top of tcache_maxclass, there is no overlap in
size classes served by Ccache and Tcache. The maximum size class, served by the ccache is set via `opt_ccache_max`. Ccache can be disabled during the configure time (`--enable-cpu-cache`) or at startup (`opt_ccache`). 

The pointers are stored and allocated out of stack, flush/fill logic is reused from Tcache. Ccache flushes pointers in LIFO order, unlike Tcache. If an allocation overflows/underflows, the thread transfers the Ccache into a 'special state', obtaining the exclusive access to the underlying data structure, and performs refill/flush.

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -156,8 +156,8 @@ arena_malloc(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind, bool zero,
 			return tcache_alloc_large(tsdn_tsd(tsdn), arena,
 			    tcache, size, ind, zero, slow_path);
 		}
-		if (likely(config_cpu_cache && size < ccache_maxclass && arena
-		    && arena_is_auto(arena))) {
+		if (likely(config_cpu_cache && size < ccache_maxclass && (!arena
+		    || arena_is_auto(arena)))) {
 			assert(!tsdn_null(tsdn));
 			return ccache_alloc(tsdn_tsd(tsdn), arena, size, ind,
 			    zero, false);
@@ -285,15 +285,14 @@ arena_dalloc_no_tcache(tsdn_t *tsdn, void *ptr) {
 JEMALLOC_ALWAYS_INLINE void
 arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
     bool slow_path) {
-	if (szind < nhbins) {
-		if (config_prof && unlikely(szind < SC_NBINS)) {
-			arena_dalloc_promoted(tsdn, ptr, tcache, slow_path);
-		} else {
-			tcache_dalloc_large(tsdn_tsd(tsdn), tcache, ptr, szind,
-			    slow_path);
-		}
-	} else if (config_cpu_cache && szind < ccache_maxind) {
-		ccache_free(tsdn_tsd(tsdn), ptr, szind, /* small= */ false);
+	if (config_prof && unlikely(szind < SC_NBINS)) {
+		arena_dalloc_promoted(tsdn, ptr, tcache, slow_path);
+	} else if (szind < nhbins) {
+		tcache_dalloc_large(tsdn_tsd(tsdn), tcache, ptr, szind,
+		    slow_path);
+	} else if (config_cpu_cache && szind < ccache_maxind &&
+	    !ccache_free(tsdn_tsd(tsdn), ptr, szind, /* small= */ false)) {
+		return;
 	} else {
 		edata_t *edata = emap_edata_lookup(tsdn, &arena_emap_global,
 		    ptr);
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -293,7 +293,7 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
 			    slow_path);
 		}
 	} else if (config_cpu_cache && szind < ccache_maxind) {
-		ccache_free(tsdn_tsd(tsdn), ptr, szind, /* small= */ true);
+		ccache_free(tsdn_tsd(tsdn), ptr, szind, /* small= */ false);
 	} else {
 		edata_t *edata = emap_edata_lookup(tsdn, &arena_emap_global,
 		    ptr);
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -10,6 +10,7 @@
 #include "jemalloc/internal/sc.h"
 #include "jemalloc/internal/sz.h"
 #include "jemalloc/internal/ticker.h"
+#include "jemalloc/internal/ccache.h"
 
 static inline arena_t *
 arena_get_from_edata(edata_t *edata) {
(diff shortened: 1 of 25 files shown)


src/ccache.c
```diff
@@ -0,0 +1,523 @@
...
+fallback:
+			return arena_malloc_hard(tsd_tsdn(tsd), arena, size,
+			    ind, zero);
+		}
+	}
+	assert(ret != NULL);
+	if (unlikely(zero)) {
+		size_t usize = sz_index2size(ind);
+		assert(tcache_salloc(tsd_tsdn(tsd), ret) == usize);
+		memset(ret, 0, usize);
```
interwq, line 299: Will want to monitor the cost of this part -- since we are targeting "medium" size classes, the zero pattern there might be different than the small ones. Imagine a recycling pattern which goes through `ccache_bin_refill` often to grab 4K-14K sizes. The refill may grab many fresh zeroed pages, but the zero info is lost when stored in ccache.

src/ccache.c
```diff
@@ -0,0 +1,523 @@
...
+	void *ret = NULL;
+	volatile rseq_t *rseq_abi = &tsd_ccache_tdatap_get(tsd)->rseq_abi;
+	int cpu;
+	ccache_t *ccache;
+	ccache_bin_t *bin;
+restart:
+	cpu = rseq_abi->cpu_id;
+	ccache = ccache_get(cpu);
+	bin = ccache_bin_get(ccache, ind);
+	__asm__ __volatile__ goto(
```
interwq, line 209: What would happen if this runs on a host w/o rseq supported kernel?

src/ccache.c
```diff
@@ -0,0 +1,523 @@
...
+	    /* ------------------- End ------------------- */
+	    "rseq_cs_postcommit%=:\n\t"
+
+	    /* ax == Allocatable and executable section */
+	    ".pushsection __rseq_abort, \"ax\"\n\t"
+	    ".byte 0x0f, 0x1f, 0x05\n\t"
+	    ".long "STRINGIFY(JEMALLOC_RSEQ_SIG)"\n\t"
+	    "rseq_cs_abort%=:\n\t"
+	    /*
+	     * Spin in case of abort - re-fetch per-cpu cache for the updated
```
interwq, line 258: Curious about this: would it simplify things much, if on failure we simply give up / skip ccache?

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -155,8 +156,11 @@ arena_malloc(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind, bool zero,
 			return tcache_alloc_large(tsdn_tsd(tsdn), arena,
 			    tcache, size, ind, zero, slow_path);
 		}
-		/* (size > tcache_maxclass) case falls through. */
-		assert(size > tcache_maxclass);
+		if (likely(size <= ccache_maxclass && arena &&
+		    arena_is_auto(arena))) {
+			return ccache_alloc(tsdn_tsd(tsdn), arena, size, ind,
```
interwq, line 161: nit: add `assert(!tsdn_null(tsdn));` before calling.

src/ccache.c
```diff
@@ -0,0 +1,523 @@
...
+
+	    /* -------------------- Begin ------------------- */
+	    "rseq_cs_start%=:\n\t"
+	    "cmpl %[cpu_id], %[initial_cpu_id]\n\t"
+	    /* Restart in case of CPU id missmatch*/
+	    "jne %l[restart]\n\t"
+
+	    /* rax = head */
+	    "movq %c[head_offset](%[bin]), %%rax\n\t"
+	    /* If in fallback mode, then jump to fallback label */
```
interwq, line 231: Do we use head==null as a flag for the fallback mode? If so let's document when / how that happens.

src/ccache.c
```diff
@@ -0,0 +1,523 @@
...
+
+bool
+ccache_free(tsd_t *tsd, void *ptr, szind_t ind, bool small) {
+	edata_t *edata = emap_edata_lookup(tsd_tsdn(tsd),
+	    &arena_emap_global, ptr);
+	arena_t *origin_arena = arena_get_from_edata(edata);
+	if (!arena_is_auto(origin_arena)) {
+		return false;
+	}
+	if (!ccache_handles_szind(ind)) {
```
interwq, line 314: nit: let's put this branch first, since it's more likely to trigger.

src/ccache.c
```diff
@@ -0,0 +1,523 @@
...
+	      [rseq_cs] "m" (rseq_abi->rseq_cs),
+	      [cpu_id] "m" (rseq_abi->cpu_id_start),
+	      [initial_cpu_id] "r" (cpu),
+	      [ccache] "rm" (ccache),
+	      [ptr] "r" (ptr)
+	    : "rax", "cc"
+	    : restart, fallback
+	 );
+
+	if (bin->head == NULL) {
```
interwq, line 391: Question: it looks this indicates both refill and flush mode. If the flushing thread gets preempted here, and another thread does a lot of activities then trigger a refill -- would that be an issue? e.g. ccache_bin_flush may encounter an empty bin, or refill encounter a full bin.

src/ccache.c
Lapenkov: There's TODO in `tsd_ccache_init` - the initialization should fail. I still haven't implemented the handler for this case. I currently picture it as follows: either we disable percpu cache in runtime, or abort if `opt.abort` is `true`.

src/ccache.c
Lapenkov: Yes, `head==null` means fallback mode. I'll document the state machine once it's finalized.

src/ccache.c
Lapenkov: Don't think it would simplify things. The complexity stems from the concurrent access to the underlying data structure (stack), and the source of concurrent access is possible concurrent refilling and attempting to operate the stack. If we remove retries, that would mean if a thread is preempted during the critical section, it can't retry the access. This only removes one jump instruction to `restart` label, everything else stays the same.

src/ccache.c
Lapenkov: That is a nice observation, I'll keep an eye on it.

src/ccache.c
Lapenkov: It should be impossible.
1) There is only one thread attempting to alloc/free from a specific cache bin at a time, in a critical section.
2) The result of the critical section is either preemption (no side effects) or a commit, which atomically changes stack->head to one of the following states: (+=1; -=1; =0).
3) Only the thread, which has changed head to 0, is capable of reaching `if (bin->head == NULL)`. Then it's the owner of the refilling/flushing.
4) Other threads, scheduled on the same CPU and entering the critical section, will see `head == null` by executing `test %%rax, %%rax` and realize that the bin is in the fallback mode, therefore there must be another thread that has put the bin into this state, and that thread is responsible for refilling/flushing the bin. Thus, the current thread must not try allocating from the ccache.
5) Once the thread, responsible for refilling/flushing, finishes the procedure, it makes the bin accessible to other threads by storing the non-zero value to `bin->head`.

Lapenkov: Actually, you were right, this was a bug. Sorry for the misleading previous reply.

Good thing is that it was caught by the fuzzy test. Making the scheduler switch more aggressively also helps.

yiguolei:
@Lapenkov  Hi any process on this? We are expecting this feature.

Lapenkov:
Alas, I don't actively participate in jemalloc's development anymore.

The main blocker for this change was that I didn't have enough time to evaluate the performance implications of the feature. In the current state of this MR, it should be ready for such a test, though. Maybe you can build jemalloc with this MR and report what performance impact you see in your software?

But generally, I'd assume that this MR is frozen unless it becomes a priority for the maintainers.

everschen:
such big feature was abandoned, any known performance issue?

everschen:
I have rebased this PR to latest dev, solve the conflict issues and merge the 16 commits into 1, I will try to do more test base on this draft PR, any comments are welcome.
https://github.com/jemalloc/jemalloc/pull/2568

interwq:
@everschen as mentioned on the other PR, we still plan to come back to this idea. Though we are focusing on huge pages first and will take some time there.

Issue #2223, opened 2022-02-13
zero_realloc option naming


davidtgoldblatt:
Now that `realloc(ptr, 0)` is UB per C23, calling the mode that acts like malloc(0) "strict" (originally meant in the vein of "strictly conforming to the standard semantics") is sort of weird. I think 5.3 will be the first release with this option; maybe we can sneak in a better name under the wire.

Maybe "malloc", so that the options are `zero_realloc:malloc|free|abort`. That has kind of a nice symmetry; in the first case, realloc(ptr, 0) is sort of like a malloc; in the second, realloc(ptr, 0) is like a free, and in the last, realloc(ptr, 0) is an abort.

I dunno, maybe that's overly cute.

interwq:
I like the `malloc|free|abort` naming, quite straightforward. It's true that 5.3 will be the first version to include it.

We have the option documented in the manual, as expected. https://github.com/jemalloc/jemalloc/blob/dev/doc/jemalloc.xml.in#L1575
But given that it's UB in C23, do you think we should make it experimental or so, maybe even undocument it @davidtgoldblatt?

davidtgoldblatt:
I think we should keep documenting it; basically every platform has *some* guarantee that they document. It's only UB because different platforms chose *different* behaviors, so that portable code can't assume anything; we should probably do the same (and just let users tell us what they expect).

It's probably also worth picking the default to try and match the platform (e.g. zero_realloc:free by default on Linux, zero_realloc:strict/malloc on BSDs, etc.)

interwq:
After the most recent discussion, possible names for the first option: `malloc`, `alloc`, `realloc_1`.

Lapenkov:
Thinking about it retrospectively, I'd say I like `realloc_1` the most, because that is literally what it does.

Lapenkov:
@interwq @davidtgoldblatt if you both like `realloc_1` and don't deem it unacceptably weird (`realloc_zero:realloc_1` indeed sounds quite weird if you don't know the context), I can change it.

davidtgoldblatt:
Honestly every time I see this I come to a different conclusion. realloc_1 is semantically different from realloc(ptr, 1), because in the former you're not allowed to use the pointer (even if it's non-null, you're not allowed to dereference it).

My final naming suggestions: `realloc_zero:allocates|deallocates|aborts`? (Or maybe "frees" instead of "deallocates"?) Describes the behavior rather than trying to summarize it in terms of a different function?

[...]

PR #2270, opened 2022-05-03
Make the default option of zero realloc match the system allocator.

interwq:
More context in #2223

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -638,6 +638,7 @@ dnl Define cpp macros in CPPFLAGS, rather than doing AC_DEFINE(macro), since the
 dnl definitions need to be seen before any headers are included, which is a pain
 dnl to make happen otherwise.
 default_retain="0"
+zero_realloc_default_free="0"
 maps_coalesce="1"
 DUMP_SYMS="${NM} -a"
 SYM_PREFIX=""
@@ -684,6 +685,7 @@ case "${host}" in
 	if test "${LG_SIZEOF_PTR}" = "3"; then
 	  default_retain="1"
 	fi
+	zero_realloc_default_free="1"
 	;;
   *-*-linux*)
 	dnl syscall(2) and secure_getenv(3) are exposed by _GNU_SOURCE.
@@ -698,6 +700,7 @@ case "${host}" in
 	if test "${LG_SIZEOF_PTR}" = "3"; then
 	  default_retain="1"
 	fi
+	zero_realloc_default_free="1"
 	;;
   *-*-kfreebsd*)
 	dnl syscall(2) and secure_getenv(3) are exposed by _GNU_SOURCE.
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -653,12 +654,14 @@ case "${host}" in
 	SOREV="${rev}.${so}"
 	sbrk_deprecated="1"
 	SYM_PREFIX="_"
+	zero_realloc_default_alloc="1"
```
davidtgoldblatt, line 657: I don't know that this is right, but I'll trust that you checked it.

configure.ac
```diff
@@ -638,6 +638,7 @@ dnl Define cpp macros in CPPFLAGS, rather than doing AC_DEFINE(macro), since the
 dnl definitions need to be seen before any headers are included, which is a pain
 dnl to make happen otherwise.
 default_retain="0"
+zero_realloc_default_alloc="0"
```
davidtgoldblatt, line 641: I might invert the direction of this test; assume it's `alloc` by default, and change to `free` only if we know it's the platform behavior

(e.g. on gnu linux it should be `free`, on musl linux it should be `alloc`; defaulting to the safe-but-leaky choice).

configure.ac
interwq: Yeah the [Mac OS man page](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/realloc.3.html) is clear: `If size is zero and ptr is not NULL, a new,
     minimum sized object is allocated and the original object is freed.`

configure.ac
interwq: Updated

configure.ac
```diff
@@ -684,6 +685,7 @@ case "${host}" in
 	if test "${LG_SIZEOF_PTR}" = "3"; then
 	  default_retain="1"
 	fi
+	zero_realloc_default_free="1"
```
interwq, line 688: Verified that Scudo is doing the "free" option: https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/scudo/scudo_allocator.cpp#L720

Lapenkov (approved):

configure.ac
```diff
@@ -698,6 +700,7 @@ case "${host}" in
...
```
davidtgoldblatt, line 703: Last suggestion I might make is to check if it's Linux *and* glibc here; e.g. musl does `alloc` by default.

interwq: Yeah I remember your comment earlier on this. However at this point I'm hesitant to make that change (plan to publish the release tomorrow), since we don't have a glibc / musl check already in place -- 10 lines above we simply declare `glibc="1"` under Linux.

davidtgoldblatt: Fair enough, seems reasonable

PR #2275, opened 2022-05-09
Fix compiling edata.h with MSVC

georgthegreat:
At the time an attempt to compile jemalloc 5.3.0 with MSVC 2019 results in the following error message:

> jemalloc/include/jemalloc/internal/edata.h:660: error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax


Lapenkov:
Out of curiosity, is it complaining while building `src/jemalloc_cpp.cpp`?

georgthegreat:
> Out of curiosity, is it complaining while building `src/jemalloc_cpp.cpp`?

Indeed, this is the source that causes problems.

Lapenkov:
Got it! Thanks for the contribution!

PR #2280, opened 2022-05-20
malloc_getcpu implementation fallback for amd64 and arm64 (macOs only

devnexen:
for now).

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_inlines_a.h
+++ b/include/jemalloc/internal/jemalloc_internal_inlines_a.h
@@ -15,8 +15,8 @@ malloc_getcpu(void) {
 #elif defined(JEMALLOC_HAVE_SCHED_GETCPU)
 	return (malloc_cpuid_t)sched_getcpu();
 #elif defined(HAVE_RDTSCP)
-	unsigned int dx;
-	asm volatile("rdtscp" : "=d"(dx) :: "memory");
+	unsigned int ax, cx, dx;
+	asm volatile("rdtscp" : "=a"(ax), "=d"(dx), "=c"(cx) ::);
 	return (malloc_cpuid_t)(dx & 0xfff);
 #elif defined(__aarch64__) && defined(__APPLE__)
 	// TODO: other oses most likely use tpidr_el0 instead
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/jemalloc_internal_inlines_a.h
```diff
@@ -14,6 +14,15 @@ malloc_getcpu(void) {
 	return GetCurrentProcessorNumber();
 #elif defined(JEMALLOC_HAVE_SCHED_GETCPU)
 	return (malloc_cpuid_t)sched_getcpu();
+#elif defined(HAVE_RDTSCP)
+	unsigned int dx;
+	asm volatile("rdtscp" : "=d"(dx) :: "memory");
```
Lapenkov: 1. Why does it have the constraint "=d" and not "=r" like the next one?
2. Why does it have "memory" clobber? It doesn't seem to modify anything other than the output operand.

devnexen: 1. for this case, we re only interested in (r)dx.
2. You re right it's pointless in this case.

include/jemalloc/internal/jemalloc_internal_inlines_a.h
```diff
@@ -14,6 +14,15 @@ malloc_getcpu(void) {
...
 	return (malloc_cpuid_t)sched_getcpu();
+#elif defined(HAVE_RDTSCP)
+	unsigned int ax, cx, dx;
+	asm volatile("rdtscp" : "=a"(ax), "=d"(dx), "=c"(cx) ::);
+	return (malloc_cpuid_t)(dx & 0xfff);
+#elif defined(__aarch64__) && defined(__APPLE__)
+	// TODO: other oses most likely use tpidr_el0 instead
+	uintptr_t c;
+	asm volatile("mrs %x0, tpidrro_el0" : "=r"(c) :: "memory");
+	return (malloc_cpuid_t)(c & (1 << 3) -1);
```
Lapenkov, line 25: 1. Did you figure this out by reading `cswitch.s`? Just wondering if it's documented somewhere.
2. Is memory clobber needed here?

Lapenkov:
Are you planning to allow per-cpu arena on MacOS after this change?

include/jemalloc/internal/jemalloc_internal_inlines_a.h
devnexen: not officially documented at all :-)

devnexen:
> Are you planning to allow per-cpu arena on MacOS after this change?

was part of the plan indeed.

Lapenkov:
Merged as https://github.com/jemalloc/jemalloc/commit/df8f7d10af15d549ab73ba807b2e14a9d7fe1cc2

PR #2284, opened 2022-06-01
install: do not use "install -v", it's not portable

0-wiz-0:
At least NetBSD's install(1) does not support this.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -555,18 +555,18 @@ endif
 install_bin:
 	$(INSTALL) -d $(BINDIR)
 	@for b in $(BINS); do \
-	$(INSTALL) -v -m 755 $$b $(BINDIR); \
+	$(INSTALL) -m 755 $$b $(BINDIR); \
 done
 
 install_include:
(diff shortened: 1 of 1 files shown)


Lapenkov:
What a pity! It's way nicer than to echo every command. But let's resort to what we had before to still keep it verbose:
https://github.com/jemalloc/jemalloc/commit/f15d8f3b416f6812ac030bc1a7aacf05927a4d7f

0-wiz-0:
Thanks for comment, @Lapenkov.
Yes, a pity indeed.
Should I change my pull request or will you just revert your commit?

Lapenkov:
> Thanks for comment, @Lapenkov. Yes, a pity indeed. Should I change my pull request or will you just revert your commit?

Right, I just reverted my commit: https://github.com/jemalloc/jemalloc/commit/df7ad8a9b6121c5c4b15bad5606b51bf734416a6

Thank you for bringing this up!

BusyJay:
Hi, can we get a release for this?

PR #2285, opened 2022-06-01
new configure option `--enable-pageid` used for Linux only for now.

devnexen:


src/pages.c
```diff
@@ -21,6 +21,13 @@
 #else
 #define PAGES_FD_TAG -1
 #endif
+#ifdef JEMALLOC_HAVE_PRCTL
+#include <sys/prctl.h>
+#ifndef PR_SET_VMA
+#define PR_SET_VMA 0x53564d41
+#define PR_SET_VMA_ANON_NAME 0
+#endif
```
Lapenkov, line 29: If `PR_SET_VMA` is undefined, doesn't that mean that the kernel doesn't support it? Maybe it should be a configuration error if `PR_SET_VMA` is undefined, but `enable-pageid` is `yes`?

src/pages.c
```diff
@@ -98,6 +105,21 @@ static int madvise_MADV_DONTNEED_zeroes_pages()
...
 #endif
 
+#ifdef JEMALLOC_PAGEID
+static int os_page_id(void *addr, size_t size, const char *name)
+{
+#ifdef JEMALLOC_HAVE_PRCTL
+	/*
+	 * While parsing `/proc/<pid>/maps` file, the block could appear as
+	 * 7f4836000000-7f4836800000 rw-p 00000000 00:00 0 [anon:jemalloc_pg_overcommit]`
+	 */
```
Lapenkov, line 115: That's very nice, I've wanted to see this feature for a long time.

src/pages.c
devnexen: at worse it s a no-op, in some cases the kernel supports it but not defined in the header (my case).

Lapenkov: But then `os_page_id` will return `-1` and fail the assertion you added.

Lapenkov:
@devnexen I highly appreciate your contributions. I think they will be even better if you follow some simple guidelines for jemalloc development:
1. Add commit description to explain the purpose of the changes. Sure, I often can figure it out by reading the PR, but that would save some time.
2. You already follow most of the implicit coding style requirements, except for this: lines are limited by 80 characters.

I'll be very grateful if you reflect these two items in your future contributions.

src/pages.c
devnexen: ah true :-)

Lapenkov:
Merged as https://github.com/jemalloc/jemalloc/commit/4fc5c4fbac156c9f44452d3f30216451711dfa18

src/pages.c
```diff
@@ -162,6 +185,11 @@ os_pages_map(void *addr, size_t size, size_t alignment, bool *commit) {
 #endif
 	assert(ret == NULL || (addr == NULL && ret != addr) || (addr != NULL &&
 	    ret == addr));
+#ifdef JEMALLOC_PAGEID
+	int n = os_page_id(ret, size,
+        os_overcommits ? "jemalloc_pg_overcommit" : "jemalloc_pg");
```
azat, line 190: What are the benefit of such naming? You can see if overcommit is enabled or not manually.

PR #2286, opened 2022-06-07
OpenBSD build fix. still no cpu affinity.

devnexen:


Lapenkov:
Need more context about what exactly it fixes and what it has to do with cpu affinity.

src/background_thread.c
```diff
@@ -80,6 +80,7 @@ background_thread_info_init(tsdn_t *tsdn, background_thread_info_t *info) {
 
 static inline bool
 set_current_thread_affinity(int cpu) {
+#ifndef __OpenBSD__
```
interwq: Maybe the other way around, like ifdef OPENBSD  return false?

PR #2287, opened 2022-06-08
Enable retain by default on macOS.

interwq:
@krallin and @brekhus identified that jemalloc w/o retain on macOS will suffer extremely high fork() overhead (can reach tens of seconds), because the cost is quadratic with the number of mappings / VM "holes".  Retain fixes the issue, at a small cost of extra VM space reserved.

cc: @devnexen


Lapenkov (approved):

devnexen:
works fine for me

PR #2291, opened 2022-06-11
enabled percpu_arena settings on macOs.

devnexen:
follow-up on #2280

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -376,6 +376,9 @@ jobs:
     - os: osx
       arch: amd64
       env: CC=gcc CXX=g++ CONFIGURE_FLAGS="--with-malloc-conf=tcache:false" EXTRA_CFLAGS="-Werror -Wno-array-bounds -Wno-unknown-warning-option -Wno-ignored-attributes -Wno-deprecated-declarations"
+    - os: osx
+      arch: amd64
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -376,6 +376,9 @@ jobs:
     - os: osx
       arch: amd64
       env: CC=gcc CXX=g++ CONFIGURE_FLAGS="--with-malloc-conf=tcache:false" EXTRA_CFLAGS="-Werror -Wno-array-bounds -Wno-unknown-warning-option -Wno-ignored-attributes -Wno-deprecated-declarations"
+    - os: osx
```
interwq, line 379: This config file is generated by ./scripts/gen_travis.py. You will want to edit it in that script to include the option for mac.

scripts/gen_travis.py
```diff
@@ -257,7 +257,6 @@ def generate_macos(arch):
 
     exclude = ([Option.as_malloc_conf(opt) for opt in (
             'dss:primary',
-            'percpu_arena:percpu',
```
interwq, line 260: Forgot to mention, you will need to run the script, which will update the config to include the change.

interwq:
Thanks for the patch!

PR #2292, opened 2022-06-13
Update building for Windows instructions

Lapenkov:
Explain how to build for Windows in INSTALL.md and remove another readme.txt in
an obscure location.


INSTALL.md
```diff
@@ -1,3 +1,4 @@
+
```
interwq: Intentional?

interwq (approved):
Looks great!

INSTALL.md
Lapenkov: Definitely not, thanks :)

PR #2296, opened 2022-06-18
CI update FreeBSD version.

devnexen:


interwq:
Thanks for the patch!

PR #2303, opened 2022-07-02
Remove duplicated words in documentation

charlesdarkwind:


interwq:
Thanks for the patch!

PR #2310, opened 2022-07-19
[WIP] Add double free detection in thread cache for debug build

izaitsevfb:
Add new runtime option `debug_dblfree_cbin_scan_limit` that specifies the limit of for number of stack entries to scan in cache bit when trying to detect double free (debug build only).

### Context:

Double free is a common pattern to cause heap corruption. Typically we expect ASAN to catch these, however it's not always feasible because of the high CPU & memory overhead of ASAN. Thus we still rely on the debug build of jemalloc to catch things sometimes.

The thread cache (tcache) is critical to performance since it serves as the common and fast path for the majority of allocations. By default the debug build has tcache on, so that it's not too slow -- however the downside is, it could easily hide double frees. Currently the sanity checks may still discover double frees, but only down the arena path where the original troubled free is in the past.

The first step of the double free checking, is to scan the existing items within the tcache and see if there's any match.

### DIFF
--- a/include/jemalloc/internal/cache_bin.h
+++ b/include/jemalloc/internal/cache_bin.h
@@ -2,6 +2,7 @@
 #define JEMALLOC_INTERNAL_CACHE_BIN_H
 
 #include "jemalloc/internal/ql.h"
+#include "jemalloc/internal/safety_check.h"
 #include "jemalloc/internal/sz.h"
 
 /*
@@ -428,6 +429,35 @@ cache_bin_full(cache_bin_t *bin) {
 }
 
 /*
+ * Scans the allocated area of the cache_bin for the given pointer up to limit.
+ * Fires safety_check_fail if the ptr is found and returns true.
+ */
+JEMALLOC_ALWAYS_INLINE bool
+cache_bin_dalloc_safety_checks(cache_bin_t *bin, void *ptr) {
+	if (!config_debug || opt_debug_double_free_max_scan == 0) {
+		return false;
+	}
+
+	cache_bin_sz_t ncached = cache_bin_ncached_get_internal(bin, false);
+	unsigned max_scan = opt_debug_double_free_max_scan < ncached
+	    ? opt_debug_double_free_max_scan
+	    : ncached;
+
+	void **cur = bin->stack_head;
+	void **limit = cur + max_scan;
+	for (; cur < limit; cur++) {
+		if (*cur == ptr) {
+			safety_check_fail(
+			    "Invalid deallocation detected: double free of "
+			    "pointer %p\n",
+			    ptr);
+			return true;
+		}
+	}
+	return false;
+}
+
+/*
  * Free an object into the given bin.  Fails only if the bin is full.
  */
 JEMALLOC_ALWAYS_INLINE bool
@@ -435,6 +465,10 @@ cache_bin_dalloc_easy(cache_bin_t *bin, void *ptr) {
 	if (unlikely(cache_bin_full(bin))) {
 		return false;
 	}
+
+        if (unlikely(cache_bin_dalloc_safety_checks(bin, ptr))) {
+                return true;
+        }
 
 	bin->stack_head--;
 	*bin->stack_head = ptr;
--- a/include/jemalloc/internal/jemalloc_internal_externs.h
+++ b/include/jemalloc/internal/jemalloc_internal_externs.h
@@ -34,6 +34,7 @@ extern malloc_init_t malloc_init_state;
 extern const char *zero_realloc_mode_names[];
 extern atomic_zu_t zero_realloc_count;
 extern bool opt_cache_oblivious;
+extern unsigned opt_debug_double_free_max_scan;
 
 /* Escape free-fastpath when ptr & mask == 0 (for sanitization purpose). */
 extern uintptr_t san_cache_bin_nonfast_mask;
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -154,6 +154,9 @@ fxp_t		opt_narenas_ratio = FXP_INIT_INT(4);
 
 unsigned	ncpus;
 
+unsigned opt_debug_double_free_max_scan =
+    SAFETY_CHECK_DOUBLE_FREE_MAX_SCAN_DEFAULT;
+
 /* Protects arenas initialization. */
 malloc_mutex_t arenas_lock;
 
@@ -1420,6 +1423,10 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			CONF_HANDLE_UNSIGNED(opt_lg_tcache_flush_large_div,
 			    "lg_tcache_flush_large_div", 1, 16,
 			    CONF_CHECK_MIN, CONF_CHECK_MAX, /* clip */ true)
+			CONF_HANDLE_UNSIGNED(opt_debug_double_free_max_scan,
+			    "debug_double_free_max_scan", 0, UINT_MAX,
+			    CONF_DONT_CHECK_MIN, CONF_DONT_CHECK_MAX,
+			    /* clip */ false)
 
 			/*
 			 * The runtime option of oversize_threshold remains
@@ -1736,6 +1743,10 @@ malloc_conf_init_check_deps(void) {
 		malloc_printf("<jemalloc>: prof_leak_error is set w/o "
 		    "prof_final.\n");
 		return true;
+	}
+	/* To emphasize in the stats output that opt is disabled when !debug. */
+	if (!config_debug) {
+		opt_debug_double_free_max_scan = 0;
 	}
 
 	return false;
--- a/test/unit/double_free.c
+++ b/test/unit/double_free.c
@@ -10,13 +10,14 @@ void fake_abort(const char *message) {
 }
 
 void
-test_large_double_free_pre(void) {
+test_double_free_pre(void) {
 	safety_check_set_abort(&fake_abort);
 	fake_abort_called = false;
 }
 
 void
-test_large_double_free_post() {
+test_double_free_post() {
+	mallctl("thread.tcache.flush", NULL, NULL, NULL, 0);
 	expect_b_eq(fake_abort_called, true, "Double-free check didn't fire.");
 	safety_check_set_abort(NULL);
 }
@@ -29,7 +30,7 @@ TEST_BEGIN(test_large_double_free_tcache) {
 	 */
 	test_skip_if(config_debug);
 
-	test_large_double_free_pre();
+	test_double_free_pre();
 	char *ptr = malloc(SC_LARGE_MINCLASS);
 	bool guarded = extent_is_guarded(tsdn_fetch(), ptr);
 	free(ptr);
@@ -43,8 +44,7 @@ TEST_BEGIN(test_large_double_free_tcache) {
 		 */
 		fake_abort_called = true;
 	}
-	mallctl("thread.tcache.flush", NULL, NULL, NULL, 0);
-	test_large_double_free_post();
+	test_double_free_post();
 }
 TEST_END
 
@@ -52,7 +52,7 @@ TEST_BEGIN(test_large_double_free_no_tcache) {
 	test_skip_if(!config_opt_safety_checks);
 	test_skip_if(config_debug);
 
-	test_large_double_free_pre();
+	test_double_free_pre();
 	char *ptr = mallocx(SC_LARGE_MINCLASS, MALLOCX_TCACHE_NONE);
 	bool guarded = extent_is_guarded(tsdn_fetch(), ptr);
 	dallocx(ptr, MALLOCX_TCACHE_NONE);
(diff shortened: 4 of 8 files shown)


include/jemalloc/internal/cache_bin.h
```diff
@@ -436,6 +461,12 @@ cache_bin_dalloc_easy(cache_bin_t *bin, void *ptr) {
 		return false;
 	}
 
+        if (config_debug && opt_debug_dblfree_cbin_scan_limit != 0) {
+                if (cache_bin_validate_double_free(bin, ptr)) {
```
interwq: With the change mentioned above, this can just become `if (cache_bin_dalloc_safety_checks(...)) {`

so that the dalloc_easy impl isn't directly exposed to the details of the safety_checks.

include/jemalloc/internal/cache_bin.h
```diff
@@ -427,6 +428,30 @@ cache_bin_full(cache_bin_t *bin) {
 	return ((uint16_t)(uintptr_t)bin->stack_head == bin->low_bits_full);
 }
 
+/*
+ * Scans the allocated area of the cache_bin for the given pointer up to limit.
+ * Fires safety_check_fail if the ptr is found and returns false.
+ */
+JEMALLOC_ALWAYS_INLINE bool
+cache_bin_validate_double_free(cache_bin_t *bin, void *ptr) {
```
interwq: nit: let's name it `cache_bin_dalloc_safety_checks()`, and add the following logic at the beginning:

```
if (!config_debug || opt_debug_double_free_max_scan == 0) {
    return false;
}
```

include/jemalloc/internal/cache_bin.h
```diff
@@ -427,6 +428,30 @@ cache_bin_full(cache_bin_t *bin) {
...
+                cur++;
+                limit--;
+        }
+
+        // ptr is already present in the stack
+        if (limit != 0 && ((uint16_t)(uintptr_t)cur != bin->low_bits_full)) {
+                safety_check_fail(
+                    "Invalid deallocation detected: double free\n");
+                return true;
+        }
```
interwq: With the `max_scan == 0` check at the beginning, we can simplify above a bit:
```
 while (limit != 0 && (uint16_t)(uintptr_t)cur != bin->low_bits_full) {
    if (*cur == ptr) {
        ...
    }
    limit--;
    cur++;
}
```

include/jemalloc/internal/cache_bin.h
```diff
@@ -427,6 +428,30 @@ cache_bin_full(cache_bin_t *bin) {
...
+        while (limit != 0 && ((uint16_t)(uintptr_t)cur != bin->low_bits_full)
+            && *cur != ptr) {
+                cur++;
+                limit--;
+        }
+
+        // ptr is already present in the stack
+        if (limit != 0 && ((uint16_t)(uintptr_t)cur != bin->low_bits_full)) {
+                safety_check_fail(
+                    "Invalid deallocation detected: double free\n");
```
interwq: Let's also output the ptr value in the failure message.

include/jemalloc/internal/jemalloc_internal_externs.h
```diff
@@ -34,6 +34,7 @@ extern malloc_init_t malloc_init_state;
 extern const char *zero_realloc_mode_names[];
 extern atomic_zu_t zero_realloc_count;
 extern bool opt_cache_oblivious;
+extern unsigned opt_debug_dblfree_cbin_scan_limit;
```
interwq: Maybe `opt_debug_double_free_max_scan`

src/jemalloc.c
```diff
@@ -1420,6 +1422,9 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			CONF_HANDLE_UNSIGNED(opt_lg_tcache_flush_large_div,
 			    "lg_tcache_flush_large_div", 1, 16,
 			    CONF_CHECK_MIN, CONF_CHECK_MAX, /* clip */ true)
+			CONF_HANDLE_UNSIGNED(opt_debug_dblfree_cbin_scan_limit,
+			    "debug_dblfree_cbin_scan_limit", 0, 1024,
+                            CONF_DONT_CHECK_MIN, CONF_CHECK_MAX, /* clip */ true)
```
interwq: No need to really limit the max, i.e. set it to UINT_MAX, and CONF_DONT_CHECK_MAX

If the user wants to set a max value to scan as many as possible, let's allow it.

src/jemalloc.c
```diff
@@ -154,6 +154,8 @@ fxp_t		opt_narenas_ratio = FXP_INIT_INT(4);
 
 unsigned	ncpus;
 
+unsigned	opt_debug_dblfree_cbin_scan_limit;
```
interwq: For setting the default value, here you can do `unsigned opt_debug_double_free_max_scan = SAFETY_CHECK_DOUBLE_FREE_MAX_SCAN_DEFAULT;`
and then in `safety_check.h` have
```
#define SAFETY_CHECK_DOUBLE_FREE_MAX_SCAN_DEFAULT 32
```

Also let's add this in https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L1734
```
if (!config_debug) {
    opt_debug_double_free_max_scan = 0;
}
```

It doesn't really affect anything, but should make the stats output clearer (that it's disabled).

test/unit/double_free.c
```diff
@@ -70,8 +70,30 @@ TEST_BEGIN(test_large_double_free_no_tcache) {
 }
 TEST_END
 
+TEST_BEGIN(test_double_free_tcache) {
+    test_skip_if(!config_debug);
+
+    test_large_double_free_pre();
```
interwq: nit: Let's remove the `large` part from the _pre and _post functions names. You will also want to skip the test if tcache is disabled: https://github.com/jemalloc/jemalloc/blob/d660683d3ddc2aaebf41a5662a6bc629be016e6d/test/unit/uaf.c#L51

test/unit/double_free.c
```diff
@@ -70,8 +70,30 @@ TEST_BEGIN(test_large_double_free_no_tcache) {
...
+    char *ptr = malloc(SC_NTINY);
```
interwq: malloc(1) is fine. SC_NTINY is actually the number of tiny size classes.

test/unit/double_free.c
```diff
@@ -70,8 +70,30 @@ TEST_BEGIN(test_large_double_free_no_tcache) {
 }
 TEST_END
 
+TEST_BEGIN(test_double_free_tcache) {
```
interwq: nit: The name `test_double_free_small_tcache`

include/jemalloc/internal/cache_bin.h
izaitsevfb: very elegant!

include/jemalloc/internal/cache_bin.h
izaitsevfb: addressed in [a903674](https://github.com/jemalloc/jemalloc/pull/2310/commits/a903674db54ec8e8fc00b588e4e4279b0e0cd965)

include/jemalloc/internal/cache_bin.h
izaitsevfb: addressed in [a903674](https://github.com/jemalloc/jemalloc/pull/2310/commits/a903674db54ec8e8fc00b588e4e4279b0e0cd965)

include/jemalloc/internal/cache_bin.h
izaitsevfb: addressed in [a903674](https://github.com/jemalloc/jemalloc/pull/2310/commits/a903674db54ec8e8fc00b588e4e4279b0e0cd965)

include/jemalloc/internal/cache_bin.h
izaitsevfb: addressed in [a903674](https://github.com/jemalloc/jemalloc/pull/2310/commits/a903674db54ec8e8fc00b588e4e4279b0e0cd965)

include/jemalloc/internal/jemalloc_internal_externs.h
izaitsevfb: addressed in [5a2502a](https://github.com/jemalloc/jemalloc/pull/2310/commits/5a2502a2dd9fb3e401373e087983f334ed35dbc4)

src/jemalloc.c
izaitsevfb: addressed in [5a2502a](https://github.com/jemalloc/jemalloc/pull/2310/commits/5a2502a2dd9fb3e401373e087983f334ed35dbc4)

src/jemalloc.c
izaitsevfb: addressed in [5a2502a](https://github.com/jemalloc/jemalloc/pull/2310/commits/5a2502a2dd9fb3e401373e087983f334ed35dbc4)

test/unit/double_free.c
izaitsevfb: addressed in [4bc176c](https://github.com/jemalloc/jemalloc/pull/2310/commits/4bc176c3002fa751f267c7a420d38a8a2df2e712)

test/unit/double_free.c
izaitsevfb: addressed in [4bc176c](https://github.com/jemalloc/jemalloc/pull/2310/commits/4bc176c3002fa751f267c7a420d38a8a2df2e712)

test/unit/double_free.c
izaitsevfb: addressed in [4bc176c](https://github.com/jemalloc/jemalloc/pull/2310/commits/4bc176c3002fa751f267c7a420d38a8a2df2e712)

davidtgoldblatt:
I probably won't do an in-depth review, but one comment from a skim: I think this is a lot more useful/powerful if it's not debug-mode-only (e.g. what's the point if you have to re-build / re-link to get the checking; if you're rebuilding anyways, you might as well rebuild with ASAN and get way better checking.). We can get that for free by folding the option check into the fast-path check. So, the cache-bin code won't read the global directly; instead it'll get passed in as a static parameter down fast paths (and a dynamic one down slow paths).

The `opt.junk` and `opt.zero` handling is perhaps a good example of how to do that.

interwq:
@davidtgoldblatt agreed; that will be one of the followup items around this topic (the other being checking the slab bitmap in addition to scanning the stack).

include/jemalloc/internal/cache_bin.h
```diff
@@ -436,6 +465,10 @@ cache_bin_dalloc_easy(cache_bin_t *bin, void *ptr) {
 		return false;
 	}
 
+        if (cache_bin_dalloc_safety_checks(bin, ptr)) {
```
interwq: nit: let's mark this `if (unlikely(...))`

include/jemalloc/internal/cache_bin.h
```diff
@@ -427,6 +428,34 @@ cache_bin_full(cache_bin_t *bin) {
...
+JEMALLOC_ALWAYS_INLINE bool
+cache_bin_dalloc_safety_checks(cache_bin_t *bin, void *ptr) {
+        if (!config_debug || opt_debug_double_free_max_scan == 0) {
+                return false;
+        }
+
+        void **cur = bin->stack_head;
+        unsigned limit = opt_debug_double_free_max_scan;
+
+        while (limit != 0 && (uint16_t)(uintptr_t)cur != bin->low_bits_empty) {
```
interwq: nit: Let's try optimizing a bit further -- for debug build it actually doesn't matter much; however we may add it to opt build later, also it can be a good example.
```
cache_bin_sz_t ncached = cache_bin_ncached_get_internal(bin, false);
unsigned max_scan = opt_debug_double_free_max_scan < ncached ? opt_debug_double_free_max_scan : ncached;

void **cur = bin->stack_head;
for (void **limit = cur + max_scan; cur < limit; cur++) {
    if (*cur == ptr) {
        ...
    }
}
```
This way we can avoid 1 branch check per iteration.

For the comments we always do `/* */`. It can be either one line, or multiple line format like you have for this function.

test/unit/double_free.c
```diff
@@ -44,15 +44,15 @@ TEST_BEGIN(test_large_double_free_tcache) {
 		fake_abort_called = true;
 	}
 	mallctl("thread.tcache.flush", NULL, NULL, NULL, 0);
```
interwq: Let's move this to `_post()`, given that it's needed in the new test as well.

test/unit/double_free.c
```diff
@@ -66,12 +66,43 @@ TEST_BEGIN(test_large_double_free_no_tcache) {
...
+        mallctl("thread.tcache.flush", NULL, NULL, NULL, 0);
+        test_double_free_post();
 }
 TEST_END
 
 int
 main(void) {
-	return test(test_large_double_free_no_tcache,
-	    test_large_double_free_tcache);
+        return test(test_large_double_free_no_tcache,
```
interwq: The indent change probably unintended? It should remain tabs.

For the new test below, the convention here is to add it in a separate line.

interwq:
Travis CI has a few failures but all on their side, none related to this diff.

test/unit/double_free.c
izaitsevfb: I screwed up the indent style in my changes (used only spaces instead of tabs + spaces), I'll reformat and force-push everything to fix that.

test/unit/double_free.c
izaitsevfb: Do I understand correctly that it's ok to call `mallctl("thread.tcache.flush", NULL, NULL, NULL, 0);`  in `test_large_double_free_no_tcache` as well?
https://github.com/jemalloc/jemalloc/blob/4bc176c3002fa751f267c7a420d38a8a2df2e712/test/unit/double_free.c#L69

interwq: Yes that's fine.

izaitsevfb: addressed in [60a3574](https://github.com/jemalloc/jemalloc/pull/2310/commits/60a357485181d0cb1f45c6c04a7641992e9178ff)

include/jemalloc/internal/cache_bin.h
izaitsevfb: addressed in [60a3574](https://github.com/jemalloc/jemalloc/pull/2310/commits/60a357485181d0cb1f45c6c04a7641992e9178ff)

include/jemalloc/internal/cache_bin.h
izaitsevfb: addressed in [60a3574](https://github.com/jemalloc/jemalloc/pull/2310/commits/60a357485181d0cb1f45c6c04a7641992e9178ff)

test/unit/double_free.c
izaitsevfb: I went through all my changes and made sure that the indentation and comment style is correct:
https://github.com/jemalloc/jemalloc/compare/60a357485181d0cb1f45c6c04a7641992e9178ff..eb070173cceeaf545c6ffb1460facb3659b3ffb3

interwq:
Tweaked a bit and merged as 36366f3c4c741723369853c923e56999716398fc with your authorship @izaitsevfb. Thank you for working on this!

PR #2315, opened 2022-08-10
Add double free detection using slab bitmap for debug build

guangli-dai:
Added 
- a sanity check for double free issue in the arena in case that the tcache has been flushed
- a corresponding unit test

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -298,6 +298,54 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
 	}
 }
 
+/* Find the region index of a pointer. */
+JEMALLOC_ALWAYS_INLINE size_t
+arena_slab_regind_impl(div_info_t* div_info, szind_t binind,
+    edata_t *slab, const void *ptr) {
+	size_t diff, regind;
+
+	/* Freeing a pointer outside the slab can cause assertion failure. */
+	assert((uintptr_t)ptr >= (uintptr_t)edata_addr_get(slab));
+	assert((uintptr_t)ptr < (uintptr_t)edata_past_get(slab));
+	/* Freeing an interior pointer can cause assertion failure. */
+	assert(((uintptr_t)ptr - (uintptr_t)edata_addr_get(slab)) %
+	    (uintptr_t)bin_infos[binind].reg_size == 0);
+
+	diff = (size_t)((uintptr_t)ptr - (uintptr_t)edata_addr_get(slab));
+
+	/* Avoid doing division with a variable divisor. */
+	regind = div_compute(div_info, diff);
+	assert(regind < bin_infos[binind].nregs);
+	return regind;
+}
+
+/* Checks whether ptr is currently active in the arena. */
+JEMALLOC_ALWAYS_INLINE bool
+arena_tcache_dalloc_small_safety_check(tsdn_t *tsdn, void *ptr) {
+	if (!config_debug) {
+		return false;
+	}
+	edata_t *edata = emap_edata_lookup(tsdn, &arena_emap_global, ptr);
+	szind_t binind = edata_szind_get(edata);
+	div_info_t div_info = arena_binind_div_info[binind];
+	/*
+	 * Calls the internal function arena_slab_regind_impl because the
+	 * safety check does not require a lock.
+	 */
+	size_t regind = arena_slab_regind_impl(&div_info, binind, edata, ptr);
+	slab_data_t *slab_data = edata_slab_data_get(edata);
+	const bin_info_t *bin_info = &bin_infos[binind];
+	assert(edata_nfree_get(edata) < bin_info->nregs);
+	if (unlikely(!bitmap_get(slab_data->bitmap, &bin_info->bitmap_info,
+	    regind))) {
+		safety_check_fail(
+		    "Invalid deallocation detected: the pointer being freed (%p) not "
+		    "currently active, possibly caused by double free bugs.\n", ptr);
+		return true;
+	}
+	return false;
+}
+
 JEMALLOC_ALWAYS_INLINE void
 arena_dalloc(tsdn_t *tsdn, void *ptr, tcache_t *tcache,
     emap_alloc_ctx_t *caller_alloc_ctx, bool slow_path) {
@@ -328,6 +376,9 @@ arena_dalloc(tsdn_t *tsdn, void *ptr, tcache_t *tcache,
 
 	if (likely(alloc_ctx.slab)) {
 		/* Small allocation. */
+		if (arena_tcache_dalloc_small_safety_check(tsdn, ptr)) {
+			return;
+		}
 		tcache_dalloc_small(tsdn_tsd(tsdn), tcache, ptr,
 		    alloc_ctx.szind, slow_path);
 	} else {
@@ -415,6 +466,9 @@ arena_sdalloc(tsdn_t *tsdn, void *ptr, size_t size, tcache_t *tcache,
 
 	if (likely(alloc_ctx.slab)) {
 		/* Small allocation. */
+		if (arena_tcache_dalloc_small_safety_check(tsdn, ptr)) {
+			return;
+		}
 		tcache_dalloc_small(tsdn_tsd(tsdn), tcache, ptr,
 		    alloc_ctx.szind, slow_path);
 	} else {
@@ -465,22 +519,7 @@ struct arena_dalloc_bin_locked_info_s {
 JEMALLOC_ALWAYS_INLINE size_t
 arena_slab_regind(arena_dalloc_bin_locked_info_t *info, szind_t binind,
     edata_t *slab, const void *ptr) {
-	size_t diff, regind;
-
-	/* Freeing a pointer outside the slab can cause assertion failure. */
-	assert((uintptr_t)ptr >= (uintptr_t)edata_addr_get(slab));
-	assert((uintptr_t)ptr < (uintptr_t)edata_past_get(slab));
-	/* Freeing an interior pointer can cause assertion failure. */
-	assert(((uintptr_t)ptr - (uintptr_t)edata_addr_get(slab)) %
-	    (uintptr_t)bin_infos[binind].reg_size == 0);
-
-	diff = (size_t)((uintptr_t)ptr - (uintptr_t)edata_addr_get(slab));
-
-	/* Avoid doing division with a variable divisor. */
-	regind = div_compute(&info->div_info, diff);
-
-	assert(regind < bin_infos[binind].nregs);
-
+	size_t regind = arena_slab_regind_impl(&info->div_info, binind, slab, ptr);
 	return regind;
 }
 
--- a/test/unit/double_free.c
+++ b/test/unit/double_free.c
@@ -21,6 +21,15 @@ test_double_free_post() {
 	safety_check_set_abort(NULL);
 }
 
+bool tcache_enabled() {
+	bool enabled;
+	size_t sz = sizeof(enabled);
+	assert_d_eq(
+	    mallctl("thread.tcache.enabled", &enabled, &sz, NULL, 0), 0,
+	    "Unexpected mallctl failure");
+	return enabled;
+}
+
 TEST_BEGIN(test_large_double_free_tcache) {
 	test_skip_if(!config_opt_safety_checks);
 	/*
@@ -72,15 +81,8 @@ TEST_END
 
 TEST_BEGIN(test_small_double_free_tcache) {
 	test_skip_if(!config_debug);
-
 	test_skip_if(opt_debug_double_free_max_scan == 0);
-
-	bool tcache_enabled;
-	size_t sz = sizeof(tcache_enabled);
-	assert_d_eq(
-	    mallctl("thread.tcache.enabled", &tcache_enabled, &sz, NULL, 0), 0,
-	    "Unexpected mallctl failure");
-	test_skip_if(!tcache_enabled);
+	test_skip_if(!tcache_enabled());
 
 	test_double_free_pre();
 	char *ptr = malloc(1);
@@ -101,10 +103,40 @@ TEST_BEGIN(test_small_double_free_tcache) {
 }
 TEST_END
 
+TEST_BEGIN(test_small_double_free_arena) {
+	test_skip_if(!config_debug);
+	test_skip_if(!tcache_enabled());
+
+	test_double_free_pre();
+	/*
+	 * Allocate one more pointer to keep the slab partially used after
+	 * flushing the cache.
+	 */
+	char *ptr1 = malloc(1);
+	char *ptr = malloc(1);
+	bool guarded = extent_is_guarded(tsdn_fetch(), ptr);
+	free(ptr);
+	if (!guarded) {
+		mallctl("thread.tcache.flush", NULL, NULL, NULL, 0);
+		free(ptr);
+	} else {
+		/*
+		 * Skip because guarded extents may unguard immediately on
+		 * deallocation, in which case the second free will crash before
+		 * reaching the intended safety check.
+		 */
+		fake_abort_called = true;
+	}
+	test_double_free_post();
+	free(ptr1);
+}
+TEST_END
+
 int
 main(void) {
 	return test(
 	    test_large_double_free_no_tcache,
 	    test_large_double_free_tcache,
-	    test_small_double_free_tcache);
+	    test_small_double_free_tcache,
+	    test_small_double_free_arena);
 }


include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,47 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
+}
+
+JEMALLOC_ALWAYS_INLINE void
+arena_dalloc_small_sanity_check(tsdn_t *tsdn, void *ptr) {
+	edata_t *edata = emap_edata_lookup(tsdn, &arena_emap_global,
+	    ptr);
+	arena_t *arena = arena_get_from_edata(edata);
+	szind_t binind = edata_szind_get(edata);
+	unsigned binshard = edata_binshard_get(edata);
+	bin_t *bin = arena_get_bin(arena, binind, binshard);
```
interwq: bin doesn't seem to be used.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,47 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
 }
 
+static inline bin_t *
+arena_get_bin(arena_t *arena, szind_t binind, unsigned binshard) {
+	bin_t *shard0 = (bin_t *)((uintptr_t)arena + arena_bin_offsets[binind]);
+	return shard0 + binshard;
+}
+
+JEMALLOC_ALWAYS_INLINE size_t
+arena_slab_regind_unlocked(div_info_t* div_info, szind_t binind,
```
interwq: Let's refactor to avoid duplicate the logic for this function. You can add a impl function which is called by the original and this new one. The impl function will contain the shared logic and asserts.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Removed.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Since arena_slab_regind only contains one extra assert compared to arena_slab_regind_unlocked, I make arena_slab_regind directly call arena_slab_regind_unlocked to reduce the complexity. Does that work?

interwq:
Discussed offline.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Updated and changed the function name to arena_slab_regind_impl

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -465,20 +514,8 @@ struct arena_dalloc_bin_locked_info_s {
...
-	    (uintptr_t)bin_infos[binind].reg_size == 0);
-
-	diff = (size_t)((uintptr_t)ptr - (uintptr_t)edata_addr_get(slab));
-
-	/* Avoid doing division with a variable divisor. */
-	regind = div_compute(&info->div_info, diff);
-
+	size_t regind = arena_slab_regind_impl(&info->div_info, binind,
+	    slab, ptr);
 	assert(regind < bin_infos[binind].nregs);
```
interwq: nit: move this line into the _impl function too.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,52 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
+	regind = div_compute(div_info, diff);
+	return regind;
+}
+
+/*
+* Checks whether ptr is already deallocated in the arena.
+* Fires safety check if it has been deallocated and returns true.
+*/
+JEMALLOC_ALWAYS_INLINE bool
+arena_dalloc_small_sanity_check(tsdn_t *tsdn, void *ptr) {
```
interwq: nit: naming -- let's stick to `_safety_check`.

Also in the comments and the PR title, let's avoid using the term "in the arena", since there are several things at the arena level. Probably say something like "using slab bitmap".

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,52 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
+	}
+	edata_t *edata = emap_edata_lookup(tsdn, &arena_emap_global,
+	    ptr);
+	szind_t binind = edata_szind_get(edata);
+	div_info_t div_info = arena_binind_div_info[binind];
+	size_t regind = arena_slab_regind_impl(&div_info, binind, edata, ptr);
+	slab_data_t *slab_data = edata_slab_data_get(edata);
+	const bin_info_t *bin_info = &bin_infos[binind];
+	assert(edata_nfree_get(edata) < bin_info->nregs);
+	if (bitmap_get(slab_data->bitmap, &bin_info->bitmap_info, regind)==0) {
```
interwq: nit: format `regind) == false)` (not using 0 because we avoid implicit conversion, and bitmap_get returns a bool).

Let's also mark this branch as `unlikely`, even though it's debug build only for now.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -328,6 +374,9 @@ arena_dalloc(tsdn_t *tsdn, void *ptr, tcache_t *tcache,
 
 	if (likely(alloc_ctx.slab)) {
 		/* Small allocation. */
+		if (arena_dalloc_small_sanity_check(tsdn, ptr)) {
+			return;
+		}
```
interwq, line 381: It feels this fits better in `tcache_dalloc_small`? It's in fact not doing the arena_dalloc_small path. Maybe move everything over there and rename to tcache_dalloc_small_saftey_check?

test/unit/double_free.c
```diff
@@ -101,10 +101,44 @@ TEST_BEGIN(test_small_double_free_tcache) {
...
+
+	bool tcache_enabled;
+	size_t sz = sizeof(tcache_enabled);
+	assert_d_eq(
+	    mallctl("thread.tcache.enabled", &tcache_enabled, &sz, NULL, 0), 0,
+	    "Unexpected mallctl failure");
+	test_skip_if(!tcache_enabled);
+
+	test_double_free_pre();
+	char *ptr1 = malloc(1);
```
interwq, line 115: nit: explain why this is needed.

test/unit/double_free.c
```diff
@@ -101,10 +101,44 @@ TEST_BEGIN(test_small_double_free_tcache) {
...
+	test_skip_if(!config_debug);
+
+	test_skip_if(opt_debug_double_free_max_scan == 0);
+
+	bool tcache_enabled;
+	size_t sz = sizeof(tcache_enabled);
+	assert_d_eq(
+	    mallctl("thread.tcache.enabled", &tcache_enabled, &sz, NULL, 0), 0,
+	    "Unexpected mallctl failure");
+	test_skip_if(!tcache_enabled);
```
interwq: nit: now that this section is used twice in this file, make it a separate function.

guangli-dai: Added a function to club the tcache enabling. Yet test_skip_if cannot be called outside the test so still keep the lines 113 and 115

test/unit/double_free.c
guangli-dai: Comment added.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Renamed but the function kept in the same place to avoid circular dependency.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Modified.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Renamed

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Moved

test/unit/double_free.c
interwq: nit: Let's do `test_skip_if(tcache_enabled());` (rename the function because enable_tcache() implies it's doing something instead of checking a state). Also probably remove the blank lines between the 3 test_skips here.

test/unit/double_free.c
```diff
@@ -101,10 +105,44 @@ TEST_BEGIN(test_small_double_free_tcache) {
...
+	test_skip_if(opt_debug_double_free_max_scan == 0);
+
+	bool tcache_enabled;
+	enable_tcache(&tcache_enabled);
+	test_skip_if(!tcache_enabled);
+
+	test_double_free_pre();
+	/* Allocate one more pointer so that the cache won't be empty after
+	 * flushing.
+	*/
```
interwq: nit: Format is off. For comments that's only one word over a line, I'd say try trimming it down to fit in one line.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,61 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
 	}
 }
 
+/*
+ * Internal.
+ *
+ * Find the region index of a pointer.
+*/
```
interwq: nit: format (the stars should align). Please check the comments format of other places too. Also no need to say "Internal" in this case, since the `_impl` name implies that too.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,61 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
+		return false;
+	}
+	edata_t *edata = emap_edata_lookup(tsdn, &arena_emap_global,
+	    ptr);
+	szind_t binind = edata_szind_get(edata);
+	div_info_t div_info = arena_binind_div_info[binind];
+	/* Calls the internal function arena_slab_regind_impl because the safety check
+	 * does not require a lock.
+	*/
+	size_t regind = arena_slab_regind_impl(&div_info, binind, edata, ptr);
```
interwq, line 335: It looks to me that, we can simply call `arena_slab_regind` here? If so we don't even have to add the _impl anymore?

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,61 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
+	slab_data_t *slab_data = edata_slab_data_get(edata);
+	const bin_info_t *bin_info = &bin_infos[binind];
+	assert(edata_nfree_get(edata) < bin_info->nregs);
+	if (unlikely(bitmap_get(slab_data->bitmap, &bin_info->bitmap_info, regind)==false)) {
+		safety_check_fail(
+		    "Invalid deallocation detected: double free of "
+		    "pointer %p\n",
+		    ptr);
```
interwq: nit: this doesn't have to take multiple lines.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: `arena_slab_regind` requires a lock since its first variable taken is `arena_dalloc_bin_locked_info_t`. For the safety check, there is no need to acquire a lock, so we are using `arena_slab_regind_impl`, which does not require a lock.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: modified.

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: modified.

test/unit/double_free.c
guangli-dai: trimmed and formatted

include/jemalloc/internal/arena_inlines_b.h
interwq: Ah I see. Now I remember our discussion. Thanks for the clarification.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,56 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
+	/*
+	 * Calls the internal function arena_slab_regind_impl because the
+	 * safety check does not require a lock.
+	 */
+	size_t regind = arena_slab_regind_impl(&div_info, binind, edata, ptr);
+	slab_data_t *slab_data = edata_slab_data_get(edata);
+	const bin_info_t *bin_info = &bin_infos[binind];
+	assert(edata_nfree_get(edata) < bin_info->nregs);
+	if (unlikely(bitmap_get(slab_data->bitmap, &bin_info->bitmap_info, regind)
+	    ==false)) {
```
interwq: nit: instead of `== false`, let's do `!bitmap_get(...)`. When it's a boolean type already, there will be no implicit conversion.

test/unit/double_free.c
```diff
@@ -21,6 +21,15 @@ test_double_free_post() {
 	safety_check_set_abort(NULL);
 }
 
+bool enable_tcache() {
```
interwq: nit: Yeah I meant to rename this function to `bool tcache_enabled()`, then the name itself shows its a state checking (i.e. `is tcache enabled?` instead of `this enables tcache`).

test/unit/double_free.c
```diff
@@ -21,6 +21,15 @@ test_double_free_post() {
...
+	bool tcache_enabled;
+	size_t sz = sizeof(tcache_enabled);
+	assert_d_eq(
+	    mallctl("thread.tcache.enabled", &tcache_enabled, &sz, NULL, 0), 0,
+	    "Unexpected mallctl failure");
+	return !tcache_enabled;
```
interwq: Shouldn't this return `tcache_enabled` instead of `!tcache_enabled`?

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Modified

test/unit/double_free.c
guangli-dai: Changed the function name, also change the name of the internal variable to avoid the function and the variable to have the same name for safety

test/unit/double_free.c
guangli-dai: Yes, `return tcache_enabled` makes more sense to me. Modified and tweaked the logics when invoking as well.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,56 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
+	/* Avoid doing division with a variable divisor. */
+	regind = div_compute(div_info, diff);
+	assert(regind < bin_infos[binind].nregs);
+	return regind;
+}
+
+/*
+* Checks whether ptr is already deallocated in the arena.
+* Fires safety check if it has been deallocated and returns true.
+*/
```
interwq: nit: the format -- stars should all align (like the section you have below).

Also probably remove the second line -- the "returns true" part can cause confusion, since in actual workload / non unit tests, the safety check failure never returns to that point at all.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -298,6 +298,56 @@ arena_dalloc_large(tsdn_t *tsdn, void *ptr, tcache_t *tcache, szind_t szind,
...
+	 * safety check does not require a lock.
+	 */
+	size_t regind = arena_slab_regind_impl(&div_info, binind, edata, ptr);
+	slab_data_t *slab_data = edata_slab_data_get(edata);
+	const bin_info_t *bin_info = &bin_infos[binind];
+	assert(edata_nfree_get(edata) < bin_info->nregs);
+	if (unlikely(!bitmap_get(slab_data->bitmap, &bin_info->bitmap_info,
+	    regind))) {
+		safety_check_fail(
+		    "Invalid deallocation detected: double free of pointer %p\n", ptr);
```
interwq: This looks to be over 80 chars? Sorry when I said make it one line, I meant to fill the line as close to 80 as possible, then move on to the next. For this one, something like this example https://github.com/jemalloc/jemalloc/blob/deb8e62a837b6dd303128a544501a7dc9677e47a/include/jemalloc/internal/arena_inlines_b.h#L226:

```
        safety_check_fail("example_message_till_80_chars______"
            "more messages here...\n", ptr);
```

The message itself, maybe `Invalid deallocation detected: the pointer being freed (%p) not currently active, possibly caused by double free bugs.`. The reason being, it can simply be a corrupted pointer (which isn't caused by double frees).

test/unit/double_free.c
```diff
@@ -101,10 +103,38 @@ TEST_BEGIN(test_small_double_free_tcache) {
 }
 TEST_END
 
+TEST_BEGIN(test_small_double_free_arena) {
+	test_skip_if(!config_debug);
+	test_skip_if(opt_debug_double_free_max_scan == 0);
+	test_skip_if(!tcache_enabled());
+
+	test_double_free_pre();
+	/* Allocate one more pointer to avoid empty cache after flushing. */
```
interwq: > avoid empty cache after flushing

The cache will indeed be empty after flushing. Maybe say "to keep the slab partially used".

interwq (commented):
Only a few formatting nits

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: It is exactly 79. But after changing the text, it will be surely over 80 so I will make it two lines.

test/unit/double_free.c
guangli-dai: Modified

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: Updated

include/jemalloc/internal/arena_inlines_b.h
guangli-dai: indeed, modified and the second line removed, since it may simply be a corrupted pointer instead of already deallocated, modified the comment as well.

test/unit/double_free.c
```diff
@@ -101,10 +103,41 @@ TEST_BEGIN(test_small_double_free_tcache) {
 }
 TEST_END
 
+TEST_BEGIN(test_small_double_free_arena) {
+	test_skip_if(!config_debug);
+	test_skip_if(opt_debug_double_free_max_scan == 0);
```
interwq: It looks we don't have to depend on the scan check for double frees?

guangli-dai: Yes, removed.

PR #2317, opened 2022-08-12
Refactor cache_bin.h 

### LINKED ISSUE #2227: Refactor the remote / cross thread cache bin stats reading
Now that bin->stack_head and bin->low_bits_full can both be modified, we have the `bool racy` propagated quite extensively in `cache_bin.h`. Since the only `racy == true` case is for stats reading only, i.e. in `cache_bin_nitems_get_remote`, and unlikely to have new use cases, we might as well specialize that single function, i.e. to contain all the complexity in a single place, instead of having to ask "is this helper function safe to use for remote access". For example, `cache_bin_empty_position_get` isn't safe for the racy case, as pointed out in #2226.

We should be able to have only `cache_bin_nitems_get_remote` worry about the racy case, and simply say that all other utility functions are only for thread local usage.

guangli-dai:
Refactor cache_bin.h so that only one function is racy. 
Resolves #2227 
Current version based on PR #2315

### DIFF
--- a/include/jemalloc/internal/cache_bin.h
+++ b/include/jemalloc/internal/cache_bin.h
@@ -195,27 +195,18 @@ cache_bin_assert_earlier(cache_bin_t *bin, uint16_t earlier, uint16_t later) {
  * be associated with the position earlier in memory.
  */
 static inline uint16_t
-cache_bin_diff(cache_bin_t *bin, uint16_t earlier, uint16_t later, bool racy) {
-	/*
-	 * When it's racy, bin->low_bits_full can be modified concurrently. It
-	 * can cross the uint16_t max value and become less than
-	 * bin->low_bits_empty at the time of the check.
-	 */
-	if (!racy) {
-		cache_bin_assert_earlier(bin, earlier, later);
-	}
+cache_bin_diff(cache_bin_t *bin, uint16_t earlier, uint16_t later) {
+	cache_bin_assert_earlier(bin, earlier, later);
 	return later - earlier;
 }
 
 /*
  * Number of items currently cached in the bin, without checking ncached_max.
- * We require specifying whether or not the request is racy or not (i.e. whether
- * or not concurrent modifications are possible).
  */
 static inline cache_bin_sz_t
-cache_bin_ncached_get_internal(cache_bin_t *bin, bool racy) {
+cache_bin_ncached_get_internal(cache_bin_t *bin) {
 	cache_bin_sz_t diff = cache_bin_diff(bin,
-	    (uint16_t)(uintptr_t)bin->stack_head, bin->low_bits_empty, racy);
+	    (uint16_t)(uintptr_t)bin->stack_head, bin->low_bits_empty);
 	cache_bin_sz_t n = diff / sizeof(void *);
 	/*
 	 * We have undefined behavior here; if this function is called from the
@@ -226,7 +217,7 @@ cache_bin_ncached_get_internal(cache_bin_t *bin, bool racy) {
 	 * fast paths.  This should still be "safe" in the sense of generating
 	 * the correct assembly for the foreseeable future, though.
 	 */
-	assert(n == 0 || *(bin->stack_head) != NULL || racy);
+	assert(n == 0 || *(bin->stack_head) != NULL);
 	return n;
 }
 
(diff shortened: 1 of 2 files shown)


include/jemalloc/internal/cache_bin.h
```diff
@@ -501,60 +488,62 @@ cache_bin_stash(cache_bin_t *bin, void *ptr) {
...
 	assert(n <= cache_bin_info_ncached_max(info));
 	return n;
 }
 
 /*
  * Obtain a racy view of the number of items currently in the cache bin, in the
  * presence of possible concurrent modifications.
+ *
+ * Note that this is the only racy function. Any other functions are assumed
+ * to be non-racy.
```
interwq: We will want to beef up the comments here -- now that all the complexity is within this scope, it should contain all the information too, e.g. the comment for different functions that were removed, some should be tweaked and added here.

include/jemalloc/internal/cache_bin.h
```diff
@@ -501,60 +488,62 @@ cache_bin_stash(cache_bin_t *bin, void *ptr) {
...
  */
 static inline void
 cache_bin_nitems_get_remote(cache_bin_t *bin, cache_bin_info_t *info,
     cache_bin_sz_t *ncached, cache_bin_sz_t *nstashed) {
-	cache_bin_sz_t n = cache_bin_ncached_get_internal(bin, /* racy */ true);
+	cache_bin_sz_t diff = bin->low_bits_empty -
```
interwq, line 546: Let's try making it very clear what's happening in this function, e.g. adding `Racy version of cache_bin_ncached_get_internal` and `Racy version of cache_bin_nstashed_get_internal` before the corresponding logic.

Also, make it clear that this function should not call other utility functions and why. For the ones that it does call, explain why it's safe to do so.

guangli-dai: Added.

include/jemalloc/internal/cache_bin.h
guangli-dai: More comments added to explain values that could be concurrently modified because it's racy.

include/jemalloc/internal/cache_bin.h
```diff
@@ -501,60 +488,74 @@ cache_bin_stash(cache_bin_t *bin, void *ptr) {
 
 /*
  * Get the number of stashed pointers.
- *
- * When called from a thread not owning the TLS (i.e. racy = true), it's
- * important to keep in mind that 'bin->stack_head' and 'bin->low_bits_full' can
- * be modified concurrently and almost none assertions about their values can be
- * made.
  */
```
interwq: nit: single line format now.

include/jemalloc/internal/cache_bin.h
```diff
@@ -501,60 +488,74 @@ cache_bin_stash(cache_bin_t *bin, void *ptr) {
...
 	assert(n <= cache_bin_info_ncached_max(info));
 	return n;
 }
 
 /*
  * Obtain a racy view of the number of items currently in the cache bin, in the
  * presence of possible concurrent modifications.
+ *
+ * Note that this is the only racy function in this header. Any other functions
+ * are assumed to be non-racy.
```
interwq: Let's add this part: `The "racy" term here means accessed from another thread (that is not the owner of the specific cache bin).  This only happens when gathering stats (read-only).  The only change because of the racy condition is that assertions based on mutable fields are omitted.`

One small thing to keep in mind -- two spaces between sentences.

include/jemalloc/internal/cache_bin.h
```diff
@@ -501,60 +488,74 @@ cache_bin_stash(cache_bin_t *bin, void *ptr) {
...
+	cache_bin_sz_t diff = bin->low_bits_empty -
+	    (uint16_t)(uintptr_t)bin->stack_head;
+	cache_bin_sz_t n = diff / sizeof(void *);
+
 	assert(n <= cache_bin_info_ncached_max(info));
 	*ncached = n;
 
-	n = cache_bin_nstashed_get_internal(bin, info, /* racy */ true);
+	/* Racy version of cache_bin_nstashed_get_internal. */
+	cache_bin_sz_t ncached_max = cache_bin_info_ncached_max(info);
```
interwq: This can be removed, along with the assertion below (identical to the one next to it).

include/jemalloc/internal/cache_bin.h
guangli-dai: Added and formatted

include/jemalloc/internal/cache_bin.h
guangli-dai: Removed

include/jemalloc/internal/cache_bin.h
guangli-dai: formatted

PR #2319, opened 2022-08-22
Make jemalloc max stack depth a runtime option

guangli-dai:
Add a runtime option to set the max stack depth for backtrace.

The configurable backtrace stack depth is helpful to use cases where the backtrace hook is used to allow users append additional frames / metadata, e.g., by adding Python stacks to be combined with native ones, in which case a longer stack trace becomes common.

### DIFF
--- a/include/jemalloc/internal/prof_types.h
+++ b/include/jemalloc/internal/prof_types.h
@@ -23,7 +23,12 @@ typedef struct prof_recent_s prof_recent_t;
  * is based on __builtin_return_address() necessarily has a hard-coded number
  * of backtrace frame handlers, and should be kept in sync with this setting.
  */
-#define PROF_BT_MAX			128
+#ifdef JEMALLOC_PROF_GCC
+#  define PROF_BT_MAX_LIMIT 256
+#else
+#  define PROF_BT_MAX_LIMIT UINT_MAX
+#endif
+#define PROF_BT_MAX_DEFAULT			128
 
 /* Initial hash table size. */
 #define PROF_CKH_MINITEMS		64
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1585,6 +1585,9 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				    - 1, CONF_DONT_CHECK_MIN, CONF_CHECK_MAX,
 				    true)
 				CONF_HANDLE_BOOL(opt_prof_accum, "prof_accum")
+				CONF_HANDLE_UNSIGNED(opt_prof_bt_max, "prof_bt_max",
+				    1, PROF_BT_MAX_LIMIT, CONF_CHECK_MIN, CONF_CHECK_MAX,
+				    /* clip */ true)
 				CONF_HANDLE_SSIZE_T(opt_lg_prof_interval,
 				    "lg_prof_interval", -1,
 				    (sizeof(uint64_t) << 3) - 1)
--- a/src/prof_data.c
+++ b/src/prof_data.c
@@ -1167,13 +1167,16 @@ prof_tdata_init_impl(tsd_t *tsd, uint64_t thr_uid, uint64_t thr_discrim,
 	cassert(config_prof);
 
 	/* Initialize an empty cache for this thread. */
-	tdata = (prof_tdata_t *)iallocztm(tsd_tsdn(tsd), sizeof(prof_tdata_t),
-	    sz_size2index(sizeof(prof_tdata_t)), false, NULL, true,
+	size_t tdata_sz = ALIGNMENT_CEILING(sizeof(prof_tdata_t), QUANTUM);
+	size_t total_sz = tdata_sz + sizeof(void *) * opt_prof_bt_max;
+	tdata = (prof_tdata_t *)iallocztm(tsd_tsdn(tsd),
+	    total_sz, sz_size2index(total_sz), false, NULL, true,
 	    arena_get(TSDN_NULL, 0, true), true);
 	if (tdata == NULL) {
 		return NULL;
 	}
 
+	tdata->vec = (void **)((uintptr_t)tdata + tdata_sz);
 	tdata->lock = prof_tdata_mutex_choose(thr_uid);
 	tdata->thr_uid = thr_uid;
 	tdata->thr_discrim = thr_discrim;
--- a/src/prof_sys.c
+++ b/src/prof_sys.c
@@ -55,9 +55,9 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 	cassert(config_prof);
 	assert(*len == 0);
 	assert(vec != NULL);
-	assert(max_len == PROF_BT_MAX);
+	assert(max_len <= PROF_BT_MAX_LIMIT);
 
-	nframes = unw_backtrace(vec, PROF_BT_MAX);
+	nframes = unw_backtrace(vec, max_len);
 	if (nframes <= 0) {
 		return;
 	}
@@ -97,13 +97,14 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 
 	cassert(config_prof);
 	assert(vec != NULL);
-	assert(max_len == PROF_BT_MAX);
+	assert(max_len <= PROF_BT_MAX_LIMIT);
 
 	_Unwind_Backtrace(prof_unwind_callback, &data);
 }
 #elif (defined(JEMALLOC_PROF_GCC))
 static void
 prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
+/* The input arg must be a constant for __builtin_return_address. */
 #define BT_FRAME(i)							\
 	if ((i) < max_len) {						\
 		void *p;						\
@@ -122,7 +123,7 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 
 	cassert(config_prof);
 	assert(vec != NULL);
-	assert(max_len == PROF_BT_MAX);
+	assert(max_len <= PROF_BT_MAX_LIMIT);
 
 	BT_FRAME(0)
 	BT_FRAME(1)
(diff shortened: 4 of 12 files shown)


src/prof_data.c
```diff
@@ -1170,7 +1170,10 @@ prof_tdata_init_impl(tsd_t *tsd, uint64_t thr_uid, uint64_t thr_discrim,
 	tdata = (prof_tdata_t *)iallocztm(tsd_tsdn(tsd), sizeof(prof_tdata_t),
 	    sz_size2index(sizeof(prof_tdata_t)), false, NULL, true,
 	    arena_get(TSDN_NULL, 0, true), true);
-	if (tdata == NULL) {
+	tdata->vec = (void *)iallocztm(tsd_tsdn(tsd), sizeof(void *) * opt_prof_bt_max,
```
interwq: No need to have two separate allocs; you can allocate tdata and the vec together (which also helps locality since they are always accessed together), and make vec point to right past tdata. That can also simplify the error handling below.

Also you will want to merge the two related commits together, since they do need each other to function properly.

src/prof_sys.c
```diff
@@ -97,13 +97,17 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
...
 
 	_Unwind_Backtrace(prof_unwind_callback, &data);
 }
 #elif (defined(JEMALLOC_PROF_GCC))
 static void
 prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
+/* According to the official document, __builtin_return_address
+ * takes in a constant, instead of a variable.
+ * Thus, we use the macro below.
+*/
```
interwq: The comments format look off. You will want to configure your editor to do it for you, i.e. 80-character max line, and comment format should be:

```
/* This is a single line comment. */

/*
 * This is a multi-line comment.  Follow sentence-ending periods with two
 * spaces.  Consider that we display code with fixed-width fonts; embrace the
 * old 20th-century convention and like it.
 */
```
From this page: https://github.com/jemalloc/jemalloc/wiki/Coding-Style

src/prof_sys.c
```diff
@@ -264,6 +268,429 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 	BT_FRAME(125)
 	BT_FRAME(126)
 	BT_FRAME(127)
+	BT_FRAME(128)
```
interwq, line 268: Let's cut the static max for the PROF_GCC to 256. Also we should return an error code, if the user attempts to set a value greater than the static max.

src/prof_sys.c
```diff
@@ -97,13 +97,17 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 
 	cassert(config_prof);
 	assert(vec != NULL);
-	assert(max_len == PROF_BT_MAX);
+	assert(max_len <= PROF_BT_MAX);
```
interwq: PROF_BT_MAX will only be meaningful with PROF_GCC now, is that right? If so let's rename it to make it clear, and tweak the logic too.

guangli-dai: Modified to PROF_BT_MAX_PROF_GCC

src/prof_sys.c
guangli-dai: Modified.

src/prof_data.c
guangli-dai: allocation approach changed and commits merged.

include/jemalloc/internal/prof_data.h
```diff
@@ -10,6 +10,7 @@ extern malloc_mutex_t prof_dump_mtx;
 extern malloc_mutex_t *gctx_locks;
 extern malloc_mutex_t *tdata_locks;
 
+extern unsigned opt_prof_bt_max;
```
interwq: This doesn't look necessary, given that it's included in `prof_externs.h`?

test/unit/prof_hook.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/sh
 
 if [ "x${enable_prof}" = "x1" ] ; then
-  export MALLOC_CONF="prof:true,prof_active:true,lg_prof_sample:0"
+  export MALLOC_CONF="prof:true,prof_active:true,lg_prof_sample:0,prof_bt_max:512"
```
interwq: 512 is above the hard limit for PROF_GCC right? That would cause warning output -- which can be an issue if the testing facility is looking for anything unexpected. Maybe we keep it under / equals to the limit here?

src/jemalloc.c
```diff
@@ -1585,6 +1585,9 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				    - 1, CONF_DONT_CHECK_MIN, CONF_CHECK_MAX,
 				    true)
 				CONF_HANDLE_BOOL(opt_prof_accum, "prof_accum")
+				CONF_HANDLE_UNSIGNED(opt_prof_bt_max, "prof_bt_max",
+				    1, INT_MAX,
```
interwq: Remind me why we don't put the static max of PROF_GCC here? Is it because we get no warning output that way?

Also we should make it `UINT_MAX`, to be combined with `CONF_DONT_CHECK_MAX`.

include/jemalloc/internal/prof_types.h
```diff
@@ -23,7 +23,8 @@ typedef struct prof_recent_s prof_recent_t;
  * is based on __builtin_return_address() necessarily has a hard-coded number
  * of backtrace frame handlers, and should be kept in sync with this setting.
  */
-#define PROF_BT_MAX			128
+#define PROF_BT_MAX_PROF_GCC        256
```
interwq: Maybe do it this way to hide the PROF_GCC detail to non-prof modules:

```
#ifdef JEMALLOC_PROF_GCC
#  define PROF_BT_MAX_LIMIT 256
#else
#  define PROF_BT_MAX_LIMIT UINT_MAX
#endif
```

Then we can use `PROF_BT_MAX_LIMIT` anywhere and avoid `ifdefs` around it.

src/prof_data.c
```diff
@@ -1167,9 +1167,11 @@ prof_tdata_init_impl(tsd_t *tsd, uint64_t thr_uid, uint64_t thr_discrim,
 	cassert(config_prof);
 
 	/* Initialize an empty cache for this thread. */
-	tdata = (prof_tdata_t *)iallocztm(tsd_tsdn(tsd), sizeof(prof_tdata_t),
-	    sz_size2index(sizeof(prof_tdata_t)), false, NULL, true,
-	    arena_get(TSDN_NULL, 0, true), true);
+	tdata = (prof_tdata_t *)iallocztm(tsd_tsdn(tsd),
+	    sizeof(prof_tdata_t) + sizeof(void *) * opt_prof_bt_max,
```
interwq: nit: `sizeof(prof_tdata_t) + sizeof(void *) * opt_prof_bt_max` shows up twice. Make a variable for better readability.

src/prof_data.c
```diff
@@ -1167,9 +1167,11 @@ prof_tdata_init_impl(tsd_t *tsd, uint64_t thr_uid, uint64_t thr_discrim,
...
+	    sz_size2index(sizeof(prof_tdata_t) + sizeof(void *) * opt_prof_bt_max),
+	    false, NULL, true, arena_get(TSDN_NULL, 0, true), true);
+	tdata->vec = (void **)((uintptr_t)tdata + sizeof(prof_tdata_t));
```
interwq: Move this to after the NULL check below. Anything initialized should be valid values; otherwise it's better to remain uninitialized.

src/prof_data.c
```diff
@@ -1167,13 +1167,15 @@ prof_tdata_init_impl(tsd_t *tsd, uint64_t thr_uid, uint64_t thr_discrim,
 	cassert(config_prof);
 
 	/* Initialize an empty cache for this thread. */
-	tdata = (prof_tdata_t *)iallocztm(tsd_tsdn(tsd), sizeof(prof_tdata_t),
-	    sz_size2index(sizeof(prof_tdata_t)), false, NULL, true,
+	size_t total_sz = sizeof(prof_tdata_t) + sizeof(void *) * opt_prof_bt_max;
```
interwq: To be safe, let's add:
```
size_t tdata_sz = ALIGNMENT_CEILING(sizeof(prof_tdata_t), QUANTUM);
size_t total_sz = tdata_sz + sizeof(void *) * opt_prof_bt_max;
...
tdata->vec = (void **)((uintptr_t)tdata + tdata_size);
```

src/prof_sys.c
```diff
@@ -97,13 +96,17 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
...
 
 	_Unwind_Backtrace(prof_unwind_callback, &data);
 }
 #elif (defined(JEMALLOC_PROF_GCC))
 static void
 prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
+/*
+ * According to the official document, __builtin_return_address
+ * takes in a constant, instead of a variable.
+ * Thus, we use the macro below.
```
interwq: Maybe simplify a bit:
```
/* The input arg must be a constant for __builtin_return_address. */
```

test/unit/prof_hook.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/sh
 
 if [ "x${enable_prof}" = "x1" ] ; then
-  export MALLOC_CONF="prof:true,prof_active:true,lg_prof_sample:0"
+  export MALLOC_CONF="prof:true,prof_active:true,lg_prof_sample:0,prof_bt_max:128"
```
interwq: Let's add a test case (or maybe an assertion in an existing test) to verify that it's the value we set here.

guangli-dai: Added an assertion in current test_prof_dump_hook

src/prof_data.c
guangli-dai: Modified

src/prof_sys.c
guangli-dai: Simplified

src/prof_sys.c
```diff
@@ -55,9 +55,8 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 	cassert(config_prof);
 	assert(*len == 0);
 	assert(vec != NULL);
-	assert(max_len == PROF_BT_MAX);
```
interwq, line 58: Let's replace with `assert(max_len <= PROF_BT_MAX_LIMIT);`

src/prof_sys.c
```diff
@@ -97,13 +96,13 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 
 	cassert(config_prof);
 	assert(vec != NULL);
-	assert(max_len == PROF_BT_MAX);
```
interwq, line 100: Same here with `assert(max_len <= PROF_BT_MAX_LIMIT);`

test/unit/prof_hook.sh
interwq: 128 happens to be the default value, which isn't the best for testing. Make it 200?

src/prof_sys.c
guangli-dai: Added

src/prof_sys.c
guangli-dai: Added

interwq:
LGTM! @guangli-dai can you also update the PR message, to explain why we are adding this feature? Given that it got removed before.

PR #2325, opened 2022-09-08
Add arena level name

guangli-dai:
Add arena level name to help identify manual arenas.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -1547,6 +1547,22 @@ arena_dss_prec_set(arena_t *arena, dss_prec_t dss_prec) {
 	return false;
 }
 
+void
+arena_name_get(arena_t *arena, char *name) {
+	char *end = (char *)memchr((void *)arena->name, '\0', ARENA_NAME_LEN);
+	assert(end != NULL);
+	size_t len = (uintptr_t)end - (uintptr_t)arena->name + 1;
+	assert(len > 0 && len <= ARENA_NAME_LEN);
+
+	strncpy(name, arena->name, len);
+}
+
+void
+arena_name_set(arena_t *arena, const char *name) {
+	strncpy(arena->name, name, ARENA_NAME_LEN);
+	arena->name[ARENA_NAME_LEN - 1] = '\0';
+}
+
 ssize_t
 arena_dirty_decay_ms_default_get(void) {
 	return atomic_load_zd(&dirty_decay_ms_default, ATOMIC_RELAXED);
@@ -1669,6 +1685,11 @@ arena_new(tsdn_t *tsdn, unsigned ind, const arena_config_t *config) {
 	/* Set arena before creating background threads. */
 	arena_set(ind, arena);
 	arena->ind = ind;
+
+	/* Init the name. */
+	malloc_snprintf(arena->name, sizeof(arena->name), "%s_%u",
+	    arena_is_auto(arena) ? "auto" : "manual", arena->ind);
+	arena->name[ARENA_NAME_LEN - 1] = '\0';
 
 	nstime_init_update(&arena->create_time);
 
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -170,6 +170,7 @@ CTL_PROTO(arena_i_dirty_decay_ms)
 CTL_PROTO(arena_i_muzzy_decay_ms)
 CTL_PROTO(arena_i_extent_hooks)
 CTL_PROTO(arena_i_retain_grow_limit)
+CTL_PROTO(arena_i_name)
 INDEX_PROTO(arena_i)
 CTL_PROTO(arenas_bin_i_size)
 CTL_PROTO(arenas_bin_i_nregs)
@@ -504,11 +505,12 @@ static const ctl_named_node_t arena_i_node[] = {
 	 * Undocumented for now, since we anticipate an arena API in flux after
 	 * we cut the last 5-series release.
 	 */
-	{NAME("oversize_threshold"), CTL(arena_i_oversize_threshold)},
-	{NAME("dirty_decay_ms"), CTL(arena_i_dirty_decay_ms)},
-	{NAME("muzzy_decay_ms"), CTL(arena_i_muzzy_decay_ms)},
-	{NAME("extent_hooks"),	CTL(arena_i_extent_hooks)},
-	{NAME("retain_grow_limit"),	CTL(arena_i_retain_grow_limit)}
+	{NAME("oversize_threshold"),	CTL(arena_i_oversize_threshold)},
+	{NAME("dirty_decay_ms"),	CTL(arena_i_dirty_decay_ms)},
+	{NAME("muzzy_decay_ms"),	CTL(arena_i_muzzy_decay_ms)},
+	{NAME("extent_hooks"),		CTL(arena_i_extent_hooks)},
+	{NAME("retain_grow_limit"),	CTL(arena_i_retain_grow_limit)},
+	{NAME("name"),			CTL(arena_i_name)}
 };
 static const ctl_named_node_t super_arena_i_node[] = {
 	{NAME(""),		CHILD(named, arena_i)}
@@ -2978,6 +2980,61 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
 	} else {
 		ret = EFAULT;
 	}
+label_return:
+	malloc_mutex_unlock(tsd_tsdn(tsd), &ctl_mtx);
+	return ret;
+}
+
+/*
+ * When writing, newp should point to a char array storing the name to be set.
+ * A name longer than ARENA_NAME_LEN will be arbitrarily cut. When reading,
+ * oldp should point to a char array whose length is no shorter than
+ * ARENA_NAME_LEN or the length of the name when it was set.
+ */
+static int
+arena_i_name_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
+    void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret;
+	unsigned arena_ind;
+	char *name;
+
+	malloc_mutex_lock(tsd_tsdn(tsd), &ctl_mtx);
+	MIB_UNSIGNED(arena_ind, 1);
+	if (arena_ind == MALLCTL_ARENAS_ALL || arena_ind >=
+	    ctl_arenas->narenas) {
+		ret = EINVAL;
+		goto label_return;
+	}
+	arena_t *arena = arena_get(tsd_tsdn(tsd), arena_ind, false);
+	if (arena == NULL) {
+		ret = EFAULT;
+		goto label_return;
+	}
+
+	if (oldp != NULL && oldlenp != NULL) {
+		/*
+		 * Read the arena name.  When reading, the input oldp should
+		 * point to an array with a length no shorter than
+		 * ARENA_NAME_LEN or the length when it was set.
+		 */
+		if (*oldlenp != sizeof(char *)) {
+			ret = EINVAL;
+			goto label_return;
+		}
+		name = *(char **)oldp;
+		arena_name_get(arena, name);
+	}
+
+	if (newp != NULL) {
+		/* Write the arena name. */
+		WRITE(name, char *);
+		if (name == NULL) {
+			ret = EINVAL;
+			goto label_return;
+		}
+		arena_name_set(arena, name);
+	}
+	ret = 0;
 label_return:
 	malloc_mutex_unlock(tsd_tsdn(tsd), &ctl_mtx);
 	return ret;
--- a/src/stats.c
+++ b/src/stats.c
@@ -42,14 +42,17 @@ const char *arena_mutex_names[mutex_prof_num_arena_mutexes] = {
 	assert(miblen_new == miblen + 1);				\
 } while (0)
 
-#define CTL_M2_GET(n, i, v, t) do {					\
+#define CTL_MIB_GET(n, i, v, t, ind) do {				\
 	size_t mib[CTL_MAX_DEPTH];					\
 	size_t miblen = sizeof(mib) / sizeof(size_t);			\
 	size_t sz = sizeof(t);						\
 	xmallctlnametomib(n, mib, &miblen);				\
-	mib[2] = (i);							\
+	mib[(ind)] = (i);							\
 	xmallctlbymib(mib, miblen, (void *)v, &sz, NULL, 0);		\
 } while (0)
+
+#define CTL_M1_GET(n, i, v, t) CTL_MIB_GET(n, i, v, t, 1)
+#define CTL_M2_GET(n, i, v, t) CTL_MIB_GET(n, i, v, t, 2)
 
 /******************************************************************************/
 /* Data. */
@@ -1042,6 +1045,8 @@ JEMALLOC_COLD
 static void
 stats_arena_print(emitter_t *emitter, unsigned i, bool bins, bool large,
     bool mutex, bool extents, bool hpa) {
+	char name[ARENA_NAME_LEN];
+	char *namep = name;
 	unsigned nthreads;
 	const char *dss;
 	ssize_t dirty_decay_ms, muzzy_decay_ms;
@@ -1059,6 +1064,10 @@ stats_arena_print(emitter_t *emitter, unsigned i, bool bins, bool large,
 	uint64_t uptime;
 
 	CTL_GET("arenas.page", &page, size_t);
+	if (i != MALLCTL_ARENAS_ALL && i != MALLCTL_ARENAS_DESTROYED) {
+		CTL_M1_GET("arena.0.name", i, (void *)&namep, const char *);
+		emitter_kv(emitter, "name", "name", emitter_type_string, &namep);
+	}
 
 	CTL_M2_GET("stats.arenas.0.nthreads", i, &nthreads, unsigned);
 	emitter_kv(emitter, "nthreads", "assigned threads",
--- a/test/include/test/test.h
+++ b/test/include/test/test.h
@@ -266,7 +266,7 @@
 #define expect_false(a, ...)	expect_b_eq(a, false, __VA_ARGS__)
 
 #define verify_str_eq(may_abort, a, b, ...) do {			\
-	if (strcmp((a), (b))) {						\
+	if (strcmp((a), (b)) != 0) {						\
 		char prefix[ASSERT_BUFSIZE];				\
 		char message[ASSERT_BUFSIZE];				\
 		malloc_snprintf(prefix, sizeof(prefix),			\
@@ -284,7 +284,7 @@
 } while (0)
 
 #define verify_str_ne(may_abort, a, b, ...) do {			\
-	if (!strcmp((a), (b))) {					\
+	if (strcmp((a), (b)) == 0) {					\
 		char prefix[ASSERT_BUFSIZE];				\
 		char message[ASSERT_BUFSIZE];				\
 		malloc_snprintf(prefix, sizeof(prefix),			\
--- a/test/unit/mallctl.c
+++ b/test/unit/mallctl.c
@@ -711,6 +711,48 @@ TEST_BEGIN(test_arena_i_dss) {
 }
 TEST_END
 
+TEST_BEGIN(test_arena_i_name) {
+	unsigned arena_ind;
+	size_t ind_sz = sizeof(arena_ind);
+	size_t mib[3];
+	size_t miblen;
+	char name_old[ARENA_NAME_LEN];
+	char *name_oldp = name_old;
+	size_t sz = sizeof(name_oldp);
+	char default_name[ARENA_NAME_LEN];
+	const char *name_new = "test name";
+	const char *super_long_name = "A name longer than ARENA_NAME_LEN";
+	size_t super_long_name_len = strlen(super_long_name);
+	assert(super_long_name_len > ARENA_NAME_LEN);
+
+	miblen = sizeof(mib)/sizeof(size_t);
+	expect_d_eq(mallctlnametomib("arena.0.name", mib, &miblen), 0,
+	    "Unexpected mallctlnametomib() error");
+
+	expect_d_eq(mallctl("arenas.create", (void *)&arena_ind, &ind_sz, NULL,
+	    0), 0, "Unexpected mallctl() failure");
+	mib[1] = arena_ind;
+
+	malloc_snprintf(default_name, sizeof(default_name), "manual_%u",
+	    arena_ind);
+	expect_d_eq(mallctlbymib(mib, miblen, (void *)&name_oldp, &sz,
+	    (void *)&name_new, sizeof(name_new)), 0,
+	    "Unexpected mallctl() failure");
+	expect_str_eq(name_old, default_name,
+	    "Unexpected default value for arena name");
+
+	expect_d_eq(mallctlbymib(mib, miblen, (void *)&name_oldp, &sz,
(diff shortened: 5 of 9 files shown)


src/ctl.c
```diff
@@ -2980,6 +2982,55 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+		/*
+		 * Read the arena name.  When reading, the input oldp must point
+		 * to an array with a length no shorter than ARENA_NAME_LEN.
+		 */
+		name = arena_name_get(arena);
+		if (name == NULL) {
+			ret = EFAULT;
+			goto label_return;
+		}
+		strncpy((char *)oldp, name, ARENA_NAME_LEN);
```
guangli-dai: While a conventional ctl function takes in a `char **`. Here, it is simplified so that the user only needs to pass in a `char *`. The main reason is that, by doing so, the user can directly pass in the array in spite of the difference between a char array and a char pointer. In this way, we can do:
```
char name_old[ARENA_NAME_LEN];
// ...
mallctlbymib(mib, miblen, (void *)name_old, &sz, NULL, 0)
```
instead of
```
char name_old[ARENA_NAME_LEN];
char *name_oldp = name_old;
// ...
mallctlbymib(mib, miblen, (void *)name_oldp, &sz, NULL, 0)
```

interwq: This logic should go into `arena_name_get`, then we will have a symmetric _get and _set (copying in and out, w/o exposing any of the arena_t internals), and `ARENA_NAME_LEN` won't have to show up in the ctl module.

src/ctl.c
```diff
@@ -2980,6 +2982,55 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+		ret = EFAULT;
+		goto label_return;
+	}
+	if (oldp != NULL && oldlenp != NULL) {
+		/*
+		 * Read the arena name.  When reading, the input oldp must point
+		 * to an array with a length no shorter than ARENA_NAME_LEN.
+		 */
+		name = arena_name_get(arena);
+		if (name == NULL) {
```
interwq: This won't be possible, unless some bug / corruption happened, right? In that case, do an assert would be fine.

src/arena.c
```diff
@@ -1546,6 +1546,19 @@ arena_dss_prec_set(arena_t *arena, dss_prec_t dss_prec) {
 	atomic_store_u(&arena->dss_prec, (unsigned)dss_prec, ATOMIC_RELEASE);
 	return false;
 }
+const char *
+arena_name_get(arena_t *arena) {
+	return arena->name;
+}
+bool
```
interwq: nit: blank line in between

src/ctl.c
```diff
@@ -2980,6 +2982,55 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+	const char *name = NULL;
+
+	READ_XOR_WRITE();
+
+	malloc_mutex_lock(tsd_tsdn(tsd), &ctl_mtx);
+	MIB_UNSIGNED(arena_ind, 1);
+	if (arena_ind == MALLCTL_ARENAS_ALL || arena_ind >=
+	    ctl_arenas->narenas) {
+		malloc_printf("Invalid arena index %d. This function only takes in "
+		"indices no larger than %d.\n", arena_ind, ctl_arenas->narenas);
```
interwq: We should not print anything, besides returning proper error codes. The user can decide to output if they'd like, however as a lower level lib let's not write anything unless have to (e.g. when terminating).

The other factor is, we should not say arena index greater than N is not valid, since it may change at any time, e.g. even before this msg makes it to the log.

src/ctl.c
```diff
@@ -2980,6 +2982,55 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+			ret = EFAULT;
+			goto label_return;
+		}
+		strncpy((char *)oldp, name, ARENA_NAME_LEN);
+		ret = 0;
+	} else {
+		/* Write the arena name. */
+		WRITE(name, char *);
+		if (arena_name_set(arena, name)) {
+			ret = EFAULT;
```
interwq: This indicates a bad input (null ptr as input arg), right? If so, let's check it here in the ctl module (and returns EINVAL if it happens), and don't require arena_name_set to return a bool.

guangli-dai: Most set functions in arena.c are returning a bool indicating whether the set is successful or not. Is there a specific reason why we do not following the convention here?

src/ctl.c
guangli-dai: Moved the logic into the get function.

src/ctl.c
guangli-dai: Yes, assertion removed.

src/arena.c
guangli-dai: formatted

src/ctl.c
guangli-dai: removed

src/arena.c
```diff
@@ -1547,6 +1547,42 @@ arena_dss_prec_set(arena_t *arena, dss_prec_t dss_prec) {
 	return false;
 }
 
+size_t
```
guangli-dai: Not that necessary to put the helper function inline here?

src/ctl.c
guangli-dai: Would be better to keep the convention so that the input is always on the stack.

src/arena.c
guangli-dai: directly include in the get function since it's not called in set right now

src/arena.c
```diff
@@ -1547,6 +1547,23 @@ arena_dss_prec_set(arena_t *arena, dss_prec_t dss_prec) {
 	return false;
 }
 
+void
+arena_name_get(arena_t *arena, char *name) {
+	char *end;
+	size_t len;
+	end = (char *)memchr((void *)arena->name, '\0', ARENA_NAME_LEN);
```
interwq: nit: `char *end = ...` and `size_t len = ...`. We are moving towards this direction.

src/arena.c
```diff
@@ -1547,6 +1547,23 @@ arena_dss_prec_set(arena_t *arena, dss_prec_t dss_prec) {
...
+	assert(end != NULL);
+	len = end - arena->name + 1;
```
interwq: nit: convert to `uintptr`s 

Also add `assert(len > 0 && len <= ARENA_NAME_LEN);`

src/arena.c
```diff
@@ -1665,6 +1682,9 @@ arena_new(tsdn_t *tsdn, unsigned ind, const arena_config_t *config) {
 		}
 	}
 
+	/* Init the name to an empty string. */
+	arena->name[0] = '\0';
```
interwq: WDYT setting the init name to `auto` and `manual`, depending on `arena_is_auto(arena)`? That way we get some useful info w/o any change to the call sites.

src/ctl.c
```diff
@@ -504,11 +505,12 @@ static const ctl_named_node_t arena_i_node[] = {
...
-	{NAME("oversize_threshold"), CTL(arena_i_oversize_threshold)},
-	{NAME("dirty_decay_ms"), CTL(arena_i_dirty_decay_ms)},
-	{NAME("muzzy_decay_ms"), CTL(arena_i_muzzy_decay_ms)},
-	{NAME("extent_hooks"),	CTL(arena_i_extent_hooks)},
-	{NAME("retain_grow_limit"),	CTL(arena_i_retain_grow_limit)}
+	{NAME("oversize_threshold"),	CTL(arena_i_oversize_threshold)},
+	{NAME("dirty_decay_ms"),		CTL(arena_i_dirty_decay_ms)},
+	{NAME("muzzy_decay_ms"),		CTL(arena_i_muzzy_decay_ms)},
+	{NAME("extent_hooks"),			CTL(arena_i_extent_hooks)},
+	{NAME("retain_grow_limit"),		CTL(arena_i_retain_grow_limit)},
```
interwq: The spacing look a bit strange. Is the rendering here off again?

src/ctl.c
```diff
@@ -2983,6 +2985,55 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+	arena_t *arena = arena_get(tsd_tsdn(tsd), arena_ind, false);
+	if (arena == NULL) {
+		ret = EFAULT;
+		goto label_return;
+	}
+	if (oldp != NULL && oldlenp != NULL) {
+		/*
+		 * Read the arena name.  When reading, the input oldp should point to
+		 * an array with a length no shorter than ARENA_NAME_LEN or the length
+		 * of the name.
```
interwq: nit: `or the length of the name when it was set`

Maybe have a comment section above the function signature -- we don't want to include it in the manual yet, however an explanation about the usage will help the developers too.

src/ctl.c
```diff
@@ -2983,6 +2985,55 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
 }
 
+static int
+arena_i_name_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
+    void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret;
+	unsigned arena_ind;
+	char *name = NULL;
+
+	READ_XOR_WRITE();
```
interwq: This check may do `goto label_return;` so it needs to be covered by the lock below.

src/ctl.c
```diff
@@ -2983,6 +2985,55 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+		ret = 0;
+	} else {
+		/* Write the arena name. */
+		WRITE(name, char *);
+		if (name == NULL) {
+			ret = EINVAL;
+			goto label_return;
+		}
+		arena_name_set(arena, name);
+		ret = 0;
```
interwq: This is shared by the two branches. Move it out to after the branch.

src/ctl.c
```diff
@@ -2983,6 +2985,55 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+		 * of the name.
+		 */
+		if (*oldlenp != sizeof(char *)) {
+			ret = EINVAL;
+			goto label_return;
+		}
+		name = *(char **)oldp;
+		arena_name_get(arena, name);
+		ret = 0;
+	} else {
```
interwq: The branch condition doesn't look right. Failing the `oldp != NULL && oldlenp != NULL` doesn't necessarily mean it's a valid write.

src/stats.c
```diff
@@ -1059,6 +1070,10 @@ stats_arena_print(emitter_t *emitter, unsigned i, bool bins, bool large,
 	uint64_t uptime;
 
 	CTL_GET("arenas.page", &page, size_t);
+	if (i != MALLCTL_ARENAS_ALL && i != MALLCTL_ARENAS_DESTROYED) {
+		CTL_M1_GET("arena.0.name", i, (void *)&namep, const char *);
+		emitter_kv(emitter, "name", "name", emitter_type_string, &namep);
```
interwq, line 1069: Can you share an example output from this?

test/unit/mallctl.c
```diff
@@ -711,6 +711,47 @@ TEST_BEGIN(test_arena_i_dss) {
...
+	    NULL, 0), 0, "Unexpected mallctl() failure");
+	expect_str_eq(name_old, "test name", "Unexpected value for arena name");
+
+	name_new = "A name longer than ARENA_NAME_LEN";
+	expect_d_eq(mallctlbymib(mib, miblen, NULL, NULL, (void *)&name_new,
+	    sizeof(name_new)), 0, "Unexpected mallctl() failure");
+
+	expect_d_eq(mallctlbymib(mib, miblen, (void *)&name_oldp, &sz,
+	    NULL, 0), 0, "Unexpected mallctl() failure");
+	expect_str_eq(name_old, "A name longer than ARENA_NAME_L",
```
interwq: This would fail if the macro value is changed. Better do a starts_with check.

src/arena.c
guangli-dai: modified

src/arena.c
guangli-dai: modified

src/ctl.c
guangli-dai: Yes, the rendering problem, I think I find a way and fix it now.

src/ctl.c
guangli-dai: Added.

src/ctl.c
guangli-dai: Moved below the lock

src/ctl.c
guangli-dai: Moved out. Besides, after changing the branch a bit, the neither read nor write situation will also reach the line. I assume that is what we want.

src/ctl.c
guangli-dai: Correct, this branch also included the "neither read nor write condition" in, which will then return EINVAL because name will be NULL. Modified to avoid this situation.

src/stats.c
guangli-dai: A sample output for arena 0:
```
arenas[0]:
name: "auto_0"
assigned threads: 1
```

test/unit/mallctl.c
guangli-dai: Changed it to a dynamic check and sticking to expect_str_eq. Does that look good?

test/unit/mallctl.c
```diff
@@ -711,6 +711,53 @@ TEST_BEGIN(test_arena_i_dss) {
...
+	unsigned arena_ind;
+	size_t ind_sz = sizeof(arena_ind);
+	const char *name_new;
+	char name_old[ARENA_NAME_LEN];
+	char *name_oldp = name_old;
+	size_t sz = sizeof(name_oldp);
+	size_t mib[3];
+	size_t miblen;
+	char super_long_name[] = "A name longer than ARENA_NAME_LEN";
+	int super_long_name_len = 33;
```
interwq: Let's do `char *super_long_name = ...` and use strlen to get this, and assert it's longer than ARENA_NAME_LEN.

src/arena.c
guangli-dai: Added

src/arena.c
```diff
@@ -1670,6 +1686,13 @@ arena_new(tsdn_t *tsdn, unsigned ind, const arena_config_t *config) {
 	arena_set(ind, arena);
 	arena->ind = ind;
 
+	/* Init the name to an empty string. */
```
interwq: nit: update the comment.

`strcpy(arena->name, arena_is_auto(arena) ? "auto" : "manual");`.

Still thinking about the default name, to make it as useful as possible. How about `auto_123/456`, where 123 is the arena id, and 456 is the total number of arenas? Might be confusing though?

src/ctl.c
```diff
@@ -2983,6 +2985,61 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+ */
+static int
+arena_i_name_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
+    void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret;
+	unsigned arena_ind;
+	char *name = NULL;
+
+	malloc_mutex_lock(tsd_tsdn(tsd), &ctl_mtx);
+	READ_XOR_WRITE();
```
interwq: Remind me why this has to be XOR again?

src/stats.c
```diff
@@ -42,6 +42,15 @@ const char *arena_mutex_names[mutex_prof_num_arena_mutexes] = {
...
 
+#define CTL_M1_GET(n, i, v, t) do {					\
+	size_t mib[CTL_MAX_DEPTH];					\
+	size_t miblen = sizeof(mib) / sizeof(size_t);			\
+	size_t sz = sizeof(t);						\
+	xmallctlnametomib(n, mib, &miblen);				\
+	mib[1] = (i);							\
+	xmallctlbymib(mib, miblen, (void *)v, &sz, NULL, 0);		\
+} while (0)
+
```
interwq, line 53: Try merging  CTL_M1_GET and _M2_GET. You can define a CTL_MIB_GET which takes the index and shared by the M1 and M2 getters.

test/unit/mallctl.c
```diff
@@ -711,6 +711,53 @@ TEST_BEGIN(test_arena_i_dss) {
...
+	    NULL, 0), 0, "Unexpected mallctl() failure");
+	expect_str_eq(name_old, "test name", "Unexpected value for arena name");
+
+	name_new = super_long_name;
+	expect_d_eq(mallctlbymib(mib, miblen, NULL, NULL, (void *)&name_new,
+	    sizeof(name_new)), 0, "Unexpected mallctl() failure");
+
+
+	if (super_long_name_len >= ARENA_NAME_LEN) {
+		super_long_name[ARENA_NAME_LEN - 1] = '\0';
```
interwq: I'd prefer keeping super_long_name as a string literal, and use strncmp below to compare so that we don't have to modify any strings.

test/unit/mallctl.c
```diff
@@ -711,6 +711,53 @@ TEST_BEGIN(test_arena_i_dss) {
...
+	expect_d_eq(mallctlbymib(mib, miblen, NULL, NULL, (void *)&name_new,
+	    sizeof(name_new)), 0, "Unexpected mallctl() failure");
+
+	expect_d_eq(mallctlbymib(mib, miblen, (void *)&name_oldp, &sz,
+	    NULL, 0), 0, "Unexpected mallctl() failure");
+	expect_str_eq(name_old, "test name", "Unexpected value for arena name");
+
+	name_new = super_long_name;
+	expect_d_eq(mallctlbymib(mib, miblen, NULL, NULL, (void *)&name_new,
+	    sizeof(name_new)), 0, "Unexpected mallctl() failure");
```
interwq: Use super_long_name directly here. Then the name_new can be made a const char * as well.

test/unit/mallctl.c
```diff
@@ -711,6 +711,53 @@ TEST_BEGIN(test_arena_i_dss) {
...
+	    NULL, 0), 0, "Unexpected mallctl() failure");
+	expect_str_eq(name_old, "manual", "Unexpected default value for arena name");
+
+	name_new = "test name";
+	expect_d_eq(mallctlbymib(mib, miblen, NULL, NULL, (void *)&name_new,
+	    sizeof(name_new)), 0, "Unexpected mallctl() failure");
+
+	expect_d_eq(mallctlbymib(mib, miblen, (void *)&name_oldp, &sz,
+	    NULL, 0), 0, "Unexpected mallctl() failure");
+	expect_str_eq(name_old, "test name", "Unexpected value for arena name");
```
interwq: nit: use name_new instead of "test name" to avoid having the same string twice.

src/arena.c
guangli-dai: Adding the total number of arenas could be confusing since it's a global info not specific to the arena. I think adding the id is a good idea: `auto_123` with 123 as the arena id.

src/ctl.c
guangli-dai: Originally we set this because of the life cycle of the name pointer. Yet now that we are copying the name instead of directly returning the pointer, there is no life cycle problem so I think we can allow read and write at the same time now. Should we do that?

src/arena.c
guangli-dai: modified

test/unit/mallctl.c
guangli-dai: done.

test/unit/mallctl.c
guangli-dai: done.

src/ctl.c
guangli-dai: modified so that it can read and write at the same time since there is no concern regarding the life cycle right now.

src/stats.c
guangli-dai: Merged.

test/unit/mallctl.c
guangli-dai: Modified and used `strncmp` to check the result

test/unit/mallctl.c
guangli-dai: modified

src/arena.c
```diff
@@ -1670,6 +1686,11 @@ arena_new(tsdn_t *tsdn, unsigned ind, const arena_config_t *config) {
 	arena_set(ind, arena);
 	arena->ind = ind;
 
+	/* Init the name. */
+	sprintf(arena->name, "%s_%u", arena_is_auto(arena) ? "auto" : "manual",
+	    arena->ind);
+	arena->name[ARENA_NAME_LEN - 1] = '\0';
```
guangli-dai, line 1692: Although the arena's index cannot be too large and we should be completely safe, added this line just to make it safe.

test/include/test/test.h
```diff
@@ -284,7 +284,7 @@
 } while (0)
 
 #define verify_str_ne(may_abort, a, b, ...) do {			\
-	if (!strcmp((a), (b))) {					\
+	if (strcmp((a), (b)) == 0) {					\
```
guangli-dai, line 287: Modified to avoid implicit cast from int to bool

src/arena.c
```diff
@@ -1670,6 +1686,11 @@ arena_new(tsdn_t *tsdn, unsigned ind, const arena_config_t *config) {
 	arena_set(ind, arena);
 	arena->ind = ind;
 
+	/* Init the name. */
+	sprintf(arena->name, "%s_%u", arena_is_auto(arena) ? "auto" : "manual",
```
guangli-dai: avoid using sprintf, maybe `malloc_snprintf`?

src/ctl.c
```diff
@@ -2983,6 +2985,61 @@ arena_i_retain_grow_limit_ctl(tsd_t *tsd, const size_t *mib,
...
+ * A name longer than ARENA_NAME_LEN will be arbitrarily cut. When reading,
+ * oldp should point to a char array whose length is no shorter than
+ * ARENA_NAME_LEN or the length of the name when it was set.
+ */
+static int
+arena_i_name_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
+    void *oldp, size_t *oldlenp, void *newp, size_t newlen) {
+	int ret;
+	unsigned arena_ind;
+	char *name = NULL;
```
interwq: nit: no need to initialize here. We try avoiding double initialization, meaning that it should be explicitly initialized down the path its needed. This way the compiler can warn us if there's any possibility of access w/o init (which won't be possible if it's simply initialized at the beginning. The `ret` in this function is a good example of following this principle.

test/unit/mallctl.c
```diff
@@ -711,6 +711,49 @@ TEST_BEGIN(test_arena_i_dss) {
...
+	expect_d_eq(mallctlnametomib("arena.0.name", mib, &miblen), 0,
+	    "Unexpected mallctlnametomib() error");
+
+	expect_d_eq(mallctl("arenas.create", (void *)&arena_ind, &ind_sz, NULL,
+	    0), 0, "Unexpected mallctl() failure");
+	mib[1] = arena_ind;
+
+	malloc_snprintf(default_name, sizeof(default_name), "manual_%u",
+	    arena_ind);
+	name_new = "test name";
```
interwq: nit: since it's a const really, move it up to init along with definition. Maybe also put the two literal strings next to each other.

src/ctl.c
guangli-dai: removed the initialization

test/unit/mallctl.c
guangli-dai: init moved up

PR #2330, opened 2022-09-14
Avoid using a lookup table for sz_s2a

LoganEvans:
Summary:
This modifies sz_s2a so that when an allocation size is compute for an arbitrary number of bytes, the code will avoid using a runtime-generated lookup table. The new code makes sz_s2a a logically constexpr function.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


LoganEvans:
The `sz_s2a_compute` function looks to be more optimal than the version I cooked up. I'm going to explore using that approach instead.

PR #2336, opened 2022-09-26
Fix safety_check segfault in double free test

divanorama:
After https://github.com/jemalloc/jemalloc/pull/2062 `large_dalloc_safety_checks` got a `edata == NULL` case but error message wasn't working for `edata == NULL`

`test/unit/double_free` was failing with `EXC_BAD_ACCESS (code=1, address=0x8)` on aarch64-darwin with clang-11

Adding `edata==NULL` case to error logging here.


interwq:
Thanks for catching this! The PR is merged as 1897f185d2c06307fefc4d8f4512eeb13c474999

PR #2339, opened 2022-09-29
update PROFILING_INTERNALS.md

jordalgo:
Expand the bad example of summing before unbiasing.


interwq:
Looks great. Thanks @jordalgo !

PR #2341, opened 2022-10-01
fix build for non linux/BSD platforms.

devnexen:


interwq:
Thanks for the patch @devnexen !

PR #2344, opened 2022-10-08
Minor fixes

madscientist:
Use `%u` not `%d` when printing unsigned values.

Take advantage of GCC's copy() facility in `__attribute__` to duplicate all the attributes of these functions.  I'm not sure about this because I'm not sure if Clang supports it.

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_defs.h.in
+++ b/include/jemalloc/internal/jemalloc_internal_defs.h.in
@@ -245,6 +245,7 @@
 /*
  * popcount*() functions to use for bitmapping.
  */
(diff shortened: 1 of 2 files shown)


madscientist:
Anyone have any thoughts on these small fixes?  They're needed for a clean build in my environments.  Cheers!

davidtgoldblatt:
Sorry for the delay -- this has been a hectic couple of weeks.

Two nits:
- We do "clean linear history", with only rebase-and-merge, so please clean up any stacks.
- In general, we should be doing this sort of feature detection (for the copy attribute) via configure scripts, rather than hard-coding results via the preprocessor.

PR #2347, opened 2022-10-14
jemalloc_internal_types.h: Use alloca if __STDC_NO_VLA__ is defined

madscientist:
No currently-available version of Visual Studio C compiler supports variable length arrays, even if it defines __STDC_VERSION__ >= C99. As far as I know Microsoft has no plans to ever support VLAs in MSVC.

The C11 standard requires that the __STDC_NO_VLA__ macro be defined if the compiler doesn't support VLAs, so fall back to alloca() if so.


interwq:
Thanks for the patch!

PR #2349, opened 2022-10-21
Enable fast thread locals for dealloc-only threads.

interwq:
Previously if a thread does only allocations, it stays on the slow path / minimal initialized state forever.  However, dealloc-only is a valid pattern for dedicated reclamation threads -- this means thread cache is disabled (no batched flush) for them, which causes high overhead and contention.

Added the condition to fully initialize TSD when a fair amount of dealloc activities are observed.

### DIFF
--- a/include/jemalloc/internal/tsd.h
+++ b/include/jemalloc/internal/tsd.h
@@ -59,6 +59,7 @@ typedef ql_elm(tsd_t) tsd_link_t;
 #define TSD_DATA_SLOW							\
     O(tcache_enabled,		bool,			bool)		\
     O(reentrancy_level,		int8_t,			int8_t)		\
+    O(min_init_state_nfetched,		uint8_t,	uint8_t)	\
     O(thread_allocated_last_event,	uint64_t,	uint64_t)	\
     O(thread_allocated_next_event,	uint64_t,	uint64_t)	\
     O(thread_deallocated_last_event,	uint64_t,	uint64_t)	\
@@ -91,6 +92,7 @@ typedef ql_elm(tsd_t) tsd_link_t;
 #define TSD_DATA_SLOW_INITIALIZER					\
     /* tcache_enabled */	TCACHE_ENABLED_ZERO_INITIALIZER,	\
     /* reentrancy_level */	0,					\
+    /* min_init_state_nfetched */	0,				\
     /* thread_allocated_last_event */	0,				\
     /* thread_allocated_next_event */	0,				\
     /* thread_deallocated_last_event */	0,				\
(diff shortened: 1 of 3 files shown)


guangli-dai (approved):

davidtgoldblatt:
Is "bytes deallocated" the right metric? I would think "num dallocs" would be useful too.

I mean, this is all heuristic anyways, but "do a single big free (of, say, a TLS region or something)" would be plausible as a libc-thread-shutdown call.

src/tsd.c
```diff
@@ -302,7 +302,21 @@ tsd_fetch_slow(tsd_t *tsd, bool minimal) {
...
+		 * tell if plenty of deallocation activities already happened,
+		 * in which case the tsd is switched to nominal.
+		 *
+		 * The 8K threshold means at most 1K of dealloc calls stay on
+		 * the slow path.
+		 *
+		 * Also see comments in free_default().
+		 */
+		if (!minimal || (*tsd_thread_deallocatedp_get(tsd) >=
+		    TSD_STATE_MINIMAL_INITIALIZED_DEALLOC_THRESHOLD)) {
```
interwq: @davidtgoldblatt : this is the workaround for the dealloc-only threads slow TSD issue we discussed, in case you want a reference.

interwq:
> Is "bytes deallocated" the right metric? I would think "num dallocs" would be useful too.
> 
> I mean, this is all heuristic anyways, but "do a single big free (of, say, a TLS region or something)" would be plausible as a libc-thread-shutdown call.

Agreed. However we don't maintain a # of calls on the fastpath anymore (the GC ticker got replaced with bytes-triggered). That's another reason why I "secretly" made the transition to full-init cannot happen on the first deallocation, no matter how large.

The current heuristic does give me a pause to say at least. An alternative, is to add a counter for tsd_min_fetch() (which is the same as # of calls, only for the min_init state), which probably is safer and more predictable. WDYT?

davidtgoldblatt:
> The current heuristic does give me a pause to say at least. An alternative, is to add a counter for tsd_min_fetch() (which is the same as # of calls, only for the min_init state), which probably is safer and more predictable. WDYT?

Yeah that assuages my nerves. But I also wouldn't complain too much if you thought it was overkill for this case.

interwq:
@davidtgoldblatt just added the counter for the # of dealloc calls under the min_init state. The change is pretty contained so worth it for the assurance.

include/jemalloc/internal/tsd.h
```diff
@@ -59,6 +59,7 @@ typedef ql_elm(tsd_t) tsd_link_t;
 #define TSD_DATA_SLOW							\
     O(tcache_enabled,		bool,			bool)		\
     O(reentrancy_level,		int8_t,			int8_t)		\
+    O(min_init_state_nfetched,		uint8_t,	uint8_t)	\
```
interwq, line 62: @guangli-dai this part just got added (see David's comments), in case you want to take another look.

guangli-dai: As discussed offline, it looks good to me. Although it's rarely possible, I will open up an issue to record the potential leak caused by the coincide of the thread shut down and the num deallocs crossing the threshold.

PR #2352, opened 2022-11-02
Add a configure option --enable-force-getenv.

interwq:
Allows the use of getenv() rather than secure_getenv() to read MALLOC_CONF. This helps in situations where hosts are under full control, and setting MALLOC_CONF is needed while also setuid.  Disabled by default.

Tested locally with `--enable-force-getenv`  / `--disable-force-getenv` / default.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1571,6 +1571,22 @@ if test "x$enable_readlinkat" = "x1" ; then
 fi
 AC_SUBST([enable_readlinkat])
 
+dnl Do not force getenv by default
+AC_ARG_ENABLE([force-getenv],
+  [AS_HELP_STRING([--enable-force-getenv], [Use getenv over secure_getenv])],
+[if test "x$enable_force_getenv" = "xno" ; then
+  enable_force_getenv="0"
+else
+  enable_force_getenv="1"
+fi
+],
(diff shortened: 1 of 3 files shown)


guangli-dai (approved):
LGTM

configure.ac
```diff
@@ -1571,6 +1571,22 @@ if test "x$enable_readlinkat" = "x1" ; then
...
+else
+  enable_force_getenv="1"
+fi
+],
+[enable_force_getenv="0"]
+)
+if test "x$enable_force_getenv" = "x1" ; then
+  AC_DEFINE([JEMALLOC_FORCE_GETENV], [ ], [ ])
+fi
+AC_SUBST([force_getenv])
```
interwq, line 1588: About the naming, considered a few options: 1) "force_getenv" which is the current, 2) "prefer_getenv" which feels like there are fallbacks to try after getenv, 3) "trust_getenv" which I liked initially but feels a bit unclear now.

Any thoughts @guangli-dai?

guangli-dai: I think force_getenv has the strongest warning it needs since users should have full control of the machine to enable it. I vote for "force_getenv".

Issue #2356, opened 2022-11-23
test/stress/cpp/microbench failing with Floating point exception after recent commit


sharkcz:
We can see a test failure on x86_64 and s390x platforms in our CI after recent commits (481bbfc9906e7744716677edd49d0d6c22556a1a was OK). Both looks the same
````
...
=== test/stress/cpp/microbench ===
100000000 iterations, malloc_free=1044987us (10.449 ns/iter), new_delete=1059986us (10.599 ns/iter), time consumption ratio=0.986:1
test_free_vs_delete (non-reentrant): pass
test/test.sh: line 34: 895548 Floating point exception(core dumped) $JEMALLOC_TEST_PREFIX ${t} /var/lib/jenkins/workspace/jemalloc/label/x86_64/ /var/lib/jenkins/workspace/jemalloc/label/x86_64/
Test harness error: test/stress/cpp/microbench w/ MALLOC_CONF=""
Use prefix to debug, e.g. JEMALLOC_TEST_PREFIX="gdb --args" sh test/test.sh test/stress/cpp/microbench
make: *** [Makefile:707: stress] Error 1
Build step 'Execute shell' marked build as failure
````
The environment is Fedora 36 on x86_64 and s390x, there is no such problem other platforms (aarch64, ppc64le) it seems.

sharkcz:
backtrace from the coredump
````
Core was generated by `test/stress/cpp/microbench /var/lib/jenkins/workspace/jemalloc/label/x86_64/ /v'.
Program terminated with signal SIGFPE, Arithmetic exception.
#0  0x000000000048f9fa in timer_ratio (a=<optimized out>, b=0x7ffc3e6f9660, buf=0x7ffc3e6f964a "0.986", buflen=6) at test/src/timer.c:32
32		n = malloc_snprintf(&buf[i], buflen-i, "%"FMTu64, t0 / t1);
(gdb) where
#0  0x000000000048f9fa in timer_ratio (a=<optimized out>, b=0x7ffc3e6f9660, buf=0x7ffc3e6f964a "0.986", buflen=6) at test/src/timer.c:32
#1  0x000000000040eb73 in compare_funcs (nwarmup=10000000, niter=100000000, func_b=0x40e610 <new_delete_array()>, name_b=0x49017d "delete_array", func_a=0x40e6e0 <malloc_free_array()>, 
    name_a=0x490119 "malloc_free_array") at test/include/test/bench.h:54
[...]

PR #2359, opened 2022-12-02
Fix dividing 0 error in stress/cpp/microbench

guangli-dai:
Per issue #2356, some CXX compilers may optimize away the new/delete operation in stress/cpp/microbench.cpp. Thus, this commit (1) bumps the time interval to 1 if it is 0, and (2) modifies the pointers in the microbench to volatile.

### DIFF
--- a/test/src/timer.c
+++ b/test/src/timer.c
@@ -28,6 +28,17 @@ timer_ratio(timedelta_t *a, timedelta_t *b, char *buf, size_t buflen) {
 	size_t i = 0;
 	size_t j, n;
 
+	/* 
+ 	* The time difference could be 0 if the two clock readings are 
+ 	* identical, either due to the operations being measured in the middle
+ 	* took very little time (or even got optimized away), or the clock 
+ 	* readings are bad / very coarse grained clock.
+ 	* Thus, bump t1 if it is 0 to avoid dividing 0. 
+ 	*/
+	if (t1 == 0) {
+	    t1 = 1;
+	}
+
 	/* Whole. */
 	n = malloc_snprintf(&buf[i], buflen-i, "%"FMTu64, t0 / t1);
 	i += n;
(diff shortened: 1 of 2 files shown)


test/src/timer.c
```diff
@@ -28,6 +28,11 @@ timer_ratio(timedelta_t *a, timedelta_t *b, char *buf, size_t buflen) {
 	size_t i = 0;
 	size_t j, n;
 
+	/* Bump t1 if it is 0 to avoid dividing 0. */
+	if (t1 == 0) {
+	    t1 = 1;
```
guangli-dai, line 39: The number is odd enough to raise the attention from the user, is any extra warning needed?

interwq: Let's certainly add more comments, e.g.

The time difference could be 0 if the two clock readings are identical, either due to the operations being measured in the middle took very little time (or even got optimized away), or the clock readings are bad / very coarse grained clock.

As for the warning output, no strong opinion either way. When it happens, the results would look obviously wrong, is that so?

guangli-dai: Yes, it should look obviously wrong, an example should look like this:
`100000000 iterations, malloc_free_array=1320029us (13.200 ns/iter), delete_array=0us (0 ns/iter), time consumption ratio=1320029:1`

davidtgoldblatt:
Doesn't need to be this diff, but probably where supported we want something like
```
static void *no_opt_ptr(void *ptr) {
#if HAVE_INLINE_ASM
  asm volatile("" : "+r"(ptr));
#else
  void *volatile dup = ptr;
  ptr = dup;
#endif
  return ptr;
}
```

The fast path is fast enough that adding extra store-load forwarding noticeably affects results.

interwq:
@guangli-dai thanks for fixing this! I'd say follow David's suggestion to do inline asm volatile in another PR; you can check this for example how that support can be detected:
https://github.com/jemalloc/jemalloc/blob/dev/configure.ac#L528

PR #2361, opened 2022-12-05
malloc_sdallocx test making using smart pointers instead

devnexen:
since we re using c++ 14 and onwards.

### DIFF
--- a/test/stress/cpp/microbench.cpp
+++ b/test/stress/cpp/microbench.cpp
@@ -5,44 +5,45 @@
 
 static void
 malloc_free(void) {
-	void *p = malloc(1);
-	expect_ptr_not_null(p, "Unexpected malloc failure");
-	free(p);
+	void* volatile p = malloc(1);
+	expect_ptr_not_null((void *)p, "Unexpected malloc failure");
+	free((void *)p);
 }
 
 static void
 new_delete(void) {
-	auto p = ::operator new(1);
-	expect_ptr_not_null(p, "Unexpected new failure");
-	::operator delete(p);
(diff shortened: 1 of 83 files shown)


test/stress/cpp/microbench.cpp
```diff
@@ -39,9 +41,9 @@ new_sized_delete(void) {
 
 static void
 malloc_sdallocx(void) {
-	void *p = malloc(1);
-	expect_ptr_not_null(p, "Unexpected malloc failure");
-        sdallocx(p, 1, 0);
+	auto sdallocfn = [&](char *p){sdallocx(p, 1, 0);};
+	std::unique_ptr<char, decltype(sdallocfn)> p {(char*)malloc(1), sdallocfn};
+	expect_ptr_not_null(p.get(), "Unexpected malloc failure");
```
interwq, line 47: @devnexen I'm not sure if this actually buys us much. Since this is the benchmark code, I'm leaning towards having anything as explicit as possible, and not involve anything behind the scene (e.g. if the stdlib includes extra sanity checks for unique_ptrs). It may also make symbols more complex, say when doing perf record / report.

devnexen: This is not buying anything but wanted to play with c++ memory wrappers a bit :)

interwq: Haha understood. For now I feel we still want the jemalloc side to be in explicit control. But it would be interesting to verify that the wrappers add 0 cost.

Closing the PR but as always thanks for sharing the thoughts!

PR #2371, opened 2023-01-19
Remove locked flag set in malloc_mutex_trylock

guangli-dai:
While locked is a flag for the lock, modifying it after a failure to gain the lock may cause unnecessary spins. Specifically, if a thread A is doing try_lock and it fails, but before it sets the flag locked to true another thread owning the lock frees it and sets locked to false. The locked is then set to true by thread A again. Other threads that are spinning to wait for the lock will keep spinning while one of them should actually stop spinning and get the lock.

### DIFF
--- a/include/jemalloc/internal/mutex.h
+++ b/include/jemalloc/internal/mutex.h
@@ -175,7 +175,6 @@ malloc_mutex_trylock(tsdn_t *tsdn, malloc_mutex_t *mutex) {
 	witness_assert_not_owner(tsdn_witness_tsdp_get(tsdn), &mutex->witness);
 	if (isthreaded) {
 		if (malloc_mutex_trylock_final(mutex)) {
-			atomic_store_b(&mutex->locked, true, ATOMIC_RELAXED);
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


davidtgoldblatt:
I like this a lot (I'm not even convinced it's an optimization even if this is doing what it's meant to; seems like it guarantees extra cache contention during unlock).

Did you do any perf testing? I wouldn't be surprised if you get put a number on this.

guangli-dai:
> Did you do any perf testing? I wouldn't be surprised if you get put a number on this.

Played with [ttest](https://locklessinc.com/downloads/t-test1.c) that was used in the [original commit](https://github.com/jemalloc/jemalloc/commit/b23336af96e6ef9efb47591ce7bf2c8a1eab866b). Did not see big differences. A direct reason could be the parameter setting. I followed the original test setting and simply increase the number of threads, i.e., `./ttest1 500 3 10000 1 100` and `./ttest1 500 10 10000 1 100`. 

Besides, this test may not reflect realistic workload patterns. Do you have any suggestions on how to benchmark the influence?

PR #2380, opened 2023-02-10
Fix thread_name updating for heap profiling.

interwq:
The current thread name reading path updates the name every time, which requires both alloc and dalloc -- and the temporary NULL value in the middle causes races where the prof dump read path gets NULLed in the middle.

Minimize the changes in this commit to isolate the bugfix testing; will also refactor the whole thread name paths later.

### DIFF
--- a/src/prof_data.c
+++ b/src/prof_data.c
@@ -451,16 +451,15 @@ prof_thread_name_alloc(tsd_t *tsd, const char *thread_name) {
 	}
 
 	size = strlen(thread_name) + 1;
-	if (size == 1) {
-		return "";
-	}
-
 	ret = iallocztm(tsd_tsdn(tsd), size, sz_size2index(size), false, NULL,
 	    true, arena_get(TSDN_NULL, 0, true), true);
 	if (ret == NULL) {
 		return NULL;
 	}
+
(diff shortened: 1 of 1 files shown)


interwq:
After debating with myself for a bit, I'm going to embed the thread name buffer into the `tdata` struct. cc @guangli-dai

src/prof_data.c
```diff
@@ -451,16 +451,13 @@ prof_thread_name_alloc(tsd_t *tsd, const char *thread_name) {
 	}
 
 	size = strlen(thread_name) + 1;
-	if (size == 1) {
-		return "";
-	}
-
 	ret = iallocztm(tsd_tsdn(tsd), size, sz_size2index(size), false, NULL,
```
guangli-dai, line 454: I would suggest either do iallocztm with parameter `zero = true` or explicitly make `ret [size - 1] = 0` just in case the name passed in does not end properly.

interwq: Good point. Added the terminator below.

guangli-dai:
LGTM!

guangli-dai (approved):

interwq:
Merging this one as a temporary fix. As mentioned will make more changes soon.

PR #2384, opened 2023-02-20
Add --with-lg-tcache-limit configuration option to allow for more than 4094 tcaches

veselink1:
In the Couchbase Server Data Service (KV Engine), we use one tcache per arena, per thread (multipies).

We have one arena per bucket, and we currently support 30 buckets, but might eventually bump this up to 100 buckets per instance.

However, because we allocate a tcache per thread as well, the number of tcache which we use ends up being num-buckets x num-threads. Tcaches are allocated when a thread decides to use a bucket's arena, allocating tcaches lazily, so not all threads will need a tcache, but most will.

jemalloc can automatically allocate and make use of a tcache, but we've found that can result in incorrect accounting of memory stats per-arena, which is a deal-breaker for us, because we rely on some of these stats to calculate per-arena memory fragmentation.

We're currently testing how this change performs in our test environments, but plan on proceeding to ship a release with those changes and jemalloc configured with `--with-lg-tcache-limit=15`, to allow for up to 32K tcaches to be created. Note that we don't expect to actually reach that limit, but something within the range of 4-7K is what we'd want to be able to run correctly for some larger machine configurations.

This change is ABI-breaking, but does not change the jemalloc API. We link to jemalloc statically, so this is not an issue for us.


interwq:
Thanks for sharing the patch @veselink1 . The changes look good to me and I get why you needed it. However I'm not sure we want to go this far, in terms of committing to the added options long term -- for example we have been talking about embedding the arena index into the radix tree (which has its own bits limitation as well) for fast "remote" arena detection. Would you be fine maintaining your own jemalloc branch for this feature? My feeling is this part isn't going to change too frequently, i.e. you won't get many merge conflicts, unless we do change the specifics around the tcache and arena bits.

PR #2389, opened 2023-03-02
Simplify the logic in ph_insert

deadalnix:
Also fixes what looks like an off by one error in the lazy aux list merge part of the code that previously never touched the last node in the aux list.

### DIFF
--- a/include/jemalloc/internal/ph.h
+++ b/include/jemalloc/internal/ph.h
@@ -318,36 +318,37 @@ ph_insert(ph_t *ph, void *phn, size_t offset, ph_cmp_t cmp) {
 	 */
 	if (ph->root == NULL) {
 		ph->root = phn;
-	} else {
-		/*
-		 * As a special case, check to see if we can replace the root.
-		 * This is practically common in some important cases, and lets
-		 * us defer some insertions (hopefully, until the point where
-		 * some of the items in the aux list have been removed, savings
-		 * us from linking them at all).
-		 */
-		if (cmp(phn, ph->root) < 0) {
-			phn_lchild_set(phn, ph->root, offset);
-			phn_prev_set(ph->root, phn, offset);
-			ph->root = phn;
-			ph->auxcount = 0;
-			return;
-		}
-		ph->auxcount++;
-		phn_next_set(phn, phn_next_get(ph->root, offset), offset);
-		if (phn_next_get(ph->root, offset) != NULL) {
-			phn_prev_set(phn_next_get(ph->root, offset), phn,
-			    offset);
-		}
-		phn_prev_set(phn, ph->root, offset);
-		phn_next_set(ph->root, phn, offset);
+		return;
 	}
-	if (ph->auxcount > 1) {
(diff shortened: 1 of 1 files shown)


include/jemalloc/internal/ph.h
```diff
@@ -318,36 +318,37 @@ ph_insert(ph_t *ph, void *phn, size_t offset, ph_cmp_t cmp) {
...
+	phn_next_set(phn, phn_next_get(ph->root, offset), offset);
+	if (phn_next_get(ph->root, offset) != NULL) {
+		phn_prev_set(phn_next_get(ph->root, offset), phn,
+		    offset);
+	}
+	phn_prev_set(phn, ph->root, offset);
+	phn_next_set(ph->root, phn, offset);
+
+	ph->auxcount++;
+	unsigned nmerges = ffs_zu(ph->auxcount);
```
guangli-dai, line 348: why `ph->auxcount - 1` is changed to `ph->auxcount`? Without the if condition, `ph_try_aux_merge_pair` will be executed even when `auxcount = 1`, which will cause a regression in the performance.

deadalnix: When `auxcount = 1` then it has no trailing zero and nothing happens. The only thing the previous check obviates is computing `ffs_zu` for `auxcount = 1` but that's one CPU instruction, and the check itself is more than one instruction, so if anything, this is doing less work now.

Now, you might argue that this will do work when `auxcount = 2` when previously no work was done, and yes, you'd be right. This work is actually necessary, there was an off by one error that left a straggler that never got merged before.

guangli-dai: I think you are right about `auxcount = 1`, I forgot jemalloc's `ffs` did an extra [-1](https://github.com/jemalloc/jemalloc/blob/12cd13cd418512d9e7596921ccdb62e25a103f87/include/jemalloc/internal/bit_util.h#L32). 

As for the case where `auxcount = 2`, the previous version will indeed leave a sibling of the root unmerged. Is this intentional to be lazy by design? @davidtgoldblatt 
I can see that we always do a `ph_merge_aux` in [`ph_remove_first`](https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/ph.h#L361), making it look like that setting `nmerges` to `ffs_zu(auxcount - 1)` aims at making the merge less aggressive?

deadalnix: I do not think we benefit from the merge being less aggressive here, because the tree gets really unbalanced as one node never acquire children. Work being frontloaded at the insert point is an intentional design decision here, as it is more common to do bulk inserts than it is to do bulk removal, so shifting work from removal to insert improves cache locality.

From my discussion with @davidtgoldblatt on gitter, he also seems to think this was an off by one error in the original code (which he himself wrote).

davidtgoldblatt: Yeah, I think that this diff is right. We can double-check the perf characteristics with a canary for extra-confidence.

guangli-dai (approved):

PR #2391, opened 2023-03-03
Do not maintain root->prev in ph_remove.

deadalnix:
`root->prev` is left undefined in other codepath, so there is no point maintaining it here.


guangli-dai (approved):
LGTM!

PR #2392, opened 2023-03-03
Do not double set next.prev when deleting a node in the paring heap.

deadalnix:
It is set on line 434 anyways, this is a dead store.


deadalnix:
Superseeded by #2393

PR #2393, opened 2023-03-03
Simplify the logic in ph_remove

deadalnix:
The logic was needlessly complex, assuming prev could be null, and was duplicated for first child/non first child.


guangli-dai (approved):
Indeed much neater and LGTM. Thank you!

PR #2394, opened 2023-03-07
update instructions for source build

jjb:
This is what worked for me with jemalloc 3, not sure about other versions

on ubuntu 20:

```bash
apt-get install -y libxslt-dev xsltproc docbook-xsl autoconf
git clone https://github.com/jemalloc/jemalloc.git \
 && cd jemalloc \
 && git checkout stable-3 \
 && ./autogen.sh --with-xslroot=/usr/share/xml/docbook/stylesheet/docbook-xsl \
 && make \
 && make dist \
 && make install
LD_PRELOAD=/usr/local/lib/libjemalloc.so
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -10,6 +10,7 @@ that might work is:
 
     ./autogen.sh
     make
+    make dist
     make install
 
 You can uninstall the installed build artifacts like this:


interwq:
Thanks for sharing the findings. The `make dist` step isn't required in the newer versions anymore.

PR #2397, opened 2023-03-10
Remove unused mutex from hpa_central

deadalnix:
As the title says.


interwq:
Good catch. Thanks @deadalnix !

PR #2368, opened 2023-01-09
Fix the rdtscp detection bug and add prefix for the macro.

guangli-dai:
1. Fix the missing definition of RDTSCP macro;
2. Fix a typo and rename the macro following the convention.

PR #2398, opened 2023-03-10
Avoid assuming the arena id in test when percpu_arena is used.

interwq:
This should unblock #2368.


interwq:
@guangli-dai this should fix the test failure you saw in #2368.

guangli-dai (approved):
LGTM!

Yet why the failure only appears with RDTSCP enabled?

interwq:
Previously the percpu arena feature wasn't detected correctly and resulted in not supported at all on Mac. Your fix enables the feature while also revealed the test issue here.

PR #2400, opened 2023-03-21
Explicit arena assignment in test_thread_idle and test_tcache_max

interwq:
Otherwise the associated arena could change with percpu arena enabled.

This should unblock #2368

### DIFF
--- a/test/unit/mallctl.c
+++ b/test/unit/mallctl.c
@@ -1097,8 +1097,12 @@ TEST_BEGIN(test_thread_idle) {
 
 	unsigned arena_ind;
 	sz = sizeof(arena_ind);
-	err = mallctl("thread.arena", &arena_ind, &sz, NULL, 0);
-	expect_d_eq(err, 0, "");
+	expect_d_eq(mallctl("arenas.create", (void *)&arena_ind, &sz, NULL, 0),
+	    0, "Unexpected mallctl() failure");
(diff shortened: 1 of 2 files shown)


interwq:
@guangli-dai two more tests fixed in this PR

test/unit/mallctl.c
```diff
@@ -1111,7 +1113,7 @@ TEST_BEGIN(test_thread_idle) {
 	 * This alloc and dalloc should leave something in the tcache, in a
 	 * small size's cache bin.
 	 */
-	void *ptr = mallocx(1, 0);
+	void *ptr = mallocx(1, MALLOCX_TCACHE_NONE);
```
guangli-dai, line 1119: Why do we not want to use the thread-specific cache?

interwq: We created a new arena, and want to allocate from that arena directly to measure the stats change, thus bypassing the tcache (since it may still contain items from other arenas). Maybe a tcache flush call at the beginning would be easier to read?

guangli-dai: I see. I would say the comments above this line is creating the confusion. Maybe just modify it to add the above explanation in?

interwq: Just updated. Also added the tcache flush to make everything strictly about the newly created arena.

guangli-dai: Looks good to me now!

guangli-dai (approved):

PR #2407, opened 2023-03-29
Rework the prof thread name feature

interwq:
Inline the storage for thread name in prof_tdata_t.

The previous approach managed the thread name in a separate buffer, which causes races because the thread name update (triggered by new samples) can happen at the same time as prof dumping (which reads the thread names) -- these two operations are under separate locks to avoid blocking each other.  Implemented the thread name storage as part of the tdata struct, which resolves the lifetime issue and also avoids internal alloc / dalloc during prof_sample.

In addition, rearranged the booleans in the tdata struct (which saved 8 bytes) -- this makes the sizeof(prof_tdata_t) remain 192 after the inlined thread name.

The new multithreaded prof_sys_thread_name test added stresses the thread name update and access paths. Verified that it caught the race being fixed here and in #2380

### DIFF
--- a/include/jemalloc/internal/prof_inlines.h
+++ b/include/jemalloc/internal/prof_inlines.h
@@ -38,6 +38,22 @@ prof_gdump_get_unlocked(void) {
 	return prof_gdump_val;
 }
 
+JEMALLOC_ALWAYS_INLINE void
+prof_thread_name_assert(prof_tdata_t *tdata) {
+	if (!config_debug) {
+		return;
+	}
+	prof_active_assert();
+
+	bool terminated = false;
+	for (unsigned i = 0; i < PROF_THREAD_NAME_MAX_LEN; i++) {
+		if (tdata->thread_name[i] == '\0') {
+			terminated = true;
+		}
+	}
+	assert(terminated);
+}
+
 JEMALLOC_ALWAYS_INLINE prof_tdata_t *
 prof_tdata_get(tsd_t *tsd, bool create) {
 	prof_tdata_t *tdata;
@@ -57,6 +73,10 @@ prof_tdata_get(tsd_t *tsd, bool create) {
 			tsd_prof_tdata_set(tsd, tdata);
 		}
 		assert(tdata == NULL || tdata->attached);
+	}
+
+	if (tdata != NULL) {
+		prof_thread_name_assert(tdata);
 	}
 
 	return tdata;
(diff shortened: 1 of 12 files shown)


include/jemalloc/internal/prof_inlines.h
```diff
@@ -255,4 +255,20 @@ prof_free(tsd_t *tsd, const void *ptr, size_t usize,
 	}
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+prof_thread_name_empty(prof_tdata_t *tdata) {
+	cassert(config_prof);
```
guangli-dai: `cassert(config_prof)` is included in `prof_active_assert()`. No need to repeat?

src/prof.c
```diff
@@ -415,10 +415,12 @@ prof_tdata_t *
 prof_tdata_reinit(tsd_t *tsd, prof_tdata_t *tdata) {
 	uint64_t thr_uid = tdata->thr_uid;
 	uint64_t thr_discrim = tdata->thr_discrim + 1;
-	char *thread_name = (tdata->thread_name != NULL) ?
-	    prof_thread_name_alloc(tsd, tdata->thread_name) : NULL;
 	bool active = tdata->active;
 
+	/* Keep a local copy of the thread name, before detaching. */
+	char thread_name[PROF_THREAD_NAME_LEN_MAX];
```
guangli-dai: Init to 0 here? It should have no danger to leave it uninitialized here since `prof_thread_name_write_tdata` will cut it at the end, but users may access garbage data.

include/jemalloc/internal/prof_inlines.h
interwq: Yep removed.

src/prof.c
interwq: The `strncpy` below actually serves as the init -- it's copying from the old tdata->thread_name which is guaranteed to be `\0` terminated (when no thread name is set, it's init to 0). Let me add more assertions to make it explicit.

guangli-dai: Sorry for being unclear. A corrupted data in the original `tdata` is what I am worried (although it shouldn't happen). Yes, an assertion is more appropriate than initialization here.

interwq: Yep added a new null-checking assert right before the `strncpy`.

guangli-dai (approved):
LGTM!

src/prof_recent.c
```diff
@@ -495,7 +495,7 @@ prof_recent_alloc_dump_node(emitter_t *emitter, prof_recent_t *node) {
 	    &node->alloc_tctx->thr_uid);
 	prof_tdata_t *alloc_tdata = node->alloc_tctx->tdata;
 	assert(alloc_tdata != NULL);
-	if (alloc_tdata->thread_name != NULL) {
+	if (!prof_thread_name_empty(alloc_tdata)) {
 		emitter_json_kv(emitter, "alloc_thread_name",
 		    emitter_type_string, &alloc_tdata->thread_name);
```
cemeyer, line 500: We are seeing a crash in `jemalloc_je_ctl_byname` -> `experimental_prof_recent_alloc_dump_ctl` -> `jemalloc_je_prof_recent_alloc_dump` as (I believe) a result of this change.

The problem is that `emitter_json_kv` -> `emitter_json_value` -> `emitter_print_value` [expects `emitter_type_string` parameters to be passed as a double pointer](https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/emitter.h#L165-L167), but because we changed `thread_name` from a `char *thread_name` to `char thread_name[]`, `&thread_name` is no longer `char **` -- it's just `char *`.

I think you might need to do something like this instead:

```
if (!prof_thread_name_empty()) {
  const char *thread_name = &alloc_tdata->thread_name;
  emitter_json_kv(emitter, "alloc_thread_name",
    emitter_type_string, &thread_name);
}
```

PR #2409, opened 2023-03-31
Disallow decay during reentrancy.

interwq:
Decay should not be triggered during reentrant calls (may cause lock order reversal / deadlocks).  Added a delay_trigger flag to the tickers to bypass decay when rentrancy_level is not zero.

### DIFF
--- a/include/jemalloc/internal/ticker.h
+++ b/include/jemalloc/internal/ticker.h
@@ -57,23 +57,27 @@ ticker_read(const ticker_t *ticker) {
 JEMALLOC_NOINLINE
 #endif
 static bool
-ticker_fixup(ticker_t *ticker) {
+ticker_fixup(ticker_t *ticker, bool delay_trigger) {
+	if (delay_trigger) {
+		ticker->tick = 0;
+		return false;
+	}
 	ticker->tick = ticker->nticks;
 	return true;
 }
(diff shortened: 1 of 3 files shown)


include/jemalloc/internal/ticker.h
```diff
@@ -150,26 +154,35 @@ ticker_geom_read(const ticker_geom_t *ticker) {
 JEMALLOC_NOINLINE
 #endif
 static bool
-ticker_geom_fixup(ticker_geom_t *ticker, uint64_t *prng_state) {
+ticker_geom_fixup(ticker_geom_t *ticker, uint64_t *prng_state,
+    bool delay_trigger) {
+	if (delay_trigger) {
+		ticker->tick = 0;
```
guangli-dai, line 160: With `tick` set to 0 here, the event will be triggered immediately after quitting the reentrancy. For cases where reentrancy occurs frequently in a sequential pattern, will the regression caused by extra events be a concern here?

interwq: This is already down the event triggering path, so there won't be any extra events. But rather the event that was supposed to be triggered during reentrancy will be delay until the next normal invocation.

guangli-dai (approved):

Issue #2412, opened 2023-04-12
Question about assertions in hpa_from_pai


ericm1024:
I've been trying to read and understand the HPA code, and I noticed these assertions in `hpa_from_pai`
```
static hpa_shard_t *
hpa_from_pai(pai_t *self) {
	assert(self->alloc = &hpa_alloc);
	assert(self->expand = &hpa_expand);
	assert(self->shrink = &hpa_shrink);
	assert(self->dalloc = &hpa_dalloc);
	return (hpa_shard_t *)self;
}
```
It looks like those assertions are re-assigning the fields every time? Is this a typo, i.e. should these be `==` instead of `=`?

Let me know if I've misunderstood something.

interwq:
It does look like a bug to me. Thanks for catching it! Please feel free to submit a PR, or let me know if you'd prefer us to do that.

PR #2415, opened 2023-04-18
Check for equality instead of assigning in asserts in hpa_from_pai.

ericm1024:
It appears like a simple typo means we're unconditionally overwriting some fields in hpa_from_pai when asserts are enabled. From hpa_shard_init, it looks like these fields have these values anyway, so this shouldn't cause bugs, but if something is wrong it seems better to have these asserts in place.

See issue #2412.


interwq:
Thanks for the fix!

PR #2416, opened 2023-04-20
Some nits in cache_bin.h

deadalnix:
Fix some indentation and use `+=`, nothing fancy.


interwq:
Thanks for the patch!

PR #2418, opened 2023-04-21
Remove errant `assert` in `arena_extent_alloc_large`

Svetlitski:
This codepath may generate deferred work when the HPA is enabled. See also [@davidtgoldblatt's relevant comment on the PR which introduced this](https://github.com/jemalloc/jemalloc/pull/2107#discussion_r699770967) which prevented a similarly incorrect `assert` from being added elsewhere.


Svetlitski:
Tagging @davidtgoldblatt since we discussed this.

davidtgoldblatt:
Looks reasonable to me, but I'll let Qi or Guangli stamp (IMO just deleting is fine, but moving it into the PAC implementation also seems reasonable).

PR #2423, opened 2023-05-01
Make `edata_cmp_summary_comp` 30% faster

Svetlitski:
`edata_cmp_summary_comp` is one of the very hottest functions, taking up 3% of all time spent inside Jemalloc. I noticed that all existing callsites rely only on the sign of the value returned by this function, so I came up with this equivalent branchless implementation which preserves this property. After empirical measurement, I have found that this implementation is 30% faster, therefore representing a 1% speed-up to the allocator as a whole.

### DIFF
--- a/include/jemalloc/internal/edata.h
+++ b/include/jemalloc/internal/edata.h
@@ -664,13 +664,20 @@ edata_cmp_summary_get(const edata_t *edata) {
 
 static inline int
 edata_cmp_summary_comp(edata_cmp_summary_t a, edata_cmp_summary_t b) {
-	int ret;
-	ret = (a.sn > b.sn) - (a.sn < b.sn);
-	if (ret != 0) {
-		return ret;
-	}
-	ret = (a.addr > b.addr) - (a.addr < b.addr);
-	return ret;
+	/*
+	 * Logically, what we're doing here is comparing based on `.sn`, and
+	 * falling back to comparing on `.addr` in the case that `a.sn == b.sn`.
+	 * We accomplish this by multiplying the result of the `.sn` comparison
+	 * by 2, so that so long as it is not 0, it will dominate the `.addr`
+	 * comparison in determining the sign of the returned result value.
+	 * The justification for doing things this way is that this is
+	 * branchless - all of the branches that would be present in a
+	 * straightforward implementation are common cases, and thus the branch
+	 * prediction accuracy is not great. As a result, this implementation
+	 * is measurably faster (by around 30%).
+	 */
+	return (2 * ((a.sn > b.sn) - (a.sn < b.sn))) +
+	       ((a.addr > b.addr) - (a.addr < b.addr));
 }
 
 static inline int
@@ -683,15 +690,11 @@ edata_snad_comp(const edata_t *a, const edata_t *b) {
 
 static inline int
 edata_esnead_comp(const edata_t *a, const edata_t *b) {
-	int ret;
-
-	ret = edata_esn_comp(a, b);
-	if (ret != 0) {
-		return ret;
-	}
-
-	ret = edata_ead_comp(a, b);
-	return ret;
+	/*
+	 * Similar to `edata_cmp_summary_comp`, we've opted for a
+	 * branchless implementation for the sake of performance.
+	 */
+	return (2 * edata_esn_comp(a, b)) + edata_ead_comp(a, b);
 }
 
 ph_proto(, edata_avail, edata_t)


include/jemalloc/internal/edata.h
```diff
@@ -664,13 +664,7 @@ edata_cmp_summary_get(const edata_t *edata) {
...
 static inline int
 edata_cmp_summary_comp(edata_cmp_summary_t a, edata_cmp_summary_t b) {
-	int ret;
-	ret = (a.sn > b.sn) - (a.sn < b.sn);
-	if (ret != 0) {
-		return ret;
-	}
-	ret = (a.addr > b.addr) - (a.addr < b.addr);
-	return ret;
+  return (2 * ((a.sn > b.sn) - (a.sn < b.sn))) + ((a.addr > b.addr) - (a.addr < b.addr));
```
interwq: nit: probably explain a bit what this is doing and why the (2*) is needed. maybe even keep the plain logic in the comment.

Svetlitski: Updated with descriptive comment explaining the logic.

interwq: A few more nits:
1) the max line length should be 80 char
2) for multi-line comments, the format should be ([pasted from here](https://github.com/jemalloc/jemalloc/wiki/Coding-Style))
```
/*
 * This is a multi-line comment.  Follow sentence-ending periods with two
 * spaces.  Consider that we display code with fixed-width fonts; embrace the
 * old 20th-century convention and like it.
 */
```
3) Probably mention in the comments that, one factor that makes this optimization necessary is the fact that all branches here are common cases, i.e. it's hard to predict the outcome of the branches.
4) For consistency, maybe also apply the same trick to `edata_esnead_comp` below? In case it becomes more popular in the future.

Svetlitski: Thank you for the feedback. I have addressed your comments in the latest revision.

mpoeter:
Nice idea! But if the sign of the return value is actually enough, wouldn't this be even simpler?
```cpp
int edata_cmp_summary_comp3(edata_cmp_summary_t a, edata_cmp_summary_t b) {
    auto ret = a.sn - b.sn;
    if (ret == 0) {
        ret = (a.addr - b.addr);
    }
    return ret;
}
```
Yes, this does contain a branch, but this branch can easily be [removed by the compiler](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAOwAbKSuADJ4DJgAcj4ARpjEEsEADqgKhE4MHt6%2BASlpGQLhkTEs8YlcwXaYDplCBEzEBNk%2BfkG2mPaOAvWNBMXRcQlJtg1NLbntCmP9EYNlw5UAlLaoXsTI7BzmAMwRyN5YANQmO27T%2BIIAdAin2CYaAIIPj9PEXg5HmOhMDQD6yBYyT%2BCh8LEaAE8/gQTv4rE8jkcvBECIFJNCjgoGKd4Y9EcjBMkCMQMUx0OhiDiXiZ/AARKlPF4or4/f6A4Ggljg4hQtBAiDfX5MAFAkFgyGk0gsoUijninkY2JLWG4h4ATmZxEwMNOtKOECYVyxJx22COsSNDGVAFp9Ybjac3ObLUsGc8NBqqPqtTCwGBdUcNMqaaqPeqfesGEcfW71TT6YyPRHiFGDVcyRSTWaLRniDa7enycQTU6c0XXTtVXTqU9mYK2aLOdzeaggWYBazheyxVyJTCmFL613G/KoTClSqa2rk6mzEcAFT6tMO03OrH55dRx1rq1LYOWKeIo%2BIqCG3NZ525jdnosly/litVhPux51zuynvNgGt5I7Dsy7smz7I4B2lBs5V7BVx2DOEXmPJgvCIaNtRNPV7SjW0LSxN1jzwL0IB9VCAyDScEWPI9CIDNNz0wwsKUfOCj3jRjERnZCCFjOkOBWWhOAAVl4PwOC0UhUE4NxrGsTE1g2TATjMHYeFIAhNG4lYAGsQD4rgrkkAAOLgzECHZ9L0nYjMkYJeI4SRBNU0TOF4BQQA0ZTVJWOBYCQPlkjoBJyEoHy/MSA5DGASoNFcrAADc8E2AA1PBMAAdwAeWSRhOCUmhaAIBJnIgWJ7NiCJISy3gSuYHlUtibQahU7heD5NhBFShhaAheysHBIxxGE3h8C1WposwZz%2BpCVQakQrYRJRTp7NoPBYmISEPCweziTwFhypWKgDGABREpS9LMsamRBBEMR2Ckc75CUNR7N0Lh9DClBJMsfQlucyAVlQIlMjG61Up2I5rQAdTEWgnM6erMhcBh3E8Vo9DCOZSnKPRUnSbosiR3JnqxwoGAGdHFg6Lo6hmCY/Ge6pah6GYSaGCpRj6am9GmPomYWCoVgUGTNgkHj%2BLs8axI4I5VD0wJrTRI5QqMI5KiuDQVf1XBCBIeTFKWXgGq0PdSE0/Srm0vT/B2QJtMqC2%2BLMaRrNs0ghJE8WnJctz%2BsN6yzFF13HM9g2VhG4h0mcSQgA%3D) (at least on x86-64).

Svetlitski:
> Nice idea! But if the sign of the return value is actually enough, wouldn't this be even simpler?
> 
> ```c++
> int edata_cmp_summary_comp3(edata_cmp_summary_t a, edata_cmp_summary_t b) {
>     auto ret = a.sn - b.sn;
>     if (ret == 0) {
>         ret = (a.addr - b.addr);
>     }
>     return ret;
> }
> ```
> 
> Yes, this does contain a branch, but this branch can easily be [removed by the compiler](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAOwAbKSuADJ4DJgAcj4ARpjEEsEADqgKhE4MHt6%2BASlpGQLhkTEs8YlcwXaYDplCBEzEBNk%2BfkG2mPaOAvWNBMXRcQlJtg1NLbntCmP9EYNlw5UAlLaoXsTI7BzmAMwRyN5YANQmO27T%2BIIAdAin2CYaAIIPj9PEXg5HmOhMDQD6yBYyT%2BCh8LEaAE8/gQTv4rE8jkcvBECIFJNCjgoGKd4Y9EcjBMkCMQMUx0OhiDiXiZ/AARKlPF4or4/f6A4Ggljg4hQtBAiDfX5MAFAkFgyGk0gsoUijninkY2JLWG4h4ATmZxEwMNOtKOECYVyxJx22COsSNDGVAFp9Ybjac3ObLUsGc8NBqqPqtTCwGBdUcNMqaaqPeqfesGEcfW71TT6YyPRHiFGDVcyRSTWaLRniDa7enycQTU6c0XXTtVXTqU9mYK2aLOdzeaggWYBazheyxVyJTCmFL613G/KoTClSqa2rk6mzEcAFT6tMO03OrH55dRx1rq1LYOWKeIo%2BIqCG3NZ525jdnosly/litVhPux51zuynvNgGt5I7Dsy7smz7I4B2lBs5V7BVx2DOEXmPJgvCIaNtRNPV7SjW0LSxN1jzwL0IB9VCAyDScEWPI9CIDNNz0wwsKUfOCj3jRjERnZCCFjOkOBWWhOAAVl4PwOC0UhUE4NxrGsTE1g2TATjMHYeFIAhNG4lYAGsQD4rgrkkAAOLgzECHZ9L0nYjMkYJeI4SRBNU0TOF4BQQA0ZTVJWOBYCQPlkjoBJyEoHy/MSA5DGASoNFcrAADc8E2AA1PBMAAdwAeWSRhOCUmhaAIBJnIgWJ7NiCJISy3gSuYHlUtibQahU7heD5NhBFShhaAheysHBIxxGE3h8C1WposwZz%2BpCVQakQrYRJRTp7NoPBYmISEPCweziTwFhypWKgDGABREpS9LMsamRBBEMR2Ckc75CUNR7N0Lh9DClBJMsfQlucyAVlQIlMjG61Up2I5rQAdTEWgnM6erMhcBh3E8Vo9DCOZSnKPRUnSbosiR3JnqxwoGAGdHFg6Lo6hmCY/Ge6pah6GYSaGCpRj6am9GmPomYWCoVgUGTNgkHj%2BLs8axI4I5VD0wJrTRI5QqMI5KiuDQVf1XBCBIeTFKWXgGq0PdSE0/Srm0vT/B2QJtMqC2%2BLMaRrNs0ghJE8WnJctz%2BsN6yzFF13HM9g2VhG4h0mcSQgA%3D) (at least on x86-64).

The implementation you have suggested is incorrect, as `.sn` and `.addr` are 64 bit values, and the return value from the function is an `int`, so the result could be truncated. To illustrate, the following program prints `0`:
```c
#include <stdio.h>
#include <stdint.h>

int main() {
	uint64_t x = 1UL<<32;
	uint64_t y = 0;
	int z = x - y;
	printf("%d\n", z);
}
```

Also, in general I'd caution against using the output of a C++ compiler to reason about C code, as there are subtle differences in what optimizations are allowed.

mpoeter:
Good point! But given that C20 specifies that [signed integral types must use two's complement](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0907r4.html) we can use the following (at least if we compile for C/C++20):
```
int edata_cmp_summary_comp3(edata_cmp_summary_t a, edata_cmp_summary_t b) {
    int64_t ret = a.sn - b.sn;
    if (ret == 0) {
        ret = (a.addr - b.addr);
    }
    auto shifted = ret >> 32;
    if (shifted != 0) {
        return shifted;
    }
    return ret;
}
```
Since "right-shift on signed integral types is an arithmetic right shift, which performs sign-extension"  (see link), `-1 >> 32 == -1`.

On [godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAOwAbKSuADJ4DJgAcj4ARpjEEsEADqgKhE4MHt6%2BASlpGQLhkTEs8YlcwXaYDplCBEzEBNk%2BfkG2mPaOAvWNBMXRcQlJtg1NLbntCmP9EYNlw5UAlLaoXsTI7BzmAMwRyN5YANQmO27T%2BIIAdAin2CYaAIIPj9PEXg5HmOhMDQD6yBYyT%2BCh8LEaAE8/gQTv4rE8jkcvBECIFJNCjgoGKd4Y9EcjBMkCMQMUx0OhiDiXiZ/AARKlPF4or4/f6A4Ggljg4hQtBAiDfX5MAFAkFgyGk0gsoUijninkY2JLWG4h4ATmZxEwMNOtKOECYVyxJx22COsSNDGVAFp9Ybjac3ObLUsGc8NBqqPqtTCwGBdUcNMqaaqPeqfesGEcfW71TT6YyPRHiFGDVcyRSTWaLRniDa7enycQTU6c0XXTtVXTqU9mYK2aLOdzeaggWYBazheyxVyJTCmFL613G/KoTClSqa2rk6mzEcAFT6tMO03OrH55dRx1rq1LYOWKeIo%2BIqCG3NZ525jdnosly/litVhPux51zuynvNgGt5I7Dsy7smz7I4B2lBs5V7BVx2DOEXmPJgvCIaNtRNPV7SjW0LSxN1jzwL0IB9VCAyDScEWPI9CIDNNz0wwsKUfOCj3jRjERnZCCFjOkOBWWhOAAVl4PwOC0UhUE4NxrGsTE1g2TATjMHYeFIAhNG4lYAGsQD4rgrkkAAOLgzECHZ9L0nYjMkYJeI4SRBNU0TOF4BQQA0ZTVJWOBYCQPlkjoBJyEoHy/MSA5DGASoNFcrAADc8E2AA1PBMAAdwAeWSRhOCUmhaAIBJnIgWJ7NiCJISy3gSuYHlUtibQahU7heD5NhBFShhaAheysHBIxxGE3h8C1WposwZz%2BpCVQakQrYRJRTp7NoPBYmISEPCweziTwFhypWKgDGABREpS9LMsamRBBEMR2Ckc75CUNR7N0Lh9DClBJMsfQlucyAVlQIlMjG61Up2I5rQAdTEWgnM6erMhcBh3E8Vo9DCOZSnKPRUnSbosiR3JnqxwoGAGdHFg6Lo6hmCY/Ge6pah6GYSaGCpRj6am9GmPomYWCoVgUGTNgkHj%2BLs8axI4I5VD0wJrTRI5QqMI5KiuDQVf1XBCBIeTFKWXgGq0PdSE0/Srm0vT/B2QJtMqC2%2BLMaRrNs0ghJE8WnJctz%2BsN6yzFF13HM9g2VhG4h0mcSQgA%3D) this produces the following code:
```
        sub     rsi, rcx
        sub     rdi, rdx
        cmove   rdi, rsi
        mov     rax, rdi
        sar     rax, 32
        cmove   eax, edi
        ret
```

PR #2430, opened 2023-05-08
Remove trailing whitespace

Svetlitski:
Additionally, added a GitHub Action to ensure no more trailing whitespace will creep in again in the future.

I'm excluding Markdown files from this check, since trailing whitespace is significant there, and also excluding `build-aux/install-sh` because there is significant trailing whitespace on the line that sets `defaultIFS`.


interwq (commented):
@Svetlitski should we land this one after #2447?

Svetlitski:
I believe they should be independent – we can land them in either order.

PR #2432, opened 2023-05-09
Fix segfault in `extent_try_coalesce_impl`

Svetlitski:
Static analysis flagged this. `extent_record` was passing `NULL` as the value for `coalesced` to `extent_try_coalesce`, which in turn passes that argument to `extent_try_coalesce_impl`, where it is written to without checking if it is `NULL`. I can confirm from reviewing the fleetwide coredump data that this was in fact being hit in production.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -822,6 +822,7 @@ static edata_t *
 extent_try_coalesce_impl(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
     ecache_t *ecache, edata_t *edata, bool *coalesced) {
 	assert(!edata_guarded_get(edata));
+	assert(coalesced != NULL);
 	/*
 	 * We avoid checking / locking inactive neighbors for large size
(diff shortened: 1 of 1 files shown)


src/extent.c
```diff
@@ -929,7 +929,7 @@ extent_record(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, ecache_t *ecache,
 	}
 	if (!ecache->delay_coalesce) {
 		edata = extent_try_coalesce(tsdn, pac, ehooks, ecache, edata,
-		    NULL);
+		    &(bool){false});
```
interwq: Good catch!

nit: let's still add a plain `bool coalesced_unused;` and pass it here. I'm a bit concerned some of the older compilers might not like it this way.

Let's also add an `assert(coalesced != NULL);` here https://github.com/jemalloc/jemalloc/blob/70344a2d38eb71a162ea19d1a4fee8f0d168588b/src/extent.c#L824

PR #2433, opened 2023-05-09
Fix segfault in rb `*_tree_remove`

Svetlitski:
Static analysis flagged this. It's possible to segfault in the `*_tree_remove` function generated by `rb_gen`, as `nodep` may still be `NULL` after the initial for loop. I can confirm from reviewing the fleetwide coredump data that this was in fact being hit in production, primarily through `tctx_tree_remove`, and much more rarely through `gctx_tree_remove`.

Giving each of these static analysis fixes their own PR per @interwq's preference, and because it makes things clearer.

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -42,6 +42,34 @@ arena_choose_maybe_huge(tsd_t *tsd, arena_t *arena, size_t size) {
 	return tsd_arena;
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+large_dalloc_safety_checks(edata_t *edata, const void *ptr, szind_t szind) {
+	if (!config_opt_safety_checks) {
+		return false;
+	}
+
+	/*
+	 * Eagerly detect double free and sized dealloc bugs for large sizes.
+	 * The cost is low enough (as edata will be accessed anyway) to be
+	 * enabled all the time.
+	 */
+	if (unlikely(edata == NULL ||
+	    edata_state_get(edata) != extent_state_active)) {
+		safety_check_fail("Invalid deallocation detected: "
+		    "pages being freed (%p) not currently active, "
+		    "possibly caused by double free bugs.", ptr);
+		return true;
+	}
+	size_t input_size = sz_index2size(szind);
+	if (unlikely(input_size != edata_usize_get(edata))) {
+		safety_check_fail_sized_dealloc(/* current_dealloc */ true, ptr,
+		    /* true_size */ edata_usize_get(edata), input_size);
+		return true;
+	}
+
+	return false;
+}
+
 JEMALLOC_ALWAYS_INLINE void
 arena_prof_info_get(tsd_t *tsd, const void *ptr, emap_alloc_ctx_t *alloc_ctx,
     prof_info_t *prof_info, bool reset_recent) {
@@ -65,6 +93,11 @@ arena_prof_info_get(tsd_t *tsd, const void *ptr, emap_alloc_ctx_t *alloc_ctx,
 	if (unlikely(!is_slab)) {
 		/* edata must have been initialized at this point. */
 		assert(edata != NULL);
+		if (reset_recent &&
+		    large_dalloc_safety_checks(edata, ptr,
+		    edata_szind_get(edata))) {
+			return;
+		}
 		large_prof_info_get(tsd, edata, prof_info, reset_recent);
 	} else {
 		prof_info->alloc_tctx = (prof_tctx_t *)(uintptr_t)1U;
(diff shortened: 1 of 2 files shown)


interwq:
Discussed offline; the crash is likely from application level heap corruption, e.g. double frees of large buffers (which are likely to be prof sampled) may end up crashing in the rb tree remove here.

@Svetlitski a few pointers for opt mode sanity checks:

1) the large dealloc check (which is probably closer to what you need here, e.g. adding a prof_sample_free_safety_checks): https://github.com/jemalloc/jemalloc/blob/d4a2b8bab10980d4677d43560f27ac9ef66cde45/include/jemalloc/internal/arena_inlines_b.h#L215

2) the size index check during tcache flush (the ones caused by sized dealloc wrong input sizes):
https://github.com/jemalloc/jemalloc/blob/d4a2b8bab10980d4677d43560f27ac9ef66cde45/src/tcache.c#L292

Feel free to let me know if any questions.

Svetlitski:
I added a safety check in line with the examples provided @interwq.

include/jemalloc/internal/rb.h
```diff
@@ -852,6 +865,7 @@ a_prefix##remove(a_rbt_type *rbtree, a_type *node) {			\
 	    }								\
 	}								\
     }									\
+    rb_remove_safety_checks(rbtree, node, nodep, __func__);		\
```
interwq: Sorry for the delay on the PR. As discussed we suspect this was caused by double free of sampled allocations -- let's add another sanity check before calling down to the remove function, to verify that the pointer being deallocated is still in an active state (not using the rb rtree, but rather its edata metadata).

This is probably a convenient place to do so: 
https://github.com/jemalloc/jemalloc/blob/86eb49b47847e48390c672371987ff4e476e53a3/include/jemalloc/internal/arena_inlines_b.h#L67

and we can add `large_dalloc_safety_checks` there, e.g.
```
if (reset_recent) {
    large_dalloc_safety_checks(edata, ptr, edata_szind_get(edata));
}
```
It's not ideal to do this in a _info_get utility function, but this looks to have all the relevant context, with the `reset_recent` flag indicating a one-time operation that can only be done on active allocations.

include/jemalloc/internal/rb.h
```diff
@@ -560,6 +560,19 @@ a_prefix##reverse_iter_filtered(a_rbt_type *rbtree, a_type *start,	\
  * the same as with the unfiltered version, with the added constraint that the
  * returned node must pass the filter.
  */
+JEMALLOC_ALWAYS_INLINE void
+rb_remove_safety_checks(void *rbtree, void *node, void *nodep, const char *function_name) {
+	if (!config_opt_safety_checks) {
+		return;
+	}
+	if (unlikely(nodep == NULL)) {
+		safety_check_fail("Invalid deallocation detected in %s: "
```
interwq: nit: add `<jemalloc>:` at the beginning. Maybe not spill out the node and tree addresses? These are not user facing and may cause confusion instead.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -65,6 +93,10 @@ arena_prof_info_get(tsd_t *tsd, const void *ptr, emap_alloc_ctx_t *alloc_ctx,
 	if (unlikely(!is_slab)) {
 		/* edata must have been initialized at this point. */
 		assert(edata != NULL);
+		if (reset_recent &&
+		    large_dalloc_safety_checks(edata, ptr, edata_szind_get(edata))) {
```
interwq: nit: over 80-char max length. With the indent maybe:
```
		if (reset_recent && large_dalloc_safety_checks(edata, ptr,
		     edata_szind_get(edata))) {
```

include/jemalloc/internal/rb.h
```diff
@@ -560,6 +560,19 @@ a_prefix##reverse_iter_filtered(a_rbt_type *rbtree, a_type *start,	\
  * the same as with the unfiltered version, with the added constraint that the
  * returned node must pass the filter.
  */
+JEMALLOC_ALWAYS_INLINE void
+rb_remove_safety_checks(const void *nodep, const char *function_name) {
+	if (!config_opt_safety_checks) {
+		return;
+	}
+	if (unlikely(nodep == NULL)) {
+		safety_check_fail("<jemalloc>: Invalid deallocation detected in %s: "
```
interwq: nit: over 80-char length too (In case you need to change your editor settings, each tab is 8 space wide).

interwq (approved):
CI looks good. Just a couple of style nits -- once you update I'll merge w/o CI.

include/jemalloc/internal/arena_inlines_b.h
Svetlitski: Ahh, sorry, had my editor set to 4-space tabs.

include/jemalloc/internal/rb.h
Svetlitski: Fixed

PR #2435, opened 2023-05-10
Fix the prof thread_name reference in prof_recent dump.

### LINKED ISSUE #2434: Crash in prof_recent_alloc_dump after #2407
We are seeing a crash in `ctl_byname` -> `experimental_prof_recent_alloc_dump_ctl` -> `prof_recent_alloc_dump` as (I believe) a result of this change.

The problem is that `emitter_json_kv` -> `emitter_json_value` -> `emitter_print_value` [expects `emitter_type_string` parameters to be passed as a double pointer](https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/emitter.h#L165-L167), but because we changed `thread_name` from a `char *thread_name` to `char thread_name[]` in #2407, `&thread_name` is no longer `char **` -- it's just `char *`.  Or at least, I think that is what is happening.

I think you might need to do something like this instead:

```
if (!prof_thread_name_empty()) {
  const char *thread_name = &alloc_tdata->thread_name;
  emitter_json_kv(emitter, "alloc_thread_name",
    emitter_type_string, &thread_name);
}
```

interwq:
As pointed out in #2434, the thread_name in prof_tdata_t was changed in #2407. This also requires an update for the prof_recent dump, specifically the emitter expects a "char **" which is fixed in this commit.

Also, add the prof_sys_thread_name feature in the prof_recent unit test, so that the combination of the two features is covered.  Verified that it catches the issue being fixed in this PR.

As a precaution, also went through all usage containing "thread_name" in the code base to make sure no similar issue exists.

Fixes #2434.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1926,6 +1926,16 @@ dnl Check if we have dlsym support.
   if test "x${je_cv_pthread_getname_np}" = "xyes" ; then
     AC_DEFINE([JEMALLOC_HAVE_PTHREAD_GETNAME_NP], [ ], [ ])
   fi
+  dnl Check if pthread_set_name_np is available with the expected API.
+  JE_COMPILABLE([pthread_set_name_np(3)], [
+#include <pthread.h>
+#include <pthread_np.h>
+], [
+  pthread_set_name_np(pthread_self(), "set_name_test");
+], [je_cv_pthread_set_name_np])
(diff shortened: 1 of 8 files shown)


cemeyer (approved):
The proposed fix looks good to me.

guangli-dai (approved):

PR #2437, opened 2023-05-11
Remove dead stores detected by static analysis

Svetlitski:
None of these are harmful, and they are almost certainly optimized away by the compiler. The motivation for fixing them anyway is that we'd like to enable static analysis as part of CI, and the first step towards that is resolving the warnings it produces at present.

### DIFF
--- a/src/malloc_io.c
+++ b/src/malloc_io.c
@@ -423,7 +423,6 @@ malloc_vsnprintf(char *str, size_t size, const char *format, va_list ap) {
 			unsigned char len = '?';
 			char *s;
 			size_t slen;
-			bool first_width_digit = true;
 			bool pad_zero = false;
 
 			f++;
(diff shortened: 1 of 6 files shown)


src/malloc_io.c
```diff
@@ -474,7 +471,6 @@ malloc_vsnprintf(char *str, size_t size, const char *format, va_list ap) {
 				assert(uwidth != UINTMAX_MAX || get_errno() !=
 				    ERANGE);
 				width = (int)uwidth;
-				first_width_digit = false;
```
interwq, line 477: To confirm, this is the dead store we are getting rid of, right?

Svetlitski: Yes. The others were pretty straightforward, but this one took me a bit to convince myself it was correct. It _is_ a dead store, but unlike the others I'm not entirely sure that just removing it is the correct thing to do.

PR #2439, opened 2023-05-12
Added autoconf to installation instructions

jimw567:
./configure command doesn't exist in the repo. It needs to be generated using autoconf first.


interwq:
The section you are looking at is for the packaged jemalloc release (the tar downloaded at https://github.com/jemalloc/jemalloc/releases/tag/5.3.0), which has `configure` shipped within.

When building your own jemalloc from github source, the instructions are in the next section following `If building from unpackaged developer sources`

PR #2443, opened 2023-05-16
Fix build on GNU/Hurd

sthibaul:


bitcoinfinancier (approved):

paravoid:
I maintain the Debian packages for jemalloc, and @sthibaul is also with the Debian project, and one of the GNU Hurd porters. Samuel also had to submit a couple of changes ([1](https://sourceware.org/git/?p=glibc.git;a=commit;h=644aa127b9b42a899a12b6ccc6644bc035c231e3), [2](https://sourceware.org/git/?p=glibc.git;a=commit;h=53da64d1cf3694a132287d1ab6e9655031124e32)) to glibc for jemalloc to work on Hurd; he's definitely the subject-matter expert here!

I've included the changes in this PR in our jemalloc (5.3.0) Debian package, and it has built successfully ([build log](https://buildd.debian.org/status/fetch.php?pkg=jemalloc&arch=hurd-i386&ver=5.3.0-2&stamp=1694537714&raw=0)) on Hurd, and is passing the test suite with no warnings.

I doubt Hurd is a priority for jemalloc, but given this works and the changes are trivial and not invasive at all, nor hurt any other ports, I was hoping you'd just merge it, so that we don't have to carry any downstream patches. Thanks!

src/background_thread.c
```diff
@@ -116,24 +116,28 @@ set_current_thread_affinity(int cpu) {
 #if defined(JEMALLOC_HAVE_SCHED_SETAFFINITY) || defined(JEMALLOC_HAVE_PTHREAD_SETAFFINITY_NP)
 #if defined(JEMALLOC_HAVE_SCHED_SETAFFINITY)
 	cpu_set_t cpuset;
-#else
+#elif defined(JEMALLOC_HAVE_PTHREAD_GETAFFINITY_NP)
```
interwq, line 119: Question: does Hurd have pthread_getaffinity_np but not pthread_setaffinity_np? Asking because L116 have the `defined(JEMALLOC_HAVE_PTHREAD_SETAFFINITY_NP)` check.

This section of the code is assuming background thread is supported, so we should not have unsupported cases workaround here. See L86 around `#ifndef JEMALLOC_BACKGROUND_THREAD`. If it's not possible to support bg thds on hurd, let's make it explicitly so. Maybe through: https://github.com/jemalloc/jemalloc/blob/72cfdce71806443f4ccdbfe10aa5d50346a3d07e/configure.ac#L2575

src/jemalloc.c
```diff
@@ -747,8 +747,11 @@ malloc_ncpus(void) {
 #  endif
 #  if defined(JEMALLOC_HAVE_SCHED_SETAFFINITY)
 		sched_getaffinity(0, sizeof(set), &set);
-#  else
+#  elif defined(JEMALLOC_HAVE_PTHREAD_GETAFFINITY_NP)
 		pthread_getaffinity_np(pthread_self(), sizeof(set), &set);
+#  else
+		CPU_ZERO(&set);
```
interwq, line 753: This will cause jemalloc to detect wrong number of CPUs, and likely will not perform well, even defeating the purpose of using jemalloc. Is there no way to do so on hurd?

interwq (commented):
Thanks for sharing the context @paravoid and @sthibaul. Added some questions.

src/background_thread.c
sthibaul: > does Hurd have pthread_getaffinity_np but not pthread_setaffinity_np?

No, it has none of the two (and not sched_setaffinity either)

> This section of the code is assuming background thread is supported

Mmm, ok, but is thread binding really necessary for background thread support?

src/jemalloc.c
sthibaul: SMP support is still WIP, so it's actually the correct number of CPUs :)
That said, `sysconf(_SC_NPROCESSORS_ONLN)` will properly report 1, so perhaps it's the `#elif defined(CPU_COUNT)` above that should rather be `#elif defined(CPU_COUNT) && (defined(JEMALLOC_HAVE_SCHED_SETAFFINITY) || defined(JEMALLOC_HAVE_PTHREAD_GETAFFINITY_NP)`

paravoid:
Hi! Any news here?

PR #2446, opened 2023-05-17
Fix remaining static analysis warnings

Svetlitski:
Fix or suppress the remaining warnings generated by static analysis. This is a necessary step before we can incorporate static analysis into CI. Where possible, I've preferred to modify the code itself instead of just disabling the warning with a magic comment, so that if we decide to use different static analysis tools in the future we will be covered against them raising similar warnings.

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_inlines_c.h
+++ b/include/jemalloc/internal/jemalloc_internal_inlines_c.h
@@ -325,6 +325,8 @@ imalloc_fastpath(size_t size, void *(fallback_alloc)(size_t)) {
 	tcache_t *tcache = tsd_tcachep_get(tsd);
 	assert(tcache == tcache_get(tsd));
 	cache_bin_t *bin = &tcache->bins[ind];
+	/* Suppress spurious warning from static analysis */
+	assert(bin != NULL);
 	bool tcache_success;
 	void *ret;
 
--- a/include/jemalloc/internal/ph.h
+++ b/include/jemalloc/internal/ph.h
@@ -127,6 +127,7 @@ phn_merge_ordered(void *phn0, void *phn1, size_t offset,
 	phn0child = phn_lchild_get(phn0, offset);
 	phn_next_set(phn1, phn0child, offset);
 	if (phn0child != NULL) {
+		/* NOLINTNEXTLINE(readability-suspicious-call-argument) */
 		phn_prev_set(phn0child, phn1, offset);
 	}
 	phn_lchild_set(phn0, phn1, offset);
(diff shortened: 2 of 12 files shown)


include/jemalloc/internal/rtree.h
```diff
@@ -299,6 +299,10 @@ rtree_leaf_elm_write(tsdn_t *tsdn, rtree_t *rtree,
 	assert((uintptr_t)contents.edata % EDATA_ALIGNMENT == 0);
 	void *bits;
 	unsigned additional;
+  /* Suppress spurious warning from static analysis */
+  if (config_debug) {
+    additional = 0;
+  }
```
interwq: nit: put this into `rtree_contents_encode`, or that still gets the warning?

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -325,6 +325,8 @@ imalloc_fastpath(size_t size, void *(fallback_alloc)(size_t)) {
 	tcache_t *tcache = tsd_tcachep_get(tsd);
 	assert(tcache == tcache_get(tsd));
 	cache_bin_t *bin = &tcache->bins[ind];
+  /* Suppress spurious warning from static analysis */
+  assert(bin != NULL);
```
interwq: nit: indent is off

include/jemalloc/internal/ph.h
```diff
@@ -127,6 +127,7 @@ phn_merge_ordered(void *phn0, void *phn1, size_t offset,
 	phn0child = phn_lchild_get(phn0, offset);
 	phn_next_set(phn1, phn0child, offset);
 	if (phn0child != NULL) {
+		/* NOLINTNEXTLINE(readability-suspicious-call-argument) */
```
interwq, line 130: Question: if we put the NOLINT into phn_prev_ / _next_set, would that work too?

include/jemalloc/internal/rtree.h
```diff
@@ -481,6 +485,10 @@ rtree_write_range_impl(tsdn_t *tsdn, rtree_t *rtree, rtree_ctx_t *rtree_ctx,
 	 */
 	void *bits;
 	unsigned additional;
+  /* Suppress spurious warning from static analysis */
+  if (config_debug) {
+    additional = 0;
+  }
```
interwq: Similar indent here and below.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
Svetlitski: Sorry, this commit was written before I had fixed my editor settings in response to your other reviews.

include/jemalloc/internal/ph.h
Svetlitski: I believe it has to be at the callsite, but even if not, I would advise leaving it as-is, because while these are false-positives, I think the lint has value in general. For example if you had code like:
```c
void *root = ...;
void *prev = ...;
...
phn_prev_set(prev, root, OFFSET);
```

Then the lint would save you from a mistake where you accidentally flipped the order of the arguments like that.

Issue #2451, opened 2023-05-24
jeprof profling 5.2.1 / 5.3.0 differences: prof_backtrace_impl


awelzel:
Hello, I've been stumbling into a difference how 5.2.1 and 5.3.0 create and present profiling information. With 5.3.0 jeprof includes `prof_backtrace_impl` dominating the profile in my case that isn't visible with 5.2.1.

This might be something on my end, but given 5.2.1 works nicely out of the box, figured I'd report in case that might be a regression. The prof_backtrace_impl also seems like an implementation detail that should not show up?

This is on Linux, Ubuntu 22.04.2 LTS, compiling 5.2.1 and 5.3.0 from a git checkout as follows:
```
$ git clean -fdx .
$ ./autogen.sh && ./configure --enable-prof --prefix=/usr/local/jemalloc-5.x.y
$ make -j6 && sudo make install
```

When profiling an application with 5.2.1, the jeprof text output is as follows, most functions being related to the application.

```
$ LD_PRELOAD=/usr/local/jemalloc-5.2.1/lib/libjemalloc.so.2 MALLOC_CONF="prof:true,prof_prefix:jeprof521.out,prof_final:true,lg_prof_sample:17" timeout 5 /opt/zeek/bin/zeek -C -i dummy0

# /usr/local/jemalloc-5.2.1/bin/jeprof /opt/zeek/bin/zeek jeprof521.out.46819.0.f.heap 
Using local file /opt/zeek/bin/zeek.
Using local file jeprof521.out.46819.0.f.heap.
Welcome to jeprof!  For help, type 'help'.
(jeprof) top
Total: 45.4 MB
     8.6  19.0%  19.0%     39.0  85.9% yyparse
     7.0  15.4%  34.4%     10.3  22.6% zeek::make_intrusive
     5.3  11.6%  46.0%      5.3  11.6% zeek::util::safe_realloc
     5.3  11.6%  57.6%      5.3  11.6% zeek::Obj::SetLocationInfo
     4.6  10.1%  67.7%      4.6  10.1% __gnu_cxx::new_allocator::allocate
     1.9   4.2%  71.9%      1.9   4.2% zeek::detail::EquivClass::EquivClass
     1.9   4.1%  76.0%      4.4   9.6% zeek::detail::NFA_Machine::AppendEpsilon
     1.8   3.9%  79.9%      2.9   6.4% zeeklex
     1.3   2.8%  82.7%      1.3   2.8% zeek::util::copy_string
[...]

PR #2452, opened 2023-05-24
jeprof pruning

awelzel:
This adds a bit of pruning to jeprof to avoid having profiles ending in `prof_backtrace_impl`.

See #2451 for a bit of motivation.


interwq:
Thanks for the patch!

PR #2456, opened 2023-05-31
Use a local variable to set the alignment for this particular allocation

zoulasc:
Instead of changing mmap_flags which makes the change permanent. This was enforcing large alignments for allocations that did not need it causing fragmentation. Reported by Andreas Gustafsson.


interwq:
Thanks for the patch!

Issue #2358, opened 2022-11-30
Heap profiling uses 16x more memory on ARM64, than it does on x64


madscientist:
We have a program that attempts to keep memory usage to a value specified by the user: it will fill its cache until the memory usage is reached then free memory.  If the system cannot release enough memory for a certain amount of time (default 5 minutes) it will kill itself.  To determine the amount of memory in use, we first use mallctlbymib() with "epoch", then we use mallctlbymib() with "stats.allocated".   We are using JEmalloc 5.3 on Linux x64 and this works fine: the system will run in a stable state for a very long time.

We are now porting to Linux ARM64, and on ARM we find that jemalloc is not able to free enough memory to stay under the limit.  For example if we set the memory limit to 2G on a 16G system it runs forever using about 2G of RAM on x64, but on ARM after a while no matter how much cache we try to free, we can't get "stats.allocated" down and we kill ourselves.

Just to note, we compile jemalloc with our code and link it statically.  We invoke `configure` with `--without-export --disable-doc --enable-prof` on both platforms, but no other options.

I've attached two sets of jemalloc stats output: one for our x64 run and the other for our ARM run.  Since they work at slightly different speeds and the x64 test runs to completion while the ARM test fails partway through, they are not fully apples-to-apples.  But you can see that on ARM the "mapped" and "retained" values are significantly larger than on x64.

I do notice that the page size on x64 is 4k while the page size on ARM is 64k.  But, should that matter?  Doesn't that just mean ARM reserves fewer pages and breaks the 64k up into more blocks?

Anyone with any insight as to why this is happening, please pass along!  I'm not sure where to go from here.

[last-jemalloc-stats.x86.txt](https://github.com/jemalloc/jemalloc/files/10119349/last-jemalloc-stats.x86.txt)
[last-jemalloc-stats.arm.txt](https://github.com/jemalloc/jemalloc/files/10119350/last-jemalloc-stats.arm.txt)

madscientist:
We have done some investigation.  It seems that jemalloc is reserving a LOT more memory that doesn't appear in any profiling output, etc. on ARM, than it does on x64.  This memory is added to the stats.allocated but since it's maintained by jemalloc we can't affect it by releasing memory ourselves.

It looks like the culprit for reserving all this memory, is this function:
```
imalloc_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
    size_t usize, szind_t ind)
```
[...]

PR #2459, opened 2023-06-05
Reduce the memory overhead of sampled small allocations

Svetlitski:
Previously, small allocations which were sampled as part of heap profiling were rounded up to `SC_LARGE_MINCLASS`. This additional memory usage becomes problematic when the page size is increased, as noted in #2358.

Small allocations are now rounded up to the nearest multiple of `PAGE` instead, reducing the memory overhead by a factor of 4 in the most extreme cases.

### DIFF
--- a/include/jemalloc/internal/arena_inlines_b.h
+++ b/include/jemalloc/internal/arena_inlines_b.h
@@ -182,23 +182,22 @@ arena_decay_tick(tsdn_t *tsdn, arena_t *arena) {
 
 JEMALLOC_ALWAYS_INLINE void *
 arena_malloc(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind, bool zero,
-    tcache_t *tcache, bool slow_path) {
+    bool slab, tcache_t *tcache, bool slow_path) {
 	assert(!tsdn_null(tsdn) || tcache == NULL);
 
 	if (likely(tcache != NULL)) {
-		if (likely(size <= SC_SMALL_MAXCLASS)) {
+		if (likely(slab)) {
+			assert(sz_can_use_slab(size));
 			return tcache_alloc_small(tsdn_tsd(tsdn), arena,
 			    tcache, size, ind, zero, slow_path);
-		}
-		if (likely(size <= tcache_maxclass)) {
+		} else if (likely(size <= tcache_maxclass)) {
 			return tcache_alloc_large(tsdn_tsd(tsdn), arena,
 			    tcache, size, ind, zero, slow_path);
 		}
 		/* (size > tcache_maxclass) case falls through. */
-		assert(size > tcache_maxclass);
 	}
 
-	return arena_malloc_hard(tsdn, arena, size, ind, zero);
+	return arena_malloc_hard(tsdn, arena, size, ind, zero, slab);
 }
 
 JEMALLOC_ALWAYS_INLINE arena_t *
--- a/include/jemalloc/internal/jemalloc_internal_inlines_c.h
+++ b/include/jemalloc/internal/jemalloc_internal_inlines_c.h
@@ -52,10 +52,12 @@ isalloc(tsdn_t *tsdn, const void *ptr) {
 }
 
 JEMALLOC_ALWAYS_INLINE void *
-iallocztm(tsdn_t *tsdn, size_t size, szind_t ind, bool zero, tcache_t *tcache,
-    bool is_internal, arena_t *arena, bool slow_path) {
+iallocztm_explicit_slab(tsdn_t *tsdn, size_t size, szind_t ind, bool zero,
+    bool slab, tcache_t *tcache, bool is_internal, arena_t *arena,
+    bool slow_path) {
 	void *ret;
 
+	assert(!slab || sz_can_use_slab(size)); /* slab && large is illegal */
 	assert(!is_internal || tcache == NULL);
 	assert(!is_internal || arena == NULL || arena_is_auto(arena));
 	if (!tsdn_null(tsdn) && tsd_reentrancy_level_get(tsdn_tsd(tsdn)) == 0) {
@@ -63,11 +65,19 @@ iallocztm(tsdn_t *tsdn, size_t size, szind_t ind, bool zero, tcache_t *tcache,
 		    WITNESS_RANK_CORE, 0);
 	}
 
-	ret = arena_malloc(tsdn, arena, size, ind, zero, tcache, slow_path);
+	ret = arena_malloc(tsdn, arena, size, ind, zero, slab, tcache, slow_path);
 	if (config_stats && is_internal && likely(ret != NULL)) {
 		arena_internal_add(iaalloc(tsdn, ret), isalloc(tsdn, ret));
 	}
 	return ret;
+}
+
+JEMALLOC_ALWAYS_INLINE void *
+iallocztm(tsdn_t *tsdn, size_t size, szind_t ind, bool zero, tcache_t *tcache,
+    bool is_internal, arena_t *arena, bool slow_path) {
+	bool slab = sz_can_use_slab(size);
+	return iallocztm_explicit_slab(tsdn, size, ind, zero, slab, tcache,
+	    is_internal, arena, slow_path);
 }
 
 JEMALLOC_ALWAYS_INLINE void *
@@ -77,10 +87,11 @@ ialloc(tsd_t *tsd, size_t size, szind_t ind, bool zero, bool slow_path) {
 }
 
 JEMALLOC_ALWAYS_INLINE void *
-ipallocztm(tsdn_t *tsdn, size_t usize, size_t alignment, bool zero,
-    tcache_t *tcache, bool is_internal, arena_t *arena) {
+ipallocztm_explicit_slab(tsdn_t *tsdn, size_t usize, size_t alignment, bool zero,
+    bool slab, tcache_t *tcache, bool is_internal, arena_t *arena) {
 	void *ret;
 
+	assert(!slab || sz_can_use_slab(usize)); /* slab && large is illegal */
 	assert(usize != 0);
 	assert(usize == sz_sa2u(usize, alignment));
 	assert(!is_internal || tcache == NULL);
@@ -88,7 +99,7 @@ ipallocztm(tsdn_t *tsdn, size_t usize, size_t alignment, bool zero,
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, 0);
 
-	ret = arena_palloc(tsdn, arena, usize, alignment, zero, tcache);
+	ret = arena_palloc(tsdn, arena, usize, alignment, zero, slab, tcache);
 	assert(ALIGNMENT_ADDR2BASE(ret, alignment) == ret);
 	if (config_stats && is_internal && likely(ret != NULL)) {
 		arena_internal_add(iaalloc(tsdn, ret), isalloc(tsdn, ret));
@@ -97,9 +108,23 @@ ipallocztm(tsdn_t *tsdn, size_t usize, size_t alignment, bool zero,
 }
 
 JEMALLOC_ALWAYS_INLINE void *
+ipallocztm(tsdn_t *tsdn, size_t usize, size_t alignment, bool zero,
+    tcache_t *tcache, bool is_internal, arena_t *arena) {
+	return ipallocztm_explicit_slab(tsdn, usize, alignment, zero,
+	    sz_can_use_slab(usize), tcache, is_internal, arena);
+}
+
+JEMALLOC_ALWAYS_INLINE void *
 ipalloct(tsdn_t *tsdn, size_t usize, size_t alignment, bool zero,
     tcache_t *tcache, arena_t *arena) {
 	return ipallocztm(tsdn, usize, alignment, zero, tcache, false, arena);
+}
+
+JEMALLOC_ALWAYS_INLINE void *
+ipalloct_explicit_slab(tsdn_t *tsdn, size_t usize, size_t alignment,
+    bool zero, bool slab, tcache_t *tcache, arena_t *arena) {
+	return ipallocztm_explicit_slab(tsdn, usize, alignment, zero, slab,
+	    tcache, false, arena);
 }
 
 JEMALLOC_ALWAYS_INLINE void *
@@ -146,7 +171,7 @@ isdalloct(tsdn_t *tsdn, void *ptr, size_t size, tcache_t *tcache,
 
 JEMALLOC_ALWAYS_INLINE void *
 iralloct_realign(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size,
-    size_t alignment, bool zero, tcache_t *tcache, arena_t *arena,
+    size_t alignment, bool zero, bool slab, tcache_t *tcache, arena_t *arena,
     hook_ralloc_args_t *hook_args) {
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, 0);
@@ -157,7 +182,8 @@ iralloct_realign(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size,
 	if (unlikely(usize == 0 || usize > SC_LARGE_MAXCLASS)) {
 		return NULL;
 	}
-	p = ipalloct(tsdn, usize, alignment, zero, tcache, arena);
+	p = ipalloct_explicit_slab(tsdn, usize, alignment, zero, slab,
+	    tcache, arena);
 	if (p == NULL) {
 		return NULL;
 	}
@@ -184,8 +210,9 @@ iralloct_realign(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size,
  * passed-around anywhere.
  */
 JEMALLOC_ALWAYS_INLINE void *
-iralloct(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size, size_t alignment,
-    bool zero, tcache_t *tcache, arena_t *arena, hook_ralloc_args_t *hook_args)
+iralloct_explicit_slab(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size,
+    size_t alignment, bool zero, bool slab, tcache_t *tcache, arena_t *arena,
+    hook_ralloc_args_t *hook_args)
 {
 	assert(ptr != NULL);
 	assert(size != 0);
@@ -199,18 +226,28 @@ iralloct(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size, size_t alignment,
 		 * and copy.
 		 */
 		return iralloct_realign(tsdn, ptr, oldsize, size, alignment,
-		    zero, tcache, arena, hook_args);
+		    zero, slab, tcache, arena, hook_args);
 	}
 
 	return arena_ralloc(tsdn, arena, ptr, oldsize, size, alignment, zero,
-	    tcache, hook_args);
+	    slab, tcache, hook_args);
+}
+
+JEMALLOC_ALWAYS_INLINE void *
+iralloct(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size, size_t alignment,
+    size_t usize, bool zero, tcache_t *tcache, arena_t *arena,
+    hook_ralloc_args_t *hook_args)
+{
+	bool slab = sz_can_use_slab(usize);
+	return iralloct_explicit_slab(tsdn, ptr, oldsize, size, alignment, zero,
+	    slab, tcache, arena, hook_args);
 }
 
 JEMALLOC_ALWAYS_INLINE void *
 iralloc(tsd_t *tsd, void *ptr, size_t oldsize, size_t size, size_t alignment,
-    bool zero, hook_ralloc_args_t *hook_args) {
-	return iralloct(tsd_tsdn(tsd), ptr, oldsize, size, alignment, zero,
-	    tcache_get(tsd), NULL, hook_args);
+    size_t usize, bool zero, hook_ralloc_args_t *hook_args) {
+	return iralloct(tsd_tsdn(tsd), ptr, oldsize, size, alignment, usize,
+	    zero, tcache_get(tsd), NULL, hook_args);
 }
 
 JEMALLOC_ALWAYS_INLINE bool
@@ -405,7 +442,7 @@ maybe_check_alloc_ctx(tsd_t *tsd, void *ptr, emap_alloc_ctx_t *alloc_ctx) {
 
 JEMALLOC_ALWAYS_INLINE bool
 prof_sample_aligned(const void *ptr) {
-        return ((uintptr_t)ptr & PAGE_MASK) == 0;
+	return ((uintptr_t)ptr & PROF_SAMPLE_ALIGNMENT_MASK) == 0;
 }
 
 JEMALLOC_ALWAYS_INLINE bool
--- a/include/jemalloc/internal/prof_inlines.h
+++ b/include/jemalloc/internal/prof_inlines.h
@@ -239,14 +239,15 @@ prof_realloc(tsd_t *tsd, const void *ptr, size_t size, size_t usize,
 }
 
 JEMALLOC_ALWAYS_INLINE size_t
-prof_sample_align(size_t orig_align) {
+prof_sample_align(size_t usize, size_t orig_align) {
 	/*
-	 * Enforce page alignment, so that sampled allocations can be identified
+	 * Enforce alignment, so that sampled allocations can be identified
 	 * w/o metadata lookup.
 	 */
 	assert(opt_prof);
-	return (opt_cache_oblivious && orig_align < PAGE) ? PAGE :
-	    orig_align;
+	return (orig_align < PROF_SAMPLE_ALIGNMENT &&
+	       (sz_can_use_slab(usize) || opt_cache_oblivious)) ?
+	           PROF_SAMPLE_ALIGNMENT : orig_align;
 }
 
 JEMALLOC_ALWAYS_INLINE bool
--- a/include/jemalloc/internal/safety_check.h
+++ b/include/jemalloc/internal/safety_check.h
@@ -3,6 +3,8 @@
 
 #define SAFETY_CHECK_DOUBLE_FREE_MAX_SCAN_DEFAULT 32
 
+#include "jemalloc/internal/pages.h"
+
 void safety_check_fail_sized_dealloc(bool current_dealloc, const void *ptr,
     size_t true_size, size_t input_size);
 void safety_check_fail(const char *format, ...);
@@ -12,22 +14,50 @@ typedef void (*safety_check_abort_hook_t)(const char *message);
 /* Can set to NULL for a default. */
 void safety_check_set_abort(safety_check_abort_hook_t abort_fn);
 
+#define REDZONE_SIZE ((size_t) 32)
+#define REDZONE_FILL_VALUE 0xBC
+
+/*
+ * Normally the redzone extends `REDZONE_SIZE` bytes beyond the end of
+ * the allocation. However, we don't let the redzone extend onto another
+ * OS page because this would impose additional overhead if that page was
+ * not already resident in memory.
+ */
+JEMALLOC_ALWAYS_INLINE const unsigned char *
+compute_redzone_end(const void *_ptr, size_t usize, size_t bumped_usize) {
+	const unsigned char *ptr = (const unsigned char *) _ptr;
+	const unsigned char *redzone_end = usize + REDZONE_SIZE < bumped_usize ?
+	    &ptr[usize + REDZONE_SIZE] : &ptr[bumped_usize];
+	const unsigned char *page_end = (const unsigned char *)
+	    ALIGNMENT_CEILING(((uintptr_t) (&ptr[usize])), os_page);
+	return redzone_end < page_end ? redzone_end : page_end;
+}
+
 JEMALLOC_ALWAYS_INLINE void
 safety_check_set_redzone(void *ptr, size_t usize, size_t bumped_usize) {
-	assert(usize < bumped_usize);
-	for (size_t i = usize; i < bumped_usize && i < usize + 32; ++i) {
-		*((unsigned char *)ptr + i) = 0xBC;
+	assert(usize <= bumped_usize);
+	const unsigned char *redzone_end =
+		compute_redzone_end(ptr, usize, bumped_usize);
+	for (unsigned char *curr = &((unsigned char *)ptr)[usize];
+	     curr < redzone_end; curr++) {
+		*curr = REDZONE_FILL_VALUE;
 	}
 }
 
 JEMALLOC_ALWAYS_INLINE void
 safety_check_verify_redzone(const void *ptr, size_t usize, size_t bumped_usize)
 {
-	for (size_t i = usize; i < bumped_usize && i < usize + 32; ++i) {
-		if (unlikely(*((unsigned char *)ptr + i) != 0xBC)) {
+	const unsigned char *redzone_end =
+		compute_redzone_end(ptr, usize, bumped_usize);
+	for (const unsigned char *curr= &((const unsigned char *)ptr)[usize];
+	     curr < redzone_end; curr++) {
+		if (unlikely(*curr != REDZONE_FILL_VALUE)) {
 			safety_check_fail("Use after free error\n");
 		}
 	}
 }
+
+#undef REDZONE_SIZE
+#undef REDZONE_FILL_VALUE
 
 #endif /*JEMALLOC_INTERNAL_SAFETY_CHECK_H */
--- a/src/arena.c
+++ b/src/arena.c
@@ -1191,7 +1191,7 @@ arena_malloc_small(tsdn_t *tsdn, arena_t *arena, szind_t binind, bool zero) {
 
 void *
 arena_malloc_hard(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind,
-    bool zero) {
+    bool zero, bool slab) {
 	assert(!tsdn_null(tsdn) || arena != NULL);
 
 	if (likely(!tsdn_null(tsdn))) {
@@ -1201,18 +1201,19 @@ arena_malloc_hard(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind,
 		return NULL;
 	}
 
-	if (likely(size <= SC_SMALL_MAXCLASS)) {
+	if (likely(slab)) {
+		assert(sz_can_use_slab(size));
 		return arena_malloc_small(tsdn, arena, ind, zero);
+	} else {
+		return large_malloc(tsdn, arena, sz_index2size(ind), zero);
 	}
-	return large_malloc(tsdn, arena, sz_index2size(ind), zero);
 }
 
 void *
 arena_palloc(tsdn_t *tsdn, arena_t *arena, size_t usize, size_t alignment,
-    bool zero, tcache_t *tcache) {
-	void *ret;
-
-	if (usize <= SC_SMALL_MAXCLASS) {
+    bool zero, bool slab, tcache_t *tcache) {
+	if (slab) {
+		assert(sz_can_use_slab(usize));
 		/* Small; alignment doesn't require special slab placement. */
 
 		/* usize should be a result of sz_sa2u() */
@@ -1223,27 +1224,26 @@ arena_palloc(tsdn_t *tsdn, arena_t *arena, size_t usize, size_t alignment,
 		 */
 		assert(alignment <= PAGE);
 
-		ret = arena_malloc(tsdn, arena, usize, sz_size2index(usize),
-		    zero, tcache, true);
+		return arena_malloc(tsdn, arena, usize, sz_size2index(usize),
+		    zero, slab, tcache, true);
 	} else {
 		if (likely(alignment <= CACHELINE)) {
-			ret = large_malloc(tsdn, arena, usize, zero);
+			return large_malloc(tsdn, arena, usize, zero);
 		} else {
-			ret = large_palloc(tsdn, arena, usize, alignment, zero);
+			return large_palloc(tsdn, arena, usize, alignment, zero);
 		}
 	}
-	return ret;
 }
 
 void
-arena_prof_promote(tsdn_t *tsdn, void *ptr, size_t usize) {
+arena_prof_promote(tsdn_t *tsdn, void *ptr, size_t usize, size_t bumped_usize) {
 	cassert(config_prof);
 	assert(ptr != NULL);
-	assert(isalloc(tsdn, ptr) == SC_LARGE_MINCLASS);
-	assert(usize <= SC_SMALL_MAXCLASS);
+	assert(isalloc(tsdn, ptr) == bumped_usize);
+	assert(sz_can_use_slab(usize));
 
 	if (config_opt_safety_checks) {
-		safety_check_set_redzone(ptr, usize, SC_LARGE_MINCLASS);
+		safety_check_set_redzone(ptr, usize, bumped_usize);
 	}
 
 	edata_t *edata = emap_edata_lookup(tsdn, &arena_emap_global, ptr);
@@ -1259,13 +1259,19 @@ static size_t
 arena_prof_demote(tsdn_t *tsdn, edata_t *edata, const void *ptr) {
 	cassert(config_prof);
 	assert(ptr != NULL);
+	size_t usize = isalloc(tsdn, ptr);
+	size_t bumped_usize = sz_sa2u(usize, PROF_SAMPLE_ALIGNMENT);
+	assert(bumped_usize <= SC_LARGE_MINCLASS &&
+	    PAGE_CEILING(bumped_usize) == bumped_usize);
+	assert(edata_size_get(edata) - bumped_usize <= sz_large_pad);
+	szind_t szind = sz_size2index(bumped_usize);
 
-	edata_szind_set(edata, SC_NBINS);
-	emap_remap(tsdn, &arena_emap_global, edata, SC_NBINS, /* slab */ false);
+	edata_szind_set(edata, szind);
+	emap_remap(tsdn, &arena_emap_global, edata, szind, /* slab */ false);
 
-	assert(isalloc(tsdn, ptr) == SC_LARGE_MINCLASS);
+	assert(isalloc(tsdn, ptr) == bumped_usize);
 
-	return SC_LARGE_MINCLASS;
+	return bumped_usize;
 }
 
 void
@@ -1282,10 +1288,10 @@ arena_dalloc_promoted(tsdn_t *tsdn, void *ptr, tcache_t *tcache,
 		 * Currently, we only do redzoning for small sampled
 		 * allocations.
 		 */
-		assert(bumped_usize == SC_LARGE_MINCLASS);
 		safety_check_verify_redzone(ptr, usize, bumped_usize);
 	}
-	if (bumped_usize <= tcache_maxclass && tcache != NULL) {
+	if (bumped_usize >= SC_LARGE_MINCLASS &&
+	    bumped_usize <= tcache_maxclass && tcache != NULL) {
 		tcache_dalloc_large(tsdn_tsd(tsdn), tcache, ptr,
 		    sz_size2index(bumped_usize), slow_path);
 	} else {
@@ -1443,28 +1449,30 @@ done:
 
 static void *
 arena_ralloc_move_helper(tsdn_t *tsdn, arena_t *arena, size_t usize,
-    size_t alignment, bool zero, tcache_t *tcache) {
+    size_t alignment, bool zero, bool slab, tcache_t *tcache) {
 	if (alignment == 0) {
 		return arena_malloc(tsdn, arena, usize, sz_size2index(usize),
-		    zero, tcache, true);
+		    zero, slab, tcache, true);
 	}
 	usize = sz_sa2u(usize, alignment);
 	if (unlikely(usize == 0 || usize > SC_LARGE_MAXCLASS)) {
 		return NULL;
 	}
-	return ipalloct(tsdn, usize, alignment, zero, tcache, arena);
+	return ipalloct_explicit_slab(tsdn, usize, alignment, zero, slab,
+	    tcache, arena);
 }
 
 void *
 arena_ralloc(tsdn_t *tsdn, arena_t *arena, void *ptr, size_t oldsize,
-    size_t size, size_t alignment, bool zero, tcache_t *tcache,
+    size_t size, size_t alignment, bool zero, bool slab, tcache_t *tcache,
     hook_ralloc_args_t *hook_args) {
 	size_t usize = alignment == 0 ? sz_s2u(size) : sz_sa2u(size, alignment);
 	if (unlikely(usize == 0 || size > SC_LARGE_MAXCLASS)) {
 		return NULL;
 	}
 
-	if (likely(usize <= SC_SMALL_MAXCLASS)) {
+	if (likely(slab)) {
+		assert(sz_can_use_slab(usize));
 		/* Try to avoid moving the allocation. */
 		UNUSED size_t newsize;
 		if (!arena_ralloc_no_move(tsdn, ptr, oldsize, usize, 0, zero,
@@ -1488,7 +1496,7 @@ arena_ralloc(tsdn_t *tsdn, arena_t *arena, void *ptr, size_t oldsize,
 	 * object.  In that case, fall back to allocating new space and copying.
 	 */
 	void *ret = arena_ralloc_move_helper(tsdn, arena, usize, alignment,
-	    zero, tcache);
+	    zero, slab, tcache);
 	if (ret == NULL) {
 		return NULL;
 	}
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2360,7 +2360,7 @@ arena_get_from_ind(tsd_t *tsd, unsigned arena_ind, arena_t **arena_p) {
 /* ind is ignored if dopts->alignment > 0. */
 JEMALLOC_ALWAYS_INLINE void *
 imalloc_no_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
-    size_t size, size_t usize, szind_t ind) {
+    size_t size, size_t usize, szind_t ind, bool slab) {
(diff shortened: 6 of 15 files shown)


src/jemalloc.c
```diff
@@ -2384,26 +2384,19 @@ imalloc_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
...
 	if (usize <= SC_SMALL_MAXCLASS) {
-		assert(((dopts->alignment == 0) ?
-		    sz_s2u(SC_LARGE_MINCLASS) :
-		    sz_sa2u(SC_LARGE_MINCLASS, dopts->alignment))
-			== SC_LARGE_MINCLASS);
-		ind_large = sz_size2index(SC_LARGE_MINCLASS);
-		size_t bumped_usize = sz_s2u(SC_LARGE_MINCLASS);
-		ret = imalloc_no_sample(sopts, dopts, tsd, bumped_usize,
-		    bumped_usize, ind_large);
+		size_t bumped_usize = PAGE_CEILING(usize);
```
Svetlitski: High-level question that applies to most of these function calls: What is the convention in Jemalloc for when and where you should apply alignment to a size? I have *not* applied it here (e.g. with `sz_sa2u` or similar) since the `arena_large_palloc` accepts the alignment as an argument. Is this correct?

src/arena.c
```diff
@@ -1259,13 +1262,21 @@ static size_t
 arena_prof_demote(tsdn_t *tsdn, edata_t *edata, const void *ptr) {
 	cassert(config_prof);
 	assert(ptr != NULL);
+	size_t usize = isalloc(tsdn, ptr);
+	size_t bumped_usize = PAGE_CEILING(usize);
+	/*
+	 * TODO(Svetlitski) This doesn't properly handle the case
+	 * where the allocation was explicitly over-aligned.
+	 */
```
Svetlitski: With regards to this comment, I'm not really sure there's a way to solve this. If the allocation was over-aligned to an alignment that is greater than `PAGE` (potentially causing the correct value of `bumped_usize` to actually be greater than`PAGE_CEILING(usize)`), I don't see how we can possibly recover that information.

src/arena.c
```diff
@@ -1282,15 +1293,9 @@ arena_dalloc_promoted(tsdn_t *tsdn, void *ptr, tcache_t *tcache,
...
-		assert(bumped_usize == SC_LARGE_MINCLASS);
 		safety_check_verify_redzone(ptr, usize, bumped_usize);
 	}
-	if (bumped_usize <= tcache_maxclass && tcache != NULL) {
-		tcache_dalloc_large(tsdn_tsd(tsdn), tcache, ptr,
-		    sz_size2index(bumped_usize), slow_path);
-	} else {
-		large_dalloc(tsdn, edata);
-	}
+	large_dalloc(tsdn, edata);
```
Svetlitski: I was running into issues when allowing the thread cache to keep these oddly sized edatas around, since they don't necessarily adhere to the invariants maintained elsewhere. Is it worth complicating the tcache logic to handle this, instead of just not caching `edatas` from sampled allocations?

src/jemalloc.c
interwq: For the "normal" / non-sampled case, whenever there's a `usize` (usable size), it takes the alignment into the size class consideration already (in case the alignment would cause the size class to change, e.g. alloc 5k but asking for 4k aligned, resulting the 8k size class being used). So the convention is translating size to usize once in `imalloc_body`, and then pass the alignment argument downstream.

Here it's a special case, because we are adding another alignment and size bumping (on top of the already computed usize and alignment). The PAGE_CEILING is very close to sa2u, except the case where alignment is 8K (asked by user input), where sa2u(9k, 8k) == 16K. So we would want to change page ceiling to `sa2u(usize, dopts->alignment)` instead.

src/arena.c
Svetlitski: I suppose we could just skip sampling allocations with a requested alignment of greater than `PAGE` in the first place.

src/jemalloc.c
```diff
@@ -2384,26 +2384,19 @@ imalloc_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
...
+		arena_t *arena;
+		if (arena_get_from_ind(tsd, dopts->arena_ind, &arena)) {
+			return NULL;
+		}
+		ret = arena_large_palloc(tsd_tsdn(tsd), arena, bumped_usize, dopts->alignment,
```
interwq: We should still go through `imalloc_no_sample` in order to handle all the requirements, e.g. if there's a specific tcache id asked. Also it's helpful to keep the high level flow same for small and large sizes, with only small being bumped up a bit. If we update the usize using sa2u above, imalloc_no_sample should work correctly?

src/jemalloc.c
```diff
@@ -3315,13 +3308,13 @@ irallocx_prof_sample(tsdn_t *tsdn, void *old_ptr, size_t old_usize,
 
 	alignment = prof_sample_align(alignment);
 	if (usize <= SC_SMALL_MAXCLASS) {
-		p = iralloct(tsdn, old_ptr, old_usize,
-		    SC_LARGE_MINCLASS, alignment, zero, tcache,
-		    arena, hook_args);
+		size_t bumped_usize = PAGE_CEILING(usize);
+		p = arena_prof_ralloc(tsdn, arena, old_ptr, old_usize, bumped_usize,
```
interwq: Similarly, bumped_usize should be computed with sa2u. With that change, which part of the `iralloct` isn't working?

src/jemalloc.c
```diff
@@ -2384,26 +2384,19 @@ imalloc_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
     size_t usize, szind_t ind) {
 	void *ret;
 
-	/*
-	 * For small allocations, sampling bumps the usize.  If so, we allocate
-	 * from the ind_large bucket.
-	 */
-	szind_t ind_large;
-
 	dopts->alignment = prof_sample_align(dopts->alignment);
```
interwq: I forgot to mention -- `prof_sample_align` needs to be updated as well, since it assumed using large size classes (which are all page aligned, outside of the `cache_oblivious` / addr randomization case). This is also why the bumped_size was using s2u instead of sa2u (which would make no difference). Now we should be enforcing that explicitly.

src/arena.c
interwq: Probably fine to not cache these. Question, we are going through the large pipeline, because these edatas are sampled and not slabs, with the only difference from large is their size is not a large size class. Is that right? I just realized that some of my previous comments were probably wrong and missed this part.

src/arena.c
interwq: > if the allocation was over-aligned to an alignment that is greater than PAGE

This should result the usize being a large size, if we use sa2u to set the bumped_usize, i.e. alignment >4k cannot be a small size class. Does that solve this problem?

Not sampling these over aligned allocations isn't really a good option -- imagine if there's a problematic function doing aligned malloc but with an unnecessarily large alignment, and it's cause high memory usage, then heap profiles wouldn't cover that part.

The only case we current would skip sampling or other periodic events, is when we are in a reentrancy situation, i.e. malloc => some user hook => malloc, then all the non essential features are bypassed in that nested malloc. This is rather a last resort to avoid circular dependencies (e.g. the first malloc might be sampled and is invoking the one of the sampling hooks).

src/jemalloc.c
interwq: I just realized that `imalloc_no_sample` cannot be used because it may go through the small size class route and allocate part of a slab to us. So it appears we would have to short cut a bit here. Let me look more into it.

src/arena.c
Svetlitski: > We are going through the large pipeline, because these edatas are sampled and not slabs, with the only difference from large is their size is not a large size class. Is that right?

Yes, this was precisely my intention, so that we don't have to break the previous invariants around size-classes throughout the rest of the codebase, and instead isolate the special-casing to the small number of sites modified in this PR.

src/arena.c
Svetlitski: That makes sense. Before one of your previous comments I hadn't realized `usize` had already been adjusted to reflect the user-requested alignment requirement by the time we get to `imalloc_sample` and the like.

src/arena.c
```diff
@@ -1223,27 +1231,22 @@ arena_palloc(tsdn_t *tsdn, arena_t *arena, size_t usize, size_t alignment,
...
+arena_prof_promote(tsdn_t *tsdn, void *ptr, size_t usize, size_t bumped_usize) {
 	cassert(config_prof);
 	assert(ptr != NULL);
-	assert(isalloc(tsdn, ptr) == SC_LARGE_MINCLASS);
+	assert(isalloc(tsdn, ptr) == bumped_usize);
 	assert(usize <= SC_SMALL_MAXCLASS);
 
 	if (config_opt_safety_checks) {
-		safety_check_set_redzone(ptr, usize, SC_LARGE_MINCLASS);
+		safety_check_set_redzone(ptr, usize, bumped_usize);
```
interwq, line 1246: One thing to consider and validate, is if the redzone here adds extra RSS, when the jemalloc page size is greater than system page size, given that it writes a bit beyond the usize range which can touch another page.

Another thing we can do is, having the set_ and check_redzone always stops at a 4K boundary.

src/jemalloc.c
Svetlitski: Similar to the reason for not going through `imalloc_no_sample`, if we call `iralloct` the allocation is treated as small.

src/arena.c
Svetlitski: I've modified it to stop at the 4 KiB boundary if that is before what would otherwise be the end of the redzone.

interwq:
After more debating, here's what I'm thinking now (if there's any conflict between this and my previous comments, this one wins):

For large sampled alloc, let's try making sure their handling remain unchanged. For example, what I mentioned about updating `prof_sample_align` to be at least PAGE, that change should be only towards small sizes, because the currently large sampled alloc with no user asked alignment can be handled out of tcache, i.e. this path:
https://github.com/jemalloc/jemalloc/blob/5832ef658975d5f2da2bdfddf55712d9fa343e30/src/jemalloc.c#L2378

and if we always apply a PAGE alignment, the above line will switch to `ipalloct` branch before it which has no tcache for large sizes.

Maybe something like this:
```
bumped_usize = sz_sa2u(existing_usize, PAGE);
if (bumped_usize <= SC_SMALL_MAXCLASS) {
    assert(dopts->alignment <= PAGE);
    dopts->alignment = PAGE;
    // new path handling small bumped_usize
   ......
} else {
    // don't change dopts->alignement; go through existing large path
}
```

I'm torn about diverging from the central allocation entry points, but it looks the current interface needs to be expanded otherwise. Maybe that should be considered as well, e.g. adding another flag for non-slabs explicitly (instead of using usize to indicate) in `ipalloct`? WDYT @Svetlitski ?

Svetlitski:
I definitely agree with avoiding any modifications to the existing large-allocation path that aren't strictly necessary. And I believe the changes in my latest force-push capture your intent with regards to `prof_sample_align` only over-aligning small allocations (except for when cache-oblivious is in effect).

With regards to diverging from the central allocation entry points, I agree the approach I've taken here is not the cleanest ideologically. I'm also hesitant to add more arguments to functions like `iralloct` that already accept 9 of them. I'll defer to your expertise, I'm happy to implement either approach.

interwq:
@Svetlitski Okay let's try the additional argument approach; in the long term it will help detect and avoid issues, with everything going through the same paths and assertions.

Agreed that it's a long list already; to minimize changes, maybe only add the new argument when required.

Svetlitski:
If you're wondering about the placement of the additional `slab` argument, I went with the convention of always placing it immediately after `zero` for consistency.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -19,6 +19,21 @@
...
+ * to satisfy an allocation depends solely on the allocation's
+ * effective size. However, this is *not* the case when an allocation
+ * is sampled for profiling, in which case you *must not* use a slab
+ * regardless of the effective size. Thus `should_use_slab` is called
+ * on the common path, but there exist `*_explicit_slab` variants of
+ * several functions below for handling the aforementioned case of
+ * sampled allocations.
+ */
+JEMALLOC_ALWAYS_INLINE bool
+should_use_slab(size_t usize) {
```
interwq: nit: `sz_should_use_slab`, also let's use `size` instead of `usize` -- because if size <= SMALL_MAXCLASS, then usize <= SMALL_MAXCLASS. Then we don't have to call s2u when using this.

src/arena.c
```diff
@@ -1201,18 +1201,16 @@ arena_malloc_hard(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind,
 		return NULL;
 	}
 
-	if (likely(size <= SC_SMALL_MAXCLASS)) {
+	if (likely(size <= SC_SMALL_MAXCLASS && slab)) {
```
interwq: ```if (likely(slab)) {``` should be sufficient? Then in the branch `assert(size <= SC_SMALL_MAXCLASS)`.

src/arena.c
```diff
@@ -1201,18 +1201,16 @@ arena_malloc_hard(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind,
...
 }
 
 void *
 arena_palloc(tsdn_t *tsdn, arena_t *arena, size_t usize, size_t alignment,
-    bool zero, tcache_t *tcache) {
-	void *ret;
-
-	if (usize <= SC_SMALL_MAXCLASS) {
+    bool zero, bool slab, tcache_t *tcache) {
+	if (usize <= SC_SMALL_MAXCLASS && slab) {
```
interwq: Similarly, `if (slab)` and then assert.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -52,8 +67,8 @@ isalloc(tsdn_t *tsdn, const void *ptr) {
 }
 
 JEMALLOC_ALWAYS_INLINE void *
-iallocztm(tsdn_t *tsdn, size_t size, szind_t ind, bool zero, tcache_t *tcache,
-    bool is_internal, arena_t *arena, bool slow_path) {
+iallocztm_explicit_slab(tsdn_t *tsdn, size_t size, szind_t ind, bool zero, bool slab,
+    tcache_t *tcache, bool is_internal, arena_t *arena, bool slow_path) {
```
interwq: Add `assert(!slab || sz_should_use_slab(size)); /* slab && large is illegal */`

WDYT about `sz_can_use_slab`, instead of `_should_`? That fits the assertion better, i.e. it would read like:
`if (slab) assert(sz_can_use_slab)`

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -148,23 +148,23 @@ arena_decay_tick(tsdn_t *tsdn, arena_t *arena) {
 
 JEMALLOC_ALWAYS_INLINE void *
 arena_malloc(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind, bool zero,
-    tcache_t *tcache, bool slow_path) {
+    bool slab, tcache_t *tcache, bool slow_path) {
 	assert(!tsdn_null(tsdn) || tcache == NULL);
 
 	if (likely(tcache != NULL)) {
-		if (likely(size <= SC_SMALL_MAXCLASS)) {
+		if (likely(size <= SC_SMALL_MAXCLASS && slab)) {
```
interwq: I'm hoping that compiler can avoid the extra branch here, for the common case after inlining? i.e. when not sampled, it would realize that the two conditions are identical?

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -148,23 +148,23 @@ arena_decay_tick(tsdn_t *tsdn, arena_t *arena) {
...
 			return tcache_alloc_small(tsdn_tsd(tsdn), arena,
 			    tcache, size, ind, zero, slow_path);
 		}
-		if (likely(size <= tcache_maxclass)) {
+		if (likely(size <= tcache_maxclass && !slab)) {
```
interwq: This looks wrong? When sampled, we may ask for size == 4K && !slab, which will be served using tcache_alloc_large incorrectly?

Svetlitski: Good catch, thanks.

include/jemalloc/internal/arena_inlines_b.h
Svetlitski: I compiled locally with `-O3` and ThinLTO, and yes it appears it the compiler will reduce this down to one branch when it can prove this is safe, like when we reach here from a call to `iallocztm`:


```asm
<jemalloc_je_iallocztm>:
;     bool is_internal, arena_t *arena, bool slow_path) {
               	push	rbp
               	push	r15
               	push	r14
               	push	r13
               	push	r12
               	push	rbx
               	sub	rsp, 0x1a8
               	mov	ebp, edx
               	mov	r13, rsi
               	mov	qword ptr [rsp + 0x10], rdi
               	mov	rsi, qword ptr [rsp + 0x1e0]
; 	if (likely(tcache != NULL)) {
               	test	r8, r8
               	je	 <L0>
               	mov	r12, r8
; 	return size <= SC_SMALL_MAXCLASS;
               	cmp	r13, 0x3800
; 		if (likely(size <= SC_SMALL_MAXCLASS && slab)) {
               	ja	 <L1>
; 	cache_bin_t *bin = &tcache->bins[binind];
               	mov	r15d, ebp
               	lea	r14, [r15 + 2*r15]
               	lea	rbx, [r12 + 8*r14]
               	add	rbx, 0x8
```

include/jemalloc/internal/jemalloc_internal_inlines_c.h
Svetlitski: We can't always do this it turns out. Or at least, in `iralloct` we should be calling:
```
bool slab = sz_can_use_slab(sz_sa2u(size, alignment))
```

Because `iralloct_realign` applies `sz_sa2u` to `usize` before passing it down, which causes the ` assert(!slab || sz_can_use_slab(usize)); /* slab && large is illegal */` to fail in `ipallocztm_explicit_slab`.

interwq: I see. In that case, let's stick to `usize`, and make sure that we always pass in an usize into it. 

[edit] Worth adding an assert(usize == sz_s2u(usize)) to verify.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -52,56 +68,81 @@ isalloc(tsdn_t *tsdn, const void *ptr) {
...
 	if (config_stats && is_internal && likely(ret != NULL)) {
 		arena_internal_add(iaalloc(tsdn, ret), isalloc(tsdn, ret));
 	}
 	return ret;
 }
 
+JEMALLOC_ALWAYS_INLINE void *
+iallocztm(tsdn_t *tsdn, size_t size, szind_t ind, bool zero, tcache_t *tcache,
+    bool is_internal, arena_t *arena, bool slow_path) {
+	bool slab = sz_can_use_slab(sz_s2u(size));
```
interwq: Can we avoid the `sz_s2u` here? If we don't that, which paths are failing the assert inside `sz_can_use_slab`?

Svetlitski: At least in the test-cases, we fail [coming from `prof_tdata_init_impl`](https://github.com/jemalloc/jemalloc/blob/90176f8a87a0b5bdb0ac4c1a515b1d9c58dc5a82/src/prof_data.c#L1137) in `test/integration/aligned_alloc` and we also fail in `a0ialloc` in `test_a0`. There might be more beyond this that we don't know about.

interwq: I'm a bit worried the extra s2u here is going to result in added instructions or less optimized code. Try adding the s2u to these call sites? If it's only a few of those, I'd say do that to make sure no regression on this path.

interwq: I just realized, this is the non-alignment case. Let's simply remove the s2u? We'd lose the assertion inside sz_can_use_slab, but that's no different from the current logic, i.e. using a non-usize to compare against SMALL_MAXCLASS.

Svetlitski: Good point. I've removed the call here and the assertion.

include/jemalloc/internal/safety_check.h
```diff
@@ -12,19 +12,40 @@ typedef void (*safety_check_abort_hook_t)(const char *message);
 /* Can set to NULL for a default. */
 void safety_check_set_abort(safety_check_abort_hook_t abort_fn);
 
+/* Return the smallest 4 KiB pagesize multiple that is >= s. */
+#define PAGE_4K_CEILING(s) (((s) + (4095)) & ~(4095))
```
interwq: nit: maybe `define PAGE_4K_CEILING(s) ALIGNMENT_CEILING(s, 4096)`

src/arena.c
```diff
@@ -1259,13 +1258,16 @@ static size_t
 arena_prof_demote(tsdn_t *tsdn, edata_t *edata, const void *ptr) {
 	cassert(config_prof);
 	assert(ptr != NULL);
+	size_t usize = isalloc(tsdn, ptr);
+	size_t bumped_usize = sz_sa2u(usize, PAGE);
```
interwq: nit: Let's do a separate "define PROF_SAMPLE_ALIGNMENT PAGE", and use it here and in prof_sample_align?

src/jemalloc.c
```diff
@@ -2371,41 +2371,31 @@ imalloc_no_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
...
+	dopts->alignment = prof_sample_align(usize, dopts->alignment);
 	if (usize <= SC_SMALL_MAXCLASS) {
-		assert(((dopts->alignment == 0) ?
-		    sz_s2u(SC_LARGE_MINCLASS) :
-		    sz_sa2u(SC_LARGE_MINCLASS, dopts->alignment))
-			== SC_LARGE_MINCLASS);
-		ind_large = sz_size2index(SC_LARGE_MINCLASS);
-		size_t bumped_usize = sz_s2u(SC_LARGE_MINCLASS);
+		size_t bumped_usize = sz_sa2u(usize, dopts->alignment);
+		szind_t bumped_ind = sz_size2index(bumped_usize);
```
interwq, line 2397: For the bumped path, it cannot be served out of tcache, right? If so, let's do so explicitly by adding `dopts->tcache_ind == TCACHE_IND_NONE`, which can probably simplify some downstream logic.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -19,6 +19,21 @@
...
+ * effective size. However, this is *not* the case when an allocation
+ * is sampled for profiling, in which case you *must not* use a slab
+ * regardless of the effective size. Thus `sz_can_use_slab` is called
+ * on the common path, but there exist `*_explicit_slab` variants of
+ * several functions below for handling the aforementioned case of
+ * sampled allocations.
+ */
+JEMALLOC_ALWAYS_INLINE bool
+sz_can_use_slab(size_t usize) {
+	return usize <= SC_SMALL_MAXCLASS;
```
interwq: nit: size instead of usize, since we don't guarantee it anymore.

include/jemalloc/internal/arena_inlines_b.h
```diff
@@ -181,11 +181,11 @@ arena_decay_tick(tsdn_t *tsdn, arena_t *arena) {
 
 JEMALLOC_ALWAYS_INLINE void *
 arena_malloc(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind, bool zero,
-    tcache_t *tcache, bool slow_path) {
+    bool slab, tcache_t *tcache, bool slow_path) {
 	assert(!tsdn_null(tsdn) || tcache == NULL);
 
 	if (likely(tcache != NULL)) {
-		if (likely(size <= SC_SMALL_MAXCLASS)) {
+		if (likely(size <= SC_SMALL_MAXCLASS && slab)) {
```
interwq: After we add the `TCACHE_IND_NONE` change below, technically we don't have to change this path. But let's still switch to `if (slab) { assert(size<=SMALL_MAX); ...` to use that as a consistent way of deciding between _small_alloc / _large_alloc. Also add an assert(!slab) before calling tcache_alloc_large.

For the context, the current logic as-is still routes the (size 4K + !slab) sampled case to `tcache_alloc_large` below incorrectly. This made me think about disallowing tcache explicitly.

src/arena.c
```diff
@@ -1201,18 +1201,18 @@ arena_malloc_hard(tsdn_t *tsdn, arena_t *arena, size_t size, szind_t ind,
 		return NULL;
 	}
 
-	if (likely(size <= SC_SMALL_MAXCLASS)) {
+	if (likely(slab)) {
+		assert(size <= SC_SMALL_MAXCLASS);
 		return arena_malloc_small(tsdn, arena, ind, zero);
 	}
 	return large_malloc(tsdn, arena, sz_index2size(ind), zero);
```
interwq: Add `assert(!slab || PAGE_CEILING(size) == size)` before calling large_malloc;

src/arena.c
```diff
@@ -1259,13 +1258,16 @@ static size_t
...
 	cassert(config_prof);
 	assert(ptr != NULL);
+	size_t usize = isalloc(tsdn, ptr);
+	size_t bumped_usize = sz_sa2u(usize, PAGE);
+	szind_t szind = sz_size2index(bumped_usize);
 
-	edata_szind_set(edata, SC_NBINS);
-	emap_remap(tsdn, &arena_emap_global, edata, SC_NBINS, /* slab */ false);
+	edata_szind_set(edata, szind);
+	emap_remap(tsdn, &arena_emap_global, edata, szind, /* slab */ true);
```
interwq: We should still do `slab = false`, right?

src/arena.c
```diff
@@ -1282,15 +1284,9 @@ arena_dalloc_promoted(tsdn_t *tsdn, void *ptr, tcache_t *tcache,
...
-		assert(bumped_usize == SC_LARGE_MINCLASS);
 		safety_check_verify_redzone(ptr, usize, bumped_usize);
 	}
-	if (bumped_usize <= tcache_maxclass && tcache != NULL) {
-		tcache_dalloc_large(tsdn_tsd(tsdn), tcache, ptr,
-		    sz_size2index(bumped_usize), slow_path);
-	} else {
-		large_dalloc(tsdn, edata);
-	}
+	large_dalloc(tsdn, edata);
```
interwq: If SC_LARGE_MINCLASS <= bumped_usize <= tcache_maxclass , we can still do tcache_dalloc_large

interwq (commented):
We should add unit testing for the small sizes, given that the behavior changed quite a bit. Either in one of the existing ones (test/unit/prof_.*), or even add a new one (prof_small?).

include/jemalloc/internal/safety_check.h
Svetlitski: Come to think of it, this should probably be `ALIGNMENT_CEILING(s, os_page)` instead of hard-coding it to 4 KiB.

src/arena.c
Svetlitski: I can just put `large_malloc` in the `else` of that `if` then if you feel that's clearer: an `assert(!slab ...)` is tautological, since you would've returned on line 1206 otherwise.

include/jemalloc/internal/arena_inlines_b.h
Svetlitski: Similarly here, if we change this to `if (slab) ...`, we may as well just make the other branch an `else if` instead of an `if`, assuming your intention here is clarity. This is instead of adding `assert(!slab)` before `tcache_alloc_large`, which will always be true since you would've returned by calling `tcache_alloc_small` otherwise.

src/arena.c
interwq: Put the large part in the else branch makes sense.

include/jemalloc/internal/arena_inlines_b.h
interwq: Agreed. The "else if" sounds good.

src/arena.c
Svetlitski: Now that you mention it, I don't think either is correct. What if `slab_sizes` was customized in the `MALLOC_CONF`? Say that you have `PAGE` set to 32 KiB, and `MALLOC_CONF=slab_sizes:32768-32768:2`. Consider for example when `usize` is 32 KiB. Setting `slab` to `false` is wrong, since all large size classes are >= 4x `PAGE`. Setting `slab` to `true` in the way that I'm currently doing it would _also_ be wrong, since the `MALLOC_CONF` specifies 2-page slabs for the 32 KiB size-class.

interwq: These edatas are indeed odd; their size is the small class, but not slabs. Setting it to non slab probably is "more" correct -- really the key here is, does it go through the arena bins or any slab "interior" management. In a way small vs large, is more about slab vs non-slab. As you have noticed, the large_alloc handles the <4x PAGE size fine, as long as it's a PAGE size. Setting it a non-slab to make sure it goes back to the large_* functions probably works fine, while setting to slab, as you mentioned, will create a different sized slab that will cause conflicts.

Svetlitski:
> We should add unit testing for the small sizes, given that the behavior changed quite a bit. Either in one of the existing ones (test/unit/prof_.*), or even add a new one (prof_small?).

Latest push addressed your comments, but I still need to write test cases.

interwq:
@Svetlitski for tests, if you haven't, try setting the sample interval like this, so that all allocations are sampled: https://github.com/jemalloc/jemalloc/blob/dev/test/unit/prof_mdump.sh#LL4C33-L4C49

src/arena.c
```diff
@@ -1259,13 +1259,16 @@ static size_t
 arena_prof_demote(tsdn_t *tsdn, edata_t *edata, const void *ptr) {
 	cassert(config_prof);
 	assert(ptr != NULL);
+	size_t usize = isalloc(tsdn, ptr);
+	size_t bumped_usize = sz_sa2u(usize, PROF_SAMPLE_ALIGNMENT);
+	szind_t szind = sz_size2index(bumped_usize);
```
interwq, line 1267: nit: assert(bumped_usize <= SC_LARGE_MINCLASS && PAGE_CEILING(bumped_usize) == bumped_usize)

test/unit/prof_small.c
```diff
@@ -0,0 +1,30 @@
+#include "test/jemalloc_test.h"
+
+TEST_BEGIN(test_profile_small_allocations) {
+	test_skip_if(!config_prof);
+	test_skip_if(!config_debug);
```
interwq: Why skipping the non debug build?

test/unit/prof_small.c
```diff
@@ -0,0 +1,30 @@
...
+		void *ptr = malloc(size);
+		assert_ptr_not_null(ptr, "Unexpected malloc failure");
+		edata_t *edata = emap_edata_lookup(TSDN_NULL,
+		    &arena_emap_global, ptr);
+		assert_ptr_not_null(edata,
+		    "Unable to find edata for allocation");
+		expect_false(edata_slab_get(edata),
+		    "Sampled small allocations should not be placed on slabs");
+		expect_ptr_eq(ptr, PAGE_ADDR2BASE(ptr),
+		    "Sampled allocations should be page-aligned");
```
interwq: Let's add a few more asserts:
- edata_base_get(edata) == ptr (can probably replace the one above)
- edata_usize_get(edata) == size
- edata_size_get == PAGE_CEILING(size)
- edata_prof_tctx_get(edata)->tdata != NULL

src/arena.c
```diff
@@ -1259,13 +1259,16 @@ static size_t
 arena_prof_demote(tsdn_t *tsdn, edata_t *edata, const void *ptr) {
 	cassert(config_prof);
 	assert(ptr != NULL);
+	size_t usize = isalloc(tsdn, ptr);
+	size_t bumped_usize = sz_sa2u(usize, PROF_SAMPLE_ALIGNMENT);
```
interwq, line 1263: Can we do this?

    size_t bumped_usize = edata_size_get(edata);
    assert(bumped_usize == sz_sa2u(usize, PROF_SAMPLE_ALIGNMENT));

interwq (commented):
Almost ready to go!

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -52,56 +52,81 @@ isalloc(tsdn_t *tsdn, const void *ptr) {
 }
 
 JEMALLOC_ALWAYS_INLINE void *
-iallocztm(tsdn_t *tsdn, size_t size, szind_t ind, bool zero, tcache_t *tcache,
-    bool is_internal, arena_t *arena, bool slow_path) {
+iallocztm_explicit_slab(tsdn_t *tsdn, size_t size, szind_t ind, bool zero,
```
interwq, line 55: For the context, if we are strictly sticking to convention, it's probably worth adding another letter here:
https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/jemalloc_internal_inlines_c.h#L37

e.g. `l: accepts a slab parameter`, then this becomes `iallocztml`.

I don't really have a strong preference either way. We talked about switching to a more verbose naming; however being partially-verbose is a bit awkward too. WDYT @Svetlitski ?

test/unit/prof_small.c
Svetlitski: Ah, I just took this from another one of the test cases. Will remove.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
Svetlitski: I personally feel that the current Hungarian notation convention is not the greatest for readability, since all these function names sort of blur together. I'd prefer to keep the longer name, but I'll defer to your judgement.

src/arena.c
Svetlitski: Either we _can't_ do this (although bizarrely the tests passed locally but not in CI), or I misunderstood your intent.

interwq: When it fails, what are the sizes?

Svetlitski: It looks like this _might_ just be a weird CI issue, since all [the failing tests](https://github.com/jemalloc/jemalloc/pull/2459/checks?check_run_id=14204970370) contain output like:
```
<jemalloc>: Invalid conf pair: prof:true
<jemalloc>: Invalid conf pair: prof_accum:true
<jemalloc>: Invalid conf pair: prof_active:false
<jemalloc>: Invalid conf pair: lg_prof_sample:0
```

interwq: Oh the invalid conf output are all from the system malloc -- i.e. the jemalloc on FreeBSD (that doesn't have --enable-prof). The `MALLOC_CONF` env var is also picked up by any tools we come across when doing the tests, so we get these harmless warnings. They will show up even when tests pass.

Since the assertion failed, for testing you can try replacing it with:
```
bumped_usize = edata_size_get(edata);
if (bumped_usize != sz_sa2u(usize, PROF_SAMPLE_ALIGNMENT)) {
    malloc_printf("...", bumped_usize, sz_sa2u, edata_size_get);
}
```

and push to trigger a CI run for debugging.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
interwq: Ok let's keep the long name for now. Next time when we are changing any of the params, it'd be the signal to commit to either the long or short format.

test/unit/prof_small.c
```diff
@@ -0,0 +1,37 @@
...
+		assert_ptr_not_null(edata,
+		    "Unable to find edata for allocation");
+		expect_false(edata_slab_get(edata),
+		    "Sampled small allocations should not be placed on slabs");
+		expect_ptr_eq(edata_base_get(edata), ptr,
+		    "Sampled allocations should be page-aligned");
+		expect_zu_eq(edata_usize_get(edata), size,
+		    "Edata usize did not match requested size");
+		expect_zu_eq(edata_size_get(edata), PAGE_CEILING(size),
+		    "Edata actual size was not a multiple of PAGE");
```
interwq: It looks this is firing because of the extra page from cache_oblivious. We should fix it from this line:
https://github.com/jemalloc/jemalloc/blob/90176f8a87a0b5bdb0ac4c1a515b1d9c58dc5a82/src/arena.c#L337

Maybe this way:
```
size_t pad_size = alignment < PAGE ? sz_large_pad : 0
size_t esize = usize + pad_size;
...
...
if (edata != NULL && pad_size != 0) {
    arena_cache_oblivious_randomize
```

Then change this branch into an assert:
https://github.com/jemalloc/jemalloc/blob/90176f8a87a0b5bdb0ac4c1a515b1d9c58dc5a82/include/jemalloc/internal/arena_inlines_b.h#L494

I believe this would fix the other assertion failure in demote as well

src/arena.c
Svetlitski: I ran it locally with `--enable-cache-oblivious` and was able to reproduce the failure. Re-running after applying your suggested change does indeed fix it.

test/unit/prof_small.c
Svetlitski: I ran it locally with `--enable-cache-oblivious` and was able to reproduce the failure. Re-running after applying your suggested change does indeed fix it.

Svetlitski: Have to add the missing assert still, so hold off on reviewing again

Svetlitski: Now we're getting failures on the realloc path. Specifically in `irallocx_prof_sample`: When you reach line 3310, if all of the following are true:
- `old_usize >= SC_LARGE_MINCLASS`
-  `usize <= SC_SMALL_MAXCLASS`
- `bumped_usize >= SC_LARGE_MINCLASS`
Then it's possible the reallocation will be done in-place via `large_ralloc_no_move`. This causes the assertion in `arena_prof_demote` to fail, because the allocation is larger than strictly necessary. I think all we need to do is change the comparison in the assert on line 1264 of arena.c from `==` to `>=`, since this situation _should_ be allowed.

src/arena.c
```diff
@@ -1443,28 +1450,29 @@ arena_ralloc_no_move(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size,
...
-    size_t size, size_t alignment, bool zero, tcache_t *tcache,
+    size_t size, size_t alignment, bool zero, bool slab, tcache_t *tcache,
     hook_ralloc_args_t *hook_args) {
 	size_t usize = alignment == 0 ? sz_s2u(size) : sz_sa2u(size, alignment);
 	if (unlikely(usize == 0 || size > SC_LARGE_MAXCLASS)) {
 		return NULL;
 	}
 
-	if (likely(usize <= SC_SMALL_MAXCLASS)) {
+	if (likely(usize <= SC_SMALL_MAXCLASS && slab)) {
```
interwq: nit: change to 

    if (likely(slab)) { assert(usize <= SC_SMALL_MAXCLASS);

test/unit/prof_small.c
interwq: Not sure I follow. In the case you described, doesn't it go through `arena_ralloc_move_helper` to realloc+move?

[edit] that assert on line 1264 is actually quite valuable IMO; besides the unit tests, that's probably the most noticeable check for what you are optimizing here. In other words, if we change the bumped size from 1x page to 4x again, that one would be the first to trigger. So I probably want to look more to see if it can be kept

Svetlitski: No (for context here `LG_PAGE` is 16 so `SC_SMALL_MAXCLASS` is 229376, and cache oblivious is enabled):

Screenshot 2023-06-13 at 3 22 10 PM

interwq: This path indicates:
```
if (oldsize >= SC_LARGE_MINCLASS
	    && usize >= SC_LARGE_MINCLASS) {
```
Is that the case you meant before?

interwq: Oh I see. The usize here is the bumped_usize, right?

interwq: I can't repro this locally. It looks CI failures are on FreeBSD?

Does the issue only show up with `cache-oblivious`?

interwq: NVM I'm able to repro now. So it's indeed only affected by the extra page from cache-oblivious. Let me look more into it.

interwq: Okay it looks the sz_large_pad is referenced a bit more than I thought, here's what I think we should do first:
1) revert the changes around `pad_size` and `arena_cache_oblivious_randomize`
2) then for the assertion in `arena_prof_demote`, account for the extra page, e.g. bumped_usize = sa2u + sz_large_pad
3) similarly, in the new prof_small test, account for the extra page as well, e.g. expect_zu_eq(edata_size_get(edata), PAGE_CEILING(size) + sz_large_pad)

This would make sure the all "large" alloc still have the extra padding page, even for the sampled / aligned small ones. It's not ideal since it still has 1 extra page that's unnecessary, but we probably should change that separately, given the PR is getting complex already.

Svetlitski: This also has issues. I feel that it's worth pointing out that all of the pain we're encountering is a consequence of using `size_t bumped_usize = edata_size_get(edata)` on line 1262 of arena.c, instead of the [earlier logic seen here](https://github.com/jemalloc/jemalloc/pull/2459#discussion_r1227285879) (though I understand this approach has better performance characteristics). In this case, the issue is that we're getting failures during stats consolidation `assert(nmalloc >= ndalloc);` on line 141 of arena.c. This appears to be because we're attributing the `alloc` and the `dalloc` for the same sampled profile to different size-classes as a direct consequence of using `size_t bumped_usize = edata_size_get(edata)`.  It's a different manifestation of the same original problem mentioned above, where we shrink the allocation in-place by reusing the existing extent which happens to be large enough.

interwq: I see. Good point. If we switch back to sa2u to compute bumped usize, does that avoid the troubles? If so let's do that instead. Maybe then assert the edata_size is at most 1 page away with cache_oblivious.

Svetlitski: Yes, switching back does avoid the troubles. 
> Maybe then assert edata_size is at most 1 page away with cache_oblivious

In the test-case or in `arena_prof_demote`?

interwq: In arena_prof_demote. It should hold, right?

Svetlitski: This does appear to work, at least in the test-cases. I find this surprising, I would've thought we'd run into the same realloc issue where a larger than necessary slab was used to satisfy the allocation, causing the assertion to fail.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -199,11 +226,21 @@ iralloct(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size, size_t alignment,
...
 	return arena_ralloc(tsdn, arena, ptr, oldsize, size, alignment, zero,
-	    tcache, hook_args);
+	    slab, tcache, hook_args);
+}
+
+JEMALLOC_ALWAYS_INLINE void *
+iralloct(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size, size_t alignment,
+    bool zero, tcache_t *tcache, arena_t *arena, hook_ralloc_args_t *hook_args)
+{
+	size_t usize = alignment == 0 ? sz_s2u(size) : sz_sa2u(size, alignment);
```
interwq: It looks the callsites of this all have `usize` available, calculated through `aligned_usize_get`. Let's pass `usize` down, to avoid duplicate this part.

include/jemalloc/internal/safety_check.h
```diff
@@ -12,19 +14,35 @@ typedef void (*safety_check_abort_hook_t)(const char *message);
 /* Can set to NULL for a default. */
 void safety_check_set_abort(safety_check_abort_hook_t abort_fn);
 
+JEMALLOC_ALWAYS_INLINE const unsigned char *
+compute_redzone_end(const void *_ptr, size_t usize, size_t bumped_usize) {
+	const unsigned char *ptr = (const unsigned char *) _ptr;
+	const unsigned char *redzone_end = usize + 32 < bumped_usize ?
+	    &ptr[usize + 32] : &ptr[bumped_usize];
+	const unsigned char *page_end = (const unsigned char *)
```
interwq, line 31: nit: add some comments about in which case and why this matters.

include/jemalloc/internal/prof_inlines.h
```diff
@@ -239,14 +239,15 @@ prof_realloc(tsd_t *tsd, const void *ptr, size_t size, size_t usize,
...
 	/*
-	 * Enforce page alignment, so that sampled allocations can be identified
+	 * Enforce alignment, so that sampled allocations can be identified
 	 * w/o metadata lookup.
 	 */
 	assert(opt_prof);
-	return (opt_cache_oblivious && orig_align < PAGE) ? PAGE :
-	    orig_align;
+	return (orig_align < PROF_SAMPLE_ALIGNMENT &&
+	       (usize <= SC_SMALL_MAXCLASS || opt_cache_oblivious)) ?
```
interwq: nit: swap to `sz_can_use_slab(usize)`

src/jemalloc.c
```diff
@@ -2372,41 +2372,38 @@ imalloc_no_sample(static_opts_t *sopts, dynamic_opts_t *dopts, tsd_t *tsd,
...
-	dopts->alignment = prof_sample_align(dopts->alignment);
-	if (usize <= SC_SMALL_MAXCLASS) {
-		assert(((dopts->alignment == 0) ?
-		    sz_s2u(SC_LARGE_MINCLASS) :
-		    sz_sa2u(SC_LARGE_MINCLASS, dopts->alignment))
-			== SC_LARGE_MINCLASS);
-		ind_large = sz_size2index(SC_LARGE_MINCLASS);
-		size_t bumped_usize = sz_s2u(SC_LARGE_MINCLASS);
+	if (sz_can_use_slab(usize)) {
+		size_t bumped_usize = sz_sa2u(usize, dopts->alignment);
```
interwq, line 2396: nit: `assert(dopt->alignment & PROF_SAMPLE_ALIGNMENT_MASK == 0)`

test/unit/prof_small.c
interwq: I might not answered the realloc question properly last week; the shrinking small size case should go down this path:
https://github.com/jemalloc/jemalloc/blob/dev/src/arena.c#L1424

In other words, for small sizes, we only keep the original buffer if the size class is the same. For large realloc, we shrink it by splitting the extent when crossing size classes.

test/unit/prof_small.c
```diff
@@ -0,0 +1,38 @@
...
+		    "Edata had null prof_tdata despite being sampled");
+		free(ptr);
+	}
+
+}
+TEST_END
+
+int
+main(void) {
+	return test(test_profile_small_allocations);
```
interwq: Let's add a prof small realloc test case as well?

interwq (commented):
Looks great! Just a few nits left.

test/unit/prof_small.c
Svetlitski: I see. There is the edge-case of sampling the reallocation of a small allocation which was itself already sampled, but with a requested size such that the size-class is the same before and after reallocation (technically the sizes don't have to exactly match, it's when the new size class and old size class round up to the same multiple of `PAGE`, but the point still stands). As the code currently stands, we don't reuse the existing, promoted extent in this case, but instead use a new one. Do you think it's worth special-casing this by adding some additional logic [here in `arena_ralloc`](https://github.com/jemalloc/jemalloc/blob/d1313313101f9df127bba08bf8fd90a849bf3b87/src/arena.c#L1485), or would you rather just let it use a new extent for simplicity's sake?

interwq: I'd consider that case to be rather uncommon, and probably not worth optimizing specifically. One other factor for not doing that, is we may still change the round-up-to-PAGE approach in the future, e.g. once we figure out the story for different edata types, it would be possible to use to the first item in each slab for samples (which is still page-aligned) to avoid a lot of the special casing.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
Svetlitski: Done. I haven't plumbed it _all_ the way through every ensuing function (e.g. through `iralloct_explicit_slab`, then to `arena_ralloc`, etc.), but if that was your intent let me know.

include/jemalloc/internal/jemalloc_internal_inlines_c.h
```diff
@@ -199,18 +226,29 @@ iralloct(tsdn_t *tsdn, void *ptr, size_t oldsize, size_t size, size_t alignment,
...
+	return iralloct_explicit_slab(tsdn, ptr, oldsize, size, alignment, zero,
+	    slab, tcache, arena, hook_args);
 }
 
 JEMALLOC_ALWAYS_INLINE void *
 iralloc(tsd_t *tsd, void *ptr, size_t oldsize, size_t size, size_t alignment,
     bool zero, hook_ralloc_args_t *hook_args) {
-	return iralloct(tsd_tsdn(tsd), ptr, oldsize, size, alignment, zero,
-	    tcache_get(tsd), NULL, hook_args);
+	size_t usize = alignment == 0 ? sz_s2u(size) : sz_sa2u(size, alignment);
```
interwq: nit: `iralloc` is not used anywhere right now. Let's just add `usize` to its arg list.

test/unit/prof_small.c
```diff
@@ -0,0 +1,78 @@
...
+TEST_END
+
+TEST_BEGIN(test_profile_small_reallocations_same_size_class) {
+	test_skip_if(!config_prof);
+
+	for (szind_t index = 0; index < SC_NBINS; index++) {
+		size_t size = sz_index2size(index);
+		void *ptr = malloc(size);
+		assert_small_allocation_sampled(ptr, size);
+		ptr = realloc(ptr, size - 1);
```
interwq, line 65: nit: we should be able to assert ptr1 (initial) == ptr2 (after realloc)

interwq (approved):

test/unit/prof_small.c
Svetlitski: It doesn't, [which is why I had asked about this edge-case](https://github.com/jemalloc/jemalloc/pull/2459#discussion_r1251237516).

interwq: Discussed offline. Kevin is right that we can't really assure the same buffer, even if both new and old are going to be sampled.

interwq:
Thanks again @Svetlitski ! Once this is up in our prod binaries, we can update and close #2358.

PR #1980, opened 2020-11-11
Add a per-arena oversize_threshold.

davidtgoldblatt:
This can let manual arenas trade off memory and CPU the way auto arenas do.

PR #2460, opened 2023-06-06
Fix the arena selection for oversized allocations.

interwq:
Use the per-arena oversize_threshold, instead of the global setting.

For the context, a per-arena oversize_threshold mallctl was added in #1980, but the arena selection part wasn't updated and was still using the opt runtime value. Fix that to unblock first. I'll add tests around this later.


Svetlitski (approved):
Logic checks out, looks good to me!

PR #2466, opened 2023-06-23
Avoid eager purging on the dedicated oversize arena when using bg thds.

interwq:
We have observed new workload patterns (namely ML training type) that cycle through oversized allocations frequently, because 1) the dataset might be sparse which is faster to go through, and 2) GPU accelerated.  As a result, the eager purging from the oversize arena becomes a bottleneck.  To offer an easy solution, allow normal purging of the oversized extents when background threads are enabled.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -1730,6 +1730,42 @@ label_error:
 	return NULL;
 }
 
+static arena_t *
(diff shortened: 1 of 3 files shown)


interwq:
@guangli-dai as we recently discussed.

guangli-dai (approved):

PR #2473, opened 2023-07-05
Fix warnings with -Wstrict-prototypes.

### LINKED ISSUE #2468: Missing function prototype in test hooks
https://github.com/jemalloc/jemalloc/blob/d1313313101f9df127bba08bf8fd90a849bf3b87/include/jemalloc/internal/test_hooks.h#L4

This line and the line below define a function pointer with a missing prototype. In these lines, `()` should be replaced with `(void)`.

There may be other places in the code with the same problem. NetBSD compiles the code using GCC's `-Wstrict-prototypes` option to detect these.

interwq:
Resolves #2468. Added -Wstrict-prototypes by default.

Besides the one case of `set_abort_called(const char *message)` in  test/unit/zero_realloc_abort, all others are trivial changes of `func(void)`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -262,6 +262,7 @@ if test "x$GCC" = "xyes" ; then
   dnl This one too.
   JE_CFLAGS_ADD([-Wno-missing-field-initializers])
   JE_CFLAGS_ADD([-Wno-missing-attributes])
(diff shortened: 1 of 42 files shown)


Svetlitski (approved):
As of the very latest push, this looks good to me!

guangli-dai (approved):

PR #2474, opened 2023-07-05
Remove unreachable code.

### LINKED ISSUE #2467: Unreachable duplicate code in bit_util.h
https://github.com/jemalloc/jemalloc/blob/d1313313101f9df127bba08bf8fd90a849bf3b87/include/jemalloc/internal/bit_util.h#L343

This call to `ffs_u` is either unreachable or is not compiled (at least since C90). What is its purpose?
For comparison, a few lines above, `ffs_u64` does not have this extra function call expression.

The function call was introduced in [this commit](https://github.com/jemalloc/jemalloc/commit/9f4ee6034c3ac6a8c8b5f9a0d76822fb2fd90c41#diff-45f14b04b911d1e0d4ca9b5e6372dafed402bac11a63ab6e3ffd91814c11ba04R202).

interwq:
Resolves #2467

### DIFF
--- a/include/jemalloc/internal/bit_util.h
+++ b/include/jemalloc/internal/bit_util.h
@@ -340,7 +340,6 @@ ffs_u32(uint32_t x) {
 #else
 #error No implementation for 32-bit ffs()
 #endif
-	return ffs_u(x);
 }
 
 static inline unsigned
@@ -350,7 +349,6 @@ fls_u32(uint32_t x) {
 #else
 #error No implementation for 32-bit fls()
 #endif
-	return fls_u(x);
 }
 
 static inline uint64_t


Svetlitski (approved):
Overall this looks good. It's not strictly necessary (and actually wouldn't have helped in this case), but if we want to go the extra mile to avoid similar issues in the future we should add `-Wunreachable-code` and `-Wunreachable-code-aggressive` to our warnings (with compiler support detection via autoconf, of course). These surprisingly are not enabled by `-Wall -Wextra`.

interwq:
I'll land this one for now. Please feel free to give `-Wunreachable-code` a try if you'd like @Svetlitski. A while ago I remember it could result in false positives for stuff like `if (const false) { ... }`, and there was old discussion like this: https://gcc.gnu.org/legacy-ml/gcc-help/2011-05/msg00360.html So if it ends up not working / not worth it, it's alright too.

interwq:
Merging. CI failures from Travis (same non responsive ppc64).

PR #2478, opened 2023-07-07
Fix inconsistent parameter names between definition/declaration pairs

Svetlitski:
For the sake of consistency, function definitions and their corresponding declarations should use the same names for parameters. I've enabled this check in static analysis to prevent this issue from occurring again in the future.

### DIFF
--- a/include/jemalloc/internal/arena_externs.h
+++ b/include/jemalloc/internal/arena_externs.h
@@ -47,9 +47,9 @@ edata_t *arena_extent_alloc_large(tsdn_t *tsdn, arena_t *arena,
 void arena_extent_dalloc_large_prep(tsdn_t *tsdn, arena_t *arena,
     edata_t *edata);
 void arena_extent_ralloc_large_shrink(tsdn_t *tsdn, arena_t *arena,
-    edata_t *edata, size_t oldsize);
+    edata_t *edata, size_t oldusize);
 void arena_extent_ralloc_large_expand(tsdn_t *tsdn, arena_t *arena,
-    edata_t *edata, size_t oldsize);
+    edata_t *edata, size_t oldusize);
 bool arena_decay_ms_set(tsdn_t *tsdn, arena_t *arena, extent_state_t state,
     ssize_t decay_ms);
 ssize_t arena_decay_ms_get(arena_t *arena, extent_state_t state);
--- a/include/jemalloc/internal/tsd.h
+++ b/include/jemalloc/internal/tsd.h
@@ -164,7 +164,7 @@ void malloc_tsd_dalloc(void *wrapper);
 tsd_t *malloc_tsd_boot0(void);
 void malloc_tsd_boot1(void);
 void tsd_cleanup(void *arg);
-tsd_t *tsd_fetch_slow(tsd_t *tsd, bool internal);
+tsd_t *tsd_fetch_slow(tsd_t *tsd, bool minimal);
 void tsd_state_set(tsd_t *tsd, uint8_t new_state);
 void tsd_slow_update(tsd_t *tsd);
 void tsd_prefork(tsd_t *tsd);
--- a/include/jemalloc/internal/bin.h
+++ b/include/jemalloc/internal/bin.h
@@ -48,7 +48,7 @@ struct bins_s {
 	bin_t *bin_shards;
 };
 
-void bin_shard_sizes_boot(unsigned bin_shards[SC_NBINS]);
+void bin_shard_sizes_boot(unsigned bin_shard_sizes[SC_NBINS]);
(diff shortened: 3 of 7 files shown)


include/jemalloc/internal/tsd.h
```diff
@@ -164,7 +164,7 @@ void malloc_tsd_dalloc(void *wrapper);
 tsd_t *malloc_tsd_boot0(void);
 void malloc_tsd_boot1(void);
 void tsd_cleanup(void *arg);
-tsd_t *tsd_fetch_slow(tsd_t *tsd, bool internal);
+tsd_t *tsd_fetch_slow(tsd_t *tsd, bool minimal);
```
interwq, line 167: Good catch. This one can be rather confusing.

include/jemalloc/internal/arena_externs.h
```diff
@@ -47,9 +47,9 @@ edata_t *arena_extent_alloc_large(tsdn_t *tsdn, arena_t *arena,
 void arena_extent_dalloc_large_prep(tsdn_t *tsdn, arena_t *arena,
     edata_t *edata);
 void arena_extent_ralloc_large_shrink(tsdn_t *tsdn, arena_t *arena,
-    edata_t *edata, size_t oldsize);
+    edata_t *edata, size_t oldusize);
```
interwq, line 50: While following this path, I realized that this probably needs to be updated following #2459
https://github.com/jemalloc/jemalloc/blob/dev/src/arena.c#L300
https://github.com/jemalloc/jemalloc/blob/dev/src/arena.c#L316

This shows in a different way how helpful your static-analysis fixes are :)

Svetlitski: Good catch. I'll open a separate PR for that.

interwq (approved):

include/jemalloc/internal/arena_externs.h
interwq: It might not be as straightforward though, since for large stats, i.e. `arena->stats.lstats`, the index starts from SC_LARGE_MINCLASS, therefore the `hindex`. We technically should update arena->bin.stats like here:
https://github.com/jemalloc/jemalloc/blob/dev/src/arena.c#L1175

but we neither hold that mutex at that moment, nor it was really allocated under that mutex. What's your opinion there?

Svetlitski: I'm not sure if there's any great answer here. If we include it in `bin.stats` that's going to potentially mislead anyone using those stats for memory accounting by under-reporting. If we do what we currently do, that's also wrong, but by over-reporting memory usage. I suppose the "proper" thing to do would be to add a new field:
```c
typedef arena_stats_large_t arena_stats_small_sampled_t;

struct arena_stats_s {
    ...
    arena_stats_small_sampled_t sampled_stats[4];
};
```
But then we have to go through all of the trouble of plumbing that through everywhere.

Svetlitski: Perhaps we could add a new field `uint64_t num_sampled` to `bin_stats_t`. That might be a better compromise. This way we could just update `nmalloc`/`ndalloc` for the corresponding bin, but we'd still have a way to track the memory overhead.

interwq: I agree that the current way is over-reporting. For the under-reporting part, not sure if I follow. I'm thinking to include it in the nmalloc of the corresponding bin, e.g. the one for 4K, 8K and 12K which is the actual size used for the sample. However it does feel wrong, because it's not really allocated through the bin mechanism.

Svetlitski: Oh, okay I see. When I said "under-reporting" I thought we would still be attributing the allocation to the bin for the non-promoted `usize` of the allocation, so that we don't artificially make certain size-classes look hotter than they actually are (i.e. imagine you have _tons_ of 8-byte allocations, enough that if you record the sampled ones as 4 K allocations it makes the 4 K bin look hot, misleading your optimization efforts).

interwq: Agree that there are two type of stats we should report:
1) what actually happens at the allocator level (this is the case before the prof overhead PR)
2) what the application actually requested (this isn't reported, before or after the prof overhead PR)

After the prof overhead PR, the stats is a bit off because right now it's neither. And if we want to still do (1), there isn't a place to report "small & non-slab" request.

Does the below sound reasonable?
1) pile up the `nmalloc` and `nrequests` into the bin stats of the corresponding page size. But probably not `curregs` because it should match the slab stats. 
2) adding a new counter for prof sampled small requests, either a total or per bin (which will only have non-zeros for page sizes)

Either way that certainly should happen separately. I'll merge this PR first.

Svetlitski: Addressed by #2486.

Issue #17, opened 2013-10-28
Reduce rtree memory usage.


jasone:
Reduce rtree memory usage by storing booleans (1 byte each) rather than pointers.  The rtree code is only used to record whether jemalloc manages a chunk of memory, so there's no need to store pointers in the rtree.

Issue #16, opened 2013-10-28
Port heap profiling to FreeBSD.


jasone:
Port heap profiling to FreeBSD.  This is very close to working already, the only issue being that the the the heap map data need to be synthesized in a format that pprof can digest.

Issue #15, opened 2013-10-28
Add atomic.h support for powerpc 32/64 architectures.


jasone:
Add atomic.h support for powerpc 32/64 architectures.  This is needed when building with old versions of gcc that lack the necessary intrinsics.

jasone:
@davidtgoldblatt, your work in #532 is likely to obsolete this.

davidtgoldblatt:
After #664, we should support power with gcc versions >= 4.1.0.

Issue #14, opened 2013-10-28
Experiment with per CPU arenas.


jasone:
Experiment with using just one arena per CPU, and using sched_getcpu(3) or equivalent as a heuristic when choosing an arena for allocation that cannot be satisfied by the thread-specific cache.  The known challenges include:
- sched_getcpu(3) is Linux-specific.  FreeBSD probably has the necessary infrastructure to implement an equivalent feature, but it does not currently exist.
- Because this strategy for arena selection cannot be implemented on all platforms, the existing arena selection strategy will have to remain as an option.
- If an application runs fewer threads than there are CPUs, a per CPU arena is a waste.  It may be important to transition from one arena selection strategy to another as the number of threads increases.

thestinger:
The best approach that I've come up with is the following:

* 1 arena per core, rather than 4 arenas per core
* assign the initial threads via round-robin, until each arena is used
* when there are no unused arenas, switch to sched_getcpu()
* store the thread ID of the last thread to allocate in the arena

The algorithm for picking an arena for allocating:

```
if thread.last_arena.last_allocator == thread.id && trylock() != fail
    pass
else:
    pick_arena_with_sched_getcpu()
    lock()
set_last_allocator()
```

This results in significantly better load balancing than jemalloc has at the moment while using 1/4 as many arenas. The dynamic load balancing works extremely well because a new "owning" thread for the arena will kick out any stragglers not yet assigned to the correct arena for their core.

(I also mentioned this at http://www.canonware.com/pipermail/jemalloc-discuss/2015-February/001014.html)

thestinger:
The usage of `trylock()` is probably unnecessary with the addition of the `last_allocator` heuristic. It could just do the following to avoid a pointless `sched_getcpu` call due to threads holding the arena lock to free memory:

```
if thread.arena.last_allocator != thread.id:
    thread.arena = pick_arena_with_sched_getcpu()
    lock()
    thread.arena.last_allocator = thread.id
else:
    lock()
```

The `last_allocator` handling would just use `memory_order_relaxed` atomics in C11.

poige:
The CIF allocator mentioned in  "Cache Index-Aware Memory Allocation", by Afek, Dice, and Morrison (ISSM 2011) „… can be configured to use the CPUID or RDTSCP instructions to select a heap. …”, so, possibly the same approach is also suitable for the matter in question. According to Wiki:
„…
[...]

PR #2484, opened 2023-07-17
Fix infinite purging loop in HPA

### LINKED ISSUE #2449: Infinite loop in hpa_shard_maybe_do_deferred_work during thread exit while using HPA
I've been experimenting with the HPA feature. It's my understanding that this feature isn't really meant to be used publicly yet, but it's been in the upstream code for a while now so I got curious. So I understand if there's an unwillingness to support it yet.

I'm using jemalloc as of SHA 9c32689e576906332d2ceaabafc2a927d152beba.

I've compiled with `--enable-stats --enable-fill --with-jemalloc-prefix=je_trunk_ --enable-debug`.

I'm running the following config.
```
char const * je_trunk_malloc_conf =
    "confirm_conf:true"            
    ",hpa:true"                    
    ;                              
```
I'm running on a linux 4.19.80 system.

I'm observing what appears to be an infinite loop in the following code when a thread exits.
```
#0  0x00007fffe89842c7 in __libc_write (fd=1, buf=0x7fff00337a20, nbytes=62) at ../sysdeps/unix/sysv/linux/write.c:27
#1  0x00007ffff5735756 in hpa_shard_maybe_do_deferred_work (tsdn=<optimized out>, shard=0x7fffd1613148, forced=<optimized out>) at /<omitted>/src/hpa.c:550
#2  0x00007ffff5733b75 in hpa_dalloc_batch (tsdn=0x7fff00341cc8, self=<optimized out>, list=0x7fff00337c98, deferred_work_generated=0x7fff00337ca7) at /<omitted>/src/hpa.c:873
#3  0x00007ffff575c906 in pai_dalloc_batch (tsdn=<optimized out>, self=0x7fff00337a20, list=0x0, deferred_work_generated=0x7fffd16130d000)
    at /<omitted>/include/jemalloc/internal/pai.h:77
#4  sec_flush_some_and_unlock (tsdn=0x7fff00341cc8, sec=0x7fffd16130d0, shard=<optimized out>) at /<omitted>/src/sec.c:153
#5  0x00007ffff575b605 in sec_shard_dalloc_and_unlock (tsdn=<optimized out>, sec=<optimized out>, shard=0x7fffd1617340, edata=<optimized out>) at /<omitted>/src/sec.c:329
#6  sec_dalloc (tsdn=0x7fff00341cc8, self=0x7fffd16130d0, edata=0x7fffd1622700, deferred_work_generated=0x7fff00337d1f) at /<omitted>/src/sec.c:349
#7  0x00007ffff574bcb0 in large_dalloc_finish_impl (tsdn=0x7fff00341cc8, arena=0x7fffd1601080, edata=0x3e) at /<omitted>/src/large.c:253
#8  je_large_dalloc_finish (tsdn=0x7fff00341cc8, edata=0x3e) at /<omitted>/src/large.c:266
#9  0x00007ffff576f701 in tcache_bin_flush_impl (tsd=<optimized out>, tcache=<optimized out>, cache_bin=0x7fff00342388, binind=<optimized out>, ptrs=<optimized out>, nflush=13, small=<optimized out>)
    at /<omitted>/src/tcache.c:460
#10 0x00007ffff576e67f in tcache_bin_flush_bottom (tsd=0x7fff00341cc8, tcache=<optimized out>, cache_bin=0x7fff00342388, binind=36, rem=<optimized out>, small=<optimized out>)
    at /<omitted>/src/tcache.c:519
#11 0x00007ffff5771328 in tcache_flush_cache (tsd=0x7fff00341cc8, tcache=0x7fff00342020) at /<omitted>/src/tcache.c:535
#12 0x00007ffff577158f in tcache_destroy (tsd=0x7fff00341cc8, tcache=0x7fff00342020, tsd_tcache=<optimized out>) at /<omitted>/src/tcache.c:809
#13 0x00007ffff5771446 in je_tcache_cleanup (tsd=0x7fff00341cc8) at /<omitted>/src/tcache.c:852
#14 0x00007ffff577520a in tsd_do_data_cleanup (tsd=<optimized out>) at /<omitted>src/tsd.c:398
#15 je_tsd_cleanup (arg=0x7fff00341cc8) at /<omitted>/src/tsd.c:424
#16 0x00007fffe8979408 in __nptl_deallocate_tsd () at pthread_create.c:300
#17 0x00007fffe897a81b in __nptl_deallocate_tsd () at ../sysdeps/nptl/futex-internal.h:200
#18 start_thread (arg=0x7fff00400700) at pthread_create.c:473
#19 0x00007fffe82caa3f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```
Specifically, it appears to be stuck in an infinite loop around the following code in hpa.c
```
while (hpa_should_purge(tsdn, shard) && nops < max_ops) {       
        purged = hpa_try_purge(tsdn, shard);                             
                                                                
        if (purged) {                                           
                nops++;                                         
        }                                                       
}                                                               
```
I stuck a few prints in the code to try to figure out what was going on (which is why the stacktrace above shows a __libc_write that's not in the upstream code)
```
--- a/src/hpa.c                                                     
+++ b/src/hpa.c                                                                   
@@ -519,6 +521,12 @@ hpa_shard_maybe_do_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard,
        if (!forced && shard->opts.deferral_allowed) {                                 
                return;                                                                
        }                                                                              
+                                                                                      
+        char buf[256];                                                                
+        sprintf(buf, "%ld hpa_shard_maybe_do_deferred_work(%p, %p, %d)\n",            
+               syscall(SYS_gettid), tsdn, shard, forced);                             
+        write(STDOUT_FILENO, buf, strlen(buf));                                       
+                                                                                      
        /*                                                                             
         * If we're on a background thread, do work so long as there's work to         
         * be done.  Otherwise, bound latency to not be *too* bad by doing at          
@@ -536,6 +544,11 @@ hpa_shard_maybe_do_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard,
                purged = false;                                                        
                while (hpa_should_purge(tsdn, shard) && nops < max_ops) {              
                        purged = hpa_try_purge(tsdn, shard);                           
+                                                                                      
+                        sprintf(buf, "%ld hpa_try_purge(%p, %p)=%d, nops=%lu\n",      
+                               syscall(SYS_gettid), tsdn, shard, purged, nops);       
+                        write(STDOUT_FILENO, buf, strlen(buf));                       
+                                                                                      
                        if (purged) {                                                  
                                nops++;                                                
                        }                                                              
```
The logs appear to show that the code is calling `hpa_should_purge` in a tight loop with the same arguments. Here's a sample of the printed output
```
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
655 hpa_try_purge(0x7fce00341cc8, 0x7fddb0013148)=0, nops=0
```
This occurs in a wide variety of test cases that all work with jemalloc 5.3.

I'm wondering it this is a known bug, or if I'm using HPA in an supported way? Is there a known-good config for HPA that's able to be shared publicly or a SHA where HPA is known to be stable that can be shared publicly?

Let me know if there's any other information I can provide, and thanks for all the excellent work on this project.

Svetlitski:
As reported in #2449, under certain circumstances it's possible to get stuck in an infinite loop attempting to purge from the HPA. We now handle this edge-case correctly.

Closes #2449.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -144,6 +144,7 @@ static void default_junk_free(void *ptr, size_t usize) {
 
 void (*JET_MUTABLE junk_alloc_callback)(void *ptr, size_t size) = &default_junk_alloc;
 void (*JET_MUTABLE junk_free_callback)(void *ptr, size_t size) = &default_junk_free;
+void (*JET_MUTABLE invalid_conf_abort)(void) = &abort;
 
 bool	opt_utrace = false;
 bool	opt_xmalloc = false;
@@ -959,7 +960,7 @@ malloc_abort_invalid_conf(void) {
 	assert(opt_abort_conf);
 	malloc_printf("<jemalloc>: Abort (abort_conf:true) on invalid conf "
 	    "value (see above).\n");
-	abort();
+	invalid_conf_abort();
 }
 
 static void
@@ -1079,6 +1080,46 @@ obtain_malloc_conf(unsigned which_source, char buf[PATH_MAX + 1]) {
 		ret = NULL;
 	}
 	return ret;
+}
+
+static void
+validate_hpa_settings(void) {
+	if (!hpa_supported() || !opt_hpa || opt_hpa_opts.dirty_mult == (fxp_t)-1) {
+		return;
+	}
+	size_t hpa_threshold = fxp_mul_frac(HUGEPAGE, opt_hpa_opts.dirty_mult) +
+	    opt_hpa_opts.hugification_threshold;
+	if (hpa_threshold > HUGEPAGE) {
+		return;
+	}
+
+	had_conf_error = true;
+	char hpa_dirty_mult[FXP_BUF_SIZE];
+	char hugification_threshold[FXP_BUF_SIZE];
+	char normalization_message[256] = {0};
+	fxp_print(opt_hpa_opts.dirty_mult, hpa_dirty_mult);
+	fxp_print(fxp_div(FXP_INIT_INT((unsigned)
+	    (opt_hpa_opts.hugification_threshold >> LG_PAGE)),
+	    FXP_INIT_INT(HUGEPAGE_PAGES)), hugification_threshold);
+	if (!opt_abort_conf) {
+		char normalized_hugification_threshold[FXP_BUF_SIZE];
+		opt_hpa_opts.hugification_threshold +=
+		    HUGEPAGE - hpa_threshold;
+		fxp_print(fxp_div(FXP_INIT_INT((unsigned)
+		    (opt_hpa_opts.hugification_threshold >> LG_PAGE)),
+		    FXP_INIT_INT(HUGEPAGE_PAGES)),
+		    normalized_hugification_threshold);
+		malloc_snprintf(normalization_message,
+		    sizeof(normalization_message), "<jemalloc>: Normalizing "
+		    "HPA settings to avoid pathological behavior, setting "
+		    "hpa_hugification_threshold_ratio: to %s.\n",
+		    normalized_hugification_threshold);
+	}
+	malloc_printf(
+	    "<jemalloc>: Invalid combination of options "
+	    "hpa_hugification_threshold_ratio: %s and hpa_dirty_mult: %s. "
+	    "These values should sum to > 1.0.\n%s", hugification_threshold,
+	    hpa_dirty_mult, normalization_message);
 }
 
 static void
@@ -1749,6 +1790,7 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
     /* Re-enable diagnostic "-Wtype-limits" */
     JEMALLOC_DIAGNOSTIC_POP
 		}
+		validate_hpa_settings();
 		if (opt_abort_conf && had_conf_error) {
 			malloc_abort_invalid_conf();
 		}
--- a/test/unit/hpa_validate_conf.c
+++ b/test/unit/hpa_validate_conf.c
@@ -1,0 +1,56 @@
+#include "test/jemalloc_test.h"
+
+static bool abort_called = false;
+static void (*default_malloc_message)(void *, const char *);
+
+static void
+mock_invalid_conf_abort(void) {
+	abort_called = true;
+}
+
+static void
+null_malloc_message(void *_1, const char* _2) {
+}
+
+TEST_BEGIN(test_hpa_validate_conf) {
+	test_skip_if(!hpa_supported());
+	void *ptr = malloc(4096);
+	/* Need to restore this here to see any possible assert messages */
+	malloc_message = default_malloc_message;
+	assert_true(abort_called,
+	     "Should have aborted due to invalid values for hpa_dirty_mult and "
+	     "hpa_hugification_threshold_ratio");
+	free(ptr);
+}
+TEST_END
+
+/*
+ * We have to set `abort_conf:true` here and not via the `MALLOC_CONF`
+ * environment variable in the associated shell script for this test. This is
+ * because when testing on FreeBSD (where Jemalloc is the system allocator) in
+ * CI configs where HPA is not supported, setting `abort_conf:true` there would
+ * result in the system Jemalloc picking this up and aborting before we could
+ * ever even launch the test.
+ */
+const char *malloc_conf = "abort_conf:true";
+
+int
+main(void) {
+	/*
+	 * OK, this is a sort of nasty hack.  We don't want to add *another*
+	 * config option for HPA (the intent is that it becomes available on
+	 * more platforms over time, and we're trying to prune back config
+	 * options generally.  But we'll get initialization errors on other
+	 * platforms if we set hpa:true in the MALLOC_CONF (even if we set
+	 * abort_conf:false as well).  So we reach into the internals and set
+	 * them directly, but only if we know that we're actually going to do
+	 * something nontrivial in the tests.
+	 */
+	if (hpa_supported()) {
+		default_malloc_message = malloc_message;
+		malloc_message = null_malloc_message;
+		opt_hpa = true;
+		invalid_conf_abort = mock_invalid_conf_abort;
+	}
+	return test_no_reentrancy(test_hpa_validate_conf);
+}
--- a/Makefile.in
+++ b/Makefile.in
@@ -225,6 +225,7 @@ TESTS_UNIT := \
 	$(srcroot)test/unit/hook.c \
 	$(srcroot)test/unit/hpa.c \
 	$(srcroot)test/unit/hpa_background_thread.c \
+	$(srcroot)test/unit/hpa_validate_conf.c \
 	$(srcroot)test/unit/hpdata.c \
 	$(srcroot)test/unit/huge.c \
 	$(srcroot)test/unit/inspect.c \
--- a/include/jemalloc/internal/jemalloc_internal_externs.h
+++ b/include/jemalloc/internal/jemalloc_internal_externs.h
@@ -25,6 +25,7 @@ extern bool opt_junk_alloc;
 extern bool opt_junk_free;
 extern void (*JET_MUTABLE junk_free_callback)(void *ptr, size_t size);
 extern void (*JET_MUTABLE junk_alloc_callback)(void *ptr, size_t size);
+extern void (*JET_MUTABLE invalid_conf_abort)(void);
 extern bool opt_utrace;
 extern bool opt_xmalloc;
 extern bool opt_experimental_infallible_new;
--- a/test/unit/hpa_background_thread.sh
+++ b/test/unit/hpa_background_thread.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-export MALLOC_CONF="hpa_dirty_mult:0,hpa_min_purge_interval_ms:50,hpa_sec_nshards:0"
+export MALLOC_CONF="hpa_dirty_mult:0.001,hpa_hugification_threshold_ratio:1.0,hpa_min_purge_interval_ms:50,hpa_sec_nshards:0"
 
--- a/test/unit/hpa_validate_conf.sh
+++ b/test/unit/hpa_validate_conf.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+export MALLOC_CONF='tcache:false,hpa_dirty_mult:0.25,hpa_hugification_threshold_ratio:0.6'


ericm1024:
Do you think it's worth adding a regression test based off the reproducer you posted in the issue?

If you think it's worth while, I'd be happy to take a stab at it.

src/hpa.c
```diff
@@ -538,6 +538,16 @@ hpa_shard_maybe_do_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard,
...
 			if (purged) {
 				nops++;
+			} else {
+				/*
+				 * It's important to explicitly break on
+				 * failure, as otherwise it's possible to end
+				 * up in an infinite loop continuously trying
+				 * to purge without actually being able to
+				 * do so.
+				 * See GitHub issue #2449.
```
interwq: nit: Let's avoid referring to github issues in the source / comments. Put in commit message like you have right now is sufficient.

Svetlitski: Actually come to think of it, I don't know that this is the right fix. I just pulled this up from a conversation on this topic with @davidtgoldblatt when I had encountered a similar issue, and he said:

> I think that combination of hpa_hugification_threshold_ratio and hpa_dirty_mult doesn't make sense
the one says "create 25 percent more dirty pages if you have to in order to hugify", and the other says "purge if you're more than 3 percent dirty (say, because you just created a bunch of dirty pages by hugifying)"
(probably there should be a warning or something if their sum is < 1)

davidtgoldblatt: I don't have a strong opinion. I guess stylistically, I'd say that it's hpa_should_purge that should be returning false in this situation, rather than having this loop decline to purge even when it's told it ought to.

Like, at its core this is a misconfiguration issue[1]. Our choices for responding to misconfiguration issues are:
1. Silently ignore the settings and pick a reasonable default as an alternative
2. Noisily ignore the settings and pick a reasonable default as an alternative
3. Crash at boot
4. Perform pathologically badly

I think all of these have their pros/cons. I'd maybe do 2 by default or 3 when abort_conf is true? Maybe what I'm saying is that this should be handled at conf parsing time (with a crash here after the fact if conf parsing didn't fix it). No strong opinions though, really.

[1] Not really blaming the users; it's defaults that are broken here.

Svetlitski: I agree. I'll start implementing the verification you've described, so please don't merge this PR yet @interwq.

Svetlitski: Implemented in the latest push. Let me know what you think, and once we're happy with everything I'll add test-cases exercising this new behavior, and fix the existing test-cases.

davidtgoldblatt (approved):
Looks reasonable to me.

interwq:
@Svetlitski back to you for CI failure (looks HPA isn't supported for these):

=== test/unit/hpa_validate_conf ===
<jemalloc>: Invalid conf pair: hpa_dirty_mult:0.25
<jemalloc>: Invalid conf pair: hpa_hugification_threshold_ratio:0.6
Abort trap (core dumped)

Svetlitski:
The single failure in Travis CI is just a PowerPC job that looks like it didn't boot successfully, so I believe we're good-to-go @interwq.

pdokoupil:
Sorry for interfering with this PR, but our experience so far is that changing the `hpa_dirty_mult` and `hpa_hugification_threshold_ratio` so they sum to >= 1.0 did not help with the infinite HPA loop. We tried different combinations of the two, and even though some were working, they were not stable enough. After more experiments, we discovered that if we run with a single thread (and single arena), the system is stable and there are no infinite loops. However, once we run with multiple threads (and arenas), we still sometimes end up in the infinite loop.

Applying the "fix" from https://github.com/jemalloc/jemalloc/issues/2449 helped us with infinite loops, but caused problems with jemalloc statistics, namely "mapped" was less than "active".

Could you perhaps provide us with more information on how to set `hpa_hugification_threshold_ratio`, `hpa_dirty_mult` (and other parameters, if needed) to make HPA work reliably? We would highly appreciate your support, thanks.

Svetlitski:
@pdokoupil

Are you still running into issues when you build Jemalloc from source with the current contents of this PR? I ask just in case, as perhaps your settings aren't getting applied (which this PR would catch).

pdokoupil:
> @pdokoupil
> 
> Are you still running into issues when you build Jemalloc from source with the current contents of this PR? I ask just in case, as perhaps your settings aren't getting applied (which this PR would catch).

Yes, I tried building and running with Svetlitski:dev with 0ca184134568da564103d6cc79ed34b8d6335abb applied on top of that. 
Here is the configuration we are using (config validator does not fail on that one)
```
<jemalloc>: -- Set conf value: metadata_thp:always
<jemalloc>: -- Set conf value: hpa:true
<jemalloc>: -- Set conf value: dss:disabled
<jemalloc>: -- Set conf value: thp:always
<jemalloc>: -- Set conf value: hpa_dirty_mult:0
<jemalloc>: -- Set conf value: hpa_hugification_threshold_ratio:1.0
<jemalloc>: -- Set conf value: confirm_conf:true
```

<details>
  <summary>**Build & runtime option settings**</summary>
  
 ### Build-time option settings
  ```
config.cache_oblivious: true
  config.debug: true
  config.fill: true
  config.lazy_lock: false
  config.malloc_conf: ""
  config.opt_safety_checks: true
  config.prof: false
  config.prof_libgcc: false
  config.prof_libunwind: false
  config.stats: true
  config.utrace: false
  config.xmalloc: false
```
### Run-time option settings
```
  opt.abort: true
  opt.abort_conf: true
  opt.cache_oblivious: true
  opt.confirm_conf: true
  opt.retain: true
  opt.dss: "disabled"
  opt.narenas: 64
  opt.percpu_arena: "disabled"
  opt.oversize_threshold: 8388608
  opt.hpa: true
  opt.hpa_slab_max_alloc: 65536
  opt.hpa_hugification_threshold: 2097152
  opt.hpa_hugify_delay_ms: 10000
  opt.hpa_min_purge_interval_ms: 5000
  opt.hpa_dirty_mult: "0.0"
  opt.hpa_sec_nshards: 4
  opt.hpa_sec_max_alloc: 32768
  opt.hpa_sec_max_bytes: 262144
  opt.hpa_sec_bytes_after_flush: 131072
  opt.hpa_sec_batch_fill_extra: 0
  opt.metadata_thp: "always"
  opt.mutex_max_spin: 600
  opt.background_thread: false (background_thread: false)
  opt.dirty_decay_ms: 10000 (arenas.dirty_decay_ms: 10000)
  opt.muzzy_decay_ms: 0 (arenas.muzzy_decay_ms: 0)
  opt.lg_extent_max_active_fit: 6
  opt.junk: "true"
  opt.zero: false
  opt.experimental_infallible_new: false
  opt.tcache: true
  opt.tcache_max: 32768
  opt.tcache_nslots_small_min: 20
  opt.tcache_nslots_small_max: 200
  opt.tcache_nslots_large: 20
  opt.lg_tcache_nslots_mul: 1
  opt.tcache_gc_incr_bytes: 65536
  opt.tcache_gc_delay_bytes: 0
  opt.lg_tcache_flush_small_div: 1
  opt.lg_tcache_flush_large_div: 1
  opt.debug_double_free_max_scan: 32
  opt.thp: "always"
  opt.stats_print: false
  opt.stats_print_opts: ""
  opt.stats_print: false
  opt.stats_print_opts: ""
  opt.stats_interval: -1
  opt.stats_interval_opts: ""
  opt.zero_realloc: "free"
```
  
</details>

If we try for example
```
<jemalloc>: -- Set conf value: hpa_dirty_mult:0.4
<jemalloc>: -- Set conf value: hpa_hugification_threshold_ratio:0.5
```

then the config validator fails as expected
```
<jemalloc>: Invalid combination of options hpa_hugification_threshold_ratio: 0.5 and hpa_dirty_mult: 0.39999389648437. These values should sum to >= 1.0.
```


Btw I have also noticed, that /tests/unit/hpa_background_thread is failing for us with
```
./hpa_background_thread
expect_purging:test/unit/hpa_background_thread.c:119: Failed assertion: (expect_deferred) == (observed_dirty_page) --> true != false:
test_hpa_background_thread_purges (non-reentrant): fail
expect_purging:test/unit/hpa_background_thread.c:119: Failed assertion: (expect_deferred) == (observed_dirty_page) --> true != false:
test_hpa_background_thread_enable_disable (non-reentrant): fail
```

so possibly this could be somehow related?

**Comment by Svetlitski** [comment]
@pdokoupil

The settings you have set are somewhat problematic (this is not your fault, none of this is documented anywhere). You really don't want to set `hpa_dirty_mult` to 0 and `hpa_hugification_threshold_ratio` to 1.0. By setting `hpa_dirty_mult` to 0, you're forcing the HPA to purge _any_ dirty memory immediately, so you're very unlikely to ever get actually hugified pages. Try walking that back even slightly, try `hpa_hugification_threshold_ratio:1.0,hpa_dirty_mult:0.03`, that should work just fine. One other note, `thp` and `hpa` are not really designed to work together (though I can understand your confusion here, because again none of this is actually documented because the HPA is still experiemental), so set `thp:default`. In other words, try this config:

```
MALLOC_CONF='hpa_hugification_threshold_ratio:1.0,hpa_dirty_mult:0.03,hpa:true,metadata_thp:always,thp:default'
```

File: test/unit/hpa_validate_conf.c [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+		malloc_message = null_malloc_message;
+		opt_hpa = true;
+		invalid_conf_abort = mock_invalid_conf_abort;
+		/*
+		 * We're forced to update the environment variable like this;
+		 * simply setting `opt_abort_conf = true` isn't enough, since
+		 * that might get overridden by the default compile-time malloc
+		 * conf settings.
+		 */
+		static char updated_malloc_conf[1025] =
```
interwq: Why not setting abort_conf using MALLOC_CONF in hpa_validate_conf.sh ? That one overwrites the compile time options.

src/jemalloc.c
```diff
@@ -1070,6 +1071,44 @@ obtain_malloc_conf(unsigned which_source, char buf[PATH_MAX + 1]) {
...
+		malloc_snprintf(normalization_message,
+		    sizeof(normalization_message), "<jemalloc>: Normalizing "
+		    "HPA settings to avoid pathological behavior, setting "
+		    "hpa_hugification_threshold_ratio: to %s.\n",
+		    normalized_hugification_threshold);
+	}
+	malloc_printf(
+	    "<jemalloc>: Invalid combination of options "
+	    "hpa_hugification_threshold_ratio: %s and hpa_dirty_mult: %s. "
+	    "These values should sum to >= 1.0.\n%s", hugification_threshold,
```
interwq: Should we ask for >1.0, instead of of >=? It looks the exact 1.0 case can still result in the infinite loop?

test/unit/hpa_validate_conf.c
Svetlitski: IIRC (this was over a week ago), it's because if you set `abort_conf` in the shell-script then on FreeBSD (because Jemalloc is the system allocator) in CI instances where HPA isn't supported it will fail before it ever even launches the test.

Svetlitski: We should probably modify our testing setup so that we use a different prefix for the tests than we do for the system Jemalloc to avoid this kind of thing (it would also help cut down on all the warning messages you get about `prof:true` being an invalid conf value from the system allocator picking up test configs).

interwq: I see. Yeah having a test only config would make sense. For the abort_conf here, maybe use the conf_harder hack? That overwrites everything and is process local. 
https://github.com/jemalloc/jemalloc/blob/07a2eab3ed5dd76657ee689326acd9ecaf1e2830/src/jemalloc.c#L1075

Svetlitski: Come to think of it, I think we can actually just use the default `malloc_conf` and not have to resort to `malloc_conf_2_conf_harder`, since the only thing we're trying to override is the `--with-malloc-conf` default, which has the lowest priority of the various ways to set the conf anyway.

src/jemalloc.c
Svetlitski: Yes, and I'll fix the `MALLOC_CONF` for the `hpa_background_thread` test-case accordingly.

Svetlitski: Ahhh wait, that test-case _depends_ on `hpa_dirty_mult` being 0

Svetlitski: I suppose I can just set `hpa_dirty_mult` to 0.001 or something.

interwq: Yeah I'd say fail early for any combination that's infinite-loop possible.

pdokoupil:
> @pdokoupil
> 
> The settings you have set are somewhat problematic (this is not your fault, none of this is documented anywhere). You really don't want to set `hpa_dirty_mult` to 0 and `hpa_hugification_threshold_ratio` to 1.0. By setting `hpa_dirty_mult` to 0, you're forcing the HPA to purge _any_ dirty memory immediately, so you're very unlikely to ever get actually hugified pages. Try walking that back even slightly, try `hpa_hugification_threshold_ratio:1.0,hpa_dirty_mult:0.03`, that should work just fine. One other note, `thp` and `hpa` are not really designed to work together (though I can understand your confusion here, because again none of this is actually documented because the HPA is still experiemental), so set `thp:default`. In other words, try this config:
> 
> ```
> MALLOC_CONF='hpa_hugification_threshold_ratio:1.0,hpa_dirty_mult:0.03,hpa:true,metadata_thp:always,thp:default'
> ```

@Svetlitski Thanks for helping out. We tried your config (only adding extra `confirm_conf:true` to that), but unfortunately our issue still remains the same. I tried reproducing the issue, but so far I was unsuccessful (I cannot share our code since it is proprietary as @ericm1024 mentioned in https://github.com/jemalloc/jemalloc/issues/2449). 
To narrow the issue at least a bit, I tried running our tests with GDB. We were using 16 threads that do some workload (allocations & deallocations) in parallel. I noticed that we hang once the test ends and we start cleaning up the threads. At that time, threads are invoking `pthread_join` and we usually hang (to be more specific, the jemalloc's background thread that tries to purge is hanging). All of this is very nondeterministic, sometimes we do not hang at all, sometimes the very first thread hangs. Overall, I would say it got a bit more stable after your changes and change in the config, but it is still there.

I am attaching (cleaned) backtrace for the case it helps anyhow, needles to say it is very similar to the one in https://github.com/jemalloc/jemalloc/issues/2449. 

 <details>
  <summary>Click to expand</summary>
  
  ### Backtrace
  ```cpp
#0  hpa_shard_maybe_do_deferred_work (tsdn=0x7ff500341cb8, shard=0x7fffe48131c8, forced=<optimized out>) at jemalloc/src/hpa.c:539
#1  0x00007ffff5b04844 in hpa_dalloc_batch (tsdn=0x7ff500341cb8, self=<optimized out>, list=0x7ff5003370f8, deferred_work_generated=0x7ff500337107) at jemalloc/src/hpa.c:868
#2  0x00007ffff5b2dbb6 in pai_dalloc_batch (tsdn=<optimized out>, self=0x7ff500335e80, list=0x0, deferred_work_generated=0x7fffe481315000) at jemalloc/include/jemalloc/internal/pai.h:81
#3  sec_flush_some_and_unlock (tsdn=0x7ff500341cb8, sec=0x7fffe4813150, shard=<optimized out>) at jemalloc/src/sec.c:153
#4  0x00007ffff5b2c875 in sec_shard_dalloc_and_unlock (tsdn=<optimized out>, sec=<optimized out>, shard=0x7fffe48173c0, edata=<optimized out>) at jemalloc/src/sec.c:329
#5  sec_dalloc (tsdn=0x7ff500341cb8, self=0x7fffe4813150, edata=0x7fffe4509e00, deferred_work_generated=0x7ff50033717f) at jemalloc/src/sec.c:349
#6  0x00007ffff5aba325 in je_arena_slab_dalloc (tsdn=0x7ff500341cb8, arena=0x7fffe4801100, slab=0x41) at jemalloc/src/arena.c:573
#7  0x00007ffff5b40d72 in tcache_bin_flush_impl (tsd=<optimized out>, tcache=<optimized out>, cache_bin=0x7ff5003420f0, binind=<optimized out>, ptrs=<optimized out>, nflush=197, small=<optimized out>) at jemalloc/src/tcache.c:477
#8  0x00007ffff5b3f91f in tcache_bin_flush_bottom (tsd=0x7ff500341cb8, tcache=<optimized out>, cache_bin=0x7ff5003420f0, binind=9, rem=<optimized out>, small=<optimized out>) at jemalloc/src/tcache.c:519
#9  0x00007ffff5b42638 in tcache_flush_cache (tsd=0x7ff500341cb8, tcache=0x7ff500342010) at jemalloc/src/tcache.c:535
#10 0x00007ffff5b4289f in tcache_destroy (tsd=0x7ff500341cb8, tcache=0x7ff500342010, tsd_tcache=<optimized out>) at jemalloc/src/tcache.c:809
#11 0x00007ffff5b42756 in je_tcache_cleanup (tsd=0x7ff500341cb8) at jemalloc/src/tcache.c:852
#12 0x00007ffff5b4651a in tsd_do_data_cleanup (tsd=<optimized out>) at jemalloc/src/tsd.c:398
#13 je_tsd_cleanup (arg=0x7ff500341cb8) at jemalloc/src/tsd.c:424
#14 0x00007fffe8b87408 in __nptl_deallocate_tsd () at pthread_create.c:300
#15 0x00007fffe8b8881b in __nptl_deallocate_tsd () at ../sysdeps/nptl/futex-internal.h:200
#16 start_thread (arg=0x7ff500400700) at pthread_create.c:473
#17 0x00007fffe84d8a3f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
  ```

  Note that lines may not match exactly in hpa.c as I added some temporary debug prints.
</details>

Any ideas what else could have gone wrong?

pdokoupil:
Followup on my previous comment:

After some more digging-in, we tried running with LLVM's ThreadSanitizer (TSAN) and got two <ins>possible</ins> races reported. I am attaching the TSAN reports below. I am not that familiar with Jemalloc's code to asses to which extent this should be related to our infinite loop issue, but removing the suspicious `je_free` from our code seem to solve the issue (at the cost of mem leaks, of course), so perhaps this could be related and helpful for further investigation.

<details>
  <summary>Click to expand TSAN reports</summary>
  
  ### Report 1
```cpp
WARNING: ThreadSanitizer: data race (pid=28)
  Read of size 8 at 0x7f2f20a16a80 by thread T8 (mutexes: write M0):
    #0 tcache_bin_flush_match jemalloc/src/tcache.c (libjemalloc.so+0xc00c0)
    #1 tcache_bin_flush_impl jemalloc/src/tcache.c:434:9 (libjemalloc.so+0xc00c0)
    #2 tcache_bin_flush_bottom jemalloc/src/tcache.c:519:2 (libjemalloc.so+0xc00c0)
    #3 je_tcache_bin_flush_small jemalloc/src/tcache.c:529:2 (libjemalloc.so+0xc00c0)
    #4 tcache_dalloc_small jemalloc/include/jemalloc/internal/tcache_inlines.h:162:3 (libjemalloc.so+0x887cc)
    #5 arena_dalloc jemalloc/include/jemalloc/internal/arena_inlines_b.h:399:3 (libjemalloc.so+0x887cc)
    #6 idalloctm jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h:158:2 (libjemalloc.so+0x887cc)
    #7 ifree jemalloc/src/jemalloc.c:2906:3 (libjemalloc.so+0x887cc)
    #8 je_free_default jemalloc/src/jemalloc.c:3010:4 (libjemalloc.so+0x887cc)
    #9 je_free_impl jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h:601:17 (libjemalloc.so+0x89798)
    #10 je_free jemalloc/src/jemalloc.c:3028:2 (libjemalloc.so+0x89798)
    #11 ...
    #12 <our stuff omitted>
    

  Previous write of size 8 at 0x7f2f20a16a80 by thread T15 (mutexes: write M1):
    #0 edata_nfree_inc jemalloc/include/jemalloc/internal/edata.h:498:16 (libjemalloc.so+0xc0131)
    #1 arena_dalloc_bin_locked_step jemalloc/include/jemalloc/internal/arena_inlines_b.h:572:2 (libjemalloc.so+0xc0131)
    #2 tcache_bin_flush_impl jemalloc/src/tcache.c:448:9 (libjemalloc.so+0xc0131)
    #3 tcache_bin_flush_bottom jemalloc/src/tcache.c:519:2 (libjemalloc.so+0xc0131)
    #4 je_tcache_bin_flush_small jemalloc/src/tcache.c:529:2 (libjemalloc.so+0xc0131)
    #5 tcache_dalloc_small jemalloc/include/jemalloc/internal/tcache_inlines.h:162:3 (libjemalloc.so+0x887cc)
    #6 arena_dalloc jemalloc/include/jemalloc/internal/arena_inlines_b.h:399:3 (libjemalloc.so+0x887cc)
    #7 idalloctm jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h:158:2 (libjemalloc.so+0x887cc)
    #8 ifree jemalloc/src/jemalloc.c:2906:3 (libjemalloc.so+0x887cc)
    #9 je_free_default jemalloc/src/jemalloc.c:3010:4 (libjemalloc.so+0x887cc)
    #10 je_free_impl jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h:601:17 (libjemalloc.so+0x89798)
    #11 je_free jemalloc/src/jemalloc.c:3028:2 (libjemalloc.so+0x89798)
    #12 ...
    #13 <our stuff omitted>

  Mutex M0 (0x7f2f3fc14d68) created at:
    #0 pthread_mutex_init /llvm-project-llvmorg-15.0.1/compiler-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:1316:3 (<omitted>>+0x262bcf)
    #1 je_malloc_mutex_init jemalloc/src/mutex.c:171:6 (libjemalloc.so+0x9d020)
    #2 je_bin_init jemalloc/src/bin.c:43:6 (libjemalloc.so+0x36bf8)
    #3 je_arena_new jemalloc/src/arena.c:1686:14 (libjemalloc.so+0x309b7)
    #4 arena_init_locked jemalloc/src/jemalloc.c:419:10 (libjemalloc.so+0x84dd6)
    #5 je_arena_init jemalloc/src/jemalloc.c:447:10 (libjemalloc.so+0x84dd6)
    #6 malloc_init_hard_a0_locked jemalloc/src/jemalloc.c:1945:6 (libjemalloc.so+0x93ad4)
    #7 malloc_init_hard jemalloc/src/jemalloc.c:2189:6 (libjemalloc.so+0x981f5)
    #8 malloc_init jemalloc/src/jemalloc.c:302:41 (libjemalloc.so+0x932a3)
    #9 jemalloc_constructor jemalloc/src/jemalloc.c:4243:2 (libjemalloc.so+0x932a3)

  Mutex M1 (0x7f2f20a14ce8) created at:
    #0 pthread_mutex_init /llvm-project-llvmorg-15.0.1/compiler-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:1316:3 (<omitted>+0x262bcf)
    #1 je_malloc_mutex_init jemalloc/src/mutex.c:171:6 (libjemalloc.so+0x9d020)
    #2 je_bin_init jemalloc/src/bin.c:43:6 (libjemalloc.so+0x36bf8)
    #3 je_arena_new jemalloc/src/arena.c:1686:14 (libjemalloc.so+0x309b7)
    #4 arena_init_locked jemalloc/src/jemalloc.c:419:10 (libjemalloc.so+0x854d4)
    #5 je_arena_choose_hard jemalloc/src/jemalloc.c:586:13 (libjemalloc.so+0x854d4)
    #6 arena_choose_impl jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_b.h:49:9 (libjemalloc.so+0xc5abc)
    #7 arena_choose jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_b.h:91:9 (libjemalloc.so+0xc2b27)
    #8 je_tsd_tcache_data_init jemalloc/src/tcache.c:740:11 (libjemalloc.so+0xc2b27)
    #9 je_tsd_tcache_enabled_data_init jemalloc/src/tcache.c:644:3 (libjemalloc.so+0xc2741)
    #10 je_tsd_fetch_slow jemalloc/src/tsd.c (libjemalloc.so+0xc75c5)
    #11 tsd_fetch_impl jemalloc/include/jemalloc/internal/tsd.h:129:10 (libjemalloc.so+0x91e63)
    #12 tsd_fetch jemalloc/include/jemalloc/internal/tsd.h:155:9 (libjemalloc.so+0x91e63)
    #13 je_mallctl jemalloc/src/jemalloc.c:3918:8 (libjemalloc.so+0x91e63)
    #14 <our stuff omitted, getting jemalloc option via mallctl>
    #15 ...
    #16 <our stuff omitted>

SUMMARY: ThreadSanitizer: data race jemalloc/src/tcache.c in tcache_bin_flush_match
  ```

  ### Report 2
```cpp
WARNING: ThreadSanitizer: data race (pid=28)
  Read of size 8 at 0x7f79b589b480 by thread T2:
    #0 edata_arena_ind_get jemalloc/include/jemalloc/internal/edata.h:260:42 (libjemalloc.so+0xbfeae)
    #1 tcache_bin_flush_impl jemalloc/src/tcache.c:350:28 (libjemalloc.so+0xbfeae)
    #2 tcache_bin_flush_bottom jemalloc/src/tcache.c:519:2 (libjemalloc.so+0xbfeae)
    #3 je_tcache_bin_flush_small jemalloc/src/tcache.c:529:2 (libjemalloc.so+0xbfeae)
    #4 tcache_dalloc_small jemalloc/include/jemalloc/internal/tcache_inlines.h:162:3 (libjemalloc.so+0x887cc)
    #5 arena_dalloc jemalloc/include/jemalloc/internal/arena_inlines_b.h:399:3 (libjemalloc.so+0x887cc)
    #6 idalloctm jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h:158:2 (libjemalloc.so+0x887cc)
    #7 ifree jemalloc/src/jemalloc.c:2906:3 (libjemalloc.so+0x887cc)
    #8 je_free_default jemalloc/src/jemalloc.c:3010:4 (libjemalloc.so+0x887cc)
    #9 je_free_impl jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h:601:17 (libjemalloc.so+0x89798)
    #10 je_free jemalloc/src/jemalloc.c:3028:2 (libjemalloc.so+0x89798)
    #11 ...
    #12 <our stuff omitted>

  Previous write of size 8 at 0x7f79b589b480 by thread T7 (mutexes: write M0):
    #0 edata_nfree_inc jemalloc/include/jemalloc/internal/edata.h:498:16 (libjemalloc.so+0xc0131)
    #1 arena_dalloc_bin_locked_step jemalloc/include/jemalloc/internal/arena_inlines_b.h:572:2 (libjemalloc.so+0xc0131)
    #2 tcache_bin_flush_impl jemalloc/src/tcache.c:448:9 (libjemalloc.so+0xc0131)
    #3 tcache_bin_flush_bottom jemalloc/src/tcache.c:519:2 (libjemalloc.so+0xc0131)
    #4 je_tcache_bin_flush_small jemalloc/src/tcache.c:529:2 (libjemalloc.so+0xc0131)
    #5 tcache_dalloc_small jemalloc/include/jemalloc/internal/tcache_inlines.h:162:3 (libjemalloc.so+0x887cc)
    #6 arena_dalloc jemalloc/include/jemalloc/internal/arena_inlines_b.h:399:3 (libjemalloc.so+0x887cc)
    #7 idalloctm jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h:158:2 (libjemalloc.so+0x887cc)
    #8 ifree jemalloc/src/jemalloc.c:2906:3 (libjemalloc.so+0x887cc)
    #9 je_free_default jemalloc/src/jemalloc.c:3010:4 (libjemalloc.so+0x887cc)
    #10 je_free_impl jemalloc/include/jemalloc/internal/jemalloc_internal_inlines_c.h:601:17 (libjemalloc.so+0x89798)
    #11 je_free jemalloc/src/jemalloc.c:3028:2 (libjemalloc.so+0x89798)
    #12 ...
    #13 <our stuff omitted>

  Mutex M0 (0x7f79b5c14d68) created at:
    #0 pthread_mutex_init /llvm-project-llvmorg-15.0.1/compiler-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:1316:3 (<omitted>+0x262bcf)
    #1 je_malloc_mutex_init jemalloc/src/mutex.c:171:6 (libjemalloc.so+0x9d020)
    #2 je_bin_init jemalloc/src/bin.c:43:6 (libjemalloc.so+0x36bf8)
    #3 je_arena_new jemalloc/src/arena.c:1686:14 (libjemalloc.so+0x309b7)
    #4 arena_init_locked jemalloc/src/jemalloc.c:419:10 (libjemalloc.so+0x84dd6)
    #5 je_arena_init jemalloc/src/jemalloc.c:447:10 (libjemalloc.so+0x84dd6)
    #6 malloc_init_hard_a0_locked jemalloc/src/jemalloc.c:1945:6 (libjemalloc.so+0x93ad4)
    #7 malloc_init_hard jemalloc/src/jemalloc.c:2189:6 (libjemalloc.so+0x981f5)
    #8 malloc_init jemalloc/src/jemalloc.c:302:41 (libjemalloc.so+0x932a3)
    #9 jemalloc_constructor jemalloc/src/jemalloc.c:4243:2 (libjemalloc.so+0x932a3)

SUMMARY: ThreadSanitizer: data race jemalloc/include/jemalloc/internal/edata.h:260:42 in edata_arena_ind_get
  ```
</details>

File: src/jemalloc.c [hunk 4]
```diff
@@ -1081,6 +1082,44 @@ obtain_malloc_conf(unsigned which_source, char buf[PATH_MAX + 1]) {
 	return ret;
 }
 
+static void
+validate_hpa_settings(void) {
+	size_t hpa_threshold = fxp_mul_frac(HUGEPAGE, opt_hpa_opts.dirty_mult) +
```
interwq, line 1090: nit: `if (!hpa_supported() || !opt_hpa) return;` at the very beginning

This means you'd also have to set hpa:true in the new test (should be fine?).

src/jemalloc.c
```diff
@@ -1081,6 +1082,44 @@ obtain_malloc_conf(unsigned which_source, char buf[PATH_MAX + 1]) {
...
+		    (opt_hpa_opts.hugification_threshold >> LG_PAGE)),
+		    FXP_INIT_INT(HUGEPAGE_PAGES)),
+		    normalized_hugification_threshold);
+		malloc_snprintf(normalization_message,
+		    sizeof(normalization_message), "<jemalloc>: Normalizing "
+		    "HPA settings to avoid pathological behavior, setting "
+		    "hpa_hugification_threshold_ratio: to %s.\n",
+		    normalized_hugification_threshold);
+	}
+	malloc_printf(
```
interwq, line 1118: Should we swap the order here, i.e. do this before the branch above? That way the !abort_conf case ends up with "<jemalloc>: Normalizing ..." which is probably easier to understand.

interwq:
@pdokoupil thanks for sharing the TSAN reports. I went over them and they appear to be false positives, which isn't surprising since we are indeed reading shared states w/o exclusive locking -- but in this case it's safe because we are holding alive pointers managed by that piece of metadata, so it's guaranteed that the state cannot be changed.

I'd suggest trying ASAN (w/o jemalloc because ASAN has its own allocator) and see if anything there.

src/jemalloc.c
Svetlitski: It is printed in that order, the call in the `if` branch is a `malloc_snprintf`.

src/jemalloc.c
Svetlitski: > This means you'd also have to set hpa:true in the new test (should be fine?).

We set `opt_hpa = true` directly before the test starts. See the comment in the test case (which was really written by @davidtgoldblatt, I just copied it from the `hpa_background_thread` test case) for an explanation of why.

interwq: I'd say also add the `opt_hpa` check in the first branch, after the `!hpa_supported()` condition.

Svetlitski: My apologies, I missed that in your first comment.

pdokoupil:
> ```c++
> edata_nfree_inc jemalloc
> ```

@interwq Thanks for your feedback and having look at the TSAN output. We tried ASAN w/o jemalloc, that works without issues. I also tried MSAN w/ jemalloc and that one fails immediately with:

```cpp
==28==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7f4aa15059c4 in je_pages_boot jemalloc/src/pages.c:820:6
    #1 0x7f4aa14d656e in malloc_init_hard_a0_locked jemalloc/src/jemalloc.c:1894:6
    #2 0x7f4aa14e59f7 in malloc_init_hard jemalloc/src/jemalloc.c:2189:6
```

Needles to say this could easily be either non-related or just another false positive..

interwq:
@pdokoupil yeah that's also a false positive, because that's the boot logic checking if MADV_DONTNEED results the pages being zeroed. So it's intentionally reading after madvise, and w/o initialization.

Do you have malloc_stats you can share? Also if you turn off HPA, any issue happens in that case?

pdokoupil:
> @pdokoupil yeah that's also a false positive, because that's the boot logic checking if MADV_DONTNEED results the pages being zeroed. So it's intentionally reading after madvise, and w/o initialization.
> 


> Do you have malloc_stats you can share? Also if you turn off HPA, any issue happens in that case?

@interwq Sure, here are the jemalloc stats (see the attachment [jemalloc_stats.txt](https://github.com/jemalloc/jemalloc/files/12296572/jemalloc_stats.txt)), I managed to reproduce it with fewer threads. To answer your question - it works fine without HPA. Let me add to this - I ran our tests a couple of times, until we ended in the "infinite loop", after 500 iterations in that loop I did `break` and printed stats soon after that (basically in test cleanup).

interwq:
@pdokoupil thanks for the stats. Given that the hpa_dirty_mult is low and hpa_hugification_threshold is high, I'm guessing that [hpa_should_purge](https://github.com/jemalloc/jemalloc/blob/dev/src/hpa.c#L537) is returning true, but [hpa_try_purge]( https://github.com/jemalloc/jemalloc/blob/254c4847e8ac263d24720aa93c2c7d410f55a239/src/hpa.c#L380) keeps failing -- could you break point when it loops, and check the context along those two operations? i.e. if should_purge is returning true correctly, or if try_purging is failing incorrectly.

pdokoupil:
> @pdokoupil thanks for the stats. Given that the hpa_dirty_mult is low and hpa_hugification_threshold is high, I'm guessing that [hpa_should_purge](https://github.com/jemalloc/jemalloc/blob/dev/src/hpa.c#L537) is returning true, but [hpa_try_purge](https://github.com/jemalloc/jemalloc/blob/254c4847e8ac263d24720aa93c2c7d410f55a239/src/hpa.c#L380) keeps failing -- could you break point when it loops, and check the context along those two operations? i.e. if should_purge is returning true correctly, or if try_purging is failing incorrectly.

@interwq Yes, it is exactly as you mention - `hpa_should_purge` returns true, while `hpa_try_purge` keeps failing. It is stuck in this inner loop, during the very first iteration of the `do .. while` that is above. `nops` is usually 1, but highest I have seen was 7. 
It seems that `hpa_adjusted_ndirty(tsdn, shard)` is returning `X` and `hpa_ndirty_max(tsdn, shard)` returns `Y` such that `X > Y` thus we leave `hpa_should_purge` returning `true` in the first `if` branch. More likely than an issue inside `hpa_should_purge`, I think it could be an issue inside `hpa_try_purge` where we return `false` at the very top since `psset_pick_purge(&shard->psset)` returns `NULL`. Looking into `psset_pick_purge` there is a call to `ssize_t ind_ssz = fb_fls(psset->purge_bitmap, PSSET_NPURGE_LISTS, PSSET_NPURGE_LISTS - 1);` that results in `ind_szz < 0`.

Btw adding some extra `malloc_printf` calls made the test seemingly more stable (could be coincidence, of course), so I am still thinking, if there could be any data race going on. However, looking into accesses on `psset->purge_bitmap` (without any deep understanding of the code) it seemed it is properly protected by `shard->mtx`, except for calls that originate from `psset_init` and `hpa_shard_destroy` which I guess is fine.

interwq:
@pdokoupil in case you haven't tried, building jemalloc with `--enable-debug` can help catching things earlier, though it slows down quite a bit so could affect timing.

If that doesn't catch anything new, can you help to grab the context when the loop happens, e.g. the notable values from hpa_should_purge and hpa_try_purge resulted them to return true / false. That should help us to better understand if it's from a corruption somewhere, or a genuine HPA bug.

pdokoupil:
> @pdokoupil in case you haven't tried, building jemalloc with `--enable-debug` can help catching things earlier, though it slows down quite a bit so could affect timing.
> 
> If that doesn't catch anything new, can you help to grab the context when the loop happens, e.g. the notable values from hpa_should_purge and hpa_try_purge resulted them to return true / false. That should help us to better understand if it's from a corruption somewhere, or a genuine HPA bug.

@interwq Thanks for reply, we already tried that, I disabled debug built once I started playing with TSAN since there were some more false positives in debug-only safety checks. Regarding the values: as I said it is `hpa_should_purge` that returns true since `adjusted_ndirty > ndirty_max` is true at the very beginning of the function and we return `true`. The exact values of `adjusted_ndirty` and `ndirty_max` fluctuate slightly, one of the examples I have seen was `adjusted_ndirty=155` and `ndirty_max=126` with  `psset_nactive=4205` and multiplier of 0.03 (but I have seen different values as well, e.g. 185 & 125, ...) The 155 was calculated as `psset_ndirty(&shard->psset) - shard->npending_purge` where the first part was, e.g. 251 and second 96. 

Then, when we get to `hpa_try_purge`, there is a call to `psset_pick_purge` which returns NULL and thus whole `hpa_try_purge` fails returning false. Investigating `psset_pick_purge` more shows that `fb_fls(psset->purge_bitmap, PSSET_NPURGE_LISTS,
	    PSSET_NPURGE_LISTS - 1);` returns -1. Note that `PSSET_NPURGE_LISTS = 128`. The `psset->purge_bitmap` seem to contain just to 64b unsigned integers, both set to 0.

pdokoupil:
Update: just trying to remove temporary unlock (and corresponding re-lock) during actual purge that was done in `hpa_try_purge` (around this comment: `/* Actually do the purging, now that the lock is dropped. */`) seem to solve the issue, or at least makes  it much more stable. I will do more runs of our different workloads tomorrow. Could someone please verify if this change is justified and valid? The unlocking seemed to be here on some purpose, I guess splitting the critical section to gain some extra performance.

pdokoupil:
@interwq Not sure if this can still be a coincidence, but I can now confirm that after applying the changes as described in my previous comment, we were not yet able to reproduce the infinite loop when running our tests and workloads.

interwq:
@pdokoupil sorry I have been occupied by a few things. Will try getting back to this next week.

PR #2485, opened 2023-07-18
Fix all optimization-inhibiting integer-to-pointer casts

Svetlitski:
Following from PR #2481, we replace all integer-to-pointer casts [which hide pointer provenance information (and thus inhibit optimizations)](https://clang.llvm.org/extra/clang-tidy/checks/performance/no-int-to-ptr.html) with equivalent operations that preserve this information. I have enabled the corresponding clang-tidy check in our static analysis CI so that we do not get bitten by this again in the future.

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_decls.h
+++ b/include/jemalloc/internal/jemalloc_internal_decls.h
@@ -105,4 +105,21 @@ isblank(int c) {
 #  undef small
 #endif
 
+/*
+ * Oftentimes we'd like to perform some kind of arithmetic to obtain
+ * a pointer from another pointer but with some offset or mask applied.
+ * Naively you would accomplish this by casting the source pointer to
+ * `uintptr_t`, performing all of the relevant arithmetic, and then casting
+ * the result to the desired pointer type. However, this has the unfortunate
+ * side-effect of concealing pointer provenance, hiding useful information for
+ * optimization from the compiler (see here for details:
+ * https://clang.llvm.org/extra/clang-tidy/checks/performance/no-int-to-ptr.html
+ * )
+ * Instead what one should do is cast the source pointer to `char *` and perform
+ * the equivalent arithmetic (since `char` of course represents one byte). But
+ * because `char *` has the semantic meaning of "string", we define this typedef
+ * simply to make it clearer where we are performing such pointer arithmetic.
+ */
+typedef char byte_t;
+
 #endif /* JEMALLOC_INTERNAL_H */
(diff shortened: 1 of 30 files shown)


include/jemalloc/internal/jemalloc_internal_decls.h
```diff
@@ -105,4 +105,6 @@ isblank(int c) {
 #  undef small
 #endif
 
+typedef char byte_t;
```
interwq, line 123: nit: add some comments about why using this instead of uintptr_t etc

include/jemalloc/internal/jemalloc_internal_types.h
```diff
@@ -99,7 +99,8 @@ typedef enum malloc_init_e malloc_init_t;
 
 /* Return the nearest aligned address at or below a. */
 #define ALIGNMENT_ADDR2BASE(a, alignment)				\
-	((void *)((uintptr_t)(a) & ((~(alignment)) + 1)))
+	((void *)(((byte_t *)(a)) - (((uintptr_t)(a)) -			\
+	    ((uintptr_t)(a) & ((~(alignment)) + 1)))))
```
interwq, line 103: I feel this can use some comments too (which should include the effective / original calculation). It's pretty counter intuitive why this would allow better codegen. Maybe also mention `ALIGNMENT_ADDR2CEILING` for their similarities.

src/background_thread.c
```diff
@@ -368,6 +368,7 @@ check_background_thread_creation(tsd_t *tsd,
 
 		pre_reentrancy(tsd, NULL);
 		int err = background_thread_create_signals_masked(&info->thread,
+			/* NOLINTNEXTLINE(performance-no-int-to-ptr) */
 		    NULL, background_thread_entry, (void *)(uintptr_t)i);
```
interwq, line 372: Given how popular this pattern is, maybe add a wrapper like `void *int_to_ptr_nolint()` and put the comment inside?

Svetlitski: If you want, sure. But I personally feel like the ugliness of this is a feature, not a bug. The lint (and the act of suppressing the lint) _should_ make you feel icky: smuggling random integers in pointers is something we should try to avoid when possible.

Svetlitski: Never mind, I think you're right and this can be improved, I will address this.

interwq: Sorry forgot to follow up here. I agree there should be some intentional ugliness, but perhaps in the form of a extra verbose function name, then within the wrapper function we can add detailed comments about why the hassle. Another factor is, some functions have a long list of args, and a wrapper would apply to the exact argument instead of the line.

Svetlitski: How do you feel about meeting in the middle (see latest force-push)? I've extracted the vast majority of offending casts into macros (instead of using magic numbers), which dramatically cuts down the number of `NOLINTNEXTLINE` comments we need.

interwq: This looks great. I didn't realize adding `NOLINTNEXTLINE` above the `#define`s would work for all references, but it certainly makes sense. Thanks for chasing down these magic numbers too!

PR #2486, opened 2023-07-19
Stop over-reporting memory usage from sampled small allocations

Svetlitski:
@interwq noticed [while reviewing an earlier PR](https://github.com/jemalloc/jemalloc/pull/2478#discussion_r1256217261) that I missed modifying this statistics accounting in line with the rest of the changes from #2459. This is now fixed, such that sampled small allocations increment the `.nmalloc`/`.ndalloc` of their effective bin size instead of over-reporting memory usage by attributing all such allocations to `SC_LARGE_MINCLASS`.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -293,34 +293,48 @@ arena_slab_reg_alloc_batch(edata_t *slab, const bin_info_t *bin_info,
 
 static void
 arena_large_malloc_stats_update(tsdn_t *tsdn, arena_t *arena, size_t usize) {
-	szind_t index, hindex;
-
 	cassert(config_stats);
 
+	szind_t index = sz_size2index(usize);
+	/* This only occurs when we have a sampled small allocation */
 	if (usize < SC_LARGE_MINCLASS) {
-		usize = SC_LARGE_MINCLASS;
+		assert(index < SC_NBINS);
+		assert(usize >= PAGE && usize % PAGE == 0);
+		bin_t *bin = arena_get_bin(arena, index, /* binshard */ 0);
+		malloc_mutex_lock(tsdn, &bin->lock);
+		bin->stats.nmalloc++;
+		malloc_mutex_unlock(tsdn, &bin->lock);
+	} else {
+		assert(index >= SC_NBINS);
+		szind_t hindex = index - SC_NBINS;
+		LOCKEDINT_MTX_LOCK(tsdn, arena->stats.mtx);
+		locked_inc_u64(tsdn, LOCKEDINT_MTX(arena->stats.mtx),
+			&arena->stats.lstats[hindex].nmalloc, 1);
+		LOCKEDINT_MTX_UNLOCK(tsdn, arena->stats.mtx);
 	}
-	index = sz_size2index(usize);
-	hindex = (index >= SC_NBINS) ? index - SC_NBINS : 0;
-
-	locked_inc_u64(tsdn, LOCKEDINT_MTX(arena->stats.mtx),
-	    &arena->stats.lstats[hindex].nmalloc, 1);
 }
 
 static void
 arena_large_dalloc_stats_update(tsdn_t *tsdn, arena_t *arena, size_t usize) {
-	szind_t index, hindex;
-
 	cassert(config_stats);
 
+	szind_t index = sz_size2index(usize);
+	/* This only occurs when we have a sampled small allocation */
 	if (usize < SC_LARGE_MINCLASS) {
-		usize = SC_LARGE_MINCLASS;
+		assert(index < SC_NBINS);
+		assert(usize >= PAGE && usize % PAGE == 0);
+		bin_t *bin = arena_get_bin(arena, index, /* binshard */ 0);
+		malloc_mutex_lock(tsdn, &bin->lock);
+		bin->stats.ndalloc++;
+		malloc_mutex_unlock(tsdn, &bin->lock);
+	} else {
+		assert(index >= SC_NBINS);
+		szind_t hindex = index - SC_NBINS;
+		LOCKEDINT_MTX_LOCK(tsdn, arena->stats.mtx);
+		locked_inc_u64(tsdn, LOCKEDINT_MTX(arena->stats.mtx),
+			&arena->stats.lstats[hindex].ndalloc, 1);
+		LOCKEDINT_MTX_UNLOCK(tsdn, arena->stats.mtx);
 	}
-	index = sz_size2index(usize);
-	hindex = (index >= SC_NBINS) ? index - SC_NBINS : 0;
-
-	locked_inc_u64(tsdn, LOCKEDINT_MTX(arena->stats.mtx),
-	    &arena->stats.lstats[hindex].ndalloc, 1);
 }
 
 static void
@@ -344,9 +358,7 @@ arena_extent_alloc_large(tsdn_t *tsdn, arena_t *arena, size_t usize,
 
 	if (edata != NULL) {
 		if (config_stats) {
-			LOCKEDINT_MTX_LOCK(tsdn, arena->stats.mtx);
 			arena_large_malloc_stats_update(tsdn, arena, usize);
-			LOCKEDINT_MTX_UNLOCK(tsdn, arena->stats.mtx);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Svetlitski:
I haven't been able to reproduce the error from CI locally yet, still trying.

interwq:
@Svetlitski the lock order issue is real --
<jemalloc>: Lock rank order reversal: arena_stats(4096) bin(4096)

So this is the stats mutex (needed when 8B atomic is not supported):
https://github.com/jemalloc/jemalloc/blob/cdb2c0e02fc303fd56aa525ef63eb71136e62b2d/src/arena.c#L346

The rank is there to make sure we have no deadlock, e.g. one thread lock bin, then lock arena_stats, the other thread doing the reverse.

So you would want to make sure the bin lock path is not under the arena_stats mutex.

src/arena.c
```diff
@@ -292,34 +292,50 @@ arena_slab_reg_alloc_batch(edata_t *slab, const bin_info_t *bin_info,
...
+	szind_t index = sz_size2index(usize);
 	if (usize < SC_LARGE_MINCLASS) {
-		usize = SC_LARGE_MINCLASS;
+		assert(index < SC_NBINS);
+		bin_t *bin = &arena->bins[index];
+		malloc_mutex_lock(tsdn, &bin->lock);
+		bin->stats.nmalloc++;
+		malloc_mutex_unlock(tsdn, &bin->lock);
+	} else {
+		szind_t hindex = (index >= SC_NBINS) ? index - SC_NBINS : 0;
```
interwq: nit: this can be just `assert(index >= SC_NBINS); hindex = index - SC_NBINS;`. Same for the large_dalloc_stats path.

src/arena.c
```diff
@@ -372,9 +384,7 @@ arena_extent_ralloc_large_shrink(tsdn_t *tsdn, arena_t *arena, edata_t *edata,
 	size_t usize = edata_usize_get(edata);
 
 	if (config_stats) {
-		LOCKEDINT_MTX_LOCK(tsdn, arena->stats.mtx);
```
interwq, line 376: One effect of moving the locking downstream, is that the ralloc now requires two lock-unlock pairs instead of one (I think this was why the locks were placed outside). Since this only applies to the realloc + no-64bit-atomics case, I don't think it's really a problem.

src/arena.c
Svetlitski: > Same for the large_dalloc_stats path

No, this isn't correct, it doesn't handle the edge-case correctly where the `bumped_usize` pushes you into `SC_LARGE_MINCLASS` but `usize` is less than `SC_LARGE_MINCLASS`.

src/arena.c
```diff
@@ -292,34 +292,51 @@ arena_slab_reg_alloc_batch(edata_t *slab, const bin_info_t *bin_info,
...
-
 	cassert(config_stats);
 
-	if (usize < SC_LARGE_MINCLASS) {
-		usize = SC_LARGE_MINCLASS;
+	/*
+	 * The key difference between this function and
+	 * `arena_large_malloc_stats_update` above is that here `usize` does
+	 * _not_ already include size promotion for sampled small allocations.
+	 */
```
interwq: Don't we already restored the usize by this point? It's done here: https://github.com/jemalloc/jemalloc/blob/dev/src/arena.c#L1269

which is before the actual dalloc logic

Svetlitski: Exactly, `usize` is no longer bumped here, but we want to attribute the stats to the size-class which was actually used to service this allocation, which _was_ bumped at the time.

interwq: Not following. I meant demote already happened before the dalloc, so the extent usize is already the bumped page sizes. It looks we can assert(usize % PAGE == 0) here.

Svetlitski: Okay, I see the real issue now. I originally assumed I had gotten something wrong in general due to unit-test failures. But I think the unit-test failure is actually an edge-case. Namely, in the `arena_reset_prof` test, the deallocation that gets initiated [on this line](https://github.com/jemalloc/jemalloc/blob/3e82f357bb218194df5ba1acee39cd6a7d6fe6f6/src/arena.c#L700) is tripping your proposed assertion.

interwq: Oh in that case, it doesn't do `arena_dalloc_promoted`, is that right?

Svetlitski: Ah actually I had just made a mistake. The other feedback I left to your response above needed to be incorporated (so we use `bumped_usize` when appropriate), and I was also missing parentheses around the assignment of `bumped_usize` so it was getting set to 0 or 1 instead of the actual value. Fixed, everything is passing now.

src/arena.c
Svetlitski: Fixed.

src/arena.c
interwq: I think you probably caught an issue with the reset operation. Can you check if `arena_dalloc_promoted` is missed, if prof is enabled while doing reset? https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/arena_inlines_b.h#L302

Svetlitski: Stepping through in a debugger reveals that `arena_prof_promote` is called many times, but `arena_prof_demote` is called many fewer. I can produce this on `dev` without the changes from this PR. So you're right, this _is_ an issue with reset. I'll look into solving it now.

Svetlitski: Opened PR #2496 to fix this.

Svetlitski: Rebased now that #2496 has been merged. Eliminated the unnecessary bumping logic here now that has been fixed.

src/arena.c
```diff
@@ -293,34 +293,44 @@ arena_slab_reg_alloc_batch(edata_t *slab, const bin_info_t *bin_info,
...
 arena_large_dalloc_stats_update(tsdn_t *tsdn, arena_t *arena, size_t usize) {
-	szind_t index, hindex;
-
 	cassert(config_stats);
 
+	szind_t index = sz_size2index(usize);
 	if (usize < SC_LARGE_MINCLASS) {
-		usize = SC_LARGE_MINCLASS;
+		assert(index < SC_NBINS);
+		bin_t *bin = &arena->bins[index];
```
interwq: We'll need to use `arena_get_bin` like here (with shard = 0):

https://github.com/jemalloc/jemalloc/blob/ea5b7bea3144cd26a63510016d778eab3ca58822/src/arena.c#L1391C1-L1391C1

That being said, it's better to rename arena_t.bins to something like `bins_use_arena_get_bin` (I can do that in a separate PR) since otherwise it's easy to miss the shard index.

src/arena.c
```diff
@@ -293,34 +293,44 @@ arena_slab_reg_alloc_batch(edata_t *slab, const bin_info_t *bin_info,
...
 static void
 arena_large_dalloc_stats_update(tsdn_t *tsdn, arena_t *arena, size_t usize) {
-	szind_t index, hindex;
-
 	cassert(config_stats);
 
+	szind_t index = sz_size2index(usize);
 	if (usize < SC_LARGE_MINCLASS) {
-		usize = SC_LARGE_MINCLASS;
+		assert(index < SC_NBINS);
```
interwq, line 324: nit: add comment saying happens with only small sampled allocations

edit: and assert that the usize is multiple of PAGE.

src/arena.c
Svetlitski: > That being said, it's better to rename arena_t.bins to something like bins_use_arena_get_bin

We can do even better. I can add the appropriate autoconf detection so that we can use [the deprecated attribute](https://clang.llvm.org/docs/AttributeReference.html#deprecated). If we declare `arena.bins` like this:
```c
struct arena_s {
    ...
    bin_t __attribute__((deprecated("Do not access this field directly, use `arena_get_bin` instead"))) bins[0];
};
```

Then when you compile, anywhere you use the field directly (and don't explicitly suppress the warning) you get a compiler warning message like this:
```
src/arena.c:303:24: warning: 'bins' is deprecated: Do not access this field directly, use `arena_get_bin` instead [-Wdeprecated-declarations]
```

And because it gets emitted as a diagnostic, this means static analysis will flag it!

interwq: The attribute sounds good. Maybe still rename bins to all_bins, to make it different from other places like in tcache->bins which can be indexed directly.

PR #2496, opened 2023-07-26
Ensured sampled allocations are properly deallocated during `arena_reset`

Svetlitski:
Discovered in the course of code-review for #2486. Sampled allocations were not being demoted before being deallocated during an `arena_reset` operation.

I confirmed that this makes the number of `arena_prof_demote` and `arena_prof_promote` calls match in the `arena_reset_prof` test-case by stepping through in a debugger, but it would be nice to somehow add a regression test for this (without resorting to introducing new atomics to count calls for these functions or something like that).

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -660,6 +660,76 @@ arena_bin_reset(tsd_t *tsd, arena_t *arena, bin_t *bin) {
 }
 
 void
+arena_prof_promote(tsdn_t *tsdn, void *ptr, size_t usize, size_t bumped_usize) {
+	cassert(config_prof);
+	assert(ptr != NULL);
+	assert(isalloc(tsdn, ptr) == bumped_usize);
+	assert(sz_can_use_slab(usize));
+
+	if (config_opt_safety_checks) {
+		safety_check_set_redzone(ptr, usize, bumped_usize);
+	}
+
+	edata_t *edata = emap_edata_lookup(tsdn, &arena_emap_global, ptr);
+
+	szind_t szind = sz_size2index(usize);
+	edata_szind_set(edata, szind);
+	emap_remap(tsdn, &arena_emap_global, edata, szind, /* slab */ false);
+
+	assert(isalloc(tsdn, ptr) == usize);
+}
+
+static size_t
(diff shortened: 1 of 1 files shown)


src/arena.c
```diff
@@ -697,7 +767,13 @@ arena_reset(tsd_t *tsd, arena_t *arena) {
 		if (config_prof && opt_prof) {
 			prof_free(tsd, ptr, usize, &alloc_ctx);
 		}
-		large_dalloc(tsd_tsdn(tsd), edata);
+		if (config_prof && opt_prof && alloc_ctx.szind < nhbins &&
```
interwq: The `alloc_ctx.szind < nhbins` check isn't necessary? We pass `tcache = NULL` to arena_dalloc_promoted_impl so it can't go through the thread cache path.

src/arena.c
```diff
@@ -697,7 +767,13 @@ arena_reset(tsd_t *tsd, arena_t *arena) {
...
+		    unlikely(alloc_ctx.szind < SC_NBINS)) {
```
interwq: nit: no need for the `unlikely` hint since this is a slow path

Svetlitski: Will fix. For context, this was just [copied directly from `arena_dalloc_large`](https://github.com/jemalloc/jemalloc/blob/b01d49664651f239fdf76774cb6de05ed7e63f4a/include/jemalloc/internal/arena_inlines_b.h#L301).

src/arena.c
```diff
@@ -697,7 +767,12 @@ arena_reset(tsd_t *tsd, arena_t *arena) {
 		if (config_prof && opt_prof) {
 			prof_free(tsd, ptr, usize, &alloc_ctx);
 		}
-		large_dalloc(tsd_tsdn(tsd), edata);
+		if (config_prof && opt_prof && alloc_ctx.szind < SC_NBINS) {
+			arena_dalloc_promoted_impl(tsd_tsdn(tsd), ptr,
+			    /* tcache */ NULL, /* slow_path */ false, edata);
```
interwq: Sorry didn't catch this last round: slow_path should be `true`.

Svetlitski: `slow_path` never gets used since we pass `NULL` for `tcache`, but sure fixed anyway.

PR #2501, opened 2023-07-31
Test on ARM as part of CI

Svetlitski:


interwq:
The new test with arm + 32bit is failing with:
```
E: Package 'gcc-multilib' has no installation candidate
E: Unable to locate package g++-multilib
E: Couldn't find any package by regex 'g++-multilib'
```

It looks some different packages for that combination: https://packages.ubuntu.com/search?keywords=gcc-multilib

Although I'm not against skipping the 32-bit build if it's non-trivial to support.

Svetlitski:
@interwq @guangli-dai Tagging you two in case one of you wants to pick this up, but I'm closing this PR.

PR #2503, opened 2023-08-01
Include the unrecognized malloc conf option in the error message.

### LINKED ISSUE #2502: Print malloc conf strings when they can't be parsed
The existing code can print key:value pairs when it doesn't recognize a key. But it doesn't print the full string when it is unable to even parse a key.

When I [reach here](https://github.com/jemalloc/jemalloc/blob/b01d49664651f239fdf76774cb6de05ed7e63f4a/src/jemalloc.c#L907) all I get is ...
```
<jemalloc>: Malformed conf string
```

The problem in this case is that it was trying to parse "narenas:1" rather than narenas:1 -- yes for some reason I don't understand those quotes made it into the string that malloc_conf_init_helper and malloc_conf_next were parsing.

I was using /bin/bash. I am not asking you to debug this for me, just give me the value of the full malloc conf string when this error occurs.

This failed for me and gets the error message above.
MALLOC_CONF=“narenas:1” bin/mysqld

This was fine:
MALLOC_CONF=narenas:1 bin/mysqld

interwq:
Previously the option causing trouble will not be printed, unless the option key:value pair format is found.

Test cases for the 3 failure cases:

```
MALLOC_CONF=narenas:1, ./test/unit/a0
<jemalloc>: Conf string ends with comma -- narenas:1,

MALLOC_CONF=narenas:1,abcd ./test/unit/a0
<jemalloc>: Conf string ends with key -- abcd

MALLOC_CONF=@@@@@ ./test/unit/a0
<jemalloc>: Malformed conf string -- @
```

Resolves #2502

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -865,6 +865,14 @@ malloc_conf_multi_sizes_next(const char **slab_size_segment_cur,
 	return false;
 }
 
+static void
+malloc_conf_format_error(const char *msg, const char *begin, const char *end) {
+	size_t len = end - begin + 1;
+	len = len > BUFERROR_BUF ? BUFERROR_BUF : len;
+
+	malloc_printf("<jemalloc>: %s -- %.*s\n", msg, (int)len, begin);
+}
+
 static bool
 malloc_conf_next(char const **opts_p, char const **k_p, size_t *klen_p,
     char const **v_p, size_t *vlen_p) {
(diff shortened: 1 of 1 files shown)


src/jemalloc.c
```diff
@@ -865,6 +865,23 @@ malloc_conf_multi_sizes_next(const char **slab_size_segment_cur,
 	return false;
 }
 
+static void
+malloc_conf_format_error(const char *msg, const char *begin, const char *end) {
```
Svetlitski, line 869: Are we able to leverage the existing `printf` infrastructure and do something like:
```c
size_t len = end - begin + 1;
len = len > BUFERROR_BUF ? BUFERROR_BUF : len;
malloc_printf("%s%.*s\n", msg, (int) len, begin);
```
Instead of hand-formatting everything? Or is `malloc_printf` not available this early in start-up?

interwq: For these warnings we usually go through `malloc_write` so that they can be redirected; see here: https://github.com/jemalloc/jemalloc/blob/dev/src/malloc_io.c#L81

Svetlitski: In that case, we could `malloc_snprintf` into a local buffer and then `malloc_write` that buffer out, if you feel like this is better than hand-formatting.

interwq: Actually you were right. malloc_printf takes care of the malloc_message as well. Let me switch to that.

Svetlitski (approved):

PR #2506, opened 2023-08-04
Prevent directly accessing an arena's bins

Svetlitski:
As [discussed in the review for an earlier PR](https://github.com/jemalloc/jemalloc/pull/2486#discussion_r1283612982), we'd like to avoid directly accessing an arena's bins, instead using the `arena_get_bin` function, which takes into account bin-shards. To ensure that we don't let code which violates this slip past review, I've added support for the `deprecated` attribute, and associated it with the field with a relevant warning message. This emits a diagnostic unless explicitly suppressed, so static analysis will catch mistakes of this kind during CI.


interwq:
LGTM. Thanks Kevin!

PR #2507, opened 2023-08-04
Optimize rtree encoding/decoding

Svetlitski:
Decoding rtree elements from their packed format takes a significant amount of time, accounting for nearly 5% of all time spent inside the allocator according to fleet telemetry. Accordingly, even small improvements here have large consequences. By using a slightly different encoding for the packed representation, we can reduce the number of instructions required for decode (for the `edata` field in particular). Experiments demonstrate this results in an approximately 0.5% to 1% performance improvement to the allocator as a whole.

A note for future maintainers: the generated code in some instances still leaves something the be desired. In particular, compilers aren't generating optimal code for accessing boolean values stored in the same register as other values (for structs that fit within a register), regardless of whether or not bitfields are used (see [LLVM issue #63942](https://github.com/llvm/llvm-project/issues/63942), which was filed directly in response to internal discussion of this issue).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1728,6 +1728,16 @@ else
   AC_DEFINE([JEMALLOC_INTERNAL_UNREACHABLE], [abort], [ ])
 fi
 
+JE_COMPILABLE([a program using _Static_assert], [
+], [
+	{
+		_Static_assert(1 == 1, "Trivially true");
+	}
+], [je_cv_gcc_static_assert])
+if test "x${je_cv_gcc_static_assert}" = "xyes" ; then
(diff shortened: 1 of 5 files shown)


Svetlitski:
Converted this PR to a draft until I come up with an appropriate fix for the failing `static_assert`s in some of our CI configs.

Svetlitski:
I'm currently waiting on results from internal performance testing to validate that the latest version of this PR (now that all of CI is passing) still confers the same performance benefits as the original version. Once that clears, I'll open this for review.

Svetlitski:
PGO-aware performance testing seems to demonstrate that this is perf neutral. This is confusing in light of the generated assembly, but there's unfortunately not enough time remaining to investigate this further.

PR #2479, opened 2023-07-10
Print test error messages in color when stderr is a terminal

Svetlitski:
As a very small quality-of-life improvement, when stderr is a terminal and supports color, print error messages from tests in red to make them stand out from the surrounding output.

PR #2509, opened 2023-08-07
Print colorful reminder for failed tests.

guangli-dai:
As discussed in #2479, this PR prints colorful reminders when a test fails, making it easier to locate failed tests from the large batch of outputs.

### DIFF
--- a/test/src/test.c
+++ b/test/src/test.c
@@ -173,13 +173,19 @@ p_test_impl(bool do_malloc_init, bool do_reentrant, test_t *t, va_list ap) {
 		}
 	}
 
-	malloc_printf("--- %s: %u/%u, %s: %u/%u, %s: %u/%u ---\n",
+	bool colored = test_counts[test_status_fail] != 0 &&
+	    isatty(STDERR_FILENO);
+	const char *color_start = colored ? "\033[1;31m" : "";
+	const char *color_end = colored ? "\033[0m" : "";
+	malloc_printf("%s--- %s: %u/%u, %s: %u/%u, %s: %u/%u ---\n%s",
+	    color_start,
 	    test_status_string(test_status_pass),
 	    test_counts[test_status_pass], test_count,
 	    test_status_string(test_status_skip),
 	    test_counts[test_status_skip], test_count,
 	    test_status_string(test_status_fail),
-	    test_counts[test_status_fail], test_count);
+	    test_counts[test_status_fail], test_count,
+	    color_end);
 
 	return ret;
 }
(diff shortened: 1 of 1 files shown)


test/src/test.c
```diff
@@ -173,13 +173,19 @@ p_test_impl(bool do_malloc_init, bool do_reentrant, test_t *t, va_list ap) {
 		}
 	}
 
+	if (test_counts[test_status_fail] != 0) {
+		malloc_printf("\033[1;31m");
```
Svetlitski: Nit: We're not formatting anything here so we can just use `malloc_write`

test/src/test.c
```diff
@@ -173,13 +173,19 @@ p_test_impl(bool do_malloc_init, bool do_reentrant, test_t *t, va_list ap) {
...
+	}
 	malloc_printf("--- %s: %u/%u, %s: %u/%u, %s: %u/%u ---\n",
 	    test_status_string(test_status_pass),
 	    test_counts[test_status_pass], test_count,
 	    test_status_string(test_status_skip),
 	    test_counts[test_status_skip], test_count,
 	    test_status_string(test_status_fail),
 	    test_counts[test_status_fail], test_count);
+	if (test_counts[test_status_fail] != 0) {
+		malloc_printf("\033[0m");
```
Svetlitski: Same nit as above, we can just use `malloc_write`.

Svetlitski (changes requested):

test/src/test.c
```diff
@@ -173,13 +173,19 @@ p_test_impl(bool do_malloc_init, bool do_reentrant, test_t *t, va_list ap) {
 		}
 	}
 
+	if (test_counts[test_status_fail] != 0) {
+		malloc_write("\033[1;31m");
+	}
 	malloc_printf("--- %s: %u/%u, %s: %u/%u, %s: %u/%u ---\n",
```
Svetlitski: Sorry I didn't think of this during review earlier, but come to think of it we should probably do all of this in a single write if possible to avoid interleaving with other output (when tests are run in parallel) and ending up with garbled colors. So maybe something alone the lines of:
```c
bool test_failed = test_counts[test_status_fail] != 0;
const char *color_start = test_failed ? "\033[1;31m" : "";
const char *color_end = test_failed ? "\033[0m" : "";
```

And add those in as `%s` at the beginning and end of the existing `malloc_printf`.

guangli-dai: Hmm did not think of the racy conditions before. `malloc_printf` itself is not atomic, so printing them in the same line may still theoretically possible lead to unexpected results? Yet the style is indeed more aligned and looks better so modified.

Svetlitski:
So one other thing to consider is when `stderr` is not a terminal and/or does not support color (imagine you're redirecting the output to a file, for example). In that case we really don't want color. We should at least be checking `isatty(STDERR_FILENO)`.

guangli-dai:
> We should at least be checking `isatty(STDERR_FILENO)`.

Thanks for the suggestion! Added. Considered `ncurses` but it does not look worthy for this case.

Svetlitski (approved):

PR #2516, opened 2023-08-10
Ensure short `read(2)`s/`write(2)`s are properly handled by IO utilities

Svetlitski:
`read(2)` and `write(2)` may read or write fewer bytes than were requested. In order to robustly ensure that all of the requested bytes are read/written, these edge-cases must be handled.

### DIFF
--- a/include/jemalloc/internal/malloc_io.h
+++ b/include/jemalloc/internal/malloc_io.h
@@ -68,7 +68,7 @@ void malloc_cprintf(write_cb_t *write_cb, void *cbopaque, const char *format,
 void malloc_printf(const char *format, ...) JEMALLOC_FORMAT_PRINTF(1, 2);
 
 static inline ssize_t
-malloc_write_fd(int fd, const void *buf, size_t count) {
+malloc_write_fd_syscall(int fd, const void *buf, size_t count) {
 #if defined(JEMALLOC_USE_SYSCALL) && defined(SYS_write)
 	/*
 	 * Use syscall(2) rather than write(2) when possible in order to avoid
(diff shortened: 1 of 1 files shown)


include/jemalloc/internal/malloc_io.h
```diff
@@ -90,7 +90,22 @@ malloc_write_fd(int fd, const void *buf, size_t count) {
...
 
 static inline ssize_t
-malloc_read_fd(int fd, void *buf, size_t count) {
+malloc_write_fd(int fd, const void *buf, size_t count) {
+	size_t bytes_written = 0;
+	while (bytes_written < count) {
+		ssize_t result = malloc_write_fd_syscall(fd,
+		    &((const byte_t *)buf)[bytes_written],
+		    count - bytes_written);
+		if (result == -1) {
```
interwq: nit:maybe `if (result < 0) return result;` so it's not possible for bytes_written to underflow

Also maybe switch to `do {} while` since we at least call once.

interwq:
BTW the CI Ubuntu upgrade seems working well. I haven't seen the powerpc failures recently.

include/jemalloc/internal/malloc_io.h
```diff
@@ -90,7 +90,22 @@ malloc_write_fd(int fd, const void *buf, size_t count) {
...
+			return -1;
+		}
+		bytes_written += result;
+	}
```
guangli-dai: `bytes_written` should not be larger than `count`? Can we assert it after the loop ends?

Svetlitski: `assert`ing anything here would be problematic – `assert` itself calls `malloc_printf` which eventually resolves to a call to this function. If we `assert` because we found an issue, it's possible we start recursing infinitely.

interwq:
re: Travis ppc64 stuck issue. I jinxed it https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/607759474

Issue #2521, opened 2023-08-15
any idea why run_test_0 failed to run? and any method to debug on this error? thanks


everschen:
gcc  -std=gnu11 -Wall -Wextra -Wsign-compare -Wundef -Wno-format-zero-length -Wpointer-arith -Wno-missing-braces -Wno-missing-field-initializers -Wno-missing-attributes -pipe -g3 -fvisibility=hidden -Wimplicit-fallthrough -Wdeprecated-declarations -O3 -funroll-loops -Werror -fno-builtin -c -D_GNU_SOURCE -D_REENTRANT -Iinclude -I../../include -DJEMALLOC_UNIT_TEST -I../../test/include -Itest/include -o test/unit/size_classes.o ../../test/unit/size_classes.c
In file included from ../../include/jemalloc/internal/cache_bin.h:8,
                 from ../../include/jemalloc/internal/tcache_structs.h:5,
                 from ../../include/jemalloc/internal/tsd_internals.h:16,
                 from ../../include/jemalloc/internal/tsd_tls.h:7,
                 from ../../include/jemalloc/internal/tsd.h:13,
                 from ../../include/jemalloc/internal/ckh.h:5,
                 from ../../include/jemalloc/internal/prof_structs.h:5,
                 from ../../include/jemalloc/internal/jemalloc_internal_includes.h:51,
                 from ../../test/include/test/jemalloc_test.h:47,
                 from ../../test/unit/size_classes.c:1:
In function ‘sz_index2size_lookup_impl’,
    inlined from ‘sz_index2size_lookup’ at ../../include/jemalloc/internal/sz.h:244:15,
    inlined from ‘sz_index2size’ at ../../include/jemalloc/internal/sz.h:252:9,
    inlined from ‘test_size_classes’ at ../../test/unit/size_classes.c:42:3:
../../include/jemalloc/internal/sz.h:239:33: error: array subscript 232 is above array bounds of ‘size_t[232]’ {aka ‘long unsigned int[232]’} [-Werror=array-bounds]
  239 |         return sz_index2size_tab[index];
In file included from ../../test/include/test/jemalloc_test.h:46:
../../include/jemalloc/internal/sz.h: In function ‘test_size_classes’:
include/jemalloc/internal/jemalloc_preamble.h:20:25: note: while referencing ‘jet_sz_index2size_tab’
   20 | #  define JEMALLOC_N(n) jet_##n
      |                         ^~~~
[...]

PR #2524, opened 2023-08-20
Fix array bounds false warning in gcc 12.3.0

everschen:
1.error: array subscript 232 is above array bounds of ‘size_t[232]’ in gcc 12.3.0 
2.it also optimizer to the code


everschen:
for more information about the issue, please refer to https://github.com/jemalloc/jemalloc/issues/2521

interwq:
Thanks for the patch!

PR #2525, opened 2023-08-20
Expose private library dependencies via pkg-config

BtbN:
When linking statically, these need to be included for linking to succeed.


interwq:
Thanks for the patch!

PR #2526, opened 2023-08-20
Expose jemalloc_prefix via pkg-config

BtbN:


interwq:
Thanks for the patch!

PR #2537, opened 2023-09-14
Allocate tcache stack from base allocator

### LINKED ISSUE #1000: THP on tcache avail array.
With opt.metadata_thp, enabling this should be beneficial as well.

interwq:
When using metadata_thp, allocate tcache bin stacks from base0, which means they
will be placed on huge pages along with other metadata, instead of mixed with
other regular allocations.

In order to do so, modified the base allocator to support limited reuse: freed
tcached stacks (from thread termination) will be returned to base0 and made
available for reuse, but no merging will be attempted since they were bump
allocated out of base blocks. These reused base extents are managed using
separately allocated base edata_t -- they are cached in base->edata_avail when
the extent is all allocated.

One tricky part is, stats updating must be skipped for such reused extents
(since they were accounted for already, and there is no purging for base). This
requires tracking the "if is reused" state explicitly and bypass the stats
updates when allocating from them.

Resolves  #1000.

### DIFF
--- a/src/base.c
+++ b/src/base.c
@@ -110,6 +110,16 @@ label_done:
 	}
 }
 
+static inline bool
+base_edata_is_reused(edata_t *edata) {
+	/*
+	 * Borrow the guarded bit to indicate if the extent is a recycled one,
+	 * i.e. the ones returned to base for reuse; currently only tcache bin
+	 * stacks.  Skips stats updating if so (needed for this purpose only).
+	 */
+	return edata_guarded_get(edata);
+}
+
 static void
 base_edata_init(size_t *extent_sn_next, edata_t *edata, void *addr,
     size_t size) {
@@ -118,7 +128,7 @@ base_edata_init(size_t *extent_sn_next, edata_t *edata, void *addr,
 	sn = *extent_sn_next;
 	(*extent_sn_next)++;
 
-	edata_binit(edata, addr, size, sn);
+	edata_binit(edata, addr, size, sn, false /* is_reused */);
 }
 
 static size_t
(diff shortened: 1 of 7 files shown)


interwq:
CI tested with metadata_thp on and off. Both working as expected.

test/unit/tcache_max.c
```diff
@@ -280,6 +303,7 @@ tcache_check(void *arg) {
 	    "Unexpected.mallctl().failure");
 	expect_zu_eq(old_tcache_max, min_tcache_max,
 	    "Unexpected value for tcache_max");
+	validate_tcache_stack(tcache);
```
guangli-dai, line 307: Reaching this point, the tcache has been rebooted multiple times. Maybe do another `validate_tcache_stack` at the beginning of `tcache_check` as well to make sure it is also using thp when first initialized.

src/base.c
```diff
@@ -491,6 +539,63 @@ base_alloc_edata(tsdn_t *tsdn, base_t *base) {
...
+	void *addr = base_alloc_impl(tsdn, base, alloc_size, alignment, &esn);
+	if (addr == NULL) {
+		edata_avail_insert(&base->edata_avail, edata);
+		return NULL;
+	}
+
+	edata_binit(edata, addr, alloc_size, esn, true /* is_reused */);
+	*(edata_t **)addr = edata;
+
+	return (byte_t *)addr + header_size;
```
guangli-dai, line 586: I am confused here. `edata` here is a allocated through `base_alloc_base_edata` while `addr` is allocated from `base_alloc_impl`. The two allocations are not supposed to be continuous, right? Why are we returning the `edata + header_size` here? 

Also `addr` is allocated a memory with `alloc_size`, which should include the size of an `edata_t`. Why do we need an extra `header_size` to be allocated there since we already have an `edata_t`?

src/base.c
```diff
@@ -185,24 +194,57 @@ base_extent_bump_alloc_helper(edata_t *edata, size_t *gap_size, size_t size,
...
-		    sz_size2index(edata_bsize_get(edata) + 1) - 1;
-		edata_heap_insert(&base->avail[index_floor], edata);
+		base_edata_heap_insert(tsdn, base, edata);
+	} else {
+		/* Freed base edata_t stored in edata_avail. */
+		edata_avail_insert(&base->edata_avail, edata);
 	}
 
-	if (config_stats) {
+	if (config_stats && !base_edata_is_reused(edata)) {
```
guangli-dai, line 248: This seems to be the main reason why we are maintaining another `edata` for the allocated stack. Why not directly use the guarded bit of the `edata` that already exists?

src/base.c
```diff
@@ -491,6 +539,63 @@ base_alloc_edata(tsdn_t *tsdn, base_t *base) {
...
+	/*
+	 * Reserve room for the header so that edata can be retrieved on dalloc.
+	 * Bump up to usize to improve reusability -- otherwise the freed stacks
+	 * will be put back into the previous size class.
+	 */
+	size_t esn, alignment, header_size;
+	b0_alloc_header_size(&header_size, &alignment);
+
+	size_t alloc_size = sz_s2u(stack_size + header_size);
+	void *addr = base_alloc_impl(tsdn, base, alloc_size, alignment, &esn);
```
guangli-dai, line 576: While `base` is mainly used internally now, the waiting time on `base->mtx` should be rather low? If every thread is using `base->mtx` now during its initialization, will this become a concern?

test/unit/tcache_max.c
interwq: Good point. I'll add a few more checkpoints.

src/base.c
interwq: Yeah the base is not accessed frequently, because of 2 factors:
1) metadata is only ever increasing (before this PR), which means once the process reaches peak usage, it won't reach this path anymore
2) the most frequent metadata (re)use, is the `edata_t` -- the [edata_cache](https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/edata_cache.h#L16) makes sure they are cached at higher level and only accessing the base when reaching a new peak.

As for the new use case, thread creation is quite heavy weight (and cannot be super frequent) so it won't be a significant factor here.

src/base.c
interwq: I should add more comments explaining the design:

Yes they are two separate allocations. Similar to the regular memory and their metadata, the edata_t tracks the info about the allocated memory / extent. One key difference is that, we cannot use rtree lookups to find the metadata on deallocation, and we need a way to do so.

That's what's happening here: we are adding a tiny header (effectively only a pointer: note that alloc_size includes a `sizeof(edata_t *)`, not the actual edata_t), which is stored right before the allocated addr -- then on deallocation, we can simply offset and retrieve the pointer, which finds us the edata_t (that's the beginning part of `b0_dalloc_tcache_stack`).

src/base.c
interwq: The edata already exists is from the initial base_block, and the allocated stack is bump allocated off that block. So there's no way to merge it back into the block because of the bump alloc -- that's why we need a new and separate edata_t to manage the returned tcache_stack (it gets `base_edata_heap_insert` into the available extents for future allocations).

In other words, the base_block bump allocates the initial tcache_stack, and when deallocating, the tcache_stack and the base_block are not guaranteed to be continuous, thus requiring to be managed separately with a new edata_t.

As for the guarded bit, yes the only reason that we use that to mark and track "reused", is to avoid double counting stats.

guangli-dai: I see. So the life cycle of this new edata is:
1. It is newed/retrieved from `edata_avail` and initialized when a tcache_stack is allocated;
2. When the stack is deallocated, representing the allocation, this edata is put into `avail` for the memory to be reused;
3. If the stack is reused up, the edata is inserted back to `edata_avail` for it to be reused.

Is my understanding correct?

Another quick question, can you remind me why we did not need an `else` above to handle `edata_bsize_get(edata) == 0` case before?

interwq: > Is my understanding correct?

Yes!

> why we did not need an else above to handle edata_bsize_get(edata) == 0 case before

Previously, all edatas were part of the base_block_t: https://github.com/jemalloc/jemalloc/blob/a442d9b895935ac872e7ccc705213537bc747c19/include/jemalloc/internal/base.h#L40

Since the # of base blocks were limited (similar to grow_retained: https://github.com/jemalloc/jemalloc/blob/a442d9b895935ac872e7ccc705213537bc747c19/include/jemalloc/internal/base.h#L65), we didn't bother reusing them. Instead every time we allocate a new block which contains a edata_t inside. Technically we could still reuse the edata_t once a block is drained, but the savings would be more like a few hundred bytes which is a bit hard to justify the complexity.

interwq:
@guangli-dai FYI it's updated (added comments and tweaked the test case).

src/base.c
guangli-dai: I see. Then I assume the original edata will also be reused with this new `else` added?

interwq: That's right. So we are also getting a tiny bit of savings from that, with this PR.

guangli-dai (approved):

interwq:
Only comments were tweaked after the last clean CI. Landing.

Issue #2255, opened 2022-04-19
add #pragma GCC system_header at jemalloc.h ?


benet96:
We found posix_memalign function has conflict clang header when compiled with clang++ (issue: https://github.com/jemalloc/jemalloc/issues/1557) and a commit (https://github.com/jemalloc/jemalloc/pull/1558/files) trying to solve it.

The same declaration in stdlib.h can pass the compilation because it has ```-isystem```. If Jemalloc has those function implementation the same as libc, we can add ```#pragma GCC system_header``` in jemalloc.h so compiler will consider jemalloc.h as system header even when user include it by ```-I

We think it's a neat patch, I used to remove the posix_memalign from jemalloc.h to resolve the confict, it works because stdlib.h has the same declaration and it will be merged to the declaration in jemalloc.h unless jemalloc.h has no declaration.

PR #2549, opened 2023-10-03
Mark jemalloc.h as system header to resolve header conflicts.

guangli-dai:
This is should help solve the header conflicts between jemalloc.h and mm_malloc.h in some cases. Planning to separate all jemalloc-specific functions in a new header to solve the conflict for good in the future.

#2255

### DIFF
--- a/include/jemalloc/jemalloc.sh
+++ b/include/jemalloc/jemalloc.sh
@@ -5,6 +5,7 @@ objroot=$1
 cat <<EOF
 #ifndef JEMALLOC_H_
 #define JEMALLOC_H_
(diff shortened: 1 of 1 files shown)


include/jemalloc/jemalloc.sh
```diff
@@ -3,6 +3,8 @@
 objroot=$1
 
 cat <<EOF
+#pragma GCC system_header
+
```
interwq: Move to after the `#define JEMALLOC_H_`. Worth making sure nothing happens when the header is already included.

PR #2555, opened 2023-10-11
Enable ncached_max setting through malloc_conf

guangli-dai:
This PR enables users to set a bin's ncached_max through malloc_conf using the same format as using mallctl. 

Based on #2530.


src/tcache.c
```diff
@@ -236,8 +251,9 @@ tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena,
 	void *ret;
 
 	assert(tcache_slow->arena != NULL);
-	unsigned nfill = cache_bin_info_ncached_max(&cache_bin->bin_info)
-	    >> tcache_slow->lg_fill_div[binind];
+	assert(!tcache_bin_disabled(binind, cache_bin, tcache_slow));
+	unsigned nfill = cache_bin_info_ncached_max_get(cache_bin,
+	    &cache_bin->bin_info) >> tcache_slow->lg_fill_div[binind];
```
interwq, line 259: @guangli-dai this is causing the ncached_max == 1 issue. nfill can be 0 if ncached_max is 1.

src/tcache.c
```diff
@@ -800,7 +809,9 @@ tcache_bin_info_compute(cache_bin_info_t tcache_bin_info[TCACHE_NBINS_MAX]) {
 	 * than tcache_nbins, no items will be cached.
 	 */
 	for (szind_t i = 0; i < TCACHE_NBINS_MAX; i++) {
-		unsigned ncached_max = tcache_ncached_max_compute(i);
+		unsigned ncached_max = opt_bin_info_is_set[i] ?
+		    opt_tcache_bin_info[i].ncached_max:
+		    tcache_ncached_max_compute(i);
```
interwq, line 830: Let's avoid the redundant computation for all threads. This part can be moved to tcache_boot with the results committing into `opt_tcache_bin_info`. Then here we can simply use the results. Worth converting this into asserts though (i.e. equals to _max_comput() if not set).

include/jemalloc/internal/tcache_externs.h
```diff
@@ -35,6 +35,11 @@ extern unsigned	global_do_not_change_tcache_nbins;
  */
 extern size_t	global_do_not_change_tcache_maxclass;
 
+/* Default bin info for each bin. */
+extern cache_bin_info_t opt_tcache_bin_info[TCACHE_NBINS_MAX];
+/* Records whether a bin's info is specified by malloc_conf. */
+extern bool opt_bin_info_is_set[TCACHE_NBINS_MAX];
```
interwq: Maybe avoid the bin_info term in the names (since we will limit it to cache_bin internal soon):
```
opt_tcache_ncached_max
opt_tcache_ncached_max_set
```
Also worth adding a getter for opt_tcache_ncached_max (e.g. `tcache_get_default_ncached_max(ind)`), to emphasize that these opt_ things should only be set during init, and no setter is provided since they should not change afterwards.

src/tcache.c
```diff
@@ -1019,14 +1021,42 @@ tcache_bins_ncached_max_write(tsd_t *tsd, char *settings, size_t len) {
...
 	} while (len_left > 0);
+	return false;
+}
 
-	arena_t *assigned_arena = tcache->tcache_slow->arena;
-	tcache_cleanup(tsd);
-	tsd_tcache_data_init_with_bin_settings(tsd, assigned_arena,
-	    tcache_bin_info);
+bool
+tcache_bins_ncached_max_write(tsd_t *tsd, char *settings, size_t len) {
```
interwq, line 1051: I remember we made this one only called if tcache is enabled?

include/jemalloc/internal/tcache_externs.h
guangli-dai: Naming it as `bin_info` mainly because its type is `cache_bin_info_t`, naming it as `ncached_max` could be a bit confusing. WDYT?

interwq: Each opt_ should correspond to its opt option name, in this case the tcache.ncached_max. The storage type could get changed later but the option likely will stay.

guangli-dai: I see. I will use the name `opt_tcache_ncached_max` then.

src/tcache.c
guangli-dai: Refactored the init so that this function will only be called in `tcache_boot` and now all auto tcache inits are done through `tsd_tcache_data_init_with_bin_settings`

src/tcache.c
guangli-dai: Yes, rebased and now it includes the change

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -50,6 +50,22 @@ tcache_bin_settings_backup(tcache_t *tcache,
...
 }
 
+static inline cache_bin_sz_t
+tcache_get_default_ncached_max(szind_t ind) {
+	assert(ind < TCACHE_NBINS_MAX);
+	return opt_tcache_ncached_max[ind].ncached_max;
+}
+
+static inline bool
+tcache_get_default_ncached_max_set(szind_t ind) {
```
interwq: Let's not provide this getter at all -- we don't want to encourage its usage. It gets set and accessed during init and never touched again afterwards.

We should be able to make opt_tcache_ncached_max_set JET_EXTERN?

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -50,6 +50,22 @@ tcache_bin_settings_backup(tcache_t *tcache,
...
+	return opt_tcache_ncached_max_set[ind];
+}
+
+static inline const cache_bin_info_t *
+tcache_get_default_ncached_max_arr(void) {
```
interwq: It looks this getter is used most often. Let's also remove `tcache_get_default_ncached_max`, and only keep this one.

The reference in tcache_bin_info_compute (which is during init) can access the data directly. We only want to provide the getter for the common and "encouraged" use cases.

edit: forgot to mention, try moving the impl of this function to tcache.c, then we can make `opt_tcache_ncached_max` static to only allow access through the getter.

guangli-dai: To be symmetrical, add getters for both `opt_tcache_ncached_max` and `opt_tcache_ncached_max_set` (although they are only used once for now) to warn people that they should not be set besides initialization. 

Also add an array getter for `opt_tcache_ncached_max` because the whole array as an input makes the auto tcache initialization unified with `tsd_tcache_data_init_with_bin_settings`. To further highlight the array should not be changed, add `const` to the getter return value and also corresponding functions' parameters.

src/tcache.c
```diff
@@ -1268,6 +1263,8 @@ tcache_boot(tsdn_t *tsdn, base_t *base) {
 	assert(global_do_not_change_tcache_maxclass <= TCACHE_MAXCLASS_LIMIT);
 	global_do_not_change_tcache_nbins =
 	    sz_size2index(global_do_not_change_tcache_maxclass) + 1;
+	/* Pre-compute default bin info. */
+	tcache_bin_info_compute(opt_tcache_ncached_max);
```
interwq, line 1297: Use the getter

src/tcache.c
```diff
@@ -952,7 +946,8 @@ tcache_enabled_set(tsd_t *tsd, bool enabled) {
 	bool was_enabled = tsd_tcache_enabled_get(tsd);
 
 	if (!was_enabled && enabled) {
-		tsd_tcache_data_init(tsd, NULL);
+		tsd_tcache_data_init_with_bin_settings(tsd, NULL,
```
interwq: We can rename this to `tsd_tcache_data_init` now, since bin settings are required always.

include/jemalloc/internal/tcache_inlines.h
interwq: > To be symmetrical, add getters for both opt_tcache_ncached_max and opt_tcache_ncached_max_set (although they are only used once for now) to warn people that they should not be set besides initialization.

See the comments I just added. Providing the getters kind of implies "they are fine to use when you want". For data used only during init and not after like `opt_tcache_ncached_max_set`, we should try making it static and limit its usage, and not having the getter.

Making the input array const is a great move.

src/jemalloc.c
```diff
@@ -1322,6 +1322,17 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				} while (vlen_left > 0);
 				CONF_CONTINUE;
 			}
+			if (CONF_MATCH("ncached_max")) {
```
interwq: How about `tcache_ncached_max` (which is similar to tcache_max intentionally)?

src/tcache.c
```diff
@@ -984,17 +993,9 @@ thread_tcache_max_set(tsd_t *tsd, size_t tcache_max) {
...
-	assert(tcache_available(tsd));
-	tcache_t *tcache = tsd_tcachep_get(tsd);
-	assert(tcache != NULL);
-	cache_bin_info_t tcache_bin_info[TCACHE_NBINS_MAX];
-	tcache_bin_settings_backup(tcache, tcache_bin_info);
-	const char *bin_settings_segment_cur = settings;
-	size_t len_left = len;
-	assert(len_left != 0);
-
+bool bin_info_settings_parse(const char *bin_settings_segment_cur,
```
interwq: Add the `tcache_` prefix. In general the namespace / module name should always be prefixed, especially for the non static ones.

src/jemalloc.c
```diff
@@ -1322,6 +1322,17 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
...
 				CONF_CONTINUE;
 			}
+			if (CONF_MATCH("tcache_ncached_max")) {
+				bool err = tcache_bin_info_default_init(
+				    v, vlen);
+				if (err) {
+					CONF_ERROR("Invalid settings for "
+					    "tcache_ncached_max", k, klen, v,
+					    vlen);
+					break;
```
interwq: This is incorrect. The bin_shard case above has a break to escape the local while loop. This one here would break the entire conf handle loop.

src/tcache.c
guangli-dai: The getter should provide a const pointer for common usage while this array is meant to be modified (initialized) here. It shouldn't use getter.

src/jemalloc.c
guangli-dai: Sounds good, modified.

src/tcache.c
```diff
@@ -977,24 +996,17 @@ thread_tcache_max_set(tsd_t *tsd, size_t tcache_max) {
 	tcache_max_set(tcache_slow, tcache_max);
 
 	if (enabled) {
-		tsd_tcache_data_init_with_bin_settings(tsd, assigned_arena,
+		tsd_tcache_data_init(tsd, assigned_arena,
```
interwq: nit: same line

include/jemalloc/internal/tcache_inlines.h
guangli-dai: Sure. I keep this one returning the array but name it as `tcache_get_default_ncached_max`, which shouldn't be confusing with it as the only getter. Currently the two arrays `opt_tcache_ncached_max` and `opt_tcache_ncached_max_set` are designed as below:
1. Both arrays are made static considering their common usage should not modify values inside;
2. Provide const getters for both arrays. Specifically, the only usage outside tcache.c is in the tests so both getters are `JET_EXTERN`;
3. Even inside `tcache.c`, still use getter if the value is not meant to be modified.
The design above should serve the purpose you mentioned well, except that we have to keep a getter for `opt_tcache_ncached_max_set` as well for testing purposes.

include/jemalloc/internal/tcache_inlines.h
guangli-dai: Arrays cannot be made `JET_EXTERN` since we want initialization to 0 here. Made it static and still offer a `JET_EXTERN` getter. I think this indicates it should only be used within `tcache.c` and for checking in tests. More details regarding the array and getter design in the comment below.

src/tcache.c
interwq: I see. Worth adding the comments to explain this.

src/jemalloc.c
guangli-dai: Thanks for pointing out! Removed.

interwq (approved):
Looks good! @guangli-dai if you have no other pending changes, I'll merge after CI passes.

guangli-dai:
> Looks good! @guangli-dai if you have no other pending changes, I'll merge after CI passes.

Sounds good! I don't have other changes and I have stress tested it yesterday with no errors (the later changes are mainly name changes and nit fixes) so I think it's good to go. I will continue to run stress tests as well just to make sure the latest version is as good as expected.

PR #2558, opened 2023-10-18
Add util.c into vs project file.

guangli-dai:


interwq:
There's also 2019 and 2022 now.

PR #2551, opened 2023-10-04
metadata usage breakdowns: tracking edata and rtree usages

nullptr0-0:

PR #2560, opened 2023-10-23
Improve the alignment and locality of bin_t and its mutex.

guangli-dai:
The new stats fields added in #2551 happened to make arena_bin->lock not cacheline aligned, which was detected during our benchmark runs. Explicitly align the bins in arena_t, and optimize the layout of malloc_mutex_t.

### DIFF
--- a/include/jemalloc/internal/arena_structs.h
+++ b/include/jemalloc/internal/arena_structs.h
@@ -98,10 +98,13 @@ struct arena_s {
 	/*
 	 * The arena is allocated alongside its bins; really this is a
 	 * dynamically sized array determined by the binshard settings.
+	 * Enforcing cacheline-alignment to minimize the number of cachelines
+	 * touched on the hot paths.
 	 */
 	JEMALLOC_WARN_ON_USAGE("Do not use this field directly. "
 	                       "Use `arena_get_bin` instead.")
-	bin_t			 all_bins[0];
+	JEMALLOC_ALIGNED(CACHELINE)
+	bin_t			all_bins[0];
 };
 
 #endif /* JEMALLOC_INTERNAL_ARENA_STRUCTS_H */
--- a/include/jemalloc/internal/mutex.h
+++ b/include/jemalloc/internal/mutex.h
@@ -32,6 +32,12 @@ struct malloc_mutex_s {
 			 * unlocking thread).
 			 */
 			mutex_prof_data_t	prof_data;
+			/*
+			 * Hint flag to avoid exclusive cache line contention
+			 * during spin waiting.  Placed along with prof_data
+			 * since it's always modified even with no contention.
+			 */
+			atomic_b_t		locked;
 #ifdef _WIN32
 #  if _WIN32_WINNT >= 0x0600
 			SRWLOCK         	lock;
@@ -46,11 +52,6 @@ struct malloc_mutex_s {
 #else
 			pthread_mutex_t		lock;
 #endif
-			/*
-			 * Hint flag to avoid exclusive cache line contention
-			 * during spin waiting
-			 */
-			atomic_b_t		locked;
 		};
 		/*
 		 * We only touch witness when configured w/ debug.  However we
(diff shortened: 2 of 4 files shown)


include/jemalloc/internal/arena_structs.h
```diff
@@ -46,6 +46,11 @@ struct arena_s {
 
 	/* Synchronization: internal. */
 	arena_stats_t		stats;
+	/*
+	 * The current size of stats is 10384B.  To avoid false sharing,
+	 * pad it to the multiple of 128B.
+	 */
+	char padding[112];
```
interwq: To confirm, the 2 new stats (https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/arena_stats.h#L55-L56) added happen to make the arena_stats_t not a multiple of 128, which in turn causes the fields below to be not cacheline aligned, and the slowdown we observed -- is that correct?

A couple of things to double check:
1) if you revert the metadata commit, but manually add `char padding[16]`. Does it cause the exact same slow down?
2) can you share the address of arena[0]->bin[sz2idx(256)].mutex, before and after the change? I suspect it crosses a cacheline boundary which means each locks touches 2 cachelines instead of 1 (this isn't exactly a false sharing)

guangli-dai: >  which in turn causes the fields below to be not cacheline aligned, and the slowdown we observed -- is that correct?

Yes.

Regarding the double check:

1. adding `char padding[16]` can reproduce the same slowdown.
2. As discussed offline: yes, with the new arena_statas_t, `a0->all_bins[0].lock.locked` has address `0x7ffff6014540` and `a0->all_bins[0].lock.prof_data.n_lock_ops` has address `0x7ffff6014510`. Thus, they are not on the same cacheline, causing more cache misses and therefore the slowdown.

interwq: @guangli-dai does move the lock hint: https://github.com/jemalloc/jemalloc/blob/756d4df2fd1b5bde025abed50c9b771376d72c6f/include/jemalloc/internal/mutex.h#L53

to between lock and prof_data also resolve the benchmark issue?

I'm considering this because, sizeof(bin_t) is 224 which is not a multiple of cacheline as well, so even if we make sure bin[0] starts aligned, there will be bins not aligned still. Making sure the data touched as little spread as possible would be the right solution -- looking into pthread_mutex_t
```
lock.__data 
$15 = {__lock = 0, __count = 0, __owner = 0, __nusers = 0, __kind = 512, __spins = 0, 
  __elision = 0, __list = {__prev = 0x0, __next = 0x0}}
```
It looks on the common / non-contended case, the last 16 bytes at least won't be accessed -- therefore moving the locked flag on our side to the beginning will group all the common-case data together.

interwq: @guangli-dai FYI this is the fix I'm proposing:
https://github.com/interwq/jemalloc/commit/54b5ed977603763a261b138b2b47caa99377d93c

Assuming it worked, feel free to include it in the PR here.

guangli-dai: It makes sense and can also properly resolve the slowdown on the tested benchmark. Included in this PR. 

Another question: Currently, the size of `malloc_mutex_t` is 112B. Should we consider align it with CACHELINE or make a sub-struct that is always cacheline aligned?

include/jemalloc/internal/mutex.h
```diff
@@ -99,33 +100,33 @@ struct malloc_mutex_s {
...
 #elif (defined(JEMALLOC_MUTEX_INIT_CB))
 #  if (defined(JEMALLOC_DEBUG))
 #     define MALLOC_MUTEX_INITIALIZER					\
-      {{{LOCK_PROF_DATA_INITIALIZER, PTHREAD_MUTEX_INITIALIZER, NULL, ATOMIC_INIT(false)}},	\
+      {{{LOCK_PROF_DATA_INITIALIZER, ATOMIC_INIT(false), PTHREAD_MUTEX_INITIALIZER, NULL}},	\
            WITNESS_INITIALIZER("mutex", WITNESS_RANK_OMIT), 0}
 #  else
 #     define MALLOC_MUTEX_INITIALIZER					\
-      {{{LOCK_PROF_DATA_INITIALIZER, PTHREAD_MUTEX_INITIALIZER, NULL, ATOMIC_INIT(false)}},	\
+      {{{LOCK_PROF_DATA_INITIALIZERNULL, ATOMIC_INIT(false), PTHREAD_MUTEX_INITIALIZER, NULL}},	\
```
interwq: my typo: LOCK_PROF_DATA_INITIALIZERNULL  
should be `LOCK_PROF_DATA_INITIALIZER`

include/jemalloc/internal/arena_structs.h
interwq: > Should we consider align it with CACHELINE or make a sub-struct that is always cacheline aligned?

That's a good question. Aligning every bin / mutex would make sure each lock resides on exactly 1 cacheline, which is a pretty nice guarantee. However it also means every bin will include some padding byte, increasing the total footprint of hot cachelines used by the bins. We can benchmark that change and see if it helps, but my intuition is that might not help much, since in prod workload most of the bins (specially the small sizes) are hot enough to stay cached. So that w/o padding, the lock crossing cacheline boundaries isn't penalized much, and w/ padding the extra line might loss some efficiency too.

interwq (approved):
Will merge once CI clears.

include/jemalloc/internal/arena_structs.h
guangli-dai: I see. The size increase caused by padding is indeed a concern. Will do some more experiments in prod to explore this.

include/jemalloc/internal/arena_structs.h
```diff
@@ -98,10 +98,12 @@ struct arena_s {
 	/*
 	 * The arena is allocated alongside its bins; really this is a
 	 * dynamically sized array determined by the binshard settings.
+	 * Enforcing cacheline-alignment to minimize the number of cachelines
+	 * touched on the hot paths.
 	 */
 	JEMALLOC_WARN_ON_USAGE("Do not use this field directly. "
 	                       "Use `arena_get_bin` instead.")
-	bin_t			 all_bins[0];
+	bin_t			 all_bins[0] JEMALLOC_ALIGNED(CACHELINE);
```
interwq: Apparently this isn't working for msvc:
https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/612074641

Try putting the align attr at the beginning, i.e.  
```JEMALLOC_ALIGNED(CACHELINE) bin_t all_bins[0];```

Comment (interwq) [reply include/jemalloc/internal/arena_structs.h]: Just realized this also needs to be updated:

https://github.com/jemalloc/jemalloc/blob/dev/src/arena.c#L1669

Should be updated to:
```
size_t arena_size = ALIGNMENT_CEILING(sizeof(arena_t), CACHELINE) +
    sizeof(bin_t) * nbins_total;
 ... // after base alloc's NULL check
assert((char *)&arena->all_bins[nbins_total -1] + sizeof(bin_t) <= (char *)arena + arena_size));
```

Please also verify that, without the ALIGNMENT_CEILING change, the assert can actually catch this issue.

guangli-dai: I see. The change in `arena_size` makes sense but the assertion may cause error in CI since we have set `all_bins` warning? Any suggestions?

interwq: That warning can be suppressed. See the example:
https://github.com/jemalloc/jemalloc/blob/756d4df2fd1b5bde025abed50c9b771376d72c6f/src/arena.c#L1715

guangli-dai: Updated but use `uintptr_t` for pointer calculation.

interwq (approved):

include/jemalloc/internal/arena_structs.h
interwq: I did some more sanity checking: the align attr of all_bins[0] properly bumped the size of arena_t to a multiple of cacheline, i.e. padding was added correctly at the end of arena_t, to make sure all_bins[0] always start at the end of arena_t. This means the ALIGNMENT_CEILING wasn't necessary since that was accounted for by sizeof(arena_t). However it doesn't hurt to still add it, in case of different behaviors from different compilers.

PR #2563, opened 2023-10-25
Reduce nesting in phn_merge_siblings using an early return.

deadalnix:
This shouldn't affect the behavior, but I find the code more readable this way.


guangli-dai:
Thank you for the patch! I will take a look and run some perf tests later this week.

guangli-dai (approved):

interwq:
Thanks for the patches @deadalnix !

deadalnix:
Cool! You might want to consider removing the `current` elements in bins ans associated logic and just pick the top of the heap now.

PR #2564, opened 2023-10-25
Fixup auxcount when a contradiction is detected.

deadalnix:
That seems like a sensible thing to try, but please benchmark and let me know.


guangli-dai:
Tested it on a heavy workload, CPU cycles actually suffer from a small regress (<1%) and no wins are observed. Unfortunately, the CPU sampling tooling is not working fine and I cannot provide more details regarding why from the experiments.

deadalnix:
I wasn't too sure about this one, but lacked a way to measure it precisely. I still think there is something there. Let's try something simpler.

guangli-dai:
The new version performs slightly better, but still neutral to small regression (~ 0.5%) across various contention-heavy workloads.

deadalnix:
Ok, let's just close this, then?

PR #2565, opened 2023-10-25
Ensure that the root of a heap is always the best element.

deadalnix:
It seems to me that the one special case of being able to delete the root when it has no children in a way that do not preserve the "root is the best element" invariant wasn't worth it, as it forces pessimisation elsewhere - for instance here in `ph_merge_aux`.

It's not clear to me that it is worth it, especially since one could rely on the root element instead of using a current field arena's bin for instance, and I'm sure there is more elsewhere.

In any case, I'll let you guys benchmark this to see if it is a terrible mistake or not. It doesn't look like it is on my end, but I don't have an army of services torturing this code in various ways to make sure.

### DIFF
--- a/include/jemalloc/internal/ph.h
+++ b/include/jemalloc/internal/ph.h
@@ -239,7 +239,7 @@ ph_merge_aux(ph_t *ph, size_t offset, ph_cmp_t cmp) {
 		phn_prev_set(phn, NULL, offset);
 		phn = phn_merge_siblings(phn, offset, cmp);
 		assert(phn_next_get(phn, offset) == NULL);
-		ph->root = phn_merge(ph->root, phn, offset, cmp);
+		phn_merge_ordered(ph->root, phn, offset, cmp);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


guangli-dai:
On heavy workloads, I do observe wins in CPU cycles to different degrees. I think the change is valid and a good idea.

guangli-dai (approved):

deadalnix:
It's great. A natural follow up tasks would be to remove the "current" slab from the arena's bins. It seems to be a leftover from the days where this was done with a red black tree instead of a pairing heap, and now that we have the best element always at the top of the heap, we get it for free, and current becomes redundant at best - but probably a liability as it makes the bins bigger and might get you to use an inferior slab.

everschen:
I am really curious about issue of pairing heap, could you please let us know the exact data example which cause this issue? thanks.

deadalnix:
There is no issue per se. This just ensures the root of the heap is always the best element. Previously, it was most of the time, except when you removed the root, and it had no children. While it makes things slightly slower when removing the root and the root has no children, this also makes it easier to get he best element, and all in all, the later is more useful.

everschen:
if i remember correctly, there is some policy like lazy policy, it will leave sort as late as possible, but still make sure return always correct one.
i had thought that this is the reason of high performance of ph, this is also the reason why it replace rb tree for the case.




---- Replied Message ----
| From | ***@***.***> |
| Date | 11/11/2023 01:18 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [jemalloc/jemalloc] Ensure that the root of a heap is always the best element. (PR #2565) |

There is no issue per se. This just ensures the root of the heap is always the best element. Previously, it was most of the time, except when you removed the root, and it had no children. While it makes things slightly slower when removing the root and the root has no children, this also makes it easier to get he best element, and all in all, the later is more useful.

deadalnix:
Yes, this is the principle behind a heap. It keeps the element in a partially sorted order, in such a way that it can retrieve the "best" one quickly.

include/jemalloc/internal/ph.h
```diff
@@ -239,7 +239,7 @@ ph_merge_aux(ph_t *ph, size_t offset, ph_cmp_t cmp) {
 		phn_prev_set(phn, NULL, offset);
 		phn = phn_merge_siblings(phn, offset, cmp);
 		assert(phn_next_get(phn, offset) == NULL);
-		ph->root = phn_merge(ph->root, phn, offset, cmp);
+		phn_merge_ordered(ph->root, phn, offset, cmp);
```
everschen, line 242: I believe the update is wrong, phn_merge_ordered means there's known order for these two node as "assert(cmp(phn0, phn1) <= 0);" shows, however here we just can't make sure the order of ph->root and phn, right?

guangli-dai:
FYI, while we are convinced that this PR will be a win, we want a more accurate number, which has to be collected after the holiday season. So we will likely merge this early next year. Thanks again for the patch!

PR #2577, opened 2023-12-04
Enable heap profiling on MacOS

nullptr0-0:


src/prof_sys.c
```diff
@@ -570,6 +574,61 @@ prof_dump_close(prof_dump_arg_t *arg) {
...
+	struct load_command *load_cmd = (struct load_command *)
+	    ((char *)header + sizeof(mach_header_t));
+	for (uint32_t i = 0; load_cmd && (i < header->ncmds); i++) {
+		if (load_cmd->cmd == LC_SEGMENT_VALUE) {
+			const segment_command_t *segment_cmd =
+			    (const segment_command_t *)load_cmd;
+			if (!strcmp(segment_cmd->segname, "__TEXT")) {
+				char buffer[PATH_MAX + 1];
+				malloc_snprintf(buffer, sizeof(buffer),
+				    "%016llx-%016llx: %s\n", segment_cmd->vmaddr + slide,
```
guangli-dai: nit: this line exceeds 80 chars, same for line 614 and 620.

guangli-dai (commented):
Can you add a brief description for the PR to summarize the PR? Besides, can you also upload a sample dump from mac?

test/unit/prof_mdump.c
```diff
@@ -169,7 +173,9 @@ TEST_BEGIN(test_mdump_maps_error) {
 
 	prof_dump_open_file_t *open_file_orig = prof_dump_open_file;
 	prof_dump_write_file_t *write_file_orig = prof_dump_write_file;
+#ifndef __APPLE__
 	prof_dump_open_maps_t *open_maps_orig = prof_dump_open_maps;
+#endif
```
interwq: Maybe keep the definition of prof_dump_open_maps_t for mac (but set prof_dump_open_maps to NULL or simply calling `NOT_REACHED()`), and then in this test we don't have to #ifdef many times?

The only feedback I have is, we should cut down the number of ifdefs, by grouping them together or making things more "compatible". But honestly I couldn't find a clean way. @nullptr0-0 if you also don't see a good way, we can leave it as is.

Issue #2576, opened 2023-12-01
`operator delete` bypasses debug logging


DanShaders:
In `jemalloc_cpp.cpp`, global C++ new/delete operators are calling `free_impl` function directly. This bypasses debug logging for `operator delete`. I think this is a totally unexpected (especially considering that `operator new` is properly logged) and, IMO, worth fixing.

(I just spent 4 hours because of that)

interwq:
Thanks for catching this @DanShaders . Fixed in #2578.

PR #2578, opened 2023-12-04
Add LOGs when entrying and exiting free and sdallocx.

guangli-dai:
Add LOGs to operator delete for better debugging experience in respond to issue #2576.

### DIFF
--- a/src/jemalloc_cpp.cpp
+++ b/src/jemalloc_cpp.cpp
@@ -112,7 +112,12 @@ template <bool IsNoExcept>
 JEMALLOC_ALWAYS_INLINE
 void *
 newImpl(std::size_t size) noexcept(IsNoExcept) {
-	return imalloc_fastpath(size, &fallbackNewImpl<IsNoExcept>);
+	LOG("core.operator_new.entry", "size: %zu", size);
+
+	void * ret = imalloc_fastpath(size, &fallbackNewImpl<IsNoExcept>);
+
+	LOG("core.operator_new.exit", "result: %p", ret);
+	return ret;
 }
 
 void *
@@ -173,21 +178,37 @@ operator new[](std::size_t size, std::align_val_t alignment, const std::nothrow_
 
 void
 operator delete(void *ptr) noexcept {
+	LOG("core.operator_delete.entry", "ptr: %p", ptr);
+
 	je_free_impl(ptr);
+
+	LOG("core.operator_delete.exit", "");
 }
 
 void
 operator delete[](void *ptr) noexcept {
+	LOG("core.operator_delete.entry", "ptr: %p", ptr);
+
 	je_free_impl(ptr);
+
+	LOG("core.operator_delete.exit", "");
 }
 
 void
 operator delete(void *ptr, const std::nothrow_t &) noexcept {
+	LOG("core.operator_delete.entry", "ptr: %p", ptr);
+
 	je_free_impl(ptr);
+
+	LOG("core.operator_delete.exit", "");
 }
 
 void operator delete[](void *ptr, const std::nothrow_t &) noexcept {
+	LOG("core.operator_delete.entry", "ptr: %p", ptr);
+
 	je_free_impl(ptr);
+
+	LOG("core.operator_delete.exit", "");
 }
 
 #if __cpp_sized_deallocation >= 201309
@@ -198,7 +219,11 @@ sizedDeleteImpl(void* ptr, std::size_t size) noexcept {
 	if (unlikely(ptr == nullptr)) {
 		return;
 	}
+	LOG("core.operator_delete.entry", "ptr: %p, size: %zu", ptr, size);
+
 	je_sdallocx_noflags(ptr, size);
+
+	LOG("core.operator_delete.exit", "");
 }
 
 void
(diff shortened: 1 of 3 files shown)


src/jemalloc_cpp.cpp
```diff
@@ -173,21 +173,38 @@ operator new[](std::size_t size, std::align_val_t alignment, const std::nothrow_
 
 void
 operator delete(void *ptr) noexcept {
+	LOG("core.free.entry", "ptr: %p", ptr);
```
interwq: The log entry should match the API / the higher level caller. Maybe "core.operator_delete.entry"?

Similarly for all other places in this file. `free` isn't the API used by the application but only the underlying implementation.

Need to include jemalloc/internal/log.h as well

guangli-dai: Sure. Updated the log messages.

src/jemalloc_cpp.cpp
```diff
@@ -224,27 +241,48 @@ alignedSizedDeleteImpl(void* ptr, std::size_t size, std::align_val_t alignment)
 	if (unlikely(ptr == nullptr)) {
 		return;
 	}
+	LOG("core.alignedSizedDeleteImpl.entry", "ptr: %p, size: %zu, flags: %d", ptr,
```
interwq: alignedSizedDeleteImpl isn't a public API and is used by `operator delete(void* ptr, std::size_t size, std::align_val_t alignment)`. Maybe:

`LOG("core.opeartor_delete.entry", "ptr: %p, size: %zu, alignment: %zu", ptr, size, alignment);`

Since we are making changes here, let's also improve a few other places:
1) Remove the LOG here:
https://github.com/jemalloc/jemalloc/blob/e4817c8d89a2a413e835c4adeab5c5c4412f9235/include/jemalloc/internal/jemalloc_internal_inlines_c.h#L581
Then add LOG into sizedDeleteImpl above, which matches the other operator delete LOGs but with a size param.

2) Similarly, move the one in imalloc_fastpath to its callers, so it can differentiate malloc vs operator new: https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/jemalloc_internal_inlines_c.h#L309

Do we need to include jemalloc/internal/log.h here?

guangli-dai: Sounds good, to confirm, the idea here is to log the highest level public api, is that correct? I will also double check other places if any.

> Do we need to include jemalloc/internal/log.h here?

That is already implicitly included through `internal_includes.h`.

interwq: > the idea here is to log the highest level public api, is that correct?

Right. That allows the log info to be better matched with the user side callsites, e.g. sized aligned delete should not log the MALLOCX flag (which is the internal impl to that API), but the input param of size and align. While the sdallocx itself logs the flag because that's what the users pass in.

guangli-dai:
Test failure is irrelevant to the changes in this PR. Maybe it's triggered by llvm updates. This PR based on dev with no meaningful change also raises same warnings: https://github.com/jemalloc/jemalloc/pull/2580

Tested the output with a sample program
```
void test_delete() {
	int *arr = new int[10];
	arr[0] = 5;
	arr[1] = 3;
	arr[2] = arr[0] + arr[1];
	delete[] arr;
}

void test_malloc() {
	int *test = (int *)malloc(sizeof(int));
	*test = 20;
	free(test);
}
int main() {
	test_delete();
	test_malloc();
	return 0;
}
```
Outputs on dev:
```
core.malloc.exit: result: 0x7fd431408000
core.calloc.entry: num: 1185, size: 1

core.calloc.exit: result: 0x7fd43140b000
core.malloc.entry: size: 16
core.malloc.exit: result: 0x7fd431410000
core.malloc.entry: size: 24
core.malloc.exit: result: 0x7fd431411000
core.malloc.entry: size: 160
core.malloc.exit: result: 0x7fd431412000
core.malloc.entry: size: 72704
core.malloc.exit: result: 0x7fd431417140
core.malloc.entry: size: 40
core.malloc.exit: result: 0x7fd431408030
core.malloc.entry: size: 4
core.malloc.exit: result: 0x7fd43142c000
core.free.entry: ptr: 0x7fd43142c000
core.free.exit: 
```

Outputs on this PR:

```
core.malloc.exit: result: 0x7efe1a208000
core.calloc.entry: num: 1185, size: 1
core.calloc.exit: result: 0x7efe1a20b000
core.malloc.entry: size: 16
core.malloc.exit: result: 0x7efe1a210000
core.malloc.entry: size: 24
core.malloc.exit: result: 0x7efe1a211000
core.malloc.entry: size: 160
core.malloc.exit: result: 0x7efe1a212000
core.malloc.entry: size: 72704
core.malloc.exit: result: 0x7efe1a217140
core.operator_new.entry: size: 40
core.operator_new.exit: result: 0x7efe1a208030
core.operator_delete.entry: ptr: 0x7efe1a208030
core.operator_delete.exit: 
core.malloc.entry: size: 4
core.malloc.exit: result: 0x7efe1a22c000
core.free.entry: ptr: 0x7efe1a22c000
core.free.exit: 
```

interwq (approved):

PR #2581, opened 2023-12-14
Fix static analysis warnings.

guangli-dai:


src/extent.c
```diff
@@ -211,6 +211,8 @@ ecache_evict(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 	case extent_state_retained:
 		extent_deregister(tsdn, pac, edata);
 		break;
+	case extent_state_transition:
+	case extent_state_merging:
```
interwq, line 214: nit: move extent_state_active down here as well

src/jemalloc.c
```diff
@@ -466,7 +466,9 @@ arena_bind(tsd_t *tsd, unsigned ind, bool internal) {
 		for (unsigned i = 0; i < SC_NBINS; i++) {
 			assert(bin_infos[i].n_shards > 0 &&
 			    bin_infos[i].n_shards <= BIN_SHARDS_MAX);
-			bins->binshard[i] = shard % bin_infos[i].n_shards;
+			assert(BIN_SHARDS_MAX < 1 << (sizeof(uint8_t) * 8));
+			bins->binshard[i] = (uint8_t)(shard %
```
interwq: nit: change the shard define above so that it's converted only once 
```
uint8_t shard = (uint8_t)atomic_fetch_add_u(...)
```

src/pac.c
```diff
@@ -403,6 +407,8 @@ pac_decay_stashed(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
 			nunmapped += npages;
 			break;
 		case extent_state_retained:
+		case extent_state_transition:
```
interwq, line 409: Similarly move down the active state.

src/prof.c
```diff
@@ -277,7 +277,8 @@ prof_sample_new_event_wait(tsd_t *tsd) {
 	 * otherwise bytes_until_sample would be 0 if u is exactly 1.0.
 	 */
 	uint64_t r = prng_lg_range_u64(tsd_prng_statep_get(tsd), 53);
-	double u = (r == 0U) ? 1.0 : (double)r * (1.0/9007199254740992.0L);
+	double u = (r == 0U) ? 1.0 : (double)((long double)r *
+	    (1.0L/9007199254740992.0L));
```
interwq, line 281: Question: should we convert the constant to double instead? What's the warn msg?

src/sz.c
```diff
@@ -92,7 +92,8 @@ static void
 sz_boot_size2index_tab(const sc_data_t *sc_data) {
 	size_t dst_max = (SC_LOOKUP_MAXCLASS >> SC_LG_TINY_MIN) + 1;
 	size_t dst_ind = 0;
-	for (unsigned sc_ind = 0; sc_ind < SC_NSIZES && dst_ind < dst_max;
+	assert(SC_NSIZES <= 1 << (sizeof(uint8_t) * 8));
+	for (uint8_t sc_ind = 0; sc_ind < SC_NSIZES && dst_ind < dst_max;
```
interwq: It looks sc_data_t has no explicit uint8 type. So the only uint8 part is `sz_size2index_tab`? If so let's add the assert and conversion to that specific part only.

src/jemalloc.c
guangli-dai: Not sure if this is a good idea, currently `(shard % bin_infos[i].n_shards)` can be casted into uint8_t safely because `bin_infos[i].n_shards <= BIN_SHARDS_MAX` and `BIN_SHARDS_MAX` should not exceed the uint8_t's max. Does the same assumption hold for `arena->binshard_next`?

src/prof.c
guangli-dai: The warn message is mainly complaining about the implicit conversion between `double` and `long double`. I would say converting constant to either double or long double works as long as there is no concern on the precision.

src/jemalloc.c
interwq: arena->binshard_next is unsigned. There we simply use it as a seed; for all we care it can be named "arena_sharding_random_seed". So casting would be intentionally dropping the higher bits. No preference either way here.

src/prof.c
interwq: I'd say try making sure the precision remains the same, i.e. if it's doing an implicit conversion of r, then make it explicit.

src/prof.c
```diff
@@ -277,7 +277,8 @@ prof_sample_new_event_wait(tsd_t *tsd) {
 	 * otherwise bytes_until_sample would be 0 if u is exactly 1.0.
 	 */
 	uint64_t r = prng_lg_range_u64(tsd_prng_statep_get(tsd), 53);
-	double u = (r == 0U) ? 1.0 : (double)r * (1.0/9007199254740992.0L);
+	double u = (r == 0U) ? 1.0 : ((double)r *
+	    (1.0/9007199254740992.0));
```
interwq: Doesn't look to be any different? Or am I missing something?

guangli-dai: I think C by default views a constant like `1.0` as double and `1.0L` as long double. The current version is all in double so it should be safe? Besides, the travis is failing but I cannot see any information regarding the error. Can you see anything from your side? Specifically, it seems to fail only on freebsd and stays the same among multiple attempts I tried.

interwq: I see. The division result is a tiny number so we may loss some precision if cast early / drop the long double. To be safe, maybe try casting after the multiply: `(double)(r * (1.0/9007199254740992.0L))` (might need convert r to long double?) I suspect this was the original intention.

guangli-dai: Sure, changed it back to the previous version where all are casted into long double and then cast the multiplication result back to double.

interwq:
The Travis CI FreeBSD failures don't look related. Merging.

PR #2583, opened 2023-12-22
Call Windows TlsGetValue2 instead of TlsGetValue when available

mcfi:
Windows 11 24H2 supports [TlsGetValue2.](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-tlsgetvalue2) Compared to TlsGetValue, TlsGetValue2 is guaranteed to not touch the last error, therefore eliminating the need to call GetLastError and SetLastError to save and restore the last error value. This has noticeable performance impact on tested CPUs measured by the following simple microbenchmark.
```
#define JEMALLOC_NO_DEMANGLE
#include <chrono>
#include <iostream>

const unsigned __int64 ITERATIONS = 0x800000;

void main() {
    static const int sizes[] =
    {
        7, 16, 32, 60, 91, 100, 120, 144, 169, 199, 255
    };
    
    auto start = std::chrono::high_resolution_clock::now();
    for (unsigned __int64 i = 0; i < ITERATIONS; i++)
    {
        for (auto sz : sizes)
        {
            void* p = je_malloc(sz);
            je_free(p);
        }
    }
    auto end = std::chrono::high_resolution_clock::now();
    std::chrono::duration<double> diff = end - start;
    std::cout << "Time: " << diff.count() << " s\n";
}

                                   Vanilla (sec)       This change (sec)      Time Delta
Intel i9-13900KF (x64)               1.12                      0.99               -11.6%
Snapdragon X1E78100 (Arm64)          1.71                      0.98               -42.7%
Snapdragon X1E78100 (Arm64EC)        1.76                      4.62              +162.5%
Windows Devkit 2023 (Arm64)          1.45                      1.32                -9.0%
Windows Devkit 2023 (Arm64EC)        1.45                      7.33              +405.5%
```
In x64 or Arm64 native cases, this PR eliminates calls against GetLastError/SetLastError so we saw noticeable time reduction. However, for Arm64EC, although GetLastError/SetLastError, are eliminated, we need to indirectly call TlsGetValue2, and Arm64EC indirect calls need to do EC bitmap check and fast-forward sequence check, so overall it's slower. Not because TlsGetValue2 is slower, but the cost of getting TlsGetValue2 is large. Therefore, this change is enabled for every architecture other than Arm64EC.

### DIFF
--- a/include/jemalloc/internal/quantum.h
+++ b/include/jemalloc/internal/quantum.h
@@ -24,7 +24,7 @@
 #  ifdef __arm__
 #    define LG_QUANTUM		3
 #  endif
-#  ifdef __aarch64__
+#  if defined(__aarch64__) || defined(_M_ARM64)
 #    define LG_QUANTUM		4
 #  endif
 #  ifdef __hppa__
--- a/include/jemalloc/internal/tsd_win.h
+++ b/include/jemalloc/internal/tsd_win.h
@@ -15,6 +15,16 @@ typedef struct {
 extern DWORD tsd_tsd;
 extern tsd_wrapper_t tsd_boot_wrapper;
 extern bool tsd_booted;
+#if defined(_M_ARM64EC)
+#define JEMALLOC_WIN32_TLSGETVALUE2 0
+#else
+#define JEMALLOC_WIN32_TLSGETVALUE2 1
+#endif
+#if JEMALLOC_WIN32_TLSGETVALUE2
+typedef LPVOID (WINAPI *TGV2)(DWORD dwTlsIndex);
+extern TGV2 tls_get_value2;
+extern HMODULE tgv2_mod;
+#endif
 
 /* Initialization/cleanup. */
 JEMALLOC_ALWAYS_INLINE bool
@@ -49,9 +59,17 @@ tsd_wrapper_set(tsd_wrapper_t *wrapper) {
 
 JEMALLOC_ALWAYS_INLINE tsd_wrapper_t *
 tsd_wrapper_get(bool init) {
-	DWORD error = GetLastError();
-	tsd_wrapper_t *wrapper = (tsd_wrapper_t *) TlsGetValue(tsd_tsd);
-	SetLastError(error);
+	tsd_wrapper_t *wrapper;
+#if JEMALLOC_WIN32_TLSGETVALUE2
+	if (tls_get_value2 != NULL) {
+		wrapper = (tsd_wrapper_t *) tls_get_value2(tsd_tsd);
+	} else
+#endif
+	{
+		DWORD error = GetLastError();
+		wrapper = (tsd_wrapper_t *) TlsGetValue(tsd_tsd);
+		SetLastError(error);
+	}
 
 	if (init && unlikely(wrapper == NULL)) {
 		wrapper = (tsd_wrapper_t *)
@@ -78,6 +96,12 @@ tsd_boot0(void) {
 	}
 	_malloc_tsd_cleanup_register(&tsd_cleanup_wrapper);
 	tsd_wrapper_set(&tsd_boot_wrapper);
+#if JEMALLOC_WIN32_TLSGETVALUE2
+	tgv2_mod = LoadLibraryA("api-ms-win-core-processthreads-l1-1-8.dll");
+	if (tgv2_mod != NULL) {
+		tls_get_value2 = (TGV2)GetProcAddress(tgv2_mod, "TlsGetValue2");
+	}
+#endif
 	tsd_booted = true;
 	return false;
 }
--- a/src/tsd.c
+++ b/src/tsd.c
@@ -25,6 +25,10 @@ bool tsd_booted = false;
 DWORD tsd_tsd;
 tsd_wrapper_t tsd_boot_wrapper = {false, TSD_INITIALIZER};
 bool tsd_booted = false;
+#if JEMALLOC_WIN32_TLSGETVALUE2
+TGV2 tls_get_value2 = NULL;
+HMODULE tgv2_mod = NULL;
+#endif
 #else
 
 /*


mcfi:
@interwq, could you please review this change?

mcfi:
@davidtgoldblatt @interwq , could you please review this change? Thanks! Let me know if you need anything.

interwq:
@mcfi thanks for the patch! I'll try getting back to you soon (likely early next week).

mcfi:
> @mcfi thanks for the patch! I'll try getting back to you soon (likely early next week).

Thanks a lot!

mcfi:
Friendly ping, @interwq @davidtgoldblatt, :)

include/jemalloc/internal/quantum.h
```diff
@@ -24,7 +24,7 @@
 #  ifdef __arm__
 #    define LG_QUANTUM		3
 #  endif
-#  ifdef __aarch64__
+#  if defined(__aarch64__) || defined(_M_ARM64)
```
guangli-dai, line 27: Why do we need to add `_M_ARM64` here?

include/jemalloc/internal/tsd_win.h
```diff
@@ -78,6 +96,12 @@ tsd_boot0(void) {
 	}
 	_malloc_tsd_cleanup_register(&tsd_cleanup_wrapper);
 	tsd_wrapper_set(&tsd_boot_wrapper);
+#if JEMALLOC_WIN32_TLSGETVALUE2
+	tgv2_mod = LoadLibraryA("api-ms-win-core-processthreads-l1-1-8.dll");
+	if (tgv2_mod != NULL) {
+		tlsgetvalue2 = (TGV2)GetProcAddress(tgv2_mod, "TlsGetValue2");
```
guangli-dai: `tlsgetvalue2` is a rather new function, but here it assumes this function always exist when `_M_ARM64EC` is not defined. That means `tlsgetvalue2` could be NULL here. Maybe regress to `TlsGetValue` for that case? Can also do a check in the configuration phase like [this](https://github.com/jemalloc/jemalloc/blob/5b72ac098abce464add567869d082f2097bd59a2/configure.ac#L405).

include/jemalloc/internal/quantum.h
mcfi: __aarch64__ is a predefined macro by clang/gcc, and _M_ARM64 is a predefined macro by MSVC/clang-cl. This change should equalize LG_QUANTUM defined for all mainstream compilers, IIRC. This is similar to line 21, or below. __x86_64__ is a macro predefined by clang/gcc, while _M_X64 is a predefined macro by MSVC/clang-cl.
`#  if (defined(__amd64__) || defined(__x86_64__) || defined(_M_X64))`

include/jemalloc/internal/tsd_win.h
mcfi: This code doesn't assume existence of TlsGetValue2. Windows APIs `LoadLibrary` and `GetProcAddress` are equivalent to Linux `dlopen` and `dlsym`, so we are basically querying the existence of the API and if so, call it.

mcfi: The reason why we don't want to call TlsGetValue2 if compiled for Arm64EC is that indirect calls in Arm64EC are expensive. Please see microbenchmark results in my PR description for more details. For more information about Arm64EC indirect call checks, please refer to https://learn.microsoft.com/en-us/windows/arm/arm64ec-abi#call-checkers

include/jemalloc/internal/quantum.h
guangli-dai: Thx for explaining. Asking because there are [other references to `__aarch64__`](https://github.com/search?q=repo%3Ajemalloc%2Fjemalloc%20__aarch64__&type=code), I think we should append `defined(_M_ARM64)` to those usage as well? I can open another PR for these separately.

include/jemalloc/internal/tsd_win.h
guangli-dai: Thanks for clarifying. I overlooked that the check is done in `tsd_wrapper_get`. I still wonder if it is possible to do the detection statically, i.e., during the configuration phase. `tsd_wrapper_get` is on the fastpath where we try to eliminate branching as much as possible. If we can get rid of `if (tlsgetvalue2 != NULL)` there, the perf may be even better.

include/jemalloc/internal/quantum.h
mcfi: There seem only four places that check __aarch64__, see below. The first two don't need a fix because SYS_open and __APPLE__ are not defined on Windows. The one in quantum.h is already handled by this PR. Only the last one needs some investigation. If you could open another PR to address it, that'll be great. Thanks.
```
E:\jemalloc>findstr /Ls __aarch64__ *.h
include\jemalloc\internal\jemalloc_internal_decls.h:#    if defined(SYS_open) && defined(__aarch64__)
include\jemalloc\internal\jemalloc_internal_inlines_a.h:#elif defined(__aarch64__) && defined(__APPLE__)
include\jemalloc\internal\quantum.h:#  if defined(__aarch64__) || defined(_M_ARM64)
include\jemalloc\internal\rtree.h:#    ifdef __aarch64__
```

include/jemalloc/internal/tsd_win.h
mcfi: Np. Unfortunately, that runtime check can't be removed because an app built with jemalloc may be installed and run on older Windows OSes that don't support TlsGetValue2.

mcfi: Plus, tlsgetvalue2 != NULL should always evaluate to true if on win11 24h2+ or false otherwise, which means it's great for CPU branch prediction and speculatively execution. It's unlikely to be a perf issue.

src/tsd.c
```diff
@@ -25,6 +25,10 @@ bool tsd_booted = false;
 DWORD tsd_tsd;
 tsd_wrapper_t tsd_boot_wrapper = {false, TSD_INITIALIZER};
 bool tsd_booted = false;
+#if JEMALLOC_WIN32_TLSGETVALUE2
+extern TGV2 tlsgetvalue2 = NULL;
```
interwq: Thanks for the patch @mcfi ! 

You probably don't mean `extern` here? Maybe rename to `tlsgetvalue2ptr` to be more distinguished from TlsGetValue2? Or is this really defined somewhere else that we are only referencing (in which case, will the extern cause trouble if it's not defined anywhere)? 

Also please squash all the commits into a single one.

interwq:
@Orvid could you help to take a look here as well?

src/tsd.c
mcfi: thank you for reviewing my change :) I removed the "extern"s you pointed out. In addition, instead of renaming tlsgetvalue2 to tlsgetvalue2ptr, I chose tls_get_value2 to make the name easier to read and consistent with other variables (no _ptr suffix). Hope it's ok to you.

Could you please advise on how to squash all commits into a single one? Should I close this PR and send out a new one?

include/jemalloc/internal/tsd_win.h
guangli-dai: I see. Agreed.

src/tsd.c
guangli-dai: `git rebase -i HEAD~15` to squash locally and `git push --force` to the remote will do the job.

interwq: @mcfi `tls_get_value2` sounds good to me. BTW it looks we don't have to keep a global tgv2_mod, instead it can simply be a stack variable?

re: squashing. Like Guangli mentioned, after doing so locally, force push to this branch (same PR) will work.

include/jemalloc/internal/tsd_win.h
```diff
@@ -78,6 +96,12 @@ tsd_boot0(void) {
 	}
 	_malloc_tsd_cleanup_register(&tsd_cleanup_wrapper);
 	tsd_wrapper_set(&tsd_boot_wrapper);
+#if JEMALLOC_WIN32_TLSGETVALUE2
+	tgv2_mod = LoadLibraryA("api-ms-win-core-processthreads-l1-1-8.dll");
```
interwq, line 100: I'm a bit concerned that this may not be forward compatible, e.g. the dll gets renamed or updated and we'd lose the support here (but in a non obvious way). Is there a different way that can detect version >= Windows 11, 24H2?

src/tsd.c
interwq: One way to optimize away the NULL pointer function check:

Having a global `tls_get_value_ptr` which is initialized to either `TlsGetValue2` (if available) or fall back to `TlsGetValue_handle_error` which is defined as below:
```
TlsGetValue_handle_error (tsd) {
        DWORD error = GetLastError();
        wrapper = (tsd_wrapper_t *) TlsGetValue(tsd);
        SetLastError(error);
}
```

Which is possible because the two variants of TLS get value share the same function signature. This way it's always an unconditional function call.

mcfi: > @mcfi `tls_get_value2` sounds good to me. BTW it looks we don't have to keep a global tgv2_mod, instead it can simply be a stack variable?
> 
> re: squashing. Like Guangli mentioned, after doing so locally, force push to this branch (same PR) will work.

With tgv_mod being a stack variable, the moment when the initialization function returns, the stack variable is gone, and in case some users need to call FreeLibrary against tgv_mod during application exit, they can't.. Speaking of that, is there any destruction code for jemalloc, if so, I can add a FreeLibrary call there to match LoadLibrary.

> One way to optimize away the NULL pointer function check:
> 
> Having a global `tls_get_value_ptr` which is initialized to either `TlsGetValue2` (if available) or fall back to `TlsGetValue_handle_error` which is defined as below:
> 
> ```
> TlsGetValue_handle_error (tsd) {
>         DWORD error = GetLastError();
>         wrapper = (tsd_wrapper_t *) TlsGetValue(tsd);
>         SetLastError(error);
> }
> ```
> 
> Which is possible because the two variants of TLS get value share the same function signature. This way it's always an unconditional function call.

This may have a bigger perf penalty on older Windows without TlsGetValue2 support if an application is built with some sort of control-flow integrity (e.g., Control-Flow Guard). Right now, TlsGetValue is called with an indirect call through a read-only import table, which doesn't have to be guarded by CFI. However, if we use a function pointer for TlsGetValue, in the worst case we add a CFI check before dereferencing that pointer, which is much more expensive than a simple conditional branch.

include/jemalloc/internal/tsd_win.h
mcfi: yes, this is forward compatible as all future windows should support this API set api-ms-win-core-processthreads-l1-1-8 and loading the API set dll instead of a forwarded concrete dll (e.g., kernel32.dll) is the preferred way since kernel32.dll might not exist on other Windows SKUs. Usually, we don't recommend checking windows versions when it comes to API support.

Orvid:
> @Orvid could you help to take a look here as well?

This looks good from my generic-windows perspective.

I do wonder though: Is there anything in how you're using these functions that forces this to need `TlsGetValue2` in the first place? Could this be implemented via `__thread` and `__declspec(thread)` instead? That would eliminate the library call entirely.

mcfi:
> > @Orvid could you help to take a look here as well?
> 
> This looks good from my generic-windows perspective.
> 
> I do wonder though: Is there anything in how you're using these functions that forces this to need `TlsGetValue2` in the first place? Could this be implemented via `__thread` and `__declspec(thread)` instead? That would eliminate the library call entirely.

That's a good point, Orvid. Although __declspec(thread) could be used here, it has some limitations on Windows versions before Vista, see https://learn.microsoft.com/en-us/cpp/parallel/thread-local-storage-tls?view=msvc-170. I'm not sure if any application using jemalloc supports needs to run on Vista.. Regardless, moving to __declspec(thread) could be realized in a separate PR if the jemalloc community decides to drop support for pre-vista OSes.

interwq:
Thanks @Orvid !

@mcfi IIUC, __declspec(thread) doesn't require the Get / SetLastError ops, and can replace the TlsGetValue approaches? If so, would you be interested in taking that part on as well? We can merge this PR first.

Also it seems `__declspec(thread)` will behave similar to the Linux TLS, as in the TLS space is pre-reserved with each thread, and will not allocate on get? If so, the below flag can be changed as well:
https://github.com/jemalloc/jemalloc/blob/5b72ac098abce464add567869d082f2097bd59a2/include/jemalloc/internal/tsd_win.h#L117

src/tsd.c
interwq: Thanks for the detailed answers @mcfi. 

> Speaking of that, is there any destruction code for jemalloc, if so, I can add a FreeLibrary call there to match LoadLibrary.

No library destruction support in jemalloc currently.

mcfi:
> Thanks @Orvid !
> 
> @mcfi IIUC, __declspec(thread) doesn't require the Get / SetLastError ops, and can replace the TlsGetValue approaches? If so, would you be interested in taking that part on as well? We can merge this PR first.
> 
> Also it seems `__declspec(thread)` will behave similar to the Linux TLS, as in the TLS space is pre-reserved with each thread, and will not allocate on get? If so, the below flag can be changed as well:
> 
> https://github.com/jemalloc/jemalloc/blob/5b72ac098abce464add567869d082f2097bd59a2/include/jemalloc/internal/tsd_win.h#L117

Sure, I can take on that after this PR gets merged first. :)

mcfi:
@interwq , looks like static analysis found some potential issues, but none of them was relevant to this PR

image

guangli-dai:
> @interwq , looks like static analysis found some potential issues, but none of them was relevant to this PR

It is fixed [here](https://github.com/jemalloc/jemalloc/pull/2698). Can you rebase? FYI, travis CI is not working now as well, need to wait for its recovery.

interwq:
@mcfi thanks for checking the CI signals. Yes that was a separate issue unrelated to this PR. It's fixed in https://github.com/jemalloc/jemalloc/pull/2698

Since this PR already had Travis run passed, I'm going to merge as is, instead of wait for Travis CI to recover.

PR #2586, opened 2024-01-02
Add support for namespace pids in heap profile names

hodgesds:
This change adds support for writing pid namespaces to the filename of a heap profile. When running with namespaces pids may reused across namespaces and if mounts are shared where profiles are written there is not a great way to differentiate profiles between pids. I couldn't get the `check-formatting.sh` script to work well as it depends on a rather older version of `clang-format` and my version `12.0.0` produced rather difference values.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1080,6 +1080,28 @@ enable_shared="1"
 )
 AC_SUBST([enable_shared])
 
+dnl Enable pid namespace support
+AC_CHECK_FUNCS([unshare])
+AC_ARG_ENABLE([pid_namespace],
+  [AS_HELP_STRING([--enable-pid-namespace], [Enable pid namespace support])],
+if test "x$enable_pid_namespace" = "xno" ; then
+  enable_pid_namespace="0"
+else
+  if test "x$ac_cv_func_unshare" != x; then
+    enable_pid_namespace="1"
+  else
+    enable_pid_namespace="0"
+  fi
+fi
+,
+enable_pid_namespace="0"
+)
+AC_SUBST([enable_pid_namespace])
+if test "x$enable_pid_namespace" = "x1"; then
+  AC_DEFINE([JEMALLOC_PID_NAMESPACE], [1], [Enable pid namespace support])
+fi
+
+
 dnl Enable static libs
 AC_ARG_ENABLE([static],
   [AS_HELP_STRING([--enable-static], [Build static libaries])],
@@ -2814,6 +2836,7 @@ AC_MSG_RESULT([static libs        : ${enable_static}])
 AC_MSG_RESULT([autogen            : ${enable_autogen}])
 AC_MSG_RESULT([debug              : ${enable_debug}])
 AC_MSG_RESULT([stats              : ${enable_stats}])
+AC_MSG_RESULT([pid_namespace      : ${enable_pid_namespace}])
 AC_MSG_RESULT([experimental_smallocx : ${enable_experimental_smallocx}])
 AC_MSG_RESULT([prof               : ${enable_prof}])
 AC_MSG_RESULT([prof-libunwind     : ${enable_prof_libunwind}])
--- a/src/pid_namespace.c
+++ b/src/pid_namespace.c
@@ -1,0 +1,20 @@
+#include "jemalloc/internal/pid_namespace_externs.h"
+
+#ifdef JEMALLOC_PID_NAMESPACE
+#include <sys/stat.h>
+
+static const char* PID_NAMESPACE_PATH = "/proc/self/ns/pid";
+static const char* PID_NAMESPACE_SEP = "pid:[";
+
+ssize_t pid_namespace() {
+  char buf[PATH_MAX + 1];
+  ssize_t linklen = readlink(PID_NAMESPACE_PATH, buf, PATH_MAX);
+  if (linklen == -1) {
+    return 0;
+  }
+  // Trim the trailing "]"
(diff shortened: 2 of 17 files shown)


src/prof_sys.c
```diff
@@ -711,26 +723,68 @@ prof_dump_filename(tsd_t *tsd, char *filename, char v, uint64_t vseq) {
 
 	assert(tsd_reentrancy_level_get(tsd) == 0);
 	const char *prefix = prof_prefix_get(tsd_tsdn(tsd));
+#ifdef JEMALLOC_PID_NAMESPACE
+	bool pid_ns_enabled = prof_pid_namespace_enabled(tsd_tsdn(tsd));
+#endif
 
 	if (vseq != VSEQ_INVALID) {
-	        /* "<prefix>.<pid>.<seq>.v<vseq>.heap" */
+#ifdef JEMALLOC_PID_NAMESPACE
```
hodgesds, line 717: Not super happy with how verbose this is with the ifdefs, would be open to cleaner implementations.

configure.ac
```diff
@@ -1080,6 +1080,28 @@ enable_shared="1"
...
 AC_SUBST([enable_shared])
 
+dnl Enable pid namespace support
+AC_CHECK_FUNCS([unshare])
+AC_ARG_ENABLE([pid_namespace],
+  [AS_HELP_STRING([--enable-pid-namespace], [Enable pid namespace support])],
+if test "x$enable_pid_namespace" = "xno" ; then
+  enable_pid_namespace="0"
+else
+  if test "x$ac_cv_func_unshare" != x; then
```
hodgesds, line 1090: Not sure if this check would be valid if doing cross compilation.

src/prof_sys.c
```diff
@@ -22,9 +22,15 @@
 #define _Unwind_Backtrace JEMALLOC_TEST_HOOK(_Unwind_Backtrace, test_hooks_libc_hook)
 #endif
 
+#ifdef JEMALLOC_PID_NAMESPACE
```
interwq: nit: no need for the ifdef here since inside the header everything is already guarded under `JEMALLOC_PID_NAMESPACE`

Also a single blank line below would be fine.

src/prof_sys.c
```diff
@@ -22,9 +22,15 @@
...
+#include "jemalloc/internal/pid_namespace_externs.h"
+#endif
+
+
 /******************************************************************************/
 
 malloc_mutex_t prof_dump_filename_mtx;
+malloc_mutex_t prof_pid_namespace_mtx;
```
interwq: No need for the mutex since opt_prof_pid_namespace is not allowed to change (unlike the filename prefix)

src/pid_namespace.c
```diff
@@ -0,0 +1,15 @@
+#include "jemalloc/internal/pid_namespace_externs.h"
+
+#ifdef JEMALLOC_PID_NAMESPACE
+#include <sys/stat.h>
+
+static const char* PID_NAMESPACE_PATH = "/proc/self/ns/pid";
+
+ssize_t pid_namespace() {
+  struct stat statbuf;
+  if (stat(PID_NAMESPACE_PATH, &statbuf) == -1) {
```
interwq: Maybe try an alternative way of reading the namespace (will readlink / readlinkat work)? One concern is that if some implementation of `stat` can allocate, then it would form a circular dependency (malloc->prof_dump->stat->malloc). While we already rely on readlink not allocating: https://github.com/jemalloc/jemalloc/blob/e4817c8d89a2a413e835c4adeab5c5c4412f9235/src/jemalloc.c#L997

test/unit/prof_pid_namespace.c
```diff
@@ -0,0 +1,76 @@
...
+}
+
+TEST_BEGIN(test_pid_namespace_dump) {
+	bool active;
+	void *p;
+
+	const char *test_prefix = TEST_PREFIX;
+
+	test_skip_if(!config_prof);
+	// test_skip_if(!config_enable_pid_namespace);
```
interwq: Not needed

interwq (commented):
Thanks for the PR @hodgesds ! A very useful feature.

src/pid_namespace.c
hodgesds: Let me double check that, good callout!

hodgesds:
I looked briefly if something similar would work for BSD jails and logging the jail id and it didn't seem like it would be that different. Not sure if it is worth adding at the same time or separately though.

davidtgoldblatt:
Why a config-time option for this? It seems like it could be purely opt-time, and that'd give one less dimension to have to test on that requires rebuilding.

hodgesds:
> Why a config-time option for this? It seems like it could be purely opt-time, and that'd give one less dimension to have to test on that requires rebuilding.

That's a fair callout. I think mostly as a mechanism for ease of downstream tools that process heap profiles that depend on the file name format. I guess maybe it is a moot point as using this feature would require recompiling anyways. Having to recompile to change the heap profile format though seems a bit excessive. Since I'm not a maintainer though I defer that decision :)

nullptr0-0:
change has been re-introduced in PR #2636 as discussed offline. @interwq

hodgesds:
> change has been re-introduced in PR #2636 as discussed offline. @interwq

Yeah, that works... thanks for picking this up!

interwq:
@hodgesds FYI the PR was tweaked a bit (removed the config time option) and merged as #2636 under your name. Thanks for the patch!

PR #2587, opened 2024-01-03
When safety_check_fail, also embed hint msg in the abort function name

interwq:
because there are cases only logging crash stack traces.

### DIFF
--- a/src/safety_check.c
+++ b/src/safety_check.c
@@ -20,6 +20,20 @@ void safety_check_set_abort(safety_check_abort_hook_t abort_fn) {
 	safety_check_abort = abort_fn;
 }
 
+/*
+ * In addition to malloc_write, also embed hint msg in the abort function name
+ * because there are cases only logging crash stack traces.
+ */
+static void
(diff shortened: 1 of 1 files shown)


src/safety_check.c
```diff
@@ -20,6 +20,20 @@ void safety_check_set_abort(safety_check_abort_hook_t abort_fn) {
 	safety_check_abort = abort_fn;
 }
 
+/*
+ * In addition to malloc_write, also embed hint msg in the abort function name
+ * because there are cases only logging crash stack traces.
+ */
+static void
+safety_check_detected_heap_corruption___run_address_sanitizer_build_to_debug(const char *buf) {
```
interwq, line 28: @davidtgoldblatt @guangli-dai @notdevil open to suggestions here

interwq:
(The CI failure will be fixed by #2581)

davidtgoldblatt:
A++ function naming, would review again.

guangli-dai (approved):
Great naming!

PR #2593, opened 2024-01-16
HPA: Allow frequent reused alloc to bypass the slab_max_alloc limit

interwq:
These requests do not concern internal fragmentation with huge pages, since the entire range is expected to be accessed.

For the context, this is only necessary for the time being, to better support larger page sizes (`--lg-page`). Once we change how size classing is done (cc: @guangli-dai), HPA will handle all sizes by default.

### DIFF
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -12,7 +12,8 @@ static edata_t *hpa_alloc(tsdn_t *tsdn, pai_t *self, size_t size,
     size_t alignment, bool zero, bool guarded, bool frequent_reuse,
     bool *deferred_work_generated);
 static size_t hpa_alloc_batch(tsdn_t *tsdn, pai_t *self, size_t size,
-    size_t nallocs, edata_list_active_t *results, bool *deferred_work_generated);
+    size_t nallocs, edata_list_active_t *results, bool frequent_reuse,
+    bool *deferred_work_generated);
 static bool hpa_expand(tsdn_t *tsdn, pai_t *self, edata_t *edata,
     size_t old_size, size_t new_size, bool zero, bool *deferred_work_generated);
 static bool hpa_shrink(tsdn_t *tsdn, pai_t *self, edata_t *edata,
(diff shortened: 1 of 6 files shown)


nullptr0-0 (approved):

guangli-dai:
LGTM! 

Not directly relevant to the change, but one quick question when checking the sec allocation stack trace. The comment [here](https://github.com/jemalloc/jemalloc/blob/b1792c80d2870c87af79d64bcca844d19345412d/src/sec.c#L194) is not accurate for now? Nothing to cache could also happen when hpa refuses to serve the request, i.e., size larger than threshold and not frequent_reuse?

guangli-dai (approved):

src/hpa.c
```diff
@@ -712,14 +715,22 @@ hpa_from_pai(pai_t *self) {
...
+    bool *deferred_work_generated) {
 	assert(nallocs > 0);
 	assert((size & PAGE_MASK) == 0);
 	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
 	    WITNESS_RANK_CORE, 0);
 	hpa_shard_t *shard = hpa_from_pai(self);
 
-	if (size > shard->opts.slab_max_alloc) {
+	/*
+	 * Allow frequent reused alloc to bypass the slab_max_alloc limit, as
```
ot: Should we use a more informative parameter name than `frequent_reuse`? In and of itself, a size being "frequently reused" does not imply no internal fragmentation; here the information we're conveying is that this request comes from the bins so there is no intrinsic slack in the allocation, because it is sized accurately.

So maybe something like `exact_size` or `no_slack`?

interwq: Agreed the `frequent_reuse` is rather not explicit. However it's already passed down to the page allocator level (the page alloc interface's alloc function like hpa_alloc added it for a different reason). So renaming only here would make it a bit inconsistent.

Given that it's only necessary until we have the full size support (the new size classing strategy by Guangli), I'd say keep the current naming for now, but let me add more details in the comments.

interwq:
@guangli-dai re: the comments in sec_batch_fill. Yeah it's only considering the case where hpa is "fully enabled". When size is over max_alloc, hpa will return NULL as well. Although in that case, one thing we can do is returning another value to differentiate OOM from other reasons like max_alloc limited. This can help sanity checks at higher level, but honestly doesn't look super helpful for now.

src/hpa.c
```diff
@@ -712,14 +715,26 @@ hpa_from_pai(pai_t *self) {
...
+	/*
+	 * frequent_use here indicates this request comes from the arena bins,
+	 * in which case it will be split into slabs, and therefore there is no
+	 * intrinsic slack in the allocation (the entire range of allocated size
+	 * will be accessed).
+	 *
+	 * In this case bypass the slab_max_alloc limit (if still within the
+	 * huge page size).  These requests do not concern internal
+	 * fragmentation with huge pages (again, the full size will be used).
+	 */
```
interwq, line 735: Updated the comments.

PR #2601, opened 2024-01-31
Print out all malloc_conf settings in stats

nullptr0-0:
- Add the malloc conf settings under the existing Runtime Option Settings in stats output
- Print out the value of global variable malloc_conf, filename referenced by the symbolic link /etc/malloc.conf, and the value of env variable MALLOC_CONF in order.

With the symbolic link and the environment variable set,
     lrwxrwxrwx 1 root root 10 Feb 12 11:28 /etc/malloc.conf -> abort:true
     MALLOC_CONF=prof:true,prof_active:true,lg_prof_sample:0,prof_stats:true,stats_print:true

Sample Output:
___ Begin jemalloc statistics ___
Version: "5.3.0-148-g287ac1775ddd07614e3f0ffddcb418d1f97e711b"
Build-time option settings
  config.cache_oblivious: true
  config.debug: true
  config.fill: true
  config.lazy_lock: false
  config.malloc_conf: ""
  config.opt_safety_checks: true
  config.prof: true
  config.prof_libgcc: true
  config.prof_libunwind: false
  config.stats: true
  config.utrace: false
  config.xmalloc: false
Run-time option settings
  Global variable malloc_conf: ""
  Symbolic link malloc.conf: "abort:true"
  Environment variable MALLOC_CONF: "prof:true,prof_active:true,lg_prof_sample:0,prof_stats:true,stats_print:true"
  opt.abort: true
  opt.abort_conf: true
  opt.cache_oblivious: true
  opt.confirm_conf: false
  opt.retain: true
  opt.dss: "secondary"
  opt.narenas: 320
  opt.percpu_arena: "disabled"
  opt.oversize_threshold: 8388608
  opt.hpa: false
  opt.hpa_slab_max_alloc: 65536
  opt.hpa_hugification_threshold: 1992294
.....

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -159,6 +159,10 @@ CTL_PROTO(opt_prof_sys_thread_name)
 CTL_PROTO(opt_prof_time_res)
 CTL_PROTO(opt_lg_san_uaf_align)
 CTL_PROTO(opt_zero_realloc)
+CTL_PROTO(opt_malloc_conf_symlink)
+CTL_PROTO(opt_malloc_conf_env_var)
+CTL_PROTO(opt_malloc_conf_global_var)
+CTL_PROTO(opt_malloc_conf_global_var_2_conf_harder)
 CTL_PROTO(tcache_create)
 CTL_PROTO(tcache_flush)
 CTL_PROTO(tcache_destroy)
@@ -426,6 +430,14 @@ static const ctl_named_node_t	config_node[] = {
 	{NAME("xmalloc"),	CTL(config_xmalloc)}
 };
 
+static const ctl_named_node_t opt_malloc_conf_node[] = {
+	{NAME("symlink"),	CTL(opt_malloc_conf_symlink)},
+	{NAME("env_var"),	CTL(opt_malloc_conf_env_var)},
+	{NAME("global_var"),	CTL(opt_malloc_conf_global_var)},
+	{NAME("global_var_2_conf_harder"),
+	    CTL(opt_malloc_conf_global_var_2_conf_harder)}
+};
+
 static const ctl_named_node_t opt_node[] = {
 	{NAME("abort"),		CTL(opt_abort)},
 	{NAME("abort_conf"),	CTL(opt_abort_conf)},
@@ -502,7 +514,8 @@ static const ctl_named_node_t opt_node[] = {
 	{NAME("lg_san_uaf_align"),	CTL(opt_lg_san_uaf_align)},
 	{NAME("zero_realloc"),	CTL(opt_zero_realloc)},
 	{NAME("debug_double_free_max_scan"),
-		CTL(opt_debug_double_free_max_scan)}
+		CTL(opt_debug_double_free_max_scan)},
+	{NAME("malloc_conf"),	CHILD(named, opt_malloc_conf)}
 };
 
 static const ctl_named_node_t	tcache_node[] = {
@@ -2229,6 +2242,17 @@ CTL_RO_NL_CGEN(config_uaf_detection, opt_lg_san_uaf_align,
     opt_lg_san_uaf_align, ssize_t)
 CTL_RO_NL_GEN(opt_zero_realloc,
     zero_realloc_mode_names[opt_zero_realloc_action], const char *)
+
+/* malloc_conf options */
+CTL_RO_NL_CGEN(opt_malloc_conf_symlink, opt_malloc_conf_symlink,
+    opt_malloc_conf_symlink, const char *)
+CTL_RO_NL_CGEN(opt_malloc_conf_env_var, opt_malloc_conf_env_var,
+    opt_malloc_conf_env_var, const char *)
+CTL_RO_NL_CGEN(je_malloc_conf, opt_malloc_conf_global_var, je_malloc_conf,
+    const char *)
+CTL_RO_NL_CGEN(je_malloc_conf_2_conf_harder,
+    opt_malloc_conf_global_var_2_conf_harder, je_malloc_conf_2_conf_harder,
+    const char *)
 
 /******************************************************************************/
 
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -57,6 +57,9 @@ const char	*je_malloc_conf_2_conf_harder
 #endif
     ;
 
+const char *opt_malloc_conf_symlink = NULL;
+const char *opt_malloc_conf_env_var = NULL;
+
 bool	opt_abort =
 #ifdef JEMALLOC_DEBUG
     true
@@ -955,7 +958,7 @@ malloc_slow_flag_init(void) {
 #define MALLOC_CONF_NSOURCES 5
 
 static const char *
-obtain_malloc_conf(unsigned which_source, char buf[PATH_MAX + 1]) {
+obtain_malloc_conf(unsigned which_source, char readlink_buf[PATH_MAX + 1]) {
 	if (config_debug) {
 		static unsigned read_source = 0;
 		/*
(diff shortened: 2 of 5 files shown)


src/jemalloc.c
```diff
@@ -57,6 +57,9 @@ const char	*je_malloc_conf_2_conf_harder
 #endif
     ;
 
+char malloc_conf_file[PATH_MAX + 1];
+char malloc_conf_env_var[PATH_MAX + 1];
```
interwq: Discussed offline with Sherry about how to avoid the two static buffers here.

nullptr0-0: updated!

src/stats.c
```diff
@@ -1430,6 +1430,43 @@ stats_general_print(emitter_t *emitter) {
...
+	 * - (Included) The string pointed to by the global variable malloc_conf
+	 * - (Included) The “name” of the file referenced by the symbolic link
+	 *     named /etc/malloc.conf
+	 * - (Included) The value of the environment variable MALLOC_CONF
+	 * - (Optional, Internal) The string pointed to by the global variable
+	 *     malloc_conf_2_conf_harder, which is hidden from the public.
+	 *
+	 * Note: The outputs are strictly ordered by priorities (low -> high).
+	 *
+	 * */
```
guangli-dai: nit: just `*/`

guangli-dai (commented):
Overall LGTM!

src/ctl.c
```diff
@@ -933,7 +946,8 @@ static const ctl_named_node_t	root_node[] = {
 	{NAME("arenas"),	CHILD(named, arenas)},
 	{NAME("prof"),		CHILD(named, prof)},
 	{NAME("stats"),		CHILD(named, stats)},
-	{NAME("experimental"),	CHILD(named, experimental)}
+	{NAME("experimental"),	CHILD(named, experimental)},
+	{NAME("malloc_conf"),	CHILD(named, malloc_conf)}
```
interwq: Let's move this to the `opt` namespace. That would also make it clearer that this is read-only (as everything under `opt` is).

src/ctl.c
```diff
@@ -920,6 +924,15 @@ static const ctl_named_node_t experimental_node[] = {
 	{NAME("thread"),	CHILD(named, experimental_thread)}
 };
 
+static const ctl_named_node_t malloc_conf_node[] = {
+	{NAME("file"),		CTL(malloc_conf_file)},
+	{NAME("env_var"),	CTL(malloc_conf_env_var)},
+	{NAME("global_var_malloc_conf"),
```
interwq: The `_malloc_conf` suffix seems redundant. Maybe `global_var` and `global_var_2_conf_harder`? So the path would look like `opt.malloc_conf.global_var`

src/jemalloc.c
```diff
@@ -1026,6 +1029,7 @@ obtain_malloc_conf(unsigned which_source, char buf[PATH_MAX + 1]) {
 			 * Do nothing; opts is already initialized to the value
 			 * of the MALLOC_CONF environment variable.
 			 */
+			malloc_conf_env_var = ret;
```
interwq: nit: update the comments above as well (the do nothing part isn't true anymore).

src/jemalloc.c
```diff
@@ -1783,13 +1787,14 @@ malloc_conf_init_check_deps(void) {
 }
 
 static void
-malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS]) {
+malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
+    char readlink_buf[PATH_MAX + 1]) {
 	const char *opts_cache[MALLOC_CONF_NSOURCES] = {NULL, NULL, NULL, NULL,
 		NULL};
-	char buf[PATH_MAX + 1];
+
```
interwq, line 1790: nit: unnecessary blank line

src/jemalloc.c
```diff
@@ -1855,7 +1860,8 @@ malloc_init_hard_a0_locked(void) {
 	if (config_prof) {
 		prof_boot0();
 	}
-	malloc_conf_init(&sc_data, bin_shard_sizes);
+	char readlink_buf[PATH_MAX + 1];
```
interwq, line 1858: Explicitly set `readlink_buf[0]` = `\0` instead of relying on the downstream to do so, given that we need `strlen` on it later.

src/jemalloc.c
```diff
@@ -1949,6 +1955,16 @@ malloc_init_hard_a0_locked(void) {
...
 
+	void *readlink_allocated = NULL;
+	size_t len = strlen(readlink_buf);
+	if (len > 0) {
+		readlink_allocated = a0ialloc(len + 1, false, true);
+	}
+	if (readlink_allocated != NULL) {
+		memcpy(readlink_allocated, readlink_buf, len + 1);
+		malloc_conf_file = readlink_allocated;
+	}
```
interwq, line 1961: Refactor a bit:
```
size_t readlink_buf_len = strlen(readlink_buf);
if (readlink_buf_len > 0) {
    void *readlink_allocated = a0ialloc(len + 1, false, true);
    memcpy(readlink_allocated, readlink_buf, len + 1);
    malloc_conf_file = readlink_allocated;
}
```

src/stats.c
```diff
@@ -1430,6 +1430,43 @@ stats_general_print(emitter_t *emitter) {
...
+	 * malloc_conf.
+	 *
+	 * Sources are documented in https://jemalloc.net/jemalloc.3.html#tuning
+	 * - (Not Included Here) The string specified via --with-malloc-conf,
+	 *     which is already printed out above as config.malloc_conf
+	 * - (Included) The string pointed to by the global variable malloc_conf
+	 * - (Included) The “name” of the file referenced by the symbolic link
+	 *     named /etc/malloc.conf
+	 * - (Included) The value of the environment variable MALLOC_CONF
+	 * - (Optional, Internal) The string pointed to by the global variable
```
interwq: nit: replace the term `internal` with unofficial or undocumented

src/stats.c
```diff
@@ -1430,6 +1430,43 @@ stats_general_print(emitter_t *emitter) {
...
+	 *
+	 * */
+#define MALLOC_CONF_WRITE(name, message)					\
+	if (je_mallctl("malloc_conf."name, (void *)&cpv, &cpsz, NULL, 0) !=	\
+	    0) {								\
+		cpv = "";							\
+	}									\
+	emitter_kv(emitter, name, message, emitter_type_string,	&cpv);
+
+	emitter_dict_begin(emitter, "malloc_conf", "Tuning settings");
```
interwq: nit: "Runtime settings"

src/ctl.c
```diff
@@ -920,6 +924,15 @@ static const ctl_named_node_t experimental_node[] = {
 	{NAME("thread"),	CHILD(named, experimental_thread)}
 };
 
+static const ctl_named_node_t malloc_conf_node[] = {
+	{NAME("file"),		CTL(malloc_conf_file)},
```
interwq: Maybe `symlink` instead of `file`?

src/jemalloc.c
nullptr0-0: Does `a0ialloc` guarantee non-null result? memcpy on NULL will result in UB?

interwq: Nope you will still want to guard memcpy with the non-NULL check

src/jemalloc.c
```diff
@@ -1855,7 +1860,8 @@ malloc_init_hard_a0_locked(void) {
 	if (config_prof) {
 		prof_boot0();
 	}
-	malloc_conf_init(&sc_data, bin_shard_sizes);
+	char readlink_buf[PATH_MAX + 1];
+	malloc_conf_init(&sc_data, bin_shard_sizes, readlink_buf);
```
guangli-dai, line 1860: Question: what if the setting is a string longer than PATH_MAX? If so, would the overflow be a concern? Can we add a test to confirm the behavior under this scenario is as expected?

interwq: We explicitly add the terminator here: https://github.com/jemalloc/jemalloc/pull/2601/files#diff-a4cb09e38cfec8141b07c291f731a8e01a17412568a852884fd921e8e521766bR1015, which means we will truncate the buffer in case of > PATH_MAX which is by design.

re: tests -- I'll leave it to Sherry to decide if a test is worth it (IMO it's probably too much trouble messing with the symlink w/o affecting other stuff in the system)

interwq (approved):
@nullptr0-0 can you please update the sample output in the summary

nullptr0-0:
@interwq Updated the new output! The malloc conf settings are moved to the existing Run-time option settings. wdyt?

image

PR #2606, opened 2024-02-13
Allow zero sized memalign to pass.

interwq:
Instead of failing on assertions.  Previously the same change was made for posix_memalign and aligned_alloc (#1554).  Make memalign behave the same way even though it's obsolete.

Reported by Conrad Meyer

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -3052,6 +3052,7 @@ je_memalign(size_t alignment, size_t size) {
 	static_opts_init(&sopts);
 	dynamic_opts_init(&dopts);
 
+	sopts.bump_empty_aligned_alloc = true;
 	sopts.min_alignment = 1;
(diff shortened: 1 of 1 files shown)


guangli-dai:
Maybe also add the test for memalign?

interwq:
> Maybe also add the test for memalign?

Usually I'd say yes. However `memalign` has been long obsolete, and we currently don't have any existing test around it. `test/integration/posix_memalign` covers the zero size case (which uses the same `bump_empty_aligned_alloc` mechanism): https://github.com/jemalloc/jemalloc/blob/f96010b7fa8ce5f83802144bdebf2bb7a6679649/test/integration/posix_memalign.c#L88 Maybe let's not add new tests in this case?

guangli-dai (approved):
Sure, that sounds good to me.

PR #2607, opened 2024-02-15
Fix Cirrus CI.

interwq:
13.0-RELEASE does not exist anymore.  "The resource 'projects/freebsd-org-cloud-dev/global/images/family/freebsd-13-0' was not found"


nullptr0-0 (approved):
lgtm.

PR #2611, opened 2024-03-04
Replace unsigned induction variable with size_t in background_threads

XChy:
Hi, everyone! 
I'm a compiler developer and work on detecting missed compiler optimization opportunities in real-world projects. Recently, a missed optimization opportunity is detected in jemalloc project. This is hard to solve in compilers, but easy to solve in original project, so I post a pr to fix it here.

Since the type of `unsigned` induction is inconsistent with `size_t` variable `max_background_threads`, LLVM vectorizes it but produces weird IR and GCC fails to recognize memset pattern.

This patch avoids unnecessary vectorizations in clang and missed recognition of memset in gcc. See also [godbolt example](https://godbolt.org/z/aoeMsjr4c) and the upstream issue https://github.com/llvm/llvm-project/issues/83724.

This patch brings semantic change actually (original program may fall into infinite loop), but I assume we don't want infinite loop here. 

If there is any problem, feel free to correct it.

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -580,7 +580,7 @@ background_threads_enable(tsd_t *tsd) {
 
 	VARIABLE_ARRAY(bool, marked, max_background_threads);
 	unsigned nmarked;
-	for (unsigned i = 0; i < max_background_threads; i++) {
+	for (size_t i = 0; i < max_background_threads; i++) {
 		marked[i] = false;
 	}
 	nmarked = 0;


interwq:
Thanks for the patch! If the mismatch there is causing troubles for the compilers, is there a warning flag we can enable to check for these cases? Otherwise we can easily introduce another one later.

 cc: @davidtgoldblatt

davidtgoldblatt:
LGTM!

davidtgoldblatt:
(would be very curious in hearing more about the broader project, too -- seems interesting)

XChy:
> If the mismatch there is causing troubles for the compilers, is there a warning flag we can enable to check for these cases? Otherwise we can easily introduce another one later.

It sounds reasonable to me, but flags like [-Wconversion](https://clang.llvm.org/docs/DiagnosticsReference.html#id190) don't help with unsigned compares now. And actually, most diagnostic warnings are about correctness instead of efficiency(At least I didn't hear about that). I think I could post an upstream issue about it if you like.

XChy:
> (would be very curious in hearing more about the broader project, too -- seems interesting)

Yeah, the project is an in-progress work following works like various superoptimizers and [DCE-based method](https://dl.acm.org/doi/10.1145/3503222.3507764) that detect missed optimization in compilers. We design a different method so that we can detect in real code.

Actually, we have detected several missed optimizations in jemalloc: https://github.com/llvm/llvm-project/issues/79696 , https://github.com/llvm/llvm-project/issues/82177 (eaiser to solve in compilers, thus no transferred downstream issues). Also, we detect in projects like Linux and QEMU.

interwq:
Thanks for the patch and sharing the context @XChy !

PR #2615, opened 2024-03-09
Update acitons/checkout and actions/upload-artifact to v4

mchoo7:


interwq:
Thanks for the patch!

PR #2620, opened 2024-03-22
Set errno to ENOMEM on rallocx() OOM failures

arter97:
realloc() and rallocx() shares path, and realloc() should set errno to ENOMEM upon OOM failures.

Fixes: ee961c23100e ("Merge realloc and rallocx pathways.")


interwq:
Thanks for the fix @arter97 !

PR #2628, opened 2024-03-28
Fix the VM over-reservation on aarch64 w/ larger pages.

### LINKED ISSUE #2624: JEMalloc on ARM uses 10x more VIRT than the same work on Intel..?
I received a bug report that our core dumping procedure on ARM was many times slower than on Intel.  A quick investigation shows that the VIRT memory for our processes on ARM is ~11x larger than the same workload on Intel, although the resident sizes are equivalent.  Here's the result of top on Intel:
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
  92905 pds       20   0 1721744   1.2g  22716 S   0.0   3.8   0:50.32 myprog
  92945 pds       20   0 1605004   1.0g  32900 S   8.7   3.3   0:51.80 myprog
```
Here's the top output for ARM, with the same workload:
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 160440 pds       20   0   17.8g   1.1g  18112 S   0.0   6.9   0:51.25 myprog
 160477 pds       20   0   17.8g 982.2m  29824 S   0.7   6.2   0:53.46 myprog
```

Here are the top-line jemalloc stats retrieved from both (I addded the commas myself for readability :) ):
```
Intel:                                      ARM:
active:                    1,069,473,792    active:                    1,027,932,160
mapped:                    1,137,229,824    mapped:                   10,157,752,320
retained:                    289,882,112    retained:                  8,766,947,328
base:                         24,828,112    base:                         18,051,512
internal:                        663,712    internal:                      1,245,184
metadata_thp:                          0    metadata_thp:                          0
tcache_bytes:                  2,685,328    tcache_bytes:                  4,754,664
tcache_stashed_bytes:                  0    tcache_stashed_bytes:                  0
resident:                  1,097,093,120    resident:                  1,049,624,576
abandoned_vm:                          0    abandoned_vm:                          0
extent_avail:                        187    extent_avail:                          9
```
I can get full stats if useful.

For reference I rebuilt my ARM program using system alloc (glibc) and I got similar numbers to Intel from top (obviously no jemalloc stats):
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 233542 pds       20   0 2323968   1.0g  17280 S   0.0   6.5   1:17.63 myprog
 233578 pds       20   0 2125632 937216  26944 S   4.1   5.8   1:14.95 myprog
```

The important details:
* Both are using JEMalloc 5.3.0 but I pulled down the latest Git `dev` HEAD version as of this week and tried it on ARM and got the same behavior.
* The ARM system is running Red Hat EL 8, which means it's using 64k page size.  The Intel system is running Ubuntu 20.04 LTS, with (obviously) 4k page size.  Unfortunately I don't have an easy way to test on ARM with 4k page sizes.
* The Intel version is built with memory profiling enabled.  The ARM version is built with memory profiling _disabled_.

Is there a reason that we should expect the ARM implementation to have such a higher mapped (9x Intel) and retained (30x Intel!) values while most other values are similar (I guess tcache_bytes is also almost 2x but not sure if that's relevant)?

I posted something on Gitter but not sure if people hang out there anymore so I filed this.

interwq:
HUGEPAGE could be larger on some platforms (e.g. 512M on aarch64 w/ 64K pages),
in which case it would cause grow_retained / exp_grow to over-reserve VMs.

Resolves #2624

### DIFF
--- a/include/jemalloc/internal/pages.h
+++ b/include/jemalloc/internal/pages.h
@@ -27,6 +27,14 @@ extern size_t	os_page;
 #define HUGEPAGE	((size_t)(1U << LG_HUGEPAGE))
 #define HUGEPAGE_MASK	((size_t)(HUGEPAGE - 1))
 
+/*
+ * Used to validate that the hugepage size is not unexpectedly high.  The huge
+ * page features (HPA, metadata_thp) are primarily designed with a 2M THP size
+ * in mind.  Much larger sizes are not tested and likely to cause issues such as
+ * bad fragmentation or simply broken.
+ */
+#define HUGEPAGE_MAX_EXPECTED_SIZE ((size_t)(16U << 20))
+
 #if LG_HUGEPAGE != 0
 #  define HUGEPAGE_PAGES (HUGEPAGE / PAGE)
 #else
--- a/src/exp_grow.c
+++ b/src/exp_grow.c
@@ -3,6 +3,12 @@
 
 void
 exp_grow_init(exp_grow_t *exp_grow) {
-	exp_grow->next = sz_psz2ind(HUGEPAGE);
+	/*
+	 * Enforce a minimal of 2M grow, which is convenient for the huge page
+	 * use cases.  Avoid using HUGEPAGE as the value though, because on some
+	 * platforms it can be very large (e.g. 512M on aarch64 w/ 64K pages).
+	 */
+	const size_t min_grow = (size_t)2 << 20;
+	exp_grow->next = sz_psz2ind(min_grow);
 	exp_grow->limit = sz_psz2ind(SC_LARGE_MAXCLASS);
 }
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1040,18 +1040,14 @@ obtain_malloc_conf(unsigned which_source, char readlink_buf[PATH_MAX + 1]) {
 	return ret;
 }
 
-static void
-validate_hpa_settings(void) {
-	if (!hpa_supported() || !opt_hpa || opt_hpa_opts.dirty_mult == (fxp_t)-1) {
-		return;
-	}
+static bool
+validate_hpa_ratios(void) {
 	size_t hpa_threshold = fxp_mul_frac(HUGEPAGE, opt_hpa_opts.dirty_mult) +
 	    opt_hpa_opts.hugification_threshold;
 	if (hpa_threshold > HUGEPAGE) {
-		return;
+		return false;
 	}
 
-	had_conf_error = true;
 	char hpa_dirty_mult[FXP_BUF_SIZE];
 	char hugification_threshold[FXP_BUF_SIZE];
 	char normalization_message[256] = {0};
@@ -1078,6 +1074,24 @@ validate_hpa_settings(void) {
 	    "hpa_hugification_threshold_ratio: %s and hpa_dirty_mult: %s. "
 	    "These values should sum to > 1.0.\n%s", hugification_threshold,
 	    hpa_dirty_mult, normalization_message);
+
+	return true;
+}
+
+static void
+validate_hpa_settings(void) {
+	if (!hpa_supported() || !opt_hpa) {
+		return;
+	}
+	if (HUGEPAGE > HUGEPAGE_MAX_EXPECTED_SIZE) {
+		had_conf_error = true;
+		malloc_printf(
+		    "<jemalloc>: huge page size (%zu) greater than expected."
+		    "May not be supported or behave as expected.", HUGEPAGE);
+	}
+	if (opt_hpa_opts.dirty_mult != (fxp_t)-1 && validate_hpa_ratios()) {
+		had_conf_error = true;
+	}
 }
 
 static void
--- a/include/jemalloc/internal/base.h
+++ b/include/jemalloc/internal/base.h
@@ -6,6 +6,12 @@
 #include "jemalloc/internal/ehooks.h"
 #include "jemalloc/internal/mutex.h"
 
+/*
+ * Alignment when THP is not enabled.  Set to constant 2M in case the HUGEPAGE
+ * value is unexpected high (which would cause VM over-reservation).
+ */
+#define BASE_BLOCK_MIN_ALIGN ((size_t)2 << 20)
+
 enum metadata_thp_mode_e {
 	metadata_thp_disabled   = 0,
 	/*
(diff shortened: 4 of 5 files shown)


include/jemalloc/internal/pages.h
```diff
@@ -27,6 +27,9 @@ extern size_t	os_page;
 #define HUGEPAGE	((size_t)(1U << LG_HUGEPAGE))
 #define HUGEPAGE_MASK	((size_t)(HUGEPAGE - 1))
 
+/* Used to validate that the hugepage size is not unexpectedly high. */
+#define HUGEPAGE_MAX_EXPECTED_SIZE ((size_t)(16U << 20))
```
interwq, line 36: The value is picked based on the info here: https://wiki.debian.org/Hugepages#Huge_pages_sizes
```
Architecture	huge page size
arm64	4K, 2M and 1G (or 64K and 512M if one builds their own kernel with CONFIG_ARM64_64K_PAGES=y)
i386	4K and 4M (2M in PAE mode)
ia64	4K, 8K, 64K, 256K, 1M, 4M, 16M, 256M
ppc64	4K and 16M
```
as it looks the values beyond 16M are all too large and not expected.

interwq:
@guangli-dai @nullptr0-0 FYI this is the fix for the arm64 VM size issue.

src/jemalloc.c
```diff
@@ -1079,6 +1075,24 @@ validate_hpa_settings(void) {
...
+
+static void
+validate_hpa_settings(void) {
+	if (!hpa_supported() || !opt_hpa) {
+		return;
+	}
+	if (HUGEPAGE > HUGEPAGE_MAX_EXPECTED_SIZE) {
+		had_conf_error = true;
+		malloc_printf(
+		    "<jemalloc>: huge page size (%zu) greater than expected.\n",
```
guangli-dai: Can we also show `HUGEPAGE_MAX_EXPECTED_SIZE` in the error message so that users know how to configure more easily?

include/jemalloc/internal/pages.h
guangli-dai: Maybe also elaborate in the comments why too high a huge-page size is harmful?

interwq: Makes sense. Added more details.

src/jemalloc.c
interwq: I don't think we should show the MAX_EXPECTED_SIZE. It's just a way of us guessing that, that setting must not be supported well. Not really meaningful as it can be larger or smaller, and the current value is simply a best guess.

Also note that the huge page size isn't easily configurable. It's tied to hardware TLBs so there's only a handful of supported values. This msg is more saying "it's not supported and don't use it here", instead of "change some setting to a lower value so it could work".

src/exp_grow.c
```diff
@@ -3,6 +3,12 @@
...
 void
 exp_grow_init(exp_grow_t *exp_grow) {
-	exp_grow->next = sz_psz2ind(HUGEPAGE);
+	/*
+	 * Enforce a minimal of 2M grow, which is convenient for the huge page
+	 * use cases.  Avoid using HUGEPAGE as the value though, because on some
+	 * platforms it can be very large (e.g. 512M on aarch64 w/ 64K pages).
+	 */
+	const size_t min_grow = (size_t)2 << 20;
+	exp_grow->next = sz_psz2ind(min_grow);
```
madscientist, line 12: As a check I applied this change (but not the others) and used the default --with-lg-hugepage value (29 on my ARM systems).  Unfortunately, with that configuration I still see VIRT reported at 10G whereas when I set --with-lg-hugepage=21 (same as intel) I get an as-expected 1.3G VIRT, in my test.

Maybe there is someplace else that needs to be addressed, as well?

interwq: @madscientist good catch! Yes I missed the case in the base allocator which has a similar HUGEPAGE value hardcoded. See the changes in `base.c` to remedy that as well. The first commit in this PR should fix the issue properly now.

interwq:
@guangli-dai @nullptr0-0 ready for review

src/jemalloc.c
guangli-dai: I see. I was thinking adding notes like you added in the comments above with `MAX_EXPECTED_SIZE` included. Since it is a guessed value, I agree that there is no need to show it. But maybe we can make the warning more specific? Something like this:

`<jemalloc>: huge page size (%zu) greater than expected and not tested. It could cause unexpected problems.`

guangli-dai (commented):
Changes in the base allocator also look good to me. The only concern I have is whether we should warn users not to turn `metadata_thp` on if the hugepage size is large.

Also, the travis test seems to fail, but I cannot see detailed logs so don't know whether the signal is real.

src/jemalloc.c
interwq: Tweaked the wording.

interwq:
@guangli-dai 

> whether we should warn users not to turn metadata_thp on if the hugepage size is large.

I debated that as well. Decided it's safer there because the potential damage is more bounded (at most 1 THP per arena) and better understood (at worst should be perf issue instead of simply broken). So did not put a stop sign over there. After all it's an opt in feature.

guangli-dai:
> I debated that as well. Decided it's safer there because the potential damage is more bounded (at most 1 THP per arena) and better understood (at worst should be perf issue instead of simply broken). So did not put a stop sign over there. After all it's an opt in feature.
@interwq Agreed, let's keep it as is then.

src/exp_grow.c
madscientist: I've tried this version and it does seem to solve the problem!  I will be away from my system for a week (going diving!) so I won't be able to test further until I get back.  Thanks for the quick update and great help!

interwq: @madscientist Thank you again and have a great trip!

interwq:
The ppc64 CI failure is real. Still trying to figure out why.

interwq:
@madscientist FYI we haven't forgotten about this and certainly intend to land the fix. There is a ppc64 CI failure somehow triggered by this change. Still haven't figured out why. I might have to come back to this in a few weeks.

madscientist:
Thanks I appreciate the effort!

guangli-dai (approved):
Looks good! Since this diff adds a validation on whether `HUGEPAGE > HUGEPAGE_MAX_EXPECTED_SIZE` when hpa is enabled, we should also add this into the documentation for users to be aware of?

interwq:
@guangli-dai re documentation: the new HPA hasn't been officially introduced yet. We will want to make it feature complete like the size classing change, and then clean up the added knobs, before making it available.

PR #2631, opened 2024-04-03
Experimental calloc implementation with using memset() on larger sizes

nullptr0-0:
- Add a run-time option **opt_calloc_madvise_threshold** that controls when shoud memset be used to zero out memory, it's by default **0**, which means,
    - `usize >= opt_calloc_madvise_threshold` always stands, so `zero_override = zero`, the original code flow is unchanged.

- When **opt_calloc_madvise_threshold** is set to a value no smaller than `SC_LARGE_MINCLASS`, in this case when `usize` is between `SC_LARGE_MINCLASS` to `opt_calloc_madvise_threshold`, `zero_override` will be set to false, which means that we won't zero out memory inside pa_alloc() (which is using madvise()), and `memset()` will be used to do the zeroing.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -359,17 +359,38 @@ arena_extent_alloc_large(tsdn_t *tsdn, arena_t *arena, size_t usize,
 
 	bool guarded = san_large_extent_decide_guard(tsdn,
 	    arena_get_ehooks(arena), esize, alignment);
-	edata_t *edata = pa_alloc(tsdn, &arena->pa_shard, esize, alignment,
-	    /* slab */ false, szind, zero, guarded, &deferred_work_generated);
 
-	if (edata != NULL) {
-		if (config_stats) {
-			arena_large_malloc_stats_update(tsdn, arena, usize);
-		}
+	/*
+	 * - if usize >= opt_calloc_madvise_threshold,
+	 *     - pa_alloc(..., zero_override = zero, ...)
+	 * - otherwise,
+	 *     - pa_alloc(..., zero_override = false, ...)
+	 *     - use memset() to zero out memory if zero == true.
+	 */
+	bool zero_override = zero && (usize >= opt_calloc_madvise_threshold);
+	edata_t *edata = pa_alloc(tsdn, &arena->pa_shard, esize, alignment,
+	    /* slab */ false, szind, zero_override, guarded,
+	    &deferred_work_generated);
+
+	if (edata == NULL) {
+		return NULL;
 	}
 
-	if (edata != NULL && sz_large_pad != 0) {
+	if (config_stats) {
+		arena_large_malloc_stats_update(tsdn, arena, usize);
+	}
+	if (sz_large_pad != 0) {
 		arena_cache_oblivious_randomize(tsdn, arena, edata, alignment);
+	}
+	/*
+	 * This branch should be put after the randomization so that the addr
+	 * returned by edata_addr_get() has already be randomized,
+	 * if cache_oblivious is enabled.
+	 */
+	if (zero && !zero_override && !edata_zeroed_get(edata)) {
+		void *addr = edata_addr_get(edata);
+		size_t usize = edata_usize_get(edata);
+		memset(addr, 0, usize);
 	}
 
 	return edata;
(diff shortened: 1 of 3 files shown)


interwq:
> Add a build-time config config_experimental_calloc

This doesn't feel necessary to me. If you want a way to keep the current behavior unchanged, I'd say using the runtime option for that (e.g. if that is set to 0). If you are worried about the extra branch, that shouldn't be much of a concern since we are talking about medium sized calloc which isn't as heavily stressed. 

To be clear, eventually we should remove the runtime option as well, and either pick a safe default of rely on another setting such as oversize threshold.

nullptr0-0:
Verify the config is setting properly:
```
./autogen.sh 
experimental-calloc : 0
========================
 ~/jemalloc (calloc_optimization)]$ JEMALLOC_TEST_PREFIX="lldb" sh test/test.sh test/unit/a0
=== test/unit/a0 ===
(lldb) target create "test/unit/a0"
(lldb) breakpoint set --name jet_a0malloc
Breakpoint 2: where = a0`jet_a0malloc at jemalloc.c:334:9, address = 0x0000000000414550
(lldb) r
Process 3264687 launched
Process 3264687 stopped
* thread #1, name = 'a0', stop reason = breakpoint 2.1
    frame #0: 0x0000000000414550 a0`jet_a0malloc(size=1) at jemalloc.c:334:9
   331 
   332  void *
   333  a0malloc(size_t size) {
-> 334          return a0ialloc(size, false, true);
   335  }
   336 
   337  void
(lldb) p config_experimental_calloc
(const bool) false
(lldb) p opt_calloc_madvise_threshold
(size_t) 16384
```

```
./autogen.sh --enable-experimental-calloc
experimental-calloc : 1
=========================
 ~/jemalloc (calloc_optimization)]$ JEMALLOC_TEST_PREFIX="lldb" sh test/test.sh test/unit/a0
=== test/unit/a0 ===
(lldb) target create "test/unit/a0"
(lldb) breakpoint set --name jet_a0malloc
Breakpoint 1: where = a0`jet_a0malloc at jemalloc.c:334:9, address = 0x0000000000414690
(lldb) r
Process 3572194 launched
Process 3572194 stopped
* thread #1, name = 'a0', stop reason = breakpoint 1.1
    frame #0: 0x0000000000414690 a0`jet_a0malloc(size=1) at jemalloc.c:334:9
   331 
   332  void *
   333  a0malloc(size_t size) {
-> 334          return a0ialloc(size, false, true);
   335  }
   336 
   337  void
(lldb) p config_experimental_calloc
(const bool) true
(lldb) p opt_calloc_madvise_threshold
(size_t) 16384
```

src/jemalloc.c
```diff
@@ -1453,6 +1455,9 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			    "debug_double_free_max_scan", 0, UINT_MAX,
 			    CONF_DONT_CHECK_MIN, CONF_DONT_CHECK_MAX,
 			    /* clip */ false)
+			CONF_HANDLE_SIZE_T(opt_calloc_madvise_threshold,
+			    "calloc_madvise_threshold", 0, SC_LARGE_MAXCLASS,
+			    CONF_DONT_CHECK_MIN, CONF_CHECK_MAX, /* clip */ false)
```
interwq, line 1460: Maybe `CONF_DONT_CHECK_MAX`? If someone wants to set it to (size_t)(-1), we should allow it.

src/arena.c
```diff
@@ -359,16 +359,32 @@ arena_extent_alloc_large(tsdn_t *tsdn, arena_t *arena, size_t usize,
...
+	if (edata == NULL) {
+		return NULL;
 	}
 
-	if (edata != NULL && sz_large_pad != 0) {
+	if (zero && !zero_override && !edata_zeroed_get(edata)) {
+		void *addr = edata_base_get(edata);
+		size_t size = edata_size_get(edata);
+		memset(addr, 0, size);
+	}
```
interwq: Discussed offline. Move this to after `arena_cache_oblivious_randomize`

src/jemalloc.c
nullptr0-0: I'm using the same conf setting range as the `oversize_threshold`.

PR #2632, opened 2024-04-04
HPA: Fix infinite purging loop

ilvokhin:
One of the condition to start purging is `hpa_hugify_blocked_by_ndirty` function call returns true. This can happen in cases where we have no dirty memory for this shard at all. In this case purging loop will be an infinite loop.

`hpa_hugify_blocked_by_ndirty` was introduced at 0f6c420, but at that time purging loop has different form and additional `break` was not required. Purging loop form was re-written at 6630c5989, but additional exit condition wasn't added there at the time.

Repo code was shared by Patrik Dokoupil at [1], I stripped it down to minimum to reproduce issue in jemalloc unit tests.

[1]: https://github.com/jemalloc/jemalloc/pull/2533

### DIFF
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -537,9 +537,16 @@ hpa_shard_maybe_do_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard,
 		purged = false;
 		while (hpa_should_purge(tsdn, shard) && nops < max_ops) {
 			purged = hpa_try_purge(tsdn, shard);
-			if (purged) {
-				nops++;
+			if (!purged) {
+				/*
+				 * It is fine if we couldn't purge as sometimes
+				 * we try to purge just to unblock
+				 * hugification, but there is maybe no dirty
+				 * pages at all at the moment.
+				 */
+				break;
 			}
+			nops++;
 		}
 		hugified = hpa_try_hugify(tsdn, shard);
 		if (hugified) {
--- a/test/unit/hpa.c
+++ b/test/unit/hpa.c
@@ -24,7 +24,7 @@ struct test_data_s {
 static hpa_shard_opts_t test_hpa_shard_opts_default = {
 	/* slab_max_alloc */
 	ALLOC_MAX,
-	/* hugification threshold */
+	/* hugification_threshold */
 	HUGEPAGE,
 	/* dirty_mult */
 	FXP_INIT_PERCENT(25),
@@ -32,6 +32,21 @@ static hpa_shard_opts_t test_hpa_shard_opts_default = {
 	false,
 	/* hugify_delay_ms */
 	10 * 1000,
+	/* min_purge_interval_ms */
+	5 * 1000,
+};
+
+static hpa_shard_opts_t test_hpa_shard_opts_purge = {
+	/* slab_max_alloc */
+	HUGEPAGE,
+	/* hugification_threshold */
+	0.9 * HUGEPAGE,
+	/* dirty_mult */
+	FXP_INIT_PERCENT(11),
+	/* deferral_allowed */
+	true,
+	/* hugify_delay_ms */
+	0,
 	/* min_purge_interval_ms */
 	5 * 1000,
 };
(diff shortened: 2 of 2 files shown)


test/unit/hpa.c
```diff
@@ -452,6 +467,36 @@ TEST_BEGIN(test_defer_time) {
...
+	    test_hpa_shard_opts_purge.hugification_threshold / PAGE + 1;
+	const size_t size = npages * PAGE;
+
+	bool deferred_work_generated = false;
+	edata_t *edata = pai_alloc(tsdn, &shard->pai, size, PAGE,
+	     /* zero */ false, /* guarded */ false, /* frequent_reuse */ false,
+	     &deferred_work_generated);
+	expect_ptr_not_null(edata, "Unexpected alloc failure");
+
+	hpa_shard_do_deferred_work(tsdn, shard);
```
guangli-dai, line 492: Cool neat reproduction! It seems the failure rate is 100% now without the fix, right?

test/unit/hpa.c
```diff
@@ -452,6 +467,36 @@ TEST_BEGIN(test_defer_time) {
 }
 TEST_END
 
+TEST_BEGIN(test_purge_no_infinite_loop) {
```
guangli-dai, line 470: While the test is a very good reproduction of the bug this PR is fixing, I am thinking that we should make it a more general test like: 1. testing a bunch of option combinations within a certain range; 2. complicate the scenario a little bit so that purging with dirty pages existing can be tested stably as well. WDYT?

If @interwq is good with the current condition, we can merge the fix first and we can generalize the test later.

guangli-dai (commented):
The fix looks good! Thanks!

src/hpa.c
```diff
@@ -537,9 +537,15 @@ hpa_shard_maybe_do_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard,
...
+			/*
+			 * It is fine if we couldn't purge as sometimes we try
+			 * to purge just to unblock hugification, but there is maybe
+			 * no dirty pages at all at the moment.
+			 */
+			if (!purged) {
+				break;
 			}
+			nops++;
 		}
```
ot, line 550: nit: maybe a bit more readable this way?
```
if (!hpa_try_purge(tsdn, shard)) {
  // comment
  break;
}
nops++
```

ilvokhin: We still need to update `purged` as it is used to terminate loop at line 550.

I guess the best we can do is this, but not sure it is more readable to be honest as line seems a bit overloaded: negation, assignment, function call and condition checking it one line. 
```
if (!(purged = hpa_try_purge(tsdn, shard))) {
        // comment;
        break;
}
```
Moved comment inside if though.

test/unit/hpa.c
ilvokhin: > It seems the failure rate is 100% now without the fix, right?

Yes, it triggers 100% at least on x86_64, with `PAGE = 4 KiB` and `HUGEPAGE = 2 MiB`.

Basically, we allocate 461 page out of 512 pages on the huge page, which is 1888256 bytes (461 * 4 KiB) with 1887436 bytes as hugification threshold (2 MiB * 0.9). So, `hpa_should_purge` returns `true`, because `hpa_hugify_blocked_by_ndirty` returns `true`. We enter purging loop and call `hpa_try_purge` to purge, but there is nothing to purge, so it returns `false`, we didn't increment `nops` and continue `while` from the beginning again.

test/unit/hpa.c
ilvokhin: Actually, I was hoping for feedback about this test :)

I am not sure how useful it is to have when bug is fixed, because it is carefully crafted for this particular bug. I searched and didn't find much regression test in the code base.

Primary goal was to have as small and at the same time clean demonstration that bug is in place and proposed patch indeed fixes it. So I intentionally didn't add any extra logic in the test.

That's being said, I am fine with removing or extend tests whatever consensus here will be.

interwq (approved):

PR #2633, opened 2024-04-05
rallocx path: only set errno on the realloc case.

interwq:


interwq:
@arter97 FYI we tweaked this a bit because the non-standard APIs (i.e. `rallocx` here) don't set errno. Updated so that errno is only set on the `realloc` path.

guangli-dai (approved):

PR #2637, opened 2024-04-12
Fix the tcache flush sanity checking around ncached and nstashed.

interwq:
When there were many items stashed, it's possible that after flushing stashed, ncached is already lower than the remain, in which case the flush can simply return at that point.

@banitag1 caught this assertion triggered with certain workloads.


guangli-dai (approved):
Discussed offline, LGTM.

PR #2638, opened 2024-04-16
[fastpath] Allow compile-time size computation.

davidtgoldblatt:
In LTO-able contexts, if size is a static constant, then so are szind and usize. But: indirecting through a lookup table obscures this fact from the compiler. Instead, we try to detect if size is static and use a constant-foldable computation if it is.


interwq:
@davidtgoldblatt we have made plan for this (waiting for buckification to fully rollout). You should sync up with us to avoid conflict like this.

davidtgoldblatt:
lol, felt a little too easy -- I wondered what the deal was

PR #2642, opened 2024-05-02
Fix background_thread creation for the oversize_arena.

interwq:
Bypassing background thread creation for the oversize_arena used to be an optimization since that arena had eager purging.  However #2466 changed the purging policy for the oversize_arena -- specifically it switched to the default decay time when background_thread is enabled.

This issue is noticable when the number of arenas is low: whenever the total # of arenas is <= 4 (which is the default max # of background threads), in which case the purging will be stalled since no background thread is created for the oversize_arena.

### DIFF
--- a/include/jemalloc/internal/arena_externs.h
+++ b/include/jemalloc/internal/arena_externs.h
@@ -103,7 +103,6 @@ void arena_nthreads_inc(arena_t *arena, bool internal);
 void arena_nthreads_dec(arena_t *arena, bool internal);
 arena_t *arena_new(tsdn_t *tsdn, unsigned ind, const arena_config_t *config);
 bool arena_init_huge(arena_t *a0);
-bool arena_is_huge(unsigned arena_ind);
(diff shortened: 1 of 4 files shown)


nullptr0-0:
the change lgtm but some CI is failing?

nullptr0-0 (approved):

guangli-dai:
Searched `arena_is_huge` and found [this](https://github.com/jemalloc/jemalloc/blob/47d69b4eabae199fa8b5d948f0043effccfbc31e/src/ctl.c#L2988). We may want to change the logic here a bit as well to avoid duplicated background threads?

interwq:
@guangli-dai good catch. That block is not needed anymore. As a result `arena_is_huge()` is removed completely. For the context it wouldn't cause duplicated threads, since the background thread module considers that case (i.e. many threads could be triggering thread creation at the same time) and checks here: https://github.com/jemalloc/jemalloc/blob/47d69b4eabae199fa8b5d948f0043effccfbc31e/src/background_thread.c#L518

PR #2643, opened 2024-05-03
Simplify the logic in tcache_gc_small.

deadalnix:
The else branch is completely unnecessary here because the then branch returns. this makes it seems like the control flow is somehow different from what follows, when it is not.


interwq:
Thanks for the PR @deadalnix !

PR #2644, opened 2024-05-03
Adjust the value width in stats output.

interwq:
Some of the values are accumulative and can reach high after running for long periods.


guangli-dai (approved):
LGTM. The changes should be pretty safe and the Travis errors are most likely irrelevant.

PR #2648, opened 2024-05-13
[WIP] tcache management optimization

nullptr0-0:
This change introduces an optional setting `opt_tcache_gc_interval_us` that could lower GC frequency by requiring a time interval between two GCs.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -1047,14 +1047,16 @@ arena_bin_choose(tsdn_t *tsdn, arena_t *arena, szind_t binind,
 
 void
 arena_cache_bin_fill_small(tsdn_t *tsdn, arena_t *arena,
-    cache_bin_t *cache_bin, szind_t binind, const cache_bin_sz_t nfill) {
+    cache_bin_t *cache_bin, szind_t binind, const cache_bin_sz_t nfill_min,
+    const cache_bin_sz_t nfill_max) {
 	assert(cache_bin_ncached_get_local(cache_bin) == 0);
-	assert(nfill != 0);
+	assert(nfill_min > 0 && nfill_min <= nfill_max);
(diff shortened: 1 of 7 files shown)


src/arena.c
```diff
@@ -1040,14 +1040,16 @@ arena_bin_choose(tsdn_t *tsdn, arena_t *arena, szind_t binind,
...
 	assert(cache_bin_ncached_get_local(cache_bin) == 0);
-	assert(nfill != 0);
+	assert(nfill_min > 0 && nfill_min <= nfill_max);
+	assert(nfill_max <= cache_bin_ncached_max_get(cache_bin));
 
 	const bin_info_t *bin_info = &bin_infos[binind];
 
-	CACHE_BIN_PTR_ARRAY_DECLARE(ptrs, nfill);
-	cache_bin_init_ptr_array_for_fill(cache_bin, &ptrs, nfill);
+	CACHE_BIN_PTR_ARRAY_DECLARE(ptrs, nfill_max);
```
nullptr0-0, line 1051: When we increase the ncached_max, this array might be subject to overflow? Shall we hard limit the batch fill size?

test/unit/arena_decay.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-export MALLOC_CONF="dirty_decay_ms:1000,muzzy_decay_ms:1000,tcache_max:1024"
+export MALLOC_CONF="dirty_decay_ms:1000,muzzy_decay_ms:1000,tcache_max:1024,tcache_gc_incr_bytes:939524096"
```
nullptr0-0, line 3: the arena_decay test is overriding nstime_update which affects the GC timer, disable GC for this test.

PR #2651, opened 2024-05-23
skip tcache GC for tcache_max unit test

nullptr0-0:
tcache_max unit test is validating the content of the tcache stack, while it could be affected by tcache GC. It's not a problem today since the allocation in the unit test is not hitting the GC trigger yet, Instead we should explicitly skip GC in this unit test.

### DIFF
--- a/test/unit/tcache_max.sh
+++ b/test/unit/tcache_max.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-export MALLOC_CONF="tcache_max:1024"
+export MALLOC_CONF="tcache_max:1024,tcache_gc_incr_bytes:939524096"


test/unit/tcache_max.c
```diff
@@ -189,6 +194,7 @@ TEST_BEGIN(test_tcache_max) {
 	test_skip_if(opt_prof);
 	test_skip_if(san_uaf_detection_enabled());
 
+	tcache_event = &tcache_event_no_ops;
```
guangli-dai: For safety, please change `tcache_event` back to its original value at the end of the test. Same below.

guangli-dai (commented):
The CI failure looks like a ppc failure caused by the CI system.

nullptr0-0:
Update the change to set a really large tcache_gc_incr_bytes value for the tcache_max test, instead of changing the source code.

PR #2652, opened 2024-05-31
Export hugepage size with `arenas.hugepage`

ilvokhin:
Hugepages mighe be used for Metadata or Hugepages Aware Allocator, but 
currently there is no way to get from jemalloc what hugepage size is used.

Export hugepage size in the same way page size is exported.


guangli-dai:
Since HUGEPAGE is defined at compile time, it sounds more reasonable to expose it as `config.hugepage`? But surprised to see we have page size exposed as `arenas.page`, making `arenas.hugepage` also fine to me. @interwq Which one do you prefer?

interwq:
@guangli-dai yeah since we already have `arenas.page`, the current one looks fine.

PR #2654, opened 2024-06-05
`throw()` -> `noexcept`

r-barnes:
`throw()` is eliminated in C++17. `noexcept`, available since C++11, should be used instead.


interwq:
Merged as #2656.

PR #2658, opened 2024-06-07
Respect `hpa_min_purge_interval_ms` option 

ilvokhin:
Currently, hugepages aware allocator backend works together with classic one as a fallback for not yet supported allocations. When background threads are enabled wake up time for classic interfere with hpa as there were no checks inside hpa purging logic to check if we are not purging too frequently. If background thread is running and `hpa_should_purge` returns true, then we will purge, even if we purged less than hpa_min_purge_interval_ms ago.

Change in `hpa_min_purge_interval_ms` handling logic is not backward compatible as it might increase memory usage. This logic change guarded by
newly introduced `hpa_strict_min_purge_interval` option.

When `hpa_strict_min_purge_interval` is true, we will purge no more than `hpa_min_purge_interval_ms`. When `hpa_strict_min_purge_interval` is
false, old purging logic behaviour is preserved.

Long term strategy migrate all users of hpa to new logic and then delete `hpa_strict_min_purge_interval` option.

### DIFF
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -382,10 +382,12 @@ hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
 	 * Make sure we respect purge interval setting and don't purge
 	 * too frequently.
 	 */
-	uint64_t since_last_purge_ms = shard->central->hooks.ms_since(
-	    &shard->last_purge);
-	if (since_last_purge_ms < shard->opts.min_purge_interval_ms) {
-	     return false;
+	if (shard->opts.strict_min_purge_interval) {
+		uint64_t since_last_purge_ms = shard->central->hooks.ms_since(
+		    &shard->last_purge);
+		if (since_last_purge_ms < shard->opts.min_purge_interval_ms) {
+		     return false;
+		}
 	}
 
 	hpdata_t *to_purge = psset_pick_purge(&shard->psset);
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -378,6 +378,16 @@ static bool
 hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
 	malloc_mutex_assert_owner(tsdn, &shard->mtx);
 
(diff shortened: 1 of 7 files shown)


src/hpa.c
```diff
@@ -378,6 +378,18 @@ static bool
 hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
 	malloc_mutex_assert_owner(tsdn, &shard->mtx);
 
+	/*
+	 * Make sure we respect purge interval setting and don't purge
```
interwq, line 382: @ilvokhin to double check, it looks before this PR, the min interval is used to determine the next bg thd wakeup here:
https://github.com/jemalloc/jemalloc/blob/90c627edb70e081e1298b79010478d2f804467f1/src/hpa.c#L943

but we can still wake up early once we reach the early wakeup threshold (that is 4M)?

ilvokhin: Yes, this is matches my understanding as well. In addition, we might wake up early due to dirty_decay_ms even without early wakeup logic on the classic side.

In other words, we will still wake up background thread at whatever times current logic say us to wake up, but we will not purge anything, because of this condition.

interwq: > we will still wake up background thread at whatever times current logic say us to wake up, but we will not purge anything

In that case, the bg thds waking up will only purge the PAC side of things, is that right? Because otherwise we might as well just enforce the next wake up, instead of still letting it wake up and then skip.

guangli-dai:
LGTM! The PR seems completed instead of a draft PR, is it ready for merging?

interwq:
LGTM too. @ilvokhin feel free to undraft it and I'll merge.

src/hpa.c
ilvokhin: > In that case, the bg thds waking up will only purge the PAC side of things, is that right?

Yes, exactly. We'll still purge PAC side.

interwq (approved):

PR #2659, opened 2024-06-10
Change macOS mmap tag to fix conflict with CoreMedia

kdrag0n:
Tag 101 is assigned to "CoreMedia Capture Data", which makes for confusing output when debugging.

To avoid conflicts, use a tag in the reserved application-specific range from 240–255 (inclusive).

All assigned tags: https://github.com/apple-oss-distributions/xnu/blob/94d3b452840153a99b38a3a9659680b2a006908e/osfmk/mach/vm_statistics.h#L773-L775

### DIFF
--- a/src/pages.c
+++ b/src/pages.c
@@ -17,7 +17,7 @@
 #include <sys/bitops.h>	/* ilog2 */
 #endif
 #ifdef JEMALLOC_HAVE_VM_MAKE_TAG
-#define PAGES_FD_TAG VM_MAKE_TAG(101U)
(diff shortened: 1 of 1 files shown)


guangli-dai:
Thanks for the patch. It looks reasonable to me. This PR shouldn't affect windows so the CI failure looks like non-relevant. @interwq Can you help trigger them again just to double check? 

As for compatibility, I don't know any other systems like mac that use VM tags and have such a convention so I think the patch is safe to be merged.

guangli-dai (approved):

interwq:
cc: @devnexen for the tag value change

PR #2663, opened 2024-06-19
A couple of simplifications in tcache.c .

deadalnix:
Adjusting the low water mark in tcache_event is not necessary as tcache_gc_{small,large} already do it and other code path do not change the low water mark anyways.

The computation for lg_fill_div in tcache_gc_small was unnecessarily complex.


deadalnix:
Sorry, the water mark adjustment is obviously necessary...

PR #2669, opened 2024-07-01
Fix NSTIME_MONOTONIC flag for win32 implementation.

burtonli:
the GetSystemTimeAsFileTime API in Win32 does not guarantee monotonic time. This time can be adjusted by the system clock, meaning it can move forward or backward due to time synchronization, daylight saving time adjustments, or manual changes.
So we should claim NSTIME_MONOTONIC as false, otherwise it causes crash at arena_maybe_decay() function in arena.c.

image

### DIFF
--- a/src/nstime.c
+++ b/src/nstime.c
@@ -172,7 +172,7 @@ nstime_ns_since(const nstime_t *past) {
 }
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


interwq:
Thanks for the fix!

PR #2670, opened 2024-07-04
Increase `emitter_print_value` buf size

ilvokhin:
There are few long options (`bin_shards` and `slab_sizes` for example) when they are specified and we emit statistics, value gets truncated.

Worst case for `bin_shards` will be `|sssss:nnn`, where `sssss` is a size class (maximum 5 digits) and `nnn` three digits for number of shards. Total number of small classes currently is 36, so at most 360 (36 * 10) characters for `bin_shards`. Approximately same calculations should be true for `slab_sizes`, so them both in total should occupy no more than 720 characters (360 * 2).

256 characters was previous `buf` size plus 720 from calculations above is 976, we can round it up to round number like 1024.

### DIFF
--- a/include/jemalloc/internal/emitter.h
+++ b/include/jemalloc/internal/emitter.h
@@ -129,7 +129,7 @@ static inline void
 emitter_print_value(emitter_t *emitter, emitter_justify_t justify, int width,
     emitter_type_t value_type, const void *value) {
 	size_t str_written;
-#define BUF_SIZE 256
+#define BUF_SIZE 1024
 #define FMT_SIZE 10
 	/*
 	 * We dynamically generate a format string to emit, to let us use the


include/jemalloc/internal/emitter.h
```diff
@@ -129,7 +129,7 @@ static inline void
 emitter_print_value(emitter_t *emitter, emitter_justify_t justify, int width,
     emitter_type_t value_type, const void *value) {
 	size_t str_written;
-#define BUF_SIZE 256
+#define BUF_SIZE 1024
```
interwq, line 132: @ilvokhin is it only the multiple sizes / ranges cases that's too long? If so we should add a helper function to split and calls the emitter multiple times (so each invocation only prints one `size:value|`. Asking because the 1K buf size on the stack is more or less approaching the "last straw" for a fiber overflow type of issue.

If it's a more common issue, what we can do it is adding a loop in the emitter here, with each iteration handling at most the BUF_SIZE length.

Let's also extract the emitter_type_string case (along with the buffer) into a separate function.

ilvokhin: Thanks for review @interwq! I tried to implement second approach by emitting at most the `BUF_SIZE` chars each iteration, but there is one challenging part: width justification support. When widths of a string is bigger than `BUF_SIZE` and we output string in multiple parts it is tricky to make justification work properly.

See ebe0b5f828 for more context for strings justification for table output.

ilvokhin:
Too much hassle to make it work properly. The complexity of the solution does not outweigh the pain of the problem itself.

include/jemalloc/internal/emitter.h
interwq: @ilvokhin good point. Agreed it's not worth the trouble and complexity to add justification here. How about we don't support justification for the long string case at all? Non default justification and super long string don't sound like a valid combination in our case. I looked through the use cases and it doesn't look like we have it -- the ones with explicit justification look to be either title or plain counters. The emitter_string_type looks all go through emitter_kv / emitter_json_kv which has emitter_justify_none.

Maybe you can try adding an assertion in `case emitter_type_string:` to verify it's always emitter_justify_none. Although in the final PR we probably should skip the assertion. It's probably not worth crashing everything if it's only a string format issue.

ilvokhin:
Took another stab here https://github.com/jemalloc/jemalloc/pull/2676 as I don't have permissions to re-open this PR.

PR #2676, opened 2024-07-18
Emit long string values without truncation

ilvokhin:
There are few long options (`bin_shards` and `slab_sizes` for example) when they are specified and we emit statistics value gets truncated.

Moved emitting logic for strings into separate `emitter_emit_str` function. It will try to emit string same way as before and if value is too long will fallback emiting rest partially with chunks of `BUF_SIZE`.

Justification for long strings (longer than `BUF_SIZE`) is not supported.

https://github.com/jemalloc/jemalloc/pull/2670 is where discussion started.

### DIFF
--- a/include/jemalloc/internal/emitter.h
+++ b/include/jemalloc/internal/emitter.h
@@ -118,6 +118,37 @@ emitter_gen_fmt(char *out_fmt, size_t out_size, const char *fmt_specifier,
 	return out_fmt;
 }
 
+static inline void
+emitter_emit_str(emitter_t *emitter, emitter_justify_t justify, int width,
+    char *fmt, size_t fmt_size, const char *str) {
+#define BUF_SIZE 256
+	char buf[BUF_SIZE];
+	size_t str_written = malloc_snprintf(buf, BUF_SIZE, "\"%s\"", str);
+	emitter_printf(emitter,
+	    emitter_gen_fmt(fmt, fmt_size, "%s", justify, width), buf);
+	if (str_written < BUF_SIZE) {
+		return;
+	}
+	/*
+	 * There is no support for long string justification at the moment as
+	 * we output them partially with multiple malloc_snprintf calls and
+	 * justufication will work correctly only withing one call.
+	 * Fortunately this is not a big concern as we don't use justufication
+	 * with long strings right now.
+	 *
+	 * We emitted leading quotation mark and trailing '\0', hence need to
+	 * exclude extra characters from str shift.
+	 */
+	str += BUF_SIZE - 2;
+	do {
+		str_written = malloc_snprintf(buf, BUF_SIZE, "%s\"", str);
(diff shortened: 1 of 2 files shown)


interwq:
@ilvokhin it looks the Windows CI failure are real, with both Travis and Appveyor report this unit test crashing:
```
=== test/unit/malloc_io ===
test_malloc_strtoumax_no_endptr (non-reentrant): pass
test_malloc_strtoumax_no_endptr (libc-reentrant): pass
test_malloc_strtoumax_no_endptr (arena_new-reentrant): pass
test_malloc_strtoumax (non-reentrant): pass
test_malloc_strtoumax (libc-reentrant): pass
test_malloc_strtoumax (arena_new-reentrant): pass
test_malloc_snprintf_truncated (non-reentrant): pass
test_malloc_snprintf_truncated (libc-reentrant): pass
test_malloc_snprintf_truncated (arena_new-reentrant): pass
Test harness error: test/unit/malloc_io w/ MALLOC_CONF=""
Use prefix to debug, e.g. JEMALLOC_TEST_PREFIX="gdb --args" sh test/test.sh test/unit/malloc_io
mingw32-make: *** [Makefile:699: check_unit] Error 1
```

test/include/test/test.h
```diff
@@ -1,4 +1,4 @@
-#define ASSERT_BUFSIZE	256
+#define ASSERT_BUFSIZE	4096
```
interwq: Is this change intentional?

ilvokhin: Kind of. Without this change not all output will fit in the buffer to print assert message, when test fails. But this is not necessary for the primary change. Let me try to back it out and check if Windows CI will recover.

ilvokhin: @interwq, you were right, that's it. I suspect it is likely because internal `MALLOC_PRINTF_BUFSIZE` is 4096 bytes as well and they might interact in non-obvious way. I'll try to understand it better if I'll have time, meanwhile remove this change from PR.

include/jemalloc/internal/emitter.h
```diff
@@ -118,6 +118,31 @@ emitter_gen_fmt(char *out_fmt, size_t out_size, const char *fmt_specifier,
...
+	    emitter_gen_fmt(fmt, fmt_size, "%s", justify, width), buf);
+	if (str_written < BUF_SIZE) {
+		return;
+	}
+	/*
+	 * We emitted leading quotation mark and trailing '\0', hence need to
+	 * exclude extra characters from str shift.
+	 */
+	str += BUF_SIZE - 2;
+	while (str_written >= BUF_SIZE) {
```
interwq: nit: change to a `do {} while ()` loop, since we execute at least once when reaching here. Also add some comments about long str output has no `justify` support.

interwq (approved):

include/jemalloc/internal/emitter.h
ilvokhin: @interwq, thanks for review! All done.

PR #2677, opened 2024-07-19
Avoid stack overflow when use VARRIABLE_ARRAY internally

nullptr0-0:


nullptr0-0:
With [4b20269](https://github.com/jemalloc/jemalloc/pull/2677/commits/4b202696f520ddde8b4b691e056069a46ca82257) assert the max varriable array size, and without [4be59bc](https://github.com/jemalloc/jemalloc/pull/2677/commits/4be59bce565b26a416b55bf2b7636d2d2e078f2f), if increase the tcached_max to 1024 like in [f2c3328](https://github.com/jemalloc/jemalloc/pull/2677/commits/f2c33282fbae401613d1371c2afd806c1c8a66e0), the assertion will fail.


[~/jemalloc (4b202696)]$ sh test/test.sh test/stress/fill_flush
=== test/stress/fill_flush ===
<jemalloc>: src/tcache.c:334: Failed assertion: "sizeof(emap_batch_lookup_result_t) * (nflush + 1) <= VARIABLE_ARRAY_SIZE_MAX"
test/test.sh: line 34: 577822 Aborted                 (core dumped) $JEMALLOC_TEST_PREFIX ${t} /home/sherrycheng/jemalloc/ /home/sherrycheng/jemalloc/
Test harness error: test/stress/fill_flush w/ MALLOC_CONF=""
Use prefix to debug, e.g. JEMALLOC_TEST_PREFIX="gdb --args" sh test/test.sh test/stress/fill_flush

nullptr0-0:
@interwq I agree that we should have a hard bound on the regular dalloc flush, instead of solely depending on half of the bin. But I still think we should have the for loop in the tcache_bin_flush_impl() level so that the call site is more relaxed, and the tcache_bin_flush_impl itself is more resilient.

include/jemalloc/internal/cache_bin.h
```diff
@@ -600,6 +600,13 @@ cache_bin_nitems_get_remote(cache_bin_t *bin, cache_bin_sz_t *ncached,
 	 */
 }
 
+/*
+ * Limit how many items can be flushed in a batch (Which is the upper bound
+ * for the nflush parameter in tcache_bin_flush_impl()).
+ * This is to avoid stack overflow when we do batch edata look up.
+ */
+#define CACHE_BIN_NFLUSH_BATCH_MAX 256
```
guangli-dai: Since we have `VARIABLE_ARRAY_SIZE_MAX`, would it be better for us to calculate this constant based on that? This could effectively avoid future changes causing conflicts between the two.

guangli-dai (commented):
It seems there are quite some other usage of `VARIABLE_ARRAY` now, like [this on stats.c](https://github.com/jemalloc/jemalloc/blob/c893fcd169fffca1b9d3156c6637a197765b82d0/src/stats.c#L1899). If the number of arenas gets large enough, it is possible for the array to exceed `VARIABLE_ARRAY_SIZE_MAX`, is that correct? 

Maybe for such use cases that could cause assertions but are far less frequently touched compared to tcache flushing, we should replace them with `VARIABLE_ARRAY_UNSAFE`?

src/tcache.c
```diff
@@ -712,22 +712,38 @@ tcache_bin_flush_impl_large(tsd_t *tsd, tcache_t *tcache, cache_bin_t *cache_bin
...
-		tcache_bin_flush_impl_small(tsd, tcache, cache_bin, binind,
-		    ptrs, nflush);
-	} else {
-		tcache_bin_flush_impl_large(tsd, tcache, cache_bin, binind,
-		    ptrs, nflush);
-	}
+	assert(ptrs != NULL && ptrs->ptr != NULL);
+	unsigned nflushed = 0;
+	cache_bin_ptr_array_t ptrs_batch;
+	(&ptrs_batch)->ptr = ptrs->ptr;
```
interwq: Currently there are two assignments for (&ptrs_batch)->ptr (the other one at the end of while loop). Let's merge the two and have only 1 line after `(&ptrs_batch)->n = (cache_bin_sz_t)nflush_batch;`

`(&ptrs_batch)->ptr = ptrs->ptr + nflushed;`

This also has a small benefit that ptrs_batch->ptr never points out of range, even the last iteration.

interwq (approved):

src/tcache.c
nullptr0-0: updated!

PR #2679, opened 2024-07-25
Utilise the thread bin to their full extent when heavily allocating.

deadalnix:
I came up with that strategy because I implemented a similar thread cache mechanism in the context of a GC. When using a GC, it is common to allocate way more than you deallocate (at least explicitly) and we benefit from being able to utilize the full extent of the cache by setting lg_fill_div to 0.

The obvious downside is that it may not leave enough space behind for free, causing churn int he cache (fill in the bin, then flush a good chunk of it to be able to free allocations). We avoid that situation by remembering if we had to flush to free during the bin's gc cycle, and if so, we do not allow lg_fill_div to go down to 0.

As far as I can tell, this strategy is also beneficial in the malloc/free situation, when the application is in an "expansion" phase, as in allocates way more than it frees, but I'd be curious to see what you guys find when trying this and if it doesn't work for you, why?

### DIFF
--- a/include/jemalloc/internal/tcache_externs.h
+++ b/include/jemalloc/internal/tcache_externs.h
@@ -49,6 +49,8 @@ size_t tcache_salloc(tsdn_t *tsdn, const void *ptr);
 void *tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
     cache_bin_t *cache_bin, szind_t binind, bool *tcache_success);
 
+void tcache_bin_dalloc_small_flush(tsd_t *tsd, tcache_t *tcache,
+    cache_bin_t *cache_bin, szind_t binind, unsigned rem);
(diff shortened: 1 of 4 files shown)


guangli-dai:
We are modifying the thread cache fill and flush part in #2685. @nullptr0-0 Any insights on this idea allowing `lg_fill_div` to be 0?

deadalnix:
As far as I can tell, both approach are not mutually exclusive (and in fact I was considering doing something akin to #2685 myself).

Obviously, there will be merge conflicts "as this".

deadalnix:
Is this superseeded by 7c996861 ?

interwq:
@deadalnix yes I believe so. The stack in #2685 (including the commit you mentioned) should improve refill and flush batching behavior. We just recently deployed it into our fleet and the results were good. Please feel free to try the current dev branch.

Thank you for the PR! I'm closing it but feel free to let us know if any issues.

PR #2685, opened 2024-08-05
Lower the frequency of tcache gc and introduce slab-aware tcache fill/flush

nullptr0-0:
### What's the change? 
- **Lowered GC Frequency**
  - **Before**:
    - tcache_event is triggered every tcache_gc_incr_bytes bytes, which could lead to very frequent GC when there is a large sized allocation.
  - **After**:
    - A timer is added to prevent consecutive GC within certain time intervals, and instead of GC only one bin, the new GC will look at multiple bins and making sure certain number of bins are flushed.

- **Slab-aware Fill and GC Flush for tcache small bins**.
  - **Before**:
    - The number of items filled at a time is fixed, i.e. (ncached_max >> lg_fill_div).
    - During GC, a bin will only be flushed if the low_water mark is not zero, and the items below 3/4 low_water at the bottom of the bin are flushed.
    - [What's the Problem?] The tcache fill and flush here is not taking locality into consideration, which makes it subject to fragmentation especially if tcache is used more aggressively.
  - **After**:
    - Instead of filling fixed number of items as requested, the lower-level will try to fulfill the request by dropping the last partially full slab as long as min filling value has been reached.
    - For GC flush, we care about both letting go idle items and badly located items (e.g. items belong to remote arenas). To achieve this, we query the associated arena nonfull slabs to get some heuristic locality info, and use it to determine which items should be let go, instead of blindly flushing bottom items.

### [WIP] How the change affects different traffic pattern?
On avg, the number of items filled at once for each sc will be increased. The reason is that, the nfill count is adjusted by looking at if the bin was ever used up between two GCs, with GC being too frequent before, the nfill is adjusted really low. A hard required GC interval makes sure that the bin has a higher chance being used up, but there are some cases that it might not cover well,
  - Burst traffic within a short period of time (i.e. less than GC interval). Since we only adjust fill count during tcache event, if there is a traffic burst within a short period time, the current mechanism will not be able to respond to that spike.
  - Similarly, if a thread sleeps most of the time, and only wakes up and runs for less than GC time interval, it's more likely to hold up memories compared to before.

### DIFF
--- a/include/jemalloc/internal/tcache_structs.h
+++ b/include/jemalloc/internal/tcache_structs.h
@@ -39,8 +39,8 @@ struct tcache_slow_s {
 	szind_t		next_gc_bin;
 	szind_t		next_gc_bin_small;
 	szind_t		next_gc_bin_large;
-	/* For small bins, fill (ncached_max >> lg_fill_div). */
-	uint8_t		lg_fill_div[SC_NBINS];
+	/* For small bins, help determine how many items to fill at a time. */
+	cache_bin_fill_ctl_t	bin_fill_ctl_do_not_access_directly[SC_NBINS];
 	/* For small bins, whether has been refilled since last GC. */
 	bool		bin_refilled[SC_NBINS];
 	/*
--- a/include/jemalloc/internal/tcache_structs.h
+++ b/include/jemalloc/internal/tcache_structs.h
@@ -33,8 +33,12 @@ struct tcache_slow_s {
 	arena_t		*arena;
 	/* The number of bins activated in the tcache. */
 	unsigned	tcache_nbins;
+	/* Last time GC has been performed.  */
+	nstime_t	last_gc_time;
 	/* Next bin to GC. */
 	szind_t		next_gc_bin;
+	szind_t		next_gc_bin_small;
+	szind_t		next_gc_bin_large;
 	/* For small bins, fill (ncached_max >> lg_fill_div). */
 	uint8_t		lg_fill_div[SC_NBINS];
 	/* For small bins, whether has been refilled since last GC. */
--- a/include/jemalloc/internal/tcache_types.h
+++ b/include/jemalloc/internal/tcache_types.h
@@ -24,5 +24,8 @@ typedef struct tcaches_s tcaches_t;
 #define TCACHE_NBINS_MAX (SC_NBINS + SC_NGROUP *			\
     (TCACHE_LG_MAXCLASS_LIMIT - SC_LG_LARGE_MINCLASS) + 1)
 #define TCACHE_GC_NEIGHBOR_LIMIT ((uintptr_t)1 << 21) /* 2M */
+#define TCACHE_GC_INTERVAL_NS ((uint64_t)10 * KQU(1000000)) /* 10ms */
+#define TCACHE_GC_SMALL_NBINS_MAX ((SC_NBINS > 8) ? (SC_NBINS >> 3) : 1)
+#define TCACHE_GC_LARGE_NBINS_MAX 1
 
 #endif /* JEMALLOC_INTERNAL_TCACHE_TYPES_H */
--- a/include/jemalloc/internal/tcache_types.h
+++ b/include/jemalloc/internal/tcache_types.h
@@ -23,5 +23,6 @@ typedef struct tcaches_s tcaches_t;
 #define TCACHE_MAXCLASS_LIMIT ((size_t)1 << TCACHE_LG_MAXCLASS_LIMIT)
 #define TCACHE_NBINS_MAX (SC_NBINS + SC_NGROUP *			\
     (TCACHE_LG_MAXCLASS_LIMIT - SC_LG_LARGE_MINCLASS) + 1)
+#define TCACHE_GC_NEIGHBOR_LIMIT ((uintptr_t)1 << 21) /* 2M */
 
 #endif /* JEMALLOC_INTERNAL_TCACHE_TYPES_H */
--- a/src/arena.c
+++ b/src/arena.c
@@ -1047,14 +1047,16 @@ arena_bin_choose(tsdn_t *tsdn, arena_t *arena, szind_t binind,
 
 void
 arena_cache_bin_fill_small(tsdn_t *tsdn, arena_t *arena,
-    cache_bin_t *cache_bin, szind_t binind, const cache_bin_sz_t nfill) {
+    cache_bin_t *cache_bin, szind_t binind, const cache_bin_sz_t nfill_min,
+    const cache_bin_sz_t nfill_max) {
 	assert(cache_bin_ncached_get_local(cache_bin) == 0);
-	assert(nfill != 0);
+	assert(nfill_min > 0 && nfill_min <= nfill_max);
+	assert(nfill_max <= cache_bin_ncached_max_get(cache_bin));
 
 	const bin_info_t *bin_info = &bin_infos[binind];
 
-	CACHE_BIN_PTR_ARRAY_DECLARE(ptrs, nfill);
-	cache_bin_init_ptr_array_for_fill(cache_bin, &ptrs, nfill);
+	CACHE_BIN_PTR_ARRAY_DECLARE(ptrs, nfill_max);
+	cache_bin_init_ptr_array_for_fill(cache_bin, &ptrs, nfill_max);
 	/*
 	 * Bin-local resources are used first: 1) bin->slabcur, and 2) nonfull
 	 * slabs.  After both are exhausted, new slabs will be allocated through
@@ -1101,13 +1103,19 @@ label_refill:
 	malloc_mutex_lock(tsdn, &bin->lock);
 	arena_bin_flush_batch_after_lock(tsdn, arena, bin, binind, &batch_flush_state);
 
-	while (filled < nfill) {
+	while (filled < nfill_min) {
 		/* Try batch-fill from slabcur first. */
 		edata_t *slabcur = bin->slabcur;
 		if (slabcur != NULL && edata_nfree_get(slabcur) > 0) {
-			unsigned tofill = nfill - filled;
-			unsigned nfree = edata_nfree_get(slabcur);
-			unsigned cnt = tofill < nfree ? tofill : nfree;
+			/*
+			 * Use up the free slots if the total filled <= nfill_max.
+			 * Otherwise, fallback to nfill_min for a more conservative
+			 * memory usage.
+			 */
+			unsigned cnt = edata_nfree_get(slabcur);
+			if (cnt + filled > nfill_max) {
+				cnt = nfill_min - filled;
+			}
 
 			arena_slab_reg_alloc_batch(slabcur, bin_info, cnt,
 			    &ptrs.ptr[filled]);
@@ -1144,7 +1152,7 @@ label_refill:
 		assert(fresh_slab == NULL);
 		assert(!alloc_and_retry);
 		break;
-	} /* while (filled < nfill) loop. */
+	} /* while (filled < nfill_min) loop. */
 
 	if (config_stats && !alloc_and_retry) {
 		bin->stats.nmalloc += filled;
@@ -1162,7 +1170,7 @@ label_refill:
 
 	if (alloc_and_retry) {
 		assert(fresh_slab == NULL);
-		assert(filled < nfill);
+		assert(filled < nfill_min);
 		assert(made_progress);
 
 		fresh_slab = arena_slab_alloc(tsdn, arena, binind, binshard,
@@ -1173,7 +1181,8 @@ label_refill:
 		made_progress = false;
 		goto label_refill;
 	}
-	assert(filled == nfill || (fresh_slab == NULL && !made_progress));
+	assert((filled >= nfill_min && filled <= nfill_max) ||
+	    (fresh_slab == NULL && !made_progress));
 
 	/* Release if allocated but not used. */
 	if (fresh_slab != NULL) {
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -121,6 +121,85 @@ tcache_gc_dalloc_postponed_event_wait(tsd_t *tsd) {
 	return TE_MIN_START_WAIT;
 }
 
+static inline void
+tcache_bin_fill_ctl_init(tcache_slow_t *tcache_slow, szind_t szind) {
+	assert(szind < SC_NBINS);
+	cache_bin_fill_ctl_t *ctl =
+	    &tcache_slow->bin_fill_ctl_do_not_access_directly[szind];
+	ctl->base = 1;
+	ctl->offset = 0;
+}
+
+static inline cache_bin_fill_ctl_t *
+tcache_bin_fill_ctl_get(tcache_slow_t *tcache_slow, szind_t szind) {
+	assert(szind < SC_NBINS);
+	cache_bin_fill_ctl_t *ctl =
+	    &tcache_slow->bin_fill_ctl_do_not_access_directly[szind];
+	assert(ctl->base > ctl->offset);
+	return ctl;
+}
+
+/*
+ * The number of items to be filled at a time for a given small bin is
+ * calculated by (ncached_max >> lg_fill_div).
+ * The actual ctl struct consists of two fields, i.e. base and offset,
+ * and the difference between the two(base - offset) is the final lg_fill_div.
+ * The base is adjusted during GC based on the traffic within a period of time,
+ * while the offset is updated in real time to handle the immediate traffic.
+ */
+static inline uint8_t
+tcache_nfill_small_lg_div_get(tcache_slow_t *tcache_slow, szind_t szind) {
+	cache_bin_fill_ctl_t *ctl = tcache_bin_fill_ctl_get(tcache_slow, szind);
+	return (ctl->base - (opt_experimental_tcache_gc ? ctl->offset : 0));
+}
+
+/*
+ * When we want to fill more items to respond to burst load,
+ * offset is increased so that (base - offset) is decreased,
+ * which in return increases the number of items to be filled.
+ */
+static inline void
+tcache_nfill_small_burst_prepare(tcache_slow_t *tcache_slow, szind_t szind) {
+	cache_bin_fill_ctl_t *ctl = tcache_bin_fill_ctl_get(tcache_slow, szind);
+	if (ctl->offset + 1 < ctl->base) {
+		ctl->offset++;
+	}
+}
+
+static inline void
+tcache_nfill_small_burst_reset(tcache_slow_t *tcache_slow, szind_t szind) {
+	cache_bin_fill_ctl_t *ctl = tcache_bin_fill_ctl_get(tcache_slow, szind);
+	ctl->offset = 0;
+}
+
+/*
+ * limit == 0: indicating that the fill count should be increased,
+ * i.e. lg_div(base) should be decreased.
+ *
+ * limit != 0: limit is set to ncached_max, indicating that the fill
+ * count should be decreased, i.e. lg_div(base) should be increased.
+ */
+static inline void
+tcache_nfill_small_gc_update(tcache_slow_t *tcache_slow, szind_t szind,
+    cache_bin_sz_t limit) {
+	cache_bin_fill_ctl_t *ctl = tcache_bin_fill_ctl_get(tcache_slow, szind);
+	if (!limit && ctl->base > 1) {
+		/*
+		 * Increase fill count by 2X for small bins.  Make sure
+		 * lg_fill_div stays greater than 1.
+		 */
+		ctl->base--;
+	} else if (limit && (limit >> ctl->base) > 1) {
+		/*
+		 * Reduce fill count by 2X.  Limit lg_fill_div such that
+		 * the fill count is always at least 1.
+		 */
+		ctl->base++;
+	}
+	/* Reset the offset for the next GC period. */
+	ctl->offset = 0;
+}
+
 static uint8_t
 tcache_gc_item_delay_compute(szind_t szind) {
 	assert(szind < SC_NBINS);
@@ -298,21 +377,19 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
 	cache_bin_sz_t low_water = cache_bin_low_water_get(cache_bin);
 	if (low_water > 0) {
 		/*
-		 * Reduce fill count by 2X.  Limit lg_fill_div such that
-		 * the fill count is always at least 1.
+		 * There is unused items within the GC period => reduce fill count.
+		 * limit field != 0 is borrowed to indicate that the fill count
+		 * should be reduced.
 		 */
-		if ((cache_bin_ncached_max_get(cache_bin) >>
-		    tcache_slow->lg_fill_div[szind]) > 1) {
-			tcache_slow->lg_fill_div[szind]++;
-		}
+		tcache_nfill_small_gc_update(tcache_slow, szind,
+		    /* limit */ cache_bin_ncached_max_get(cache_bin));
 	} else if (tcache_slow->bin_refilled[szind]) {
 		/*
-		 * Increase fill count by 2X for small bins.  Make sure
-		 * lg_fill_div stays greater than 0.
+		 * There has been refills within the GC period => increase fill count.
+		 * limit field set to 0 is borrowed to indicate that the fill count
+		 * should be increased.
 		 */
-		if (tcache_slow->lg_fill_div[szind] > 1) {
-			tcache_slow->lg_fill_div[szind]--;
-		}
+		tcache_nfill_small_gc_update(tcache_slow, szind, /* limit */ 0);
 		tcache_slow->bin_refilled[szind] = false;
 	}
 	assert(!tcache_slow->bin_refilled[szind]);
@@ -526,7 +603,7 @@ tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena,
 	assert(tcache_slow->arena != NULL);
 	assert(!tcache_bin_disabled(binind, cache_bin, tcache_slow));
 	cache_bin_sz_t nfill = cache_bin_ncached_max_get(cache_bin)
-	    >> tcache_slow->lg_fill_div[binind];
+	    >> tcache_nfill_small_lg_div_get(tcache_slow, binind);
 	if (nfill == 0) {
 		nfill = 1;
 	}
@@ -534,6 +611,7 @@ tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena,
 	    /* nfill_min */ opt_experimental_tcache_gc ?
 	    ((nfill >> 1) + 1) : nfill, /* nfill_max */ nfill);
 	tcache_slow->bin_refilled[binind] = true;
+	tcache_nfill_small_burst_prepare(tcache_slow, binind);
 	ret = cache_bin_alloc(cache_bin, tcache_success);
 
 	return ret;
@@ -1059,6 +1137,7 @@ tcache_bin_flush_bottom(tsd_t *tsd, tcache_t *tcache, cache_bin_t *cache_bin,
 void
 tcache_bin_flush_small(tsd_t *tsd, tcache_t *tcache, cache_bin_t *cache_bin,
     szind_t binind, unsigned rem) {
+	tcache_nfill_small_burst_reset(tcache->tcache_slow, binind);
 	tcache_bin_flush_bottom(tsd, tcache, cache_bin, binind, rem,
 	    /* small */ true);
 }
@@ -1233,7 +1312,7 @@ tcache_init(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
 	    &cur_offset);
 	for (unsigned i = 0; i < tcache_nbins; i++) {
 		if (i < SC_NBINS) {
-			tcache_slow->lg_fill_div[i] = 1;
+			tcache_bin_fill_ctl_init(tcache_slow, i);
 			tcache_slow->bin_refilled[i] = false;
 			tcache_slow->bin_flush_delay_items[i]
 			    = tcache_gc_item_delay_compute(i);
--- a/src/tcache.c
+++ b/src/tcache.c
@@ -283,7 +283,7 @@ tcache_gc_small_bin_shuffle(cache_bin_t *cache_bin, cache_bin_sz_t nremote,
 	}
 }
 
-static void
+static bool
 tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
     szind_t szind) {
 	/*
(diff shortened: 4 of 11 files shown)


interwq:
> Similarly, if a thread sleeps most of the time, and only wakes up and runs for less than GC time interval, it's more likely to hold up memories compared to before.

I'm not fully convinced this is the case. Say for a periodic background job, each time it wakes up and process a X amount of data (assuming X >= the timer-checking bytes interval), then with the new time based GC, we still reach down the GC path for each run, right? However if each run we still handle the exact same # of bins, we will cover fewer bins -- we discussed using the actual time interval observed to determine if more bins should be covered, should we add that as well?

src/arena.c
```diff
@@ -1101,13 +1103,21 @@ arena_cache_bin_fill_small(tsdn_t *tsdn, arena_t *arena,
 	malloc_mutex_lock(tsdn, &bin->lock);
 	arena_bin_flush_batch_after_lock(tsdn, arena, bin, binind, &batch_flush_state);
 
-	while (filled < nfill) {
+	while (filled < nfill_max) {
+		if (filled >= nfill_min) {
```
interwq: Effectively, this is the same as changing to `while (filled < nfill_min) `, which makes more sense I feel?

nullptr0-0:
>  Say for a periodic background job, each time it wakes up and process a X amount of data (assuming X >= the timer-checking bytes interval), then with the new time based GC, we still reach down the GC path for each run.

Yes, but if each wake up is less than the timer interval, the GC will be performed at the next wake up VS before that GC could happen at the current wake up itself?

src/arena.c
```diff
@@ -1101,13 +1103,21 @@ arena_cache_bin_fill_small(tsdn_t *tsdn, arena_t *arena,
...
 		/* Try batch-fill from slabcur first. */
 		edata_t *slabcur = bin->slabcur;
 		if (slabcur != NULL && edata_nfree_get(slabcur) > 0) {
-			unsigned tofill = nfill - filled;
-			unsigned nfree = edata_nfree_get(slabcur);
-			unsigned cnt = tofill < nfree ? tofill : nfree;
+			/* Use up the free slots if the total filled <= nfill_max. */
+			unsigned cnt = edata_nfree_get(slabcur);
+			if (cnt + filled > nfill_max) {
+				cnt = nfill_min - filled;
```
interwq, line 1117: Did you meant to have `cnt = nfill_max - filled`? The `nfill_min - filled` would imply we always fill to min?

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
 	}
+
+	if (nflush == ncached)	goto label_flush;
+
+	/* Query arena binshard to get heuristic locality info. */
+	tsdn_t *tsdn = tsd_tsdn(tsd);
+	bin_t *bin = arena_bin_choose(tsdn, tcache_slow->arena, szind, NULL);
+	assert(bin != NULL);
+
+	void *addr = NULL;
```
interwq: Avoid double initialization.
```
void *addr = (slab != NULL) ? edata_addr_get(slab) : NULL;
```
at L183

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
 	     tcache_slow->lg_fill_div[szind]) > 1) {
 		tcache_slow->lg_fill_div[szind]++;
 	}
+
+	if (nflush == ncached)	goto label_flush;
```
interwq: Always separate line with brackets.

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+		addr = edata_addr_get(slab);
+	}
+	assert(addr != NULL || slab == NULL);
+	malloc_mutex_unlock(tsdn, &bin->lock);
+	if (addr == NULL)	goto label_flush;
+
+	/* The slab address range that the queried addr belongs to. */
+	uintptr_t slab_min = (uintptr_t)addr;
+	uintptr_t slab_max = slab_min + bin_infos[szind].slab_size;
+	/* The exp_grow->next starts from HUGEPAGE. */
```
interwq: This is not intuitive and needs more explanation.

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+	uintptr_t addr_min = slab_min > HUGEPAGE ? (slab_min - HUGEPAGE) : 0;
+	uintptr_t addr_max = slab_min < ((uintptr_t)(-1) - HUGEPAGE) ?
+	    (slab_min + HUGEPAGE) : (uintptr_t)(-1);
```
interwq: UINTPTR_MAX instead of -1

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+	/* Scan the entire bin to count the number of remote pointers. */
+	for (void **cur = head; cur < head + ncached; cur++) {
+		assert(*cur != NULL);
+		uintptr_t addr_cur = (uintptr_t)*cur;
+		nremote += (cache_bin_sz_t)(addr_cur < addr_min ||
+		    addr_cur >= addr_max);
+		nremote_hard += (cache_bin_sz_t)(addr_cur < slab_min ||
+		    addr_cur >= slab_max);
+	}
+	assert(nremote_hard >= nremote && nremote_hard <= ncached);
```
interwq: The nremote and nremote_hard names are rather confusing. Maybe nremote and nremote_further, with the latter being the current nremote?

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+	if (nremote < nflush) {
+		nremote = nremote_hard;
+		addr_min = slab_min;
+		addr_max = slab_max;
+	}
+	/*
+	 * Update the nflush to the larger value between the intended flush count
+	 * and the number of remote ptrs.
+	 */
+	nflush = nremote > nflush ? nremote : nflush;
```
interwq:
```
if (nremote > nflush) { 
    nflush = nremote;
}
```

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+	assert(nflush < ncached || nremote == ncached);
```
interwq: I don't follow this assertion. We have nflush <= ncached always. So this is saying, if (nflush == ncached), everything must be remote?

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+
+	/*
+	 * Move the remote points to the bottom of the bin for flushing.
+	 * As long as moved to the bottom, the order of these nremote ptrs
+	 * does not matter, since they are going to be flushed anyway.
+	 * The rest of the ptrs are moved to the top of the bin, and their
+	 * relative order is not maintained.
+	 */
+	void **swap = NULL;
+	cache_bin_sz_t ntop = ncached - nremote, cnt = 0;
```
interwq, line 301: Why `nremote` here? After we update nflush, we should not reference nremote anymore?

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+				*swap = tmp;
+				swap++;
+				assert(swap <= cur);
+				assert((uintptr_t)*swap < addr_min ||
+				    (uintptr_t)*swap >= addr_max);
+			}
+			continue;
+		}
+
+		assert(!not_remote);
```
interwq: Do an else branch, and remove this assert.

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+			assert(*(head + cnt) != NULL && (
+			    (uintptr_t)*(head + cnt) < addr_min ||
+			    (uintptr_t)*(head + cnt) >= addr_max));
+			void *tmp = *cur;
+			*cur = *(head + cnt);
+			*(head + cnt) = tmp;
+			cnt++;
+		}
+	}
+	assert(cnt == ntop);
```
interwq, line 355: There are non trivial scanning and reordering from the two loops above. We need another sanity checking loop that verifies the results (which should only include assertion in the body).

src/tcache.c
```diff
@@ -169,6 +167,137 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+			if (swap != NULL) {
+				assert(swap < cur);
+				assert((uintptr_t)*swap < addr_min ||
+				    (uintptr_t)*swap >= addr_max);
+				void *tmp = *cur;
+				*cur = *swap;
+				*swap = tmp;
+				swap++;
+				assert(swap <= cur);
+				assert((uintptr_t)*swap < addr_min ||
```
interwq: I think I get the logic, but it's not easy to follow. To double check: the current approach is trying to maintain the same relative order for the remaining items, is that right? If we don't care about that goal, it can be simplified such as a single loop moving from both ends?

src/tcache.c
nullptr0-0: In L171, we have if (nflush == ncached) goto label_flush; so that if the number of items to be flushed is already equal to ncached, we don't need to go through the locality route at all.

So when we enter the logic of locality flush, nflush should be less than ncached. Unless after locality check, that all the items are remote, and in that case we round the nflush up to nremote, leading nflush == ncached.

src/tcache.c
nullptr0-0: We only update nflush when nflush < nremote, in the case when nflush > nremote, we need to flush some local pointers as well to meet the required number of flush.

src/tcache.c
nullptr0-0: Will the for loop be compiled away in opt build if only assertion in the loop body?

src/tcache.c
```diff
@@ -134,6 +134,151 @@ tcache_gc_item_delay_compute(szind_t szind) {
...
+	assert(addr != NULL);
+	/* The slab address range that the provided addr belongs to. */
+	uintptr_t slab_min = (uintptr_t)addr;
+	uintptr_t slab_max = slab_min + bin_infos[szind].slab_size;
+	/*
+	 * When growing retained virtual memory, it's increased exponentially,
+	 * starting from HUGEPAGE, so that the total number of disjoint virtual
+	 * memory ranges retained by each shard is limited.
+	 */
+	uintptr_t shard_min_heuristic = (slab_min > HUGEPAGE) ?
```
guangli-dai: Would the assumption of starting from `HUGEPAGE` stay true with this [change](https://github.com/jemalloc/jemalloc/pull/2628)?

src/tcache.c
```diff
@@ -169,6 +312,61 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+	 * When the new tcache gc is not enabled, or simply the entire bin needs
+	 * to be flushed, flush the bottom nflush items directly.
+	 */
+	if (!opt_experimental_tcache_gc || nflush == ncached) {
+		goto label_flush;
+	}
+
+	/* Query arena binshard to get heuristic locality info. */
+	void *addr = tcache_gc_small_heuristic_addr_get(tsd, tcache_slow, szind);
+	if (addr == NULL) {
```
guangli-dai, line 424: What does it mean when `addr == NULL`? Does that imply there is no items cached in the bin?

test/unit/arena_decay.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-export MALLOC_CONF="dirty_decay_ms:1000,muzzy_decay_ms:1000,tcache_max:1024"
+export MALLOC_CONF="dirty_decay_ms:1000,muzzy_decay_ms:1000,tcache_max:1024,tcache_gc_incr_bytes:939524096"
```
guangli-dai: Can you add a comment on why this setting for future tracing?

guangli-dai:
Adding a runtime option to guard this is great! Did you run any experiments to confirm that with the option off, the perf would be neutral?

nullptr0-0:
> Adding a runtime option to guard this is great! Did you run any experiments to confirm that with the option off, the perf would be neutral?

Yes, unicorn and servicelab looks neutral.

src/tcache.c
nullptr0-0: yeah, with that change the value will be changed to the min_grow value, wondering if there is a clean way to share that value instead of hardcode in both places.

src/tcache.c
nullptr0-0: We try to take slabcur first, if not available, check the slabs_nonfull heap, if still empty, the addr returns NULL, meaning that the current shard has no partially used slabs.

src/tcache.c
guangli-dai: Maybe put the constant in `jemalloc_internal_decls.h`. But if this PR is landing first, should leave the change to PR #2628.

src/tcache.c
guangli-dai: I see. Since the case is rare and the heuristic way used to identify the locality cannot be trivially applied, it will just be flushing as before, right? I assume this corner case won't affect the locality much as well since it should rarely be touched?

nullptr0-0: Yes, if no locality info can be found, it will just flush the bottom nflush items as before.

src/tcache.c
interwq: > Will the for loop be compiled away in opt build if only assertion in the loop body?

Yes it will be optimized away, as long as the compiler can prove it's not an infinite loop.

src/tcache.c
```diff
@@ -134,6 +134,151 @@ tcache_gc_item_delay_compute(szind_t szind) {
...
+	for (void **cur = head + ntop; cur < head + ncached; cur++) {
+		assert(*cur != NULL);
+		/* Early break if all non-remote ptrs have been moved. */
+		if (cnt == ntop) {
+			break;
+		}
+		if ((uintptr_t)*cur >= addr_min && (uintptr_t)*cur < addr_max) {
+			assert(*(head + cnt) != NULL && (
+			    (uintptr_t)*(head + cnt) < addr_min ||
+			    (uintptr_t)*(head + cnt) >= addr_max));
```
guangli-dai: Judging whether a pointer is remote or local is used multiple times. Maybe put it into a function? This makes it neater.

src/tcache.c
```diff
@@ -134,6 +134,151 @@ tcache_gc_item_delay_compute(szind_t szind) {
...
+tcache_gc_small_nremote_get(cache_bin_t *cache_bin, void *addr,
+    uintptr_t *addr_min, uintptr_t *addr_max, szind_t szind, size_t nflush) {
+	assert(addr != NULL);
+	/* The slab address range that the provided addr belongs to. */
+	uintptr_t slab_min = (uintptr_t)addr;
+	uintptr_t slab_max = slab_min + bin_infos[szind].slab_size;
+	/*
+	 * When growing retained virtual memory, it's increased exponentially,
+	 * starting from HUGEPAGE, so that the total number of disjoint virtual
+	 * memory ranges retained by each shard is limited.
```
interwq, line 251: nit: shard is probably not the right term here. It usually refers to the unit for multiplexing. It looks here we are talking about the region that's close to the heuristic addr? Maybe `neighbor`?

src/tcache.c
```diff
@@ -169,6 +312,61 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+	/*
+	 * Move the remote points to the bottom of the bin for flushing.
+	 * As long as moved to the bottom, the order of these nremote ptrs
+	 * does not matter, since they are going to be flushed anyway.
+	 * The rest of the ptrs are moved to the top of the bin, and their
+	 * relative order is maintained.
+	 */
+	tcache_gc_small_bin_shuffle(cache_bin, nremote, addr_min, addr_max);
+
+label_flush:
```
interwq, line 464: I like the flow of this function. Thanks for the refactoring this @nullptr0-0 . One last nit is, update the first line / comment in this function to say also attempt to flush remote items.

include/jemalloc/internal/tcache_types.h
```diff
@@ -23,5 +23,6 @@ typedef struct tcaches_s tcaches_t;
 #define TCACHE_MAXCLASS_LIMIT ((size_t)1 << TCACHE_LG_MAXCLASS_LIMIT)
 #define TCACHE_NBINS_MAX (SC_NBINS + SC_NGROUP *			\
     (TCACHE_LG_MAXCLASS_LIMIT - SC_LG_LARGE_MINCLASS) + 1)
+#define TCACHE_GC_SMALL_NBINS_MAX (SC_NBINS >> 3)
```
interwq: To be extra safe: `((SC_NBINS > 8) ? SC_NBINS / 8 : 1)`

src/tcache.c
```diff
@@ -155,8 +155,8 @@ tcache_gc_small_heuristic_addr_get(tsd_t *tsd, tcache_slow_t *tcache_slow,
 
 static inline cache_bin_sz_t
 tcache_gc_small_nremote_get(cache_bin_t *cache_bin, void *addr,
-    uintptr_t *addr_min, uintptr_t *addr_max, szind_t szind, size_t nflush) {
-	assert(addr != NULL);
```
interwq: Probably unintentional change?

src/tcache.c
```diff
@@ -364,22 +377,45 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
 	tcache_gc_small_bin_shuffle(cache_bin, nremote, addr_min, addr_max);
 
 label_flush:
-	assert(nflush > 0 && nflush <= ncached);
+	if (nflush == 0) {
+		return false;
```
interwq, line 467: We should be able to `assert(low_water == 0)` here?

Also question: is it true that the changes made in this function has no logical change, but only refactor to facilitate the time driven GC?

src/tcache.c
```diff
@@ -364,22 +377,45 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+	}
+	unsigned nrem = (unsigned)(cache_bin_ncached_get_local(cache_bin) -
+	    low_water + (low_water >> 2));
+	tcache_bin_flush_large(tsd, tcache, cache_bin, szind, nrem);
+	return true;
+}
+
+static bool
+tcache_gc(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
+    szind_t szind, bool is_small) {
```
interwq: szind and is_small are redundant. We do similar thing down the fast path to avoid redundant calculation, but don't need to in other cases. Let's only pass in szind, and decide is_small within this function.

src/tcache.c
```diff
@@ -390,40 +426,68 @@ tcache_event(tsd_t *tsd) {
...
 	}
 
-	tcache_bin_flush_stashed(tsd, tcache, cache_bin, szind, is_small);
-	cache_bin_sz_t low_water = cache_bin_low_water_get(cache_bin);
-	if (low_water > 0) {
-		if (is_small) {
-			tcache_gc_small(tsd, tcache_slow, tcache, szind);
-		} else {
-			tcache_gc_large(tsd, tcache_slow, tcache, szind);
+	nstime_t now;
```
interwq, line 534: Slightly different steps that's simpler:
```
nstime_t now;
nstime_copy(&now, &tcache_slow->last_gc_time);
nstime_update(&now);
```
Then This way you won't have to handle non monotonic clock yourself.

Also use these facilities to read the time instead of direct access, for additional sanity checks:
https://github.com/jemalloc/jemalloc/blob/8c54637f8c7a98bbaec6ee38229a904bbf22170c/src/nstime.c#L60

src/tcache.c
```diff
@@ -390,40 +426,68 @@ tcache_event(tsd_t *tsd) {
...
-	} else if (is_small && tcache_slow->bin_refilled[szind]) {
-		assert(low_water == 0);
-		/*
-		 * Increase fill count by 2X for small bins.  Make sure
-		 * lg_fill_div stays greater than 0.
-		 */
-		if (tcache_slow->lg_fill_div[szind] > 1) {
-			tcache_slow->lg_fill_div[szind]--;
+		assert(ind_gc_small < SC_NBINS);
+		if ((gc_small_nbins += tcache_gc(tsd, tcache_slow, tcache,
```
interwq: tcache_gc returns a bool. Let's break this into separate lines. Maybe:
```
if (tcache_gc(...)) {
    gc_small_nbins++;
}
```
and then move the TCACHE_GC_SMALL_NBINS_MAX check to the for loop 
`i < small_nbins && gc_small_nbins < TCACHE_GC_SMALL_NBINS_MAX`

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -195,11 +195,12 @@ tcache_dalloc_small(tsd_t *tsd, tcache_t *tcache, void *ptr, szind_t binind,
...
 
 	if (unlikely(!cache_bin_dalloc_easy(bin, ptr))) {
-		if (unlikely(tcache_bin_disabled(binind, bin,
-		    tcache->tcache_slow))) {
+		tcache_slow_t *tcache_slow = tcache->tcache_slow;
+		if (unlikely(tcache_bin_disabled(binind, bin, tcache_slow))) {
 			arena_dalloc_small(tsd_tsdn(tsd), ptr);
 			return;
 		}
+		tcache_slow->lg_fill_div_offset[binind] = 0;
```
interwq: Move this into `tcache_bin_flush_small` which is a slow path? Putting it here as inlined may cause unwanted side effects on the fast path.

src/tcache.c
nullptr0-0: @interwq wdty?

src/tcache.c
nullptr0-0: correct! Basically given a low_water value to guide how much to flush, before the change, it's flushing bottom 3/4 low water; now it's moving remote ptrs to the bottom first, then flushing bottom max(nremote, 3/4 low water) items.

src/tcache.c
interwq: Maybe a hardcoded 2M is fine here? It's a huristic and doesn't really have to match the grow (which only starts with 2M but increases fast).

src/tcache.c
```diff
@@ -134,10 +134,155 @@ tcache_gc_item_delay_compute(szind_t szind) {
...
+	void **head = cache_bin->stack_head;
+	cache_bin_sz_t n_remote_slab = 0, n_remote_neighbor = 0;
+	cache_bin_sz_t ncached = cache_bin_ncached_get_local(cache_bin);
+	for (void **cur = head; cur < head + ncached; cur++) {
+		n_remote_slab += (cache_bin_sz_t)tcache_gc_is_addr_remote(*cur,
+		    slab_min, slab_max);
+		n_remote_neighbor += (cache_bin_sz_t)tcache_gc_is_addr_remote(*cur,
+		    neighbor_min, neighbor_max);
+	}
+	assert(n_remote_slab >= n_remote_neighbor && n_remote_slab <= ncached);
```
interwq: We can only assert both <= ncached, not to each other, because the slab size is dynamic as well (even though it's unlikely we want 2M slabs).

(but I think the logic below is fine, as it assume slab < 2M is the common case)

src/tcache.c
```diff
@@ -169,12 +312,70 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+	 * Move the remote points to the bottom of the bin for flushing.
+	 * As long as moved to the bottom, the order of these nremote ptrs
+	 * does not matter, since they are going to be flushed anyway.
+	 * The rest of the ptrs are moved to the top of the bin, and their
+	 * relative order is maintained.
+	 */
+	tcache_gc_small_bin_shuffle(cache_bin, nremote, addr_min, addr_max);
+
+label_flush:
+	assert(nflush > 0 && nflush <= ncached);
```
interwq: It took me a while to realize that `nflush > 0` is made sure by low_water > 0:
https://github.com/jemalloc/jemalloc/blob/5b72ac098abce464add567869d082f2097bd59a2/src/tcache.c#L207

Maybe put an assert close to the beginning of this function of nflush > 0

src/tcache.c
```diff
@@ -379,10 +397,31 @@ tcache_gc_large(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
...
+		return false;
+	}
+	unsigned nrem = (unsigned)(cache_bin_ncached_get_local(cache_bin) -
+	    low_water + (low_water >> 2));
+	tcache_bin_flush_large(tsd, tcache, cache_bin, szind, nrem);
+	return true;
+}
+
+static bool
+tcache_gc(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
```
interwq: nit: Maybe `tcache_gc_one_bin` or `tcache_try_gc`? The current name seems too generic

src/tcache.c
```diff
@@ -306,8 +306,10 @@ tcache_gc_small(tsd_t *tsd, tcache_slow_t *tcache_slow, tcache_t *tcache,
 		if (tcache_slow->lg_fill_div[szind] > 1) {
 			tcache_slow->lg_fill_div[szind]--;
 		}
+		tcache_slow->lg_fill_div_offset[szind] = 0;
 		tcache_slow->bin_refilled[szind] = false;
 	}
+	assert(tcache_slow->lg_fill_div_offset[szind] == 0);
```
interwq: How is this guaranteed? It seems we reset it either on flush, or right here 2 lines above. If it's indeed guaranteed, worth adding some comments explain why.

Since lg_fill_div and _offset are tied together, we should add helpers to handle both together, e.g. getter to return the effective lg_div, and helper to increment or decrement lg_fill_div which handles the lg_fill_div_offset internally. Then rename `lg_fill_div` itself to a longer name to make clear not access directly.

src/tcache.c
interwq: Never mind here. It got changed in the next commit.

src/tcache.c
nullptr0-0: makes sense, I'll add some comment on the special case when slab > 2M.

src/tcache.c
guangli-dai: I vote for `tcache_gc_one_bin`. While most functions return true when it fails, this function returns true when the bin specified is flushed. Maybe add a comment for this.

src/tcache.c
```diff
@@ -516,15 +518,22 @@ tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena,
...
+	    >> (lg_div - (opt_experimental_tcache_gc ? lg_offset : 0));
 	if (nfill == 0) {
 		nfill = 1;
 	}
 	arena_cache_bin_fill_small(tsdn, arena, cache_bin, binind,
 	    /* nfill_min */ opt_experimental_tcache_gc ?
 	    ((nfill >> 1) + 1) : nfill, /* nfill_max */ nfill);
 	tcache_slow->bin_refilled[binind] = true;
+	if (lg_offset + 1 < lg_div) {
+		tcache_slow->lg_fill_div_offset[binind]++;
```
guangli-dai: I think it's worth adding a comment here to explain how `lg_fill_div_offset` is updated and the reason why we don't simply reduce one to `lg_fill_div`.

guangli-dai:
While the default of `opt_experimental_tcache_gc` is false, maybe explicitly set that to true in some tests so that the new feature can be tested in the CI?

nullptr0-0:
> While the default of `opt_experimental_tcache_gc` is false, maybe explicitly set that to true in some tests so that the new feature can be tested in the CI?

I can simply change the opt_experimental_tcache_gc to true just for the sake of triggering CI and switch it back once all tests cleared?

guangli-dai:
> I can simply change the opt_experimental_tcache_gc to true just for the sake of triggering CI and switch it back once all tests cleared?

I think that works for now.

src/tcache.c
nullptr0-0: It's similar to `bin_refilled`, when `low_water > 0`, there was no fill between GC, so the offset was not changed since last GC; otherwise when there was refill between GC, as inside the `else if (tcache_slow->bin_refilled[szind])` branch, we reset both `refill` to `false` and `lg_fill_div_offset` to 0, so it's again 0.

include/jemalloc/internal/tcache_structs.h
```diff
@@ -39,8 +39,13 @@ struct tcache_slow_s {
...
 	szind_t		next_gc_bin_large;
-	/* For small bins, fill (ncached_max >> lg_fill_div). */
-	uint8_t		lg_fill_div[SC_NBINS];
+	/*
+	 * For small bins, used to calculate how many items to fill at a time.
+	 * Containing two fields, with the high 8 bits being the base, and the
+	 * low 8 bits being the offset.
+	 * The final nfill is calculated by (ncached_max >> (base - offset)).
+	 */
+	uint16_t	bin_fill_ctl[SC_NBINS];
```
interwq: You are on the right track. Let's make this a struct (with 2 uint8_t inside) to make it as explicit as possible. That should also simplify things a bit, since you won't have to manually shift things around anymore.

interwq: e.g. 
```
struct bin_fill_ctl_s {
  uint8 base;
  uint8 offset;
}
...
// then in tcache_slow
    bin_fill_ctl_t bin_fill_ctl_dont_access_directly[SC_NBINS];
```

src/tcache.c
```diff
@@ -121,6 +121,80 @@ tcache_gc_dalloc_postponed_event_wait(tsd_t *tsd) {
...
+ * The actual ctl struct consists of two fields, i.e. base and offset,
+ * and the difference between the two(base - offset) is the final lg_fill_div.
+ * The base is adjusted during GC based on the traffic within a period of time,
+ * while the offset is updated in real time to handle the immediate traffic.
+ */
+static inline uint8_t
+tcache_nfill_small_lg_div_get(tcache_slow_t *tcache_slow, szind_t szind) {
+	cache_bin_fill_ctl_t *ctl =
+	    &tcache_slow->bin_fill_ctl_do_not_access_directly[szind];
+	assert(ctl->base > ctl->offset);
```
interwq: Add a getter function, e.g. 

```
cache_bin_fill_ctl_get(tcache_slow) {
    cache_bin_fill_ctl_t *ctl = &tcache_slow->bin_fill_ctl_do_not_access_directly[szind];
    assert(ctl->base > ctl->offset);
    return ctl;
}
```
When we have the do_not_access_directly fields, we should limit where they are referenced. Typically only in getter, setter and init functions. You also don't have to repeat the asserts in the callers.

src/tcache.c
```diff
@@ -121,6 +121,80 @@ tcache_gc_dalloc_postponed_event_wait(tsd_t *tsd) {
...
+		 * the fill count is always at least 1.
+		 */
+		ctl->base++;
+	}
+}
+
+/* Update the offset, which is used to handle burst loads. */
+static inline void
+tcache_nfill_small_burst_adjust(tcache_slow_t *tcache_slow, szind_t szind,
+    bool may_have_burst) {
```
interwq: Let's split this into the two functions? e.g. tcache_nfill_small_burst_reset and tcache_nfill_small_burst_trigger (or _prepare)?

We are not sharing much logic in the current adjust function, and it's a bit confusing since we have another _adjust function right above it.

interwq (approved):

guangli-dai (approved):

PR #2686, opened 2024-08-06
Hugepages Aware Allocator: fixes for purging logic special cases

ilvokhin:
Commit 867c6dd7dc introduced a bug. Intention was to limit number of purges and respect `hpa_min_purge_interval_ms` option, but instead we ended up purging at most one page at a time. This PR is attempt to fix this and provide a way to have backward compatible behaviour to the buggy version, because turns out despite the bug this accidental algorithm proved to have properties useful in certain cases.

- Logic of `hpa_shard_maybe_do_deferred_work` was simplified without noticeable changes in behaviour.
- Check for `hpa_strict_min_purge_interval` was moved out of `hpa_try_purge`. We'll check only once, before we start to purge.
- Option `hpa_max_purge_nslabs` was added to make it possible to preserve previous behaviour if needed.
- Extended unit tests to cover newly added logic.

As HPA is in active development stage there is no guarantee for stability of options. This is especially true for `hpa_strict_min_purge_interval` and `experimental_hpa_max_purge_nhp`. It will make sense to get rid of them as soon as we figure out how end version of purging algorithm should look like.

### DIFF
--- a/include/jemalloc/internal/hpa_opts.h
+++ b/include/jemalloc/internal/hpa_opts.h
@@ -56,7 +56,7 @@ struct hpa_shard_opts_s {
 	 * This is an option to provide backward compatibility for staged rollout of
 	 * purging logic fix.
 	 */
-	bool strict_min_purge_interval;
+	bool experimental_strict_min_purge_interval;
 
 	/*
 	 * Maximum number of hugepages to purge on each purging attempt.
@@ -83,7 +83,7 @@ struct hpa_shard_opts_s {
 	10 * 1000,							\
 	/* min_purge_interval_ms */					\
 	5 * 1000,							\
-	/* strict_min_purge_interval */					\
+	/* experimental_strict_min_purge_interval */			\
 	false,								\
 	/* experimental_max_purge_nhp */				\
 	-1								\
--- a/include/jemalloc/internal/hpa_opts.h
+++ b/include/jemalloc/internal/hpa_opts.h
@@ -57,6 +57,11 @@ struct hpa_shard_opts_s {
 	 * purging logic fix.
 	 */
 	bool strict_min_purge_interval;
+
+	/*
+	 * Maximum number of hugepages to purge on each purging attempt.
+	 */
+	ssize_t experimental_max_purge_nhp;
 };
 
 #define HPA_SHARD_OPTS_DEFAULT {					\
@@ -79,7 +84,9 @@ struct hpa_shard_opts_s {
 	/* min_purge_interval_ms */					\
 	5 * 1000,							\
 	/* strict_min_purge_interval */					\
-	false								\
+	false,								\
+	/* experimental_max_purge_nhp */				\
+	-1								\
 }
 
 #endif /* JEMALLOC_INTERNAL_HPA_OPTS_H */
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -378,18 +378,6 @@ static bool
 hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
 	malloc_mutex_assert_owner(tsdn, &shard->mtx);
 
-	/*
-	 * Make sure we respect purge interval setting and don't purge
-	 * too frequently.
-	 */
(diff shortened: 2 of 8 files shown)


include/jemalloc/internal/hpa_opts.h
```diff
@@ -57,6 +57,11 @@ struct hpa_shard_opts_s {
 	 * purging logic fix.
 	 */
 	bool strict_min_purge_interval;
+
+	/*
+	 * Maximum number of slabs to purge on each purging attempt.
+	 */
+	ssize_t max_purge_nslabs;
```
interwq: Since this is only for backward compatibility with intention to be removed, let's prefix the name with `experimental_`. It adds a small benefit of not being treated as invalid options: https://github.com/jemalloc/jemalloc/blob/a25b9b8ba91881964be3083db349991bbbbf1661/src/jemalloc.c#L935

src/hpa.c
```diff
@@ -532,41 +532,41 @@ hpa_shard_maybe_do_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard,
...
 		}
 		malloc_mutex_assert_owner(tsdn, &shard->mtx);
+		nops++;
+	}
+
+	/*
+	 * Try to hugify at least once, even if we out of operations to make at
+	 * least some progress on hugification even at worst case.
+	 */
+	while (hpa_try_hugify(tsdn, shard) && nops < max_ops) {
```
interwq, line 589: The change means we prioritize purging over hugifying (which makes sense to me). Did you try this in a workload to see how much difference it makes, for both purging rate and hugifying rate?

interwq:
CI failure not relevant. Travis ppc64 jobs have not been working for a while (e.g. https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/624863898). We should consider removing it.

include/jemalloc/internal/hpa_opts.h
ilvokhin: Thanks for suggestion, done. Do you think it make sense to do the same for `strict_min_purge_interval` option? It is also supposed to be temporary.

src/hpa.c
ilvokhin: > Did you try this in a workload to see how much difference it makes, for both purging rate and hugifying rate?

Yes, tried on top-3 biggest workload available to me, which are used HPA. There is no measurable differences in behaviour.

ilvokhin:
Thanks for review, @interwq.

Renamed to `hpa_max_purge_nslabs` option to `experimental_hpa_max_purge_nhp`. To highlight it is experimental and to avoid ambiguity with slabs from small allocations.

include/jemalloc/internal/hpa_opts.h
interwq: Feel free to rename strict_min_purge_interval as well. However if it's already used in production, leave it as-is works too.

ilvokhin: Let me rename it for consistency, we'll need to change config anyway to apply a new option, so we can rename another one as well.

test/unit/hpa_background_thread.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-export MALLOC_CONF="hpa_dirty_mult:0.001,hpa_hugification_threshold_ratio:1.0,hpa_min_purge_interval_ms:50,hpa_strict_min_purge_interval:true,hpa_sec_nshards:0"
+export MALLOC_CONF="hpa_dirty_mult:0.001,hpa_hugification_threshold_ratio:1.0,hpa_min_purge_interval_ms:50,hpa_sec_nshards:0"
```
ilvokhin, line 3: `hpa_strict_min_purge_interval` removed, because we have a unit test for it and here it actually doesn't make any difference.

interwq (approved):

ilvokhin:
Added additional checks into test/unit/mallctl.c to check `experimental_hpa_strict_min_purge_interval` and `experimental_hpa_max_purge_nhp` are handled by `mallctl` as expected.

PR #2690, opened 2024-08-12
Allow disabling config from file or envvar through a build flag

roblabla:
This adds a new autoconf flag, `--disable-user-config`, which disables reading the configuration from /etc/malloc.conf or the MALLOC_CONF environment variable. This can be useful when integrating jemalloc in a binary that internally handles all aspects of the configuration and shouldn't be impacted by ambient change in the environment.

In particular, I'm working on a program that runs with elevated privileges. I'd like to use jemalloc as our allocator, but would like to avoid increasing the attack surface as much as possible.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1345,6 +1345,23 @@ if test "x$enable_stats" = "x1" ; then
 fi
 AC_SUBST([enable_stats])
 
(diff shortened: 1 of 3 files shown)


interwq:
Thank you for the PR @roblabla ! Merging

PR #2695, opened 2024-08-23
Better trigger race condition in bin_batching unit test

nullptr0-0:
The race condition in bin_batching unit test is not triggered when cross compile for aarch64, the change forces thread switches in order to simulate the race condition the test case is expecting.


interwq (approved):

guangli-dai (approved):
Failing tests are irrelevant ppc tests. They have been failing for a while. Will open a PR to disable them.

PR #2697, opened 2024-08-23
Remove tests for ppc64 on Travic CI.

guangli-dai:
Removing tests for ppc64 on Travis since they have been down for a while.


scripts/gen_travis.py
```diff
@@ -316,7 +316,6 @@ def main():
         # generate_freebsd(AMD64),
 
         generate_linux(AMD64),
-        generate_linux(PPC64LE),
```
interwq, line 319: Maybe comment this out instead of removing. Plus add comment explaining why.

PR #2699, opened 2024-08-28
Add travis tests for arm64.

guangli-dai:
Adding travis tests for ARM64 on usual cases. This adds 30 new tests: 8 new tests for Windows, 13 new tests for Linux, and 9 new tests for macOS.

### DIFF
--- a/scripts/gen_travis.py
+++ b/scripts/gen_travis.py
@@ -246,6 +246,9 @@ def generate_linux(arch):
     if arch == PPC64LE:
         # Avoid 32 bit builds and clang on PowerPC
         exclude = (CROSS_COMPILE_32BIT, CLANG,)
+    if arch == ARM64:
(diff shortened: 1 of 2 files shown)


scripts/gen_travis.py
```diff
@@ -318,8 +319,10 @@ def main():
         generate_linux(AMD64),
         # PPC tests on travis has been down for a while, disable it for now.
         # generate_linux(PPC64LE),
+        generate_linux(ARM64),
```
interwq, line 324: It looks only this is taking effect, i.e. arm64 on Linux with 13 new tests added. Maybe remove the other two unsupported platforms.

Also it looks 32-bit is not supported:
`Package gcc-multilib is not available, but is referred to by another package.`

So you may want to avoid that combination as well.

PR #2702, opened 2024-08-31
Use MSVC __declspec(thread) to implement TSD on Windows

mcfi:
MSVC supports thread-local storage through __declspec(thread), and thus we can use it to implement TSD on Windows. Compared to the original approach, we don't need manually manage thread-local memory allocation/deallocation, and getting thread-local data is directly inlined and doesn't have to go through TlsGetValue/TlsGetValue2, which helps improve performance.

Since __declspec(thread) has some limitations on pre-Vista Windows, we still keep the old TSD implementation just in case applications using jemalloc need to support pre-Vista Windows versions.

I used the same microbenchmark on [PR2583](https://github.com/jemalloc/jemalloc/pull/2583) to showcase the performance improvement, see below.

```
#define JEMALLOC_NO_DEMANGLE
#include <chrono>
#include <iostream>

const unsigned __int64 ITERATIONS = 0x800000;

void main() {
    static const int sizes[] =
    {
        7, 16, 32, 60, 91, 100, 120, 144, 169, 199, 255
    };
    
    auto start = std::chrono::high_resolution_clock::now();
    for (unsigned __int64 i = 0; i < ITERATIONS; i++)
    {
        for (auto sz : sizes)
        {
            void* p = je_malloc(sz);
            je_free(p);
        }
    }
    auto end = std::chrono::high_resolution_clock::now();
    std::chrono::duration<double> diff = end - start;
    std::cout << "Time: " << diff.count() << " s\n";
}

                                   Vanilla (sec)       This change (sec)      Time Delta
Intel i9-13900KF (x64)               0.99                      0.91               -8.1%
Snapdragon X1E78100 (Arm64)          0.98                      0.93               -5.1%
Snapdragon X1E78100 (Arm64EC)        1.76                      0.93              -47.2%
```

### DIFF
--- a/include/jemalloc/internal/tsd_win.h
+++ b/include/jemalloc/internal/tsd_win.h
@@ -7,10 +7,16 @@
 #include "jemalloc/internal/tsd_internals.h"
 #include "jemalloc/internal/tsd_types.h"
 
+/* val should always be the first field of tsd_wrapper_t since accessing
+   val is the common path and having val as the first field makes it possible
+   that converting a pointer to tsd_wrapper_t to a pointer to val is no more
+   than a type cast. */
 typedef struct {
-	bool initialized;
 	tsd_t val;
+	bool initialized;
 } tsd_wrapper_t;
+
+#if defined(JEMALLOC_LEGACY_WINDOWS_SUPPORT) || !defined(_MSC_VER)
 
 extern DWORD tsd_tsd;
 extern tsd_wrapper_t tsd_boot_wrapper;
@@ -165,3 +171,69 @@ tsd_set(tsd_t *val) {
 	}
 	wrapper->initialized = true;
 }
+
+#else // defined(JEMALLOC_LEGACY_WINDOWS_SUPPORT) || !defined(_MSC_VER)
+
+#define JEMALLOC_TSD_TYPE_ATTR(type) __declspec(thread) type
+
+extern JEMALLOC_TSD_TYPE_ATTR(tsd_wrapper_t) tsd_wrapper_tls;
+extern bool tsd_booted;
+
+/* Initialization/cleanup. */
+JEMALLOC_ALWAYS_INLINE bool
+tsd_cleanup_wrapper(void) {
+	if (tsd_wrapper_tls.initialized) {
+		tsd_wrapper_tls.initialized = false;
+		tsd_cleanup(&tsd_wrapper_tls.val);
+		if (tsd_wrapper_tls.initialized) {
+			/* Trigger another cleanup round. */
+			return true;
+		}
+	}
+	return false;
+}
+
+/* Initialization/cleanup. */
+JEMALLOC_ALWAYS_INLINE bool
+tsd_boot0(void) {
+	_malloc_tsd_cleanup_register(tsd_cleanup_wrapper);
+	tsd_booted = true;
+	return false;
+}
+
+JEMALLOC_ALWAYS_INLINE void
+tsd_boot1(void) {
+	/* Do nothing. */
+}
+
+JEMALLOC_ALWAYS_INLINE bool
+tsd_boot(void) {
+	return tsd_boot0();
+}
+
+JEMALLOC_ALWAYS_INLINE bool
+tsd_booted_get(void) {
+	return tsd_booted;
+}
+
+JEMALLOC_ALWAYS_INLINE bool
+tsd_get_allocates(void) {
+	return false;
+}
+
+/* Get/set. */
+JEMALLOC_ALWAYS_INLINE tsd_t *
+tsd_get(bool init) {
+	return &(tsd_wrapper_tls.val);
+}
+
+JEMALLOC_ALWAYS_INLINE void
+tsd_set(tsd_t *val) {
+	assert(tsd_booted);
+	if (likely(&(tsd_wrapper_tls.val) != val)) {
+		tsd_wrapper_tls.val = (*val);
+	}
+	tsd_wrapper_tls.initialized = true;
+}
+
+#endif // defined(JEMALLOC_LEGACY_WINDOWS_SUPPORT) || !defined(_MSC_VER)
--- a/src/tsd.c
+++ b/src/tsd.c
@@ -22,8 +22,12 @@ JEMALLOC_TSD_TYPE_ATTR(tsd_t) tsd_tls = TSD_INITIALIZER;
 pthread_key_t tsd_tsd;
 bool tsd_booted = false;
 #elif (defined(_WIN32))
+#if defined(JEMALLOC_LEGACY_WINDOWS_SUPPORT) || !defined(_MSC_VER)
 DWORD tsd_tsd;
-tsd_wrapper_t tsd_boot_wrapper = {false, TSD_INITIALIZER};
+tsd_wrapper_t tsd_boot_wrapper = {TSD_INITIALIZER, false};
+#else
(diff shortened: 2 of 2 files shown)


mcfi:
@interwq @guangli-dai could you guys please take a look?

mcfi:
@Orvid, thanks for your suggestion! This is the PR leveraging __declspec(thread), and could you please also help review? Thanks.

Orvid:
This looks good from my side :)

interwq:
Thanks @Orvid.

@mcfi can you please look into the Windows CI failures? e.g. `test/unit/tsd`

The Travis run: https://app.travis-ci.com/github/jemalloc/jemalloc/builds/272163021

and AppVeyor: https://ci.appveyor.com/project/jasone/jemalloc/builds/50530714

mcfi:
> Thanks @Orvid.
> 
> @mcfi can you please look into the Windows CI failures? e.g. `test/unit/tsd`
> 
> The Travis run: https://app.travis-ci.com/github/jemalloc/jemalloc/builds/272163021
> 
> and AppVeyor: https://ci.appveyor.com/project/jasone/jemalloc/builds/50530714

Thanks, @interwq, my latest push should fix all the failures.

mcfi:
@interwq , it looks like the failing checks were because of limited CI capacity instead of this code change. Could you please shed some light on when the community capacity will be restored or if you could manually restart those failing merge checks with CI capacity limit overriden?

include/jemalloc/internal/tsd_win.h
```diff
@@ -8,10 +8,12 @@
 #include "jemalloc/internal/tsd_types.h"
 
 typedef struct {
-	bool initialized;
 	tsd_t val;
+	bool initialized;
```
interwq, line 16: Unintentional change? For the context, `tsd_t` is rather large and may contain stuff rarely touched at its end (TSD_DATA_SLOWER). So putting the init state at the beginning can avoid touching a remote page.

include/jemalloc/internal/tsd_win.h
```diff
@@ -165,3 +167,80 @@ tsd_set(tsd_t *val) {
...
+/* Get/set. */
+JEMALLOC_ALWAYS_INLINE tsd_t *
+tsd_get(bool init) {
+	return &(tsd_wrapper_tls.val);
+}
+
+JEMALLOC_ALWAYS_INLINE void
+tsd_set(tsd_t *val) {
+	assert(tsd_booted);
+	if (likely(&(tsd_wrapper_tls.val) != val)) {
```
interwq, line 233: nit: `tsd_set` is a slow path (only reached during thread init and shutdown), so no need to add the `likely` hint.

src/tsd.c
```diff
@@ -23,7 +23,11 @@ pthread_key_t tsd_tsd;
 bool tsd_booted = false;
 #elif (defined(_WIN32))
 DWORD tsd_tsd;
-tsd_wrapper_t tsd_boot_wrapper = {false, TSD_INITIALIZER};
+#if defined(JEMALLOC_LEGACY_WINDOWS_SUPPORT) || !defined(_MSC_VER)
+tsd_wrapper_t tsd_boot_wrapper = {TSD_INITIALIZER, false};
+#else
+JEMALLOC_TSD_TYPE_ATTR(tsd_wrapper_t) tsd_wrapper_tls = { TSD_INITIALIZER, false };
```
interwq, line 29: I'm a bit lost here. How does the underlying storage work here? Does this already reserve the space for each thread? Or it still needs to do TlsGet / SetValue? In tsd_boot0 there is TlsAlloc, which to my understanding isn't needed for __declspec( thread )?

interwq:
Thanks for checking the CI @mcfi . You don't need to worry about Cirrus CI here (which we only use as backups for BSD). The main CI is Travis (plus AppVeyor in case of Windows).

include/jemalloc/internal/tsd_win.h
mcfi: This was intended because accessing initialized is not the common path, while the common path is tsd_get. Without this change, we need to do tsd_wrapper + ptr_size to get the pointer to val. With this change, the + ptr_size operation is not needed anymore. Sure, accessing initialized may touch a remote page, but that's only done when tsd is set up.

include/jemalloc/internal/tsd_win.h
mcfi: You are right, but likely is added in other similar places, so I kept it there. If you insist, I can remove it, please let me know.

src/tsd.c
mcfi: The way __declspec(thread) works is that variables marked with __declspec(thread) will have a section in the PE file (exe/dll), see https://learn.microsoft.com/en-us/windows/win32/debug/pe-format. The Windows loader looks at the PE file and dynamically allocates a thread-specific region for each newly created thread and copies the initial data from the PE file, and the tls region is reachable from TEB (see https://learn.microsoft.com/en-us/windows/win32/api/winternl/ns-winternl-teb). The MSVC simply emits code that traverses TEB to access the tls block. Simply put, although not 100% accurate, you may consider the compiler inlines TlsGetValue when accessing __declspec(thread). This removes the need to manually manage tls memory allocation/deallocation, and eliminates TlsGetValue calls.

interwq: Thanks for the detailed answer. If the space for __declspec(thread) is reserved by the loader, doesn't that mean TlsAlloc (in tsd_boot0) isn't needed anymore?

include/jemalloc/internal/tsd_win.h
interwq: Makes sense. Can you please add some comments about avoiding the pointer shift to get the val?

mcfi: Sure, done.

src/tsd.c
mcfi: I see your point, thanks. Yes, TlsAlloc is not needed here, sorry I missed it.. To be clear, TlsAlloc doesn't allocate memory for thread-local objects. Instead, it merely allocates an `index` that every thread can use to successfully call TlsSetValue to index a tls object. TlsAlloc is similar to pthread_setspecific, except that it doesn't re-arm the thread clean up routine. That's why on windows, we use a wrapper object with an additional `initialized` field, which indicates whether we need to call thread cleanup routine again.

mcfi:
> Thanks for checking the CI @mcfi . You don't need to worry about Cirrus CI here (which we only use as backups for BSD). The main CI is Travis (plus AppVeyor in case of Windows).

Great, thanks. Looks like both Travis CI and AppVeyor have passed after my last push.

mcfi:
@interwq , sigh, I pushed a new empty iteration to trigger CI tests. All of them passed except some linux tests in Travis CI possibly due to test infra issues.. Anyway, my change shouldn't have any material impact on non-windows systems, and all previous linux CI tests passed. The change should be ready for merge if there's no more comments or concerns.

mcfi:
Friendly ping, @interwq :)

interwq:
@mcfi sorry for the delay here. We are currently occupied by a production issue on hhvm. I'll try getting back to you in a week or so.

re: the CI issue. Yeah it seems cause by the infra issues of Travis. Please feel free to ignore these.

mcfi:
@interwq , do you have time to review this PR next week? That would help us bring the change to another major project that leverages jemalloc. Thanks.

mcfi:
Thanks so much, @interwq !

interwq:
LGTM. Thank you @mcfi for driving and pushing the TLS effort.

Please feel free to ping if any questions. I'm interested in hearing your results as well if you can share.

Issue #2540, opened 2023-09-15
Test failure on macOS Sonoma


fxcoudert:
Hi,

jemalloc 5.3.0 builds on macOS Sonoma (release candidate), but it has a test failure on both Intel and ARM on this macOS version (which we did not see in earlier versions).

```
  test_alignment_errors:test/integration/aligned_alloc.c:24: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 0
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 9
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 17
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 33
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 65
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 129
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 257
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 513
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 1025
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 2049
  test_alignment_errors:test/integration/aligned_alloc.c:32: Failed assertion: (p != ((void*)0) || get_errno() != 22) == (0) --> true != false: Expected error for invalid alignment 4097
[...]

PR #2704, opened 2024-09-07
Add attribute to compile some tests in aligned_alloc.c with no optimization

yanyir:
Building jemalloc with Clang 15+ results in errors during the check phase (#2540). Specifically, the tests `test_alignment_errors` and `test_oom_errors` fail because of Clang's optimizations. Adding  `__attribute__((optnone))` before these tests can prevent the optimizations, allowing the tests to pass.

### DIFF
--- a/test/integration/aligned_alloc.c
+++ b/test/integration/aligned_alloc.c
@@ -13,7 +13,10 @@ purge(void) {
 	    "Unexpected mallctl error");
 }
 
-__attribute__((optnone)) TEST_BEGIN(test_alignment_errors) {
+#if __clang__
+__attribute__((optnone))
+#endif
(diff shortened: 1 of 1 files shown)


interwq:
@yanyir from the context in https://github.com/jemalloc/jemalloc/issues/2540 , it looks to be the compiler not setting errno, which I'm not sure is a valid optimization.

@fxcoudert have you heard any update from the bug report?

fxcoudert:
I have confirmed (on my reduced test case) that the issue was fixed with Xcode 15.3 and later.

thesamesam:
See https://github.com/jemalloc/jemalloc/issues/2823.

PR #2712, opened 2024-09-13
Add safe frame-pointer backtrace unwinder

nslingerland:
This diff adds a frame-pointer based backtrace unwinder that can be used safely in environments that may mix code built with and withut frame pointers. Safe means that unwinding will never segfault even if code built without frame pointers is encountered. Currently it is only supported on Linux, but can be extended to other platforms.

Jemalloc's sampled heap profiling must be enabled from the start of a process. For long-running processes switching to frame-pointer based unwinding can be significant. At Meta switching from libunwind to the frame-pointer unwinder in this diff for a production service reduced unwinding overhead by ~75%.

# Background

Jemalloc currently offers three unwiding strategies:

| method  | macro | type | notes | url |
| -------- | ------- | ----- | ----- |-----|
| libunwind  | `JEMALLOC_PROF_LIBUNWIND`   | DWARF | | [link](https://github.com/libunwind/libunwind) |
| libgcc | `JEMALLOC_PROF_LIBGCC`    |  DWARF | `_Unwind_Backtrace()` | [link](https://github.com/gcc-mirror/gcc/blob/master/libgcc/unwind.inc) |
| gcc    | `JEMALLOC_PROF_GCC`   | frame pointer | `__builtin_frame_address()`, `__builtin_return_address()` | [link](https://gcc.gnu.org/onlinedocs/gcc/Return-Address.html) |

## Frame pointer unwinding
- Fast as it is effectively walking a linked list of stack frames.*
- Requires that all code be built with `-fno-omit-frame-pointer` and `-mno-omit-leaf-frame-pointer` to work correctly.
- If a call stack mixes code with and without frame pointers unwinding will be unreliable. Typically the result is truncated backtraces often ending with invalid stack frame addresses. It is also possible for frame pointer unwinding to result is a segfault when code without framepointers is encountered as registers / stack values not corresponding to stack frames will be dereferenced.

## DWARF based unwinding (e.g. libunwind)
- Works for all code - whether or not built with frame pointers.
- Requires debug information (unstripped binaries).
- Significantly slower than frame pointer based unwinding.*

(*it turns out that jemalloc's existing gcc intrinsics based frame pointer unwinder is actually slower than libunwind (DWARF) due to O(N^2) code generation)

It is increasingly common for Linux environments to be built with frame pointers enabled by default (https://www.brendangregg.com/blog/2024-03-17/the-return-of-the-frame-pointers.html). At Meta this has also long been the default for Linux hosts and binaries built in-house. Still it is possible to encounter third-party code built without frame pointers on production hosts (e.g. Intel Math Kernel Library (MKL), Nvidia CUDA libraries) - so until now the default has been libunwind.

As the documentation for the `__builtin_frame_address()` gcc intrinsic calls out:

> Calling __builtin_frame_address with a value of 0 yields the frame address of the current function, a value of 1 yields the frame address of the caller of the current function, and so forth.
> ...
> Calling this function with a nonzero argument can have unpredictable effects, including crashing the calling program.

# Safe Unwinding

In order to safely unwind using frame pointers the unwinder needs to know the valid bounds for the stack of the current thread. The end of the stack is the current stack pointer address - which can always safely be found with `__builtin_frame_address(0)`. On Linux the stack always grows down from the initial start address. During unwinding the current frame address must always be less than the start of the stack - and must be greater than the previous stack frame address.

A helper function `prof_thread_stack_start()` is added that determines the valid stack start address for the current thread. The stack start address for the main thread on Linux is available in `/proc/<pid>/stats` ([man page](https://man7.org/linux/man-pages/man5/proc_pid_stat.5.html)). For non-main threads `pthread_attr_getstack()` ([man page](https://man7.org/linux/man-pages/man3/pthread_attr_getstack.3p.html)) might be used but because it allocates memory on the heap it isn't a good option for jemalloc backtracing due to reentrancy concerns. Instead we can search for the memory mapping containing the current stack end address in `/proc/<pid>/task/<tid>/maps`. The highest address of the containing mapping is effectively the stack start address.

# Usage

A new configuration option `--enable-prof-frameptr` is added along with a `JEMALLOC_PROF_FRAME_POINTER` macro.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1418,6 +1418,33 @@ if test "x$backtrace_method" = "x" -a "x$enable_prof_libunwind" = "x1" ; then
   fi
 fi
 
+if test `uname -s` = "Linux"
+then
+  AC_ARG_ENABLE([prof-frameptr],
+    [AS_HELP_STRING([--enable-prof-frameptr], [Use optimized frame pointer unwinder for backtracing (Linux only)])],
+  [if test "x$enable_prof_frameptr" = "xno" ; then
+    enable_prof_frameptr="0"
+  else
+    enable_prof_frameptr="1"
+    if test "x$enable_prof" = "x0" ; then
+      AC_MSG_ERROR([--enable-prof-frameptr should only be used with --enable-prof])
+    fi
+  fi
+  ],
+  [enable_prof_frameptr="0"]
+  )
+  if test "x$backtrace_method" = "x" -a "x$enable_prof_frameptr" = "x1" \
+      -a "x$GCC" = "xyes" ; then
+    JE_CFLAGS_ADD([-fno-omit-frame-pointer])
+    backtrace_method="frame pointer linux"
+    AC_DEFINE([JEMALLOC_PROF_FRAME_POINTER], [ ], [ ])
+  else
+    enable_prof_frameptr="0"
+  fi
+else
+  enable_prof_frameptr="0"
+fi
+
 AC_ARG_ENABLE([prof-libgcc],
   [AS_HELP_STRING([--disable-prof-libgcc],
   [Do not use libgcc for backtracing])],
@@ -2817,6 +2844,7 @@ AC_MSG_RESULT([stats              : ${enable_stats}])
 AC_MSG_RESULT([experimental_smallocx : ${enable_experimental_smallocx}])
 AC_MSG_RESULT([prof               : ${enable_prof}])
 AC_MSG_RESULT([prof-libunwind     : ${enable_prof_libunwind}])
+AC_MSG_RESULT([prof-frameptr      : ${enable_prof_frameptr}])
 AC_MSG_RESULT([prof-libgcc        : ${enable_prof_libgcc}])
 AC_MSG_RESULT([prof-gcc           : ${enable_prof_gcc}])
 AC_MSG_RESULT([fill               : ${enable_fill}])
--- a/src/prof_sys.c
+++ b/src/prof_sys.c
@@ -3,6 +3,7 @@
 
 #include "jemalloc/internal/buf_writer.h"
 #include "jemalloc/internal/ctl.h"
+#include "jemalloc/internal/malloc_io.h"
 #include "jemalloc/internal/prof_data.h"
 #include "jemalloc/internal/prof_sys.h"
 
@@ -98,6 +99,45 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 
 	_Unwind_Backtrace(prof_unwind_callback, &data);
 }
+#elif (defined(JEMALLOC_PROF_FRAME_POINTER))
+JEMALLOC_DIAGNOSTIC_PUSH
+JEMALLOC_DIAGNOSTIC_IGNORE_FRAME_ADDRESS
+static void
+prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
+  // stack_start - highest possible valid stack address (assumption: stacks grow downward)
+  //   stack_end - current stack frame and lowest possible valid stack address
+  //               (all earlier frames will be at higher addresses than this)
+
+  // always safe to get the current stack frame address
+  void** stack_end = (void**)__builtin_frame_address(0);
+  if (stack_end == NULL) {
+    *len = 0;
+    return;
+  }
+
+  static __thread void **stack_start = (void **)0;  // thread local
+  if (stack_start == 0 || stack_end >= stack_start) {
+    stack_start = (void**)prof_thread_stack_start((uintptr_t)stack_end);
+  }
+
+  if (stack_start == 0 || stack_end >= stack_start) {
+    *len = 0;
+    return;
+  }
+
+  unsigned ii = 0;
+  void** fp = (void**)stack_end;
+  while (fp < stack_start && ii < max_len) {
+    vec[ii++] = fp[1];
+    void** fp_prev = fp;
+    fp = fp[0];
+    if (unlikely(fp <= fp_prev)) { // sanity check forward progress
+      break;
+    }
+  }
+  *len = ii;
+}
+JEMALLOC_DIAGNOSTIC_POP
 #elif (defined(JEMALLOC_PROF_GCC))
 JEMALLOC_DIAGNOSTIC_PUSH
 JEMALLOC_DIAGNOSTIC_IGNORE_FRAME_ADDRESS
@@ -484,7 +524,7 @@ prof_getpid(void) {
 #endif
 }
 
-long
+static long
 prof_get_pid_namespace() {
 	long ret = 0;
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -139,6 +139,7 @@ any of the following arguments (not a definitive list) to 'configure':
     in the following list that appears to function correctly:
 
     + libunwind      (requires --enable-prof-libunwind)
+    + frame pointer  (requires --enable-prof-frameptr)
     + libgcc         (unless --disable-prof-libgcc)
     + gcc intrinsics (unless --disable-prof-gcc)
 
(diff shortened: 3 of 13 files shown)


src/prof_sys_linux.c
```diff
@@ -0,0 +1,153 @@
...
+            buf,
+            "%*d (%*[^)]) %*c %*d %*d %*d %*d %*d %*u %*u %*u %*u %*u %*u %*u %*d %*d %*d %*d %*d %*d %*u %*u %*d %*u %*u %*u %ld",
+            &stack_start) != 1) {
+    }
+  }
+  malloc_close(fd);
+  return stack_start;
+}
+
+uintptr_t prof_thread_stack_start(uintptr_t stack_end) {
```
interwq: Question: is it true that this is the part that makes it Linux specific? If other platforms would like to utilize the framepointer based unwinder, as long as they implement this part, the rest of the mechanism would work?

nslingerland: The determination of stack address range - yes, but also the implementation of  `prof_backtrace_impl()` is making assumptions about stack layout and growth direction that are not universally true - though quite common (the gcc intrinsics hide the details of that). 

It might be possible to fold the address checks into the existing `gcc` intrinsics unwinder - they would still only be available for linux (for now) of course - but it might be easier to add support for new platforms.

nslingerland: Actually - looking at the generated code for gcc intrinsics based `prof_backtrace_impl()` we probably should avoid that where possible.

# Codegen

The gcc intrinsics approach uses `__builtin_frame_address()` to sanity check the frame - along with `__builtin_return_address()` to get the actual return address that is the unwound stack frame. The problem is that for each of those the compiler (Clang) doesn't seem to realize the relationship between `BT_FRAME(N)` and `BT_FRAME(N+1)` and the generated code becomes O(N^2).

## Clang (trunk)
https://godbolt.org/z/cs7sjPchn

Clang is re-walking the stack twice on every `BT_FRAME(N)`:

`BT_FRAME(4)`
```
        mov     rax, qword ptr [rbp]       # __builtin_frame_address()
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        cmp     qword ptr [rax], 0
        je      .LBB0_31
        mov     rax, qword ptr [rbp]       # __builtin_return_address()
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax + 8]
        test    rax, rax
        je      .LBB0_31
        mov     qword ptr [rdi + 32], rax
        mov     dword ptr [rsi], 5
        cmp     edx, 5
        je      .LBB0_31
```

`BT_FRAME(8)`
```
        mov     rax, qword ptr [rbp]       # __builtin_frame_address()
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        cmp     qword ptr [rax], 0
        je      .LBB0_31
        mov     rax, qword ptr [rbp]       # __builtin_return_address()
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax]
        mov     rax, qword ptr [rax + 8]
        test    rax, rax
        je      .LBB0_31
        mov     qword ptr [rdi + 64], rax
        mov     dword ptr [rsi], 9
        cmp     edx, 9
        je      .LBB0_31
```

## GCC (trunk)
https://godbolt.org/z/KoYEhsx7o

GCC does better - it appears to at least partially combines subsequent instances of `BT_FRAME(N)` but is still re-walking the stack each iteration (but not twice like Clang).

# Benchmark

Testing with a small benchmark that unwinds 40 frame stacks in a loop:

libunwind: 284.00us
gcc: 404.59us
fp-linux: 222.82us

The benchmark bears out the inefficient codegen with the gcc intrinsics path (this was using Clang which is our default compiler).

nslingerland: Back to your original question - is it possible to make the `fp-linux` unwinder work for other platforms - I think so.  It would be better to have a less Linux-specific naming in that case.

src/prof_sys.c
```diff
@@ -484,7 +524,7 @@ prof_getpid(void) {
 #endif
 }
 
-long
+static long
```
guangli-dai, line 527: Why do we need the change here?

guangli-dai (commented):
I also compare this with current unwinders on a stress test. The time taken is:
libunwind: 3873.91s
libgcc: 3957.97s
frameptr: 3873.96s

With some noise in the stress test, I think it matches the perf expectation?

src/prof_sys.c
nslingerland: I think I noticed that `prof_get_pid_namespace()` was unused outside of this file so might as well not export it. This change is not strictly necessary  to this diff.

nslingerland: @guangli-dai  - It seems reasonable if those are overall numbers - especially if the stress test isn't dominated by stack unwinding time (expected if sampling is working).

Can you tell me how to run the stress test?

guangli-dai:
Can you also add `prof_frameptr` to the stats [here](https://github.com/jemalloc/jemalloc/blob/4f4fd424477142ee9962fcf4e4cd0349d4e6e4d3/src/stats.c#L1469)? This should benefit the rollout a lot. Other changes look good to me.

src/prof_sys.c
nslingerland: I profiled a 1 minute portion of the microbenchmark test and found that `prof_backtrace_impl()` takes about 0.05% inclusive time with libunwind - so I think the benchmark results make sense. From earlier measurements we expect the frameptr version to take ~25% of the time - or in absolute terms a difference of 3873.91s * 0.0005 * 0.75 = 1.45s which seems like it could be within the noise of the test (as you mentioned). I will run a few more benchmark runs to understand the variation run-to-run.

guangli-dai (approved):

src/prof_sys.c
nslingerland: For 5x runs on my machine here's what I saw:

Average user time: 3596.75s
Average user time: 3635.13s
Average user time: 3624.83s
Average user time: 3610.44s
Average user time: 3618.15s

which confirms that the frameptr unwinder savings in the microbenchmark case are well within the noise.

configure.ac
```diff
@@ -1418,6 +1418,34 @@ if test "x$backtrace_method" = "x" -a "x$enable_prof_libunwind" = "x1" ; then
...
+      AC_MSG_ERROR([--enable-prof-frameptr should only be used with --enable-prof])
+    fi
+  fi
+  ],
+  [enable_prof_frameptr="0"]
+  )
+  if test "x$backtrace_method" = "x" -a "x$enable_prof_frameptr" = "x1" \
+      -a "x$GCC" = "xyes" ; then
+    JE_CFLAGS_ADD([-fno-omit-frame-pointer])
+    JE_CFLAGS_ADD([-mno-omit-leaf-frame-pointer])
```
interwq: FYI it looks the flag might not be present on some targets, e.g. on ARM as mentioned here: https://github.com/Jguer/yay/issues/2411#issuecomment-2028471940

Since it's opt in so probably fine as is.

Question: is it true that, if we don't include these, we will not get any meaningful result, because jemalloc functions are always the top of stack, and without valid frame pointers we cannot walk through to application frames?

nslingerland: From what I've seen lack of `-mno-omit-leaf-frame-pointer` doesn't break frame pointer based unwinding entirely (as long as `-fno-omit-frame-pointer` is still used - that's the critical one).

Without `-mno-omit-leaf-frame-pointer` frame pointer based unwinding can result in confusing stacks for leaf functions because a new frame isn't added by the compiler for leaf function calls. I'd say that's maybe not a big issue here as it would manifest only for jemalloc internals.

interwq: I see. In that case, do you think we should skip `-mno-omit-leaf-frame-pointer`, for cases where it's not supported (assuming `-fno-omit-frame-pointer` is more widely available)? Like you mentioned the last frame is always jemalloc internal and we even try to filter them out: https://github.com/jemalloc/jemalloc/blob/dev/bin/jeprof.in#L2956

nslingerland: It's probably easiest to not use it anywhere. I'll remove the flag.

PR #2718, opened 2024-09-20
Fix the locked flag for malloc_mutex_trylock().

interwq:


interwq:
@guangli-dai ready for review.

The added sanity check helped catching the missing state tracking from the pthread_cond_wait case.

interwq:
Static analysis caught an existing missing init value issue. Fixed by the last commit in this PR.

include/jemalloc/internal/mutex.h
```diff
@@ -216,7 +221,6 @@ malloc_mutex_lock(tsdn_t *tsdn, malloc_mutex_t *mutex) {
 	if (isthreaded) {
 		if (malloc_mutex_trylock_final(mutex)) {
 			malloc_mutex_lock_slow(mutex);
-			atomic_store_b(&mutex->locked, true, ATOMIC_RELAXED);
```
guangli-dai, line 219: Good catch! Maybe comment the idea that "the thread holding the lock should be the one updating `mutex->locked`" somewhere to prevent future mistakes on updating it? Its definition [here](https://github.com/jemalloc/jemalloc/blob/db4f0e71820017039f09e5acc04b554826e304fd/include/jemalloc/internal/mutex.h#L40) is a good candidate.

src/background_thread.c
```diff
@@ -153,6 +153,26 @@ set_current_thread_affinity(int cpu) {
...
+static int
+background_thread_cond_wait(background_thread_info_t *info,
+    struct timespec *ts) {
+	int ret;
+
+	/*
+	 * pthread_cond_wait drops and re-acquires the mutex internally, w/o
+	 * going through our wrapper.  Update the locked state explicitly.
+	 */
+	atomic_store_b(&info->mtx.locked, false, ATOMIC_RELAXED);
```
guangli-dai, line 165: When is the `locked` used?

interwq: It's not used in this sequence (cond_wait), but because it's locking-unlocking w/o updating the flag in our wrapper, it's causing confusion for other threads, and will trigger assertions (that's how it's found).

include/jemalloc/internal/mutex.h
interwq: Added comments there.

guangli-dai (approved):

interwq:
CI run successful except the Travis ARM64 unrelated to PR.

PR #2719, opened 2024-09-21
Assert the mutex is locked within malloc_mutex_assert_owner().

interwq:
The background thread locking during boot issue is caught by the newly added assertion.

A followup on #2718

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -819,7 +819,6 @@ background_thread_boot1(tsdn_t *tsdn, base_t *base) {
 	}
 	max_background_threads = opt_max_background_threads;
 
-	background_thread_enabled_set(tsdn, opt_background_thread);
 	if (malloc_mutex_init(&background_thread_lock,
 	    "background_thread_global",
 	    WITNESS_RANK_BACKGROUND_THREAD_GLOBAL,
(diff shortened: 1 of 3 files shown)


guangli-dai:
The failure in [Cirrus CI](https://github.com/jemalloc/jemalloc/pull/2719/checks?check_run_id=30461909566) looks real and relevant. Can you take a look?

interwq:
@guangli-dai I missed the `isthreaded` check there. Fixed now.

src/background_thread.c
```diff
@@ -850,7 +849,8 @@ background_thread_boot1(tsdn_t *tsdn, base_t *base) {
 		background_thread_info_init(tsdn, info);
 		malloc_mutex_unlock(tsdn, &info->mtx);
 	}
+	/* Using _impl to bypass the locking check during init. */
+	background_thread_enabled_set_impl(opt_background_thread);
```
guangli-dai, line 853: The initialization above (all the background thread relevant inits) is free from racy conditions and putting this setting before or after the initialization is fine, is that correct?

interwq: Right. All the `_boot()` functions are covered by the malloc_init lock.

interwq: (but I moved it to the end of the function, because there are still error exit paths in the middle, and technically we should not mark it as enabled unless all init succeeded)

guangli-dai (approved):

PR #2720, opened 2024-09-23
Fix compilation with MSVC 2022

roblabla:
While working on fixing the windows support of tikv-jemallocator, I hit a compilation error in jemalloc that I hadn't seen before.

On MSVC, `log` is an intrinsic that doesn't require libm. However, AC_SEARCH_LIBS does not successfully detect this, as it will try to compile a program using the wrong signature for log. Newer versions of MSVC CL detects this and rejects the program with the following messages:

```
conftest.c(40): warning C4391: 'char log()': incorrect return type for intrinsic function, expected 'double'
conftest.c(44): error C2168: 'log': too few actual parameters for intrinsic function
```

Since log is always available on MSVC (it's been around since the dawn of time), we can simply always assume it's there if MSVC is detected.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -878,11 +878,26 @@ AC_SUBST([DUMP_SYMS])
 AC_SUBST([CC_MM])
 
 dnl Determine whether libm must be linked to use e.g. log(3).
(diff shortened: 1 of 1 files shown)


interwq:
Thanks for the patch!

PR #2733, opened 2024-10-04
Add support for clock_gettime_nsec_np()

### LINKED ISSUE #2653: Microbenchmark uses wrong timebase on some Darwin systems
On an M2 Apple silicon macbook, microbenchmark reports 0.1-0.2 ns per iteration.

Source code in nstime.c shows mach_absolute_time() is used. This function does not return nanoseconds but 'ticks'.
Developer doc advises to use the equivalent clock_gettime_nsec_np(CLOCK_UPTIME_RAW))

With abovementioned change, iterations now show around 5-8 ns, a figure that seems sensible to me and matches wall-clock times.

See https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time

interwq:
Prefer clock_gettime_nsec_np(CLOCK_UPTIME_RAW) to mach_absolute_time().

Resolves #2653

### DIFF
--- a/src/nstime.c
+++ b/src/nstime.c
@@ -201,11 +201,22 @@ nstime_get(nstime_t *time) {
 	clock_gettime(CLOCK_MONOTONIC, &ts);
 	nstime_init2(time, ts.tv_sec, ts.tv_nsec);
 }
+#elif defined(JEMALLOC_HAVE_CLOCK_GETTIME_NSEC_NP)
+#  define NSTIME_MONOTONIC true
+static void
+nstime_get(nstime_t *time) {
+	nstime_init(time, clock_gettime_nsec_np(CLOCK_UPTIME_RAW));
+}
 #elif defined(JEMALLOC_HAVE_MACH_ABSOLUTE_TIME)
 #  define NSTIME_MONOTONIC true
 static void
 nstime_get(nstime_t *time) {
-	nstime_init(time, mach_absolute_time());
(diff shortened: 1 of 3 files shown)


src/nstime.c
```diff
@@ -201,6 +201,12 @@ nstime_get(nstime_t *time) {
 	clock_gettime(CLOCK_MONOTONIC, &ts);
 	nstime_init2(time, ts.tv_sec, ts.tv_nsec);
 }
+#elif defined(JEMALLOC_HAVE_CLOCK_GETTIME_NSEC_NP)
+#  define NSTIME_MONOTONIC true
+static void
+nstime_get(nstime_t *time) {
+	nstime_init(time, clock_gettime_nsec_np(CLOCK_UPTIME_RAW));
+}
```
guangli-dai, line 209: While `clock_gettime_nsec_np` is preferred, the way we use `mach_absolute_time` should also be updated like [the second example here](https://developer.apple.com/library/archive/qa/qa1398/_index.html) for correctness on older mac os:
```
static mach_timebase_info_data_t    sTimebaseInfo;
if ( sTimebaseInfo.denom == 0 ) {
        (void) mach_timebase_info(&sTimebaseInfo);
}
nstime_init(time, mach_absolute_time() * sTimebaseInfo.numer / sTimebaseInfo.denom);
```

interwq: Fixed

guangli-dai (approved):

PR #2747, opened 2024-10-25
Split `stats_arena_hpa_shard_print` function

ilvokhin:
Make multiple functions from `stats_arena_hpa_shard_print` for readability and ease of change in the future.

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -831,12 +831,16 @@ stats_arena_extents_print(emitter_t *emitter, unsigned i) {
 }
 
 static void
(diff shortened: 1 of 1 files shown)


guangli-dai:
LGTM, just to double check, can you confirm that the stats output is exactly the same after this reconfiguration?

ilvokhin:
> can you confirm that the stats output is exactly the same after this reconfiguration?

Yes, absolutely, output is the same before and after the change.

guangli-dai (approved):

PR #2754, opened 2024-11-08
Add configure check for gettid() presence

### LINKED ISSUE #2740: Issue under "old" Linux distros due to gettid() not being implemented in glibc before v2.30
I encountered an issue when linking jemalloc to a program built under an "old" Linux distro (I use 4-5 years old build systems so that my program can run under all current distros in use).

The problem stems from the use of the gettid() function introduced by commit edc1576f03d15a22b968828b68a074d9be6e5cc0

This function is not available in glibc versions prior to v2.30 and the corresponding code therefore needs to be amended or disabled when built on system with older glibc versions.

As a quick workaround for this issue, and seeing as this new code only affects profiling, I simply patched src/prof_stack_range.c at line 7 to read:
`#if defined (__linux__) && (__GLIBC_MAJOR__ > 2 || __GLIBC_MINOR__ >= 30)`
However, you might want to implement a better fix...

sharkcz:
The gettid() function is available on Linux in glibc only since version 2.30. There are supported distributions that still use older glibc version. Thus add a configure check if the gettid() function is available and extend the check in src/prof_stack_range.c so it's skipped also when gettid() isn't available.

Fixes: https://github.com/jemalloc/jemalloc/issues/2740

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2695,6 +2695,15 @@ if test "x${je_cv_pthread_mutex_adaptive_np}" = "xyes" ; then
   AC_DEFINE([JEMALLOC_HAVE_PTHREAD_MUTEX_ADAPTIVE_NP], [ ], [ ])
 fi
 
+JE_COMPILABLE([gettid], [
+#include <unistd.h>
+], [
+  int tid = gettid();
+], [je_cv_gettid])
+if test "x${je_cv_gettid}" = "xyes" ; then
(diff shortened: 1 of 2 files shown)


interwq:
Thanks for the fix @sharkcz !

cc: @nslingerland since `gettid` is only used by the new unwinder. Maybe we should add a check here as well: https://github.com/jemalloc/jemalloc/blob/dev/configure.ac#L1451

PR #2761, opened 2024-11-14
Expose `psset` state stats

ilvokhin:
When evaluating changes in HPA logic, it is useful to know internal `hpa_shard` state. Great deal of this state is `psset`. Some of the `psset` stats was available, but in disaggregated form, which is not very convenient. This commit exposed `psset` counters to `mallctl` and malloc stats dumps.

Example of how malloc stats dump will look like after the change.

```
HPA shard stats:
      Pageslabs: 14899 (4354 huge, 10545 nonhuge)
      Active pages: 6708166 (2228917 huge, 4479249 nonhuge)
      Dirty pages: 233816 (331 huge, 233485 nonhuge)
      Retained pages: 686306
      Purge passes: 8730 (10 / sec)
      Purges: 127501 (146 / sec)
      Hugeifies: 4358 (5 / sec)
      Dehugifies: 4 (0 / sec)
```

Pageslabs, active pages, dirty pages and retained pages are rows added by this change.

### DIFF
--- a/include/jemalloc/internal/psset.h
+++ b/include/jemalloc/internal/psset.h
@@ -22,6 +22,12 @@
 #define PSSET_NPSIZES 64
 
 /*
+ * We store non-hugefied and hugified pageslabs metadata separately.
+ * [0] corresponds to non-hugified and [1] to hugified pageslabs.
+ */
+#define PSSET_NHUGE 2
+
+/*
  * We keep two purge lists per page size class; one for hugified hpdatas (at
  * index 2*pszind), and one for the non-hugified hpdatas (at index 2*pszind +
  * 1).  This lets us implement a preference for purging non-hugified hpdatas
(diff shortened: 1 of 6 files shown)


test/unit/mallctl.c
```diff
@@ -1001,6 +1001,59 @@ TEST_BEGIN(test_stats_arenas) {
...
+TEST_BEGIN(test_stats_arenas_hpa_shard_counters) {
+#define TEST_STATS_ARENAS_HPA_SHARD_COUNTERS(t, name) do {		\
+	t name;								\
+	size_t sz = sizeof(t);						\
+	expect_d_eq(mallctl("stats.arenas.0.hpa_shard."#name,		\
+	    (void *)&name, &sz,						\
+	    NULL, 0), 0, "Unexpected mallctl() failure");		\
+} while (0)
+
+	TEST_STATS_ARENAS_HPA_SHARD_COUNTERS(size_t, npageslabs);
```
guangli-dai, line 1016: The tests added here report errors under both Linux and mac:
https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/628155342
https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/628155393

Please check

include/jemalloc/internal/psset.h
```diff
@@ -44,21 +50,36 @@ struct psset_bin_stats_s {
...
+	psset_bin_stats_t merged;
+
+	/*
+	 * Below are the same stats, but aggregated by different
+	 * properties of pageslabs: huginess or fullness.
+	 */
+
+	/* Non-huge and huge slabs. */
+	psset_bin_stats_t slabs[PSSET_NHUGE];
+
```
guangli-dai, line 67: While the layout of newly added field looks reasonably to me based on either the size increase and the hotness, can you check the size increase of `psset_s ` with this change just for the record?

guangli-dai (commented):
Can you share the list of stats added and also a sample malloc stats showing the changes?

test/unit/mallctl.c
ilvokhin: Thanks for catching that! I missed there is a `--disable-stats` mode, guarded new tests, when this mode is on.

include/jemalloc/internal/psset.h
ilvokhin: `sizeof(psset_t)` was 5280 byte before the change, and 5328 with this change, which make sense as we added two `psset_bin_stats_t` with three 8 byte integers and `merged` field was just moved from `psset_t` to `psset_stats_t`, which is still member of `psset_t`.

However, not sure we should care much about cache effects to be honest, this is page level allocator, which is guarded by mutex on the already relatively cold path comparing to arena code and even colder comparing to thread cache. So, even in the worst case we shouldn't observe significant performance changes, but I'll check anyway.

guangli-dai: Yes, that's true, I have also run tests which confirm that CPU and memory are neutral.

test/unit/mallctl.c
guangli-dai: For better readability, maybe simply skip the tests if `config_stats` is false, i.e., `test_skip_if(!config_stats)` like [here](https://github.com/jemalloc/jemalloc/blob/b82333fdec6e5833f88780fcf1fc50b799268e1b/test/unit/mallctl.c#L1232)

ilvokhin: I was thinking about that as well, but choose this option to align with `test_mallctl_opt`, but I don't have strong preference, replaced to `test_skip_if(!config_stats)`.

guangli-dai (approved):

PR #2762, opened 2024-11-15
Remove validation for HPA ratios

ilvokhin:
Config validation was introduced at 3aae792b with main intention to fix infinite purging loop, but it didn't actually fix the underlying problem, just masked it. Later 47d69b4ea was merged to address the same problem.

Options `hpa_dirty_mult` and `hpa_hugification_threshold` have different application dimensions: `hpa_dirty_mult` applied to active memory on the shard, but `hpa_hugification_threshold` is a threshold for single pageslab (hugepage). It doesn't make much sense to sum them up together.

While it is true that too high value of `hpa_dirty_mult` and too low value of `hpa_hugification_threshold` can lead to pathological behaviour, it is true for other options as well. Poor configurations might lead to suboptimal and sometimes completely unacceptable behaviour and that's OK, that is exactly the reason why they are called poor.

There are other mechanism exist to prevent extreme behaviour, when we hugified and then immediately purged page, see
`hpa_hugify_blocked_by_ndirty` function, which exist to prevent exactly this case.

Lastly, `hpa_dirty_mult + hpa_hugification_threshold >= 1` constraint is too tight and prevents a lot of valid configurations.


guangli-dai (approved):
Discussed offline. Concluded three reasons for removing the validation:
1. it was a mitigation for solving the infinite loop problem;
2. the restriction is too tight and is blocking the testing of some valid configurations, @ilvokhin has tried some configurations banned by the validation and they have fine results;
3. we should still keep some validation but not by adding `hpa_dirty_mult` and `hp_hugification_threshold` up. While exploring, we should remove this limit to benefit tests. Will add a validation back once we figure out some config combinations are unreasonable for anyone to try.

PR #2770, opened 2024-12-12
Disable psset test when hugepage size is too large.

guangli-dai:
With a large hugepage size, the tests of `psset` would cause stack overflow. Currently, HPA is not ready for serving large hugepage sizes such as 512MB, so disabling this test should be fair.

The first commit comes from PR #2771 just to enable the CI tests to make sure this test is fixed as expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -297,6 +297,9 @@ jobs:
       env: CC=gcc CXX=g++ CONFIGURE_FLAGS="--with-lg-page=16" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
     - os: linux
       arch: arm64
+      env: CC=gcc CXX=g++ CONFIGURE_FLAGS="--with-lg-page=16 --with-lg-hugepage=29" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
+    - os: linux
+      arch: arm64
       env: CC=gcc CXX=g++ CONFIGURE_FLAGS="--with-malloc-conf=tcache:false" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
     - os: linux
       arch: arm64
--- a/scripts/gen_travis.py
+++ b/scripts/gen_travis.py
@@ -96,6 +96,15 @@ class Option(object):
         return (isinstance(obj, Option) and obj.type == self.type
                 and obj.value == self.value)
 
+    def __repr__(self):
+        type_names = {
+            Option.Type.COMPILER: 'COMPILER',
+            Option.Type.COMPILER_FLAG: 'COMPILER_FLAG',
+            Option.Type.CONFIGURE_FLAG: 'CONFIGURE_FLAG',
+            Option.Type.MALLOC_CONF: 'MALLOC_CONF',
+            Option.Type.FEATURE: 'FEATURE'
+        }
+        return f"Option({type_names[self.type]}, {repr(self.value)})"
 
 # The 'default' configuration is gcc, on linux, with no compiler or configure
 # flags.  We also test with clang, -m32, --enable-debug, --enable-prof,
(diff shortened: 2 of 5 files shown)


nullptr0-0 (approved):

ilvokhin:
Curious, why 25 constant was chosen? Is there any motivation behind it?

Likely it would be beneficial for future generations to have some reasoning behind it.

Currently, we have `HUGEPAGE_MAX_EXPECTED_SIZE`, which is `16U << 20` ~ `LG_HUGEPAGE=24`.
https://github.com/jemalloc/jemalloc/blob/6786934280392e71a1e14d48b331d4eca58550a7/include/jemalloc/internal/pages.h#L30-L36

Probably, it make sense to reuse the same check here. For example, move `HUGEPAGE > HUGEPAGE_MAX_EXPECTED_SIZE` into a separate function and call it in `psset` tests or call `hpa_supported` (this I like a little bit less, because `psset` logic test doesn't really depends in hugepages support).

.travis.yml
```diff
@@ -295,6 +295,9 @@ jobs:
     - os: linux
       arch: arm64
       env: CC=gcc CXX=g++ CONFIGURE_FLAGS="--with-lg-page=16" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
+    - os: linux
+      arch: arm64
+      env: CC=gcc CXX=g++ CONFIGURE_FLAGS="--with-lg-page=16 --with-lg-hugepage-29" EXTRA_CFLAGS="-Werror -Wno-array-bounds"
```
ilvokhin: Probably should be `--with-lg-hugepage=29` instead of `--with-lg-hugepage-29`. Here and below in few places.

scripts/gen_travis.py
```diff
@@ -125,7 +134,9 @@ def __eq__(self, obj):
     '--disable-libdl',
     '--enable-opt-safety-checks',
     '--with-lg-page=16',
+    "--with-lg-page=16 --with-lg-hugepage-29",
```
ilvokhin: All other options have single quotes, but last one have double quotes.

ilvokhin: Also, I am not an expert in Travis config to be honest, so this might be wrong. My understanding is that we generate combinations from `configure_flag_unusuals` to generate and test all unusual configurations of combinations. If this is true, probably we can omit `--with-lg-page=16` part and add only `--with-lg-hugepage=29` here, so combination `--with-lg-page=16 --with-lg-hugepage=29` will be generated as well.

scripts/gen_travis.py
```diff
@@ -125,7 +134,9 @@ def __eq__(self, obj):
...
 )]
+LARGE_HUGEPAGE = Option.as_configure_flag("--with-lg-page=16 --with-lg-hugepage-29")
```
ilvokhin: `--with-lg-hugepage-29` -> `--with-lg-hugepage=29`

ilvokhin (commented):
Thanks for extending CI configs as well!

Core change looks good to me, left couple of comments about CI config changes that might be useful.

.travis.yml
guangli-dai: Thanks, made a typo in the script, just fixed it.

scripts/gen_travis.py
guangli-dai: >  add only --with-lg-hugepage=29 here, so combination --with-lg-page=16 --with-lg-hugepage=29 will be generated as well

For arm, not really. The number of unusual setting combined together is controlled by `max_unusual_opts` [here](https://github.com/jemalloc/jemalloc/blob/dev/scripts/gen_travis.py#L243) and it is set to 1 for arm (because we don't want that many arm tests for now). Still this raises an interesting question, should we just test with 512MB hugepage or should we couple it with 64KB pages. While in jemalloc, the two should be independent and that's why we have 64KB page tests alone before, I don't have a preference here. Please suggest.

Also FYI, I intend to put the travis change configuration in another PR because I think it may need some more discussions like this. Let's redirect the discussion [there](https://github.com/jemalloc/jemalloc/pull/2771). I am adding that commit in this PR mainly for testing purposes so that we know the test is fixed properly.

ilvokhin (approved):

PR #2771, opened 2024-12-13
Enable large hugepage tests for arm64 on Travis

guangli-dai:
Add large hugepage size (512MB) for arm64 tests on Travis.


guangli-dai:
Following the discussion [here](https://github.com/jemalloc/jemalloc/pull/2770#discussion_r1887214985) with @ilvokhin , main questions to answer are:
1. do we want to couple 64KB pages and 512MB pages into one test?
2. following the first question, maybe we should add more tests for arm64 on travis?

guangli-dai:
The new travis test results in [no response](https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/629092318), this might be a real failure. Will do more investigation.

PR #2773, opened 2024-12-18
Added prof_threshold allocation callback

metaduv:
Unlike `prof_sample` which is supported only with profiling mode active, `prof_threshold` is intended to be an always-supported allocation callback with much less overhead. The usage of the threshold allows performance critical callers to change program execution based on the callback: e.g. drop caches when memory becomes high or to predict the program is about to OOM ahead of time using peak memory watermarks.

### DIFF
--- a/include/jemalloc/internal/prof_externs.h
+++ b/include/jemalloc/internal/prof_externs.h
@@ -11,6 +11,7 @@ extern bool opt_prof_active;
 extern bool opt_prof_thread_active_init;
 extern unsigned opt_prof_bt_max;
 extern size_t opt_lg_prof_sample;    /* Mean bytes between samples. */
+extern size_t opt_experimental_lg_prof_threshold;    /* Mean bytes between thresholds. */
 extern ssize_t opt_lg_prof_interval; /* lg(prof_interval). */
 extern bool opt_prof_gdump;          /* High-water memory dumping. */
 extern bool opt_prof_final;          /* Final profile dumping. */
@@ -66,6 +67,9 @@ prof_sample_hook_t prof_sample_hook_get(void);
 
 void prof_sample_free_hook_set(prof_sample_free_hook_t hook);
 prof_sample_free_hook_t prof_sample_free_hook_get(void);
+
+void prof_threshold_hook_set(prof_threshold_hook_t hook);
+prof_threshold_hook_t prof_threshold_hook_get(void);
 
 /* Functions only accessed in prof_inlines.h */
 prof_tdata_t *prof_tdata_init(tsd_t *tsd);
--- a/include/jemalloc/internal/prof_threshold.h
+++ b/include/jemalloc/internal/prof_threshold.h
@@ -1,0 +1,11 @@
+#ifndef JEMALLOC_INTERNAL_THRESHOLD_EVENT_H
+#define JEMALLOC_INTERNAL_THRESHOLD_EVENT_H
+
+#include "jemalloc/internal/tsd_types.h"
+
+/* The activity-triggered hooks. */
+uint64_t prof_threshold_new_event_wait(tsd_t *tsd);
+uint64_t prof_threshold_postponed_event_wait(tsd_t *tsd);
+void prof_threshold_event_handler(tsd_t *tsd, uint64_t elapsed);
+
+#endif /* JEMALLOC_INTERNAL_THRESHOLD_EVENT_H */
--- a/include/jemalloc/internal/thread_event.h
+++ b/include/jemalloc/internal/thread_event.h
@@ -56,6 +56,7 @@ void tsd_te_init(tsd_t *tsd);
 #define ITERATE_OVER_ALL_EVENTS						\
     E(tcache_gc,		(opt_tcache_gc_incr_bytes > 0), true)	\
     E(prof_sample,		(config_prof && opt_prof), true)  	\
+    E(prof_threshold,		config_stats, true)  			\
     E(stats_interval,		(opt_stats_interval >= 0), true)   	\
     E(tcache_gc_dalloc,		(opt_tcache_gc_incr_bytes > 0), false)	\
     E(peak_alloc,		config_stats, true)			\
--- a/msvc/projects/vc2015/jemalloc/jemalloc.vcxproj
+++ b/msvc/projects/vc2015/jemalloc/jemalloc.vcxproj
@@ -83,6 +83,7 @@
     <ClCompile Include="..\..\..\..\src\prof_recent.c" />
     <ClCompile Include="..\..\..\..\src\prof_stats.c" />
     <ClCompile Include="..\..\..\..\src\prof_sys.c" />
+    <ClCompile Include="..\..\..\..\src\prof_threshold.c" />
     <ClCompile Include="..\..\..\..\src\psset.c" />
     <ClCompile Include="..\..\..\..\src\rtree.c" />
     <ClCompile Include="..\..\..\..\src\safety_check.c" />
--- a/msvc/projects/vc2017/jemalloc/jemalloc.vcxproj.filters
+++ b/msvc/projects/vc2017/jemalloc/jemalloc.vcxproj.filters
@@ -133,6 +133,9 @@
     <ClCompile Include="..\..\..\..\src\prof_sys.c">
       <Filter>Source Files</Filter>
     </ClCompile>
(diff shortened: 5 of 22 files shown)


msvc/projects/vc2015/jemalloc/jemalloc.vcxproj
```diff
@@ -83,6 +83,7 @@
     <ClCompile Include="..\..\..\..\src\prof_recent.c" />
     <ClCompile Include="..\..\..\..\src\prof_stats.c" />
     <ClCompile Include="..\..\..\..\src\prof_sys.c" />
+    <ClCompile Include="..\..\..\..\src\prof_threshold.c" />
```
ilvokhin, line 86: Not related to this change, but I'll ask anyway: seems like we are not building jemalloc with MSVC in CI. At least I could find a job for it. I think there is a high chance this file is out of date for some time and not working anymore.

@interwq, do you think it would be reasonable to remove MSVC project files entirely?

include/jemalloc/internal/thread_event.h
```diff
@@ -56,6 +56,7 @@ void tsd_te_init(tsd_t *tsd);
 #define ITERATE_OVER_ALL_EVENTS						\
     E(tcache_gc,		(opt_tcache_gc_incr_bytes > 0), true)	\
     E(prof_sample,		(config_prof && opt_prof), true)  	\
+    E(prof_threshold,		config_stats, true)  	\
```
ilvokhin: Trailing slash in under indented here?

test/unit/prof_threshold.c
```diff
@@ -0,0 +1,103 @@
...
+static void reset_test_config() {
+	hook_calls = 0;
+	last_peak = 0;
+	alloc_baseline = last_alloc; // We run the test multiple times
+	last_alloc = 0;
+	threshold_bytes = 1 << opt_experimental_lg_prof_threshold;
+	chunk_size = threshold_bytes / ALLOC_ITERATIONS_IN_THRESHOLD;
+}
+
+static void expectThresholdCalls(int calls) {
```
ilvokhin:
```suggestion
static void expect_threshold_calls(int calls) {
```

test/unit/prof_threshold.c
```diff
@@ -0,0 +1,103 @@
...
+	expect_zu_eq(hook_calls, calls, "Hook called the right amount of times");
+	expect_true(last_peak < chunk_size * 2, "We allocate chunk_size at a time");
+	expect_true(last_alloc >= threshold_bytes * calls + alloc_baseline, "Crosses");
```
ilvokhin: For a clearer error message purposes.

```suggestion
	expect_u64_lt(last_peak, chunk_size * 2, "We allocate chunk_size at a time");
	expect_u64_ge(last_alloc, threshold_bytes * calls + alloc_baseline, "Crosses");
```

test/unit/prof_threshold.c
```diff
@@ -0,0 +1,103 @@
...
+}
+
+static void allocateChunks(int chunks) {
```
ilvokhin:
```suggestion
static void allocate_chunks(int chunks) {
```

test/unit/prof_threshold.c
```diff
@@ -0,0 +1,103 @@
+#include "test/jemalloc_test.h"
+
+// Test config (set in reset_test_config)
```
ilvokhin: Here and below in couple of places `//` -> `/**/`, for consistency with other files.

```suggestion
/* Test config (set in reset_test_config) */
```

test/unit/prof_threshold_small.c
```diff
@@ -0,0 +1,4 @@
+#include "test/jemalloc_test.h"
+#define PROF_THRESHOLD_SMALL_C_
```
ilvokhin: I couldn't find reference to `PROF_THRESHOLD_SMALL_C_` anywhere. Maybe something is missing in `test/unit/prof_threshold.c` file to change behaviour with this macro?

test/unit/prof_threshold.c
```diff
@@ -0,0 +1,103 @@
...
+static void allocateChunks(int chunks) {
+	for (int i = 0; i < chunks; i++) {
+		void* p = mallocx(chunk_size, 0);
+		expect_ptr_not_null(p, "Failed to allocate");
+		free(p);
+	}
+}
+
+TEST_BEGIN(test_prof_threshold_hook) {
+	// Test setting and reading the hook (both value and null)
```
ilvokhin: Does it make sense to move read and write logic and check for it to it is own test to have single property we are trying to test in each test?

test/unit/prof_threshold.c
```diff
@@ -0,0 +1,103 @@
+#include "test/jemalloc_test.h"
+
+// Test config (set in reset_test_config)
+const int ALLOC_ITERATIONS_IN_THRESHOLD = 10;
```
ilvokhin: nit: Just to make it a bit more C-style.

```suggestion
#define ALLOC_ITERATIONS_IN_THRESHOLD 10;
```

test/unit/prof_threshold.c
metaduv: Thanks for the suggestion, I think it does. I was going by the similar prof_hook/prof_sample where we ended up putting them in the same file. https://github.com/jemalloc/jemalloc/blob/dev/test/unit/prof_hook.c#L291

msvc/projects/vc2015/jemalloc/jemalloc.vcxproj
interwq: > do you think it would be reasonable to remove MSVC project files entirely?

Indeed we are only using the msvc compiler (cl.exe) in the CI. I'd say to keep the project files still, since over the years there were people helping to update these files, which should indicate they do care about importing the project that way.

@metaduv Related to this, we should also update the msvc 2019 & 2022 projects similarly: https://github.com/jemalloc/jemalloc/tree/dev/msvc/projects

include/jemalloc/internal/prof_externs.h
```diff
@@ -11,6 +11,7 @@ extern bool opt_prof_active;
 extern bool opt_prof_thread_active_init;
 extern unsigned opt_prof_bt_max;
 extern size_t opt_lg_prof_sample;    /* Mean bytes between samples. */
+extern size_t opt_experimental_lg_prof_threshold;    /* Mean bytes between thresholds. */
```
interwq, line 14: Do we want to restrict it to log scale? Maybe allowing a plain size would be fine? e.g. if a binary wants a 10G threshold

include/jemalloc/internal/prof_threshold.h
```diff
@@ -0,0 +1,13 @@
+#ifndef JEMALLOC_INTERNAL_THRESHOLD_EVENT_H
+#define JEMALLOC_INTERNAL_THRESHOLD_EVENT_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/tsd_types.h"
+
+/* Manual hooks. */
```
interwq: Looks not needed anymore?

interwq (commented):
BTW it looks the commit msg was lost probably during a merge.

include/jemalloc/internal/prof_externs.h
metaduv: My goal was to keep parity with prof_sample for ease of switching back and forth / consistency. In both cases the hook gets called after 2^N bytes were allocated.

msvc/projects/vc2017/jemalloc/jemalloc.vcxproj.filters
```diff
@@ -133,6 +133,9 @@
     <ClCompile Include="..\..\..\..\src\prof_sys.c">
       <Filter>Source Files</Filter>
     </ClCompile>
+    <ClCompile Include="..\..\..\..\src\prof_threshold.c">
+      <Filter>Source Files</Filter>
+    </ClCompile>
```
interwq, line 138: Similar to here, two more files also need updating:
msvc/projects/vc2019/jemalloc/jemalloc.vcxproj.filters
msvc/projects/vc2022/jemalloc/jemalloc.vcxproj.filters

interwq (commented):
One more thing: we should also merge the two commits together. Try `git rebase -i dev`.

PR #2776, opened 2025-01-08
Add profiling stats for backtrace unwind

nslingerland:
Add malloc stats for profiling / unwinding.

Example of newly added output for a small test program that does allocations with MALLOC_CONF set as:

```
$ MALLOC_CONF=prof:true,lg_prof_sample:15,stats_print:true,stats_print_opts:mda
$ ./tool/run_stats.sh 
...
___ Begin jemalloc statistics ___
...
Profiling stats:
  Backtrace Count: 3929605 (654934 / sec)
  Backtrace Time (ns): 2015039350 total (512 avg)
...
```

```
$ MALLOC_CONF=prof:true,lg_prof_sample:15,stats_print:true,stats_print_opts:mdaJ
$ ./tool/run_stats.sh 
...
{
  "jemalloc": {
...
    "prof": {
      "thread_active_init": true,
      "active": true,
      "gdump": false,
      "interval": 0,
      "lg_sample": 15,
      "backtrace": {
        "backtrace_count": 3865607,
        "backtrace_total_time_ns": 2013999702,
        "backtrace_avg_time_ns": 521
      }
    },
...
```

### DIFF
--- a/include/jemalloc/internal/prof_stats.h
+++ b/include/jemalloc/internal/prof_stats.h
@@ -17,4 +17,11 @@ void prof_stats_dec(tsd_t *tsd, szind_t ind, size_t size);
 void prof_stats_get_live(tsd_t *tsd, szind_t ind, prof_stats_t *stats);
 void prof_stats_get_accum(tsd_t *tsd, szind_t ind, prof_stats_t *stats);
 
+extern atomic_u64_t prof_backtrace_count;
+extern atomic_u64_t prof_backtrace_time_ns;
+#ifdef JEMALLOC_PROF_FRAME_POINTER
+extern atomic_u64_t prof_stack_range_count;
+extern atomic_u64_t prof_stack_range_time_ns;
+#endif
+
 #endif /* JEMALLOC_INTERNAL_PROF_STATS_H */
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -351,6 +351,12 @@ CTL_PROTO(stats_resident)
 CTL_PROTO(stats_mapped)
 CTL_PROTO(stats_retained)
 CTL_PROTO(stats_zero_reallocs)
+CTL_PROTO(stats_prof_backtrace_count)
+CTL_PROTO(stats_prof_backtrace_time_ns)
+#ifdef JEMALLOC_PROF_FRAME_POINTER
+CTL_PROTO(stats_prof_stack_range_count)
+CTL_PROTO(stats_prof_stack_range_time_ns)
+#endif
 CTL_PROTO(experimental_hooks_install)
 CTL_PROTO(experimental_hooks_remove)
 CTL_PROTO(experimental_hooks_prof_backtrace)
@@ -956,6 +962,12 @@ static const ctl_named_node_t stats_node[] = {
 	{NAME("mutexes"),	CHILD(named, stats_mutexes)},
 	{NAME("arenas"),	CHILD(indexed, stats_arenas)},
 	{NAME("zero_reallocs"),	CTL(stats_zero_reallocs)},
+	{NAME("prof_backtrace_count"), CTL(stats_prof_backtrace_count)},
+	{NAME("prof_backtrace_time_ns"), CTL(stats_prof_backtrace_time_ns)},
+#ifdef JEMALLOC_PROF_FRAME_POINTER
+	{NAME("prof_stack_range_count"), CTL(stats_prof_stack_range_count)},
+	{NAME("prof_stack_range_time_ns"), CTL(stats_prof_stack_range_time_ns)},
+#endif
 };
 
 static const ctl_named_node_t experimental_hooks_node[] = {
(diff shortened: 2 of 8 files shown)


src/prof_sys.c
```diff
@@ -23,6 +23,28 @@
...
+
+#ifdef JEMALLOC_PROF_FRAME_POINTER
+atomic_zu_t prof_fp_stack_range_count = ATOMIC_INIT(0);
+atomic_u64_t prof_fp_stack_range_time_ns = ATOMIC_INIT(0);
+#endif
+
+static inline uint64_t timestamp_ns(void) {
+#if defined (__linux__)
+  // On Linux CLOCK_MONOTONIC/CLOCK_MONOTONIC_RAW are higher precision than
+  // CLOCK_MONOTONIC_COARSE (nstime default) or CLOCK_REALTIME (prof_time_res_high).
```
ilvokhin: Could you please use `/**/`-style comments here, instead of `//` for consistency with surrounding code?

ilvokhin: Could you please expand a comment providing a explanation **why** we need a higher precision clock here?

Side note: on my machine `CLOCK_REALTIME` and `CLOCK_MONOTONIC` have same resolution (precision) of 1 ns, measured by `clock_getres` calls. Is there any chance we can get away using `CLOCK_REALTIME`? `timestamp_ns` calls made code a bit cumbersome, because we can not use `nstime_t` primitives.

src/prof_sys.c
```diff
@@ -23,6 +23,28 @@
...
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC_RAW, &ts);
+  return ts.tv_sec * 1000000000ULL + ts.tv_nsec;
+#else
+  nstime_t ts;
+  nstime_init_update(&ts);
+  return ts.ns;
```
ilvokhin: `ns` is an implementation detail, I think we can call `nstime_ns` here instead.

src/ctl.c
```diff
@@ -3822,6 +3834,20 @@ CTL_RO_CGEN(config_stats, stats_background_thread_run_interval,
...
+CTL_RO_CGEN(config_stats, stats_prof_backtrace_time_ns,
+    atomic_load_u64(&prof_backtrace_time_ns, ATOMIC_RELAXED), uint64_t)
+
+#ifdef JEMALLOC_PROF_FRAME_POINTER
+CTL_RO_CGEN(config_stats, stats_prof_fp_stack_range_count,
+    atomic_load_zu(&prof_fp_stack_range_count, ATOMIC_RELAXED), size_t)
+
+CTL_RO_CGEN(config_stats, stats_prof_fp_stack_range_time_ns,
+    atomic_load_u64(&prof_fp_stack_range_time_ns, ATOMIC_RELAXED), uint64_t)
+#endif
```
ilvokhin: Is there any chance you could add a unit-test for this counters to `test/unit/mallctl.c` to prevent regression in the future?

src/prof_sys.c
```diff
@@ -458,13 +489,25 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 
 void
 prof_backtrace(tsd_t *tsd, prof_bt_t *bt) {
+  uint64_t ts;
+  if (config_stats) {
+    ts = timestamp_ns();
+  }
+
```
ilvokhin: Something wrong with code formatting here, seems like indent is 2 spaces instead of tabs.

src/prof_sys.c
```diff
@@ -458,13 +489,25 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
...
 	pre_reentrancy(tsd, NULL);
 	prof_backtrace_hook(bt->vec, &bt->len, opt_prof_bt_max);
 	post_reentrancy(tsd);
+
+  if (config_stats) {
+    uint64_t delta = timestamp_ns() - ts;
+    atomic_fetch_add_u64(&prof_backtrace_time_ns, delta, ATOMIC_RELAXED);
+
+    atomic_fetch_add_zu(&prof_backtrace_count, 1, ATOMIC_RELAXED);
+  }
```
ilvokhin: Same formatting problem here.

src/stats.c
```diff
@@ -1763,6 +1763,48 @@ stats_general_print(emitter_t *emitter) {
...
 		    emitter_type_ssize, &ssv);
 
+		size_t backtrace_count = atomic_load_zu(&prof_backtrace_count, ATOMIC_RELAXED);
+		uint64_t backtrace_time_ns = atomic_load_u64(&prof_backtrace_time_ns, ATOMIC_RELAXED);
+		uint64_t backtrace_avg_time_ns = backtrace_count > 0 ? backtrace_time_ns / backtrace_count : 0;
+#ifdef JEMALLOC_PROF_FRAME_POINTER
+		size_t fp_stack_range_count = atomic_load_zu(&prof_fp_stack_range_count, ATOMIC_RELAXED);
+		uint64_t fp_stack_range_time_ns = atomic_load_u64(&prof_fp_stack_range_time_ns, ATOMIC_RELAXED);
+		uint64_t fp_avg_stack_range_time_ns = fp_stack_range_count > 0 ? fp_stack_range_time_ns / fp_stack_range_count : 0;
+#endif
```
ilvokhin: We are getting this values directly from variables here, seems like `CTL_GET` (getting values from `mallctl`) are used in the surrounding code.

src/prof_sys.c
```diff
@@ -117,7 +139,16 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
 
   static __thread void **stack_start = (void **)0;  // thread local
   if (stack_start == 0 || stack_end >= stack_start) {
+    uint64_t ts;
+    if (config_stats) {
+      atomic_fetch_add_zu(&prof_fp_stack_range_count, 1, ATOMIC_RELAXED);
+      ts = timestamp_ns();
```
ilvokhin: Is there any chance we can add unit tests for this counters?

include/jemalloc/internal/prof_stats.h
```diff
@@ -17,4 +17,11 @@ void prof_stats_dec(tsd_t *tsd, szind_t ind, size_t size);
 void prof_stats_get_live(tsd_t *tsd, szind_t ind, prof_stats_t *stats);
 void prof_stats_get_accum(tsd_t *tsd, szind_t ind, prof_stats_t *stats);
 
+extern atomic_zu_t prof_backtrace_count;
```
ilvokhin: nit: this counter is monotonically increases over time and probably quite fast. Probably, it make sense to make it as `atomic_u64_t` to be on the safe side on 32 bit platforms.

ilvokhin (changes requested):
Seems there are build failures on Windows in CI.

src/prof_sys.c
nslingerland: I was worried that `nstime` uses `CLOCK_MONOTONIC_COARSE` by default - which has 1ms resolution on my machine. Unwinding times are going to be much shorter than that - and since we're trying to accumulate the total time spent doing callstack unwinding it wouldn't produce useful results with that default.

That being said it seems our jemalloc is configured to use `prof_time_resolution:high` - and that's sufficient.

PR #2779, opened 2025-01-13
Add autoconf options to enable sanitizers

ilvokhin:
This commit allows to enable sanitizers with autoconf options, instead of modifying `CFLAGS`, `CXXFLAGS` and `LDFLAGS` directly.

* `--enable-tsan` option to enable Thread Sanitizer.
* `--enable-ubsan` option to enable Undefined Behaviour Sanitizer.
* And `--enable-asan` to enable Address Sanitizer.

End goal is to speedup development by finding problems quickly, early and easier. Eventually, when all current issues will be fixed, we can enable sanitizers in CI. Fortunately, there are not a lot of problems we need to fix.

Address Sanitizer is a bit controversial, because it replaces memory allocator, so we decided to left it out for a while.

Below are couple of examples of how tests look like under different sanitizers at the moment.

```
$  ../configure --enable-tsan --enable-debug
<...>
asan               : 0
tsan               : 1
ubsan              : 0
$ make -j`nproc` check
<...>
  Thread T13 (tid=332043, running) created by main thread at:
    #0 pthread_create <null> (libtsan.so.0+0x61748)
    #1 thd_create ../test/src/thd.c:25 (bin_batching+0x5631ca)
    #2 stress_run ../test/unit/bin_batching.c:148
(bin_batching+0x40364c)
    #3 test_races ../test/unit/bin_batching.c:249
(bin_batching+0x403d79)
    #4 p_test_impl ../test/src/test.c:149 (bin_batching+0x562811)
    #5 p_test_no_reentrancy ../test/src/test.c:213
(bin_batching+0x562d35)
    #6 main ../test/unit/bin_batching.c:268 (bin_batching+0x40417e)

SUMMARY: ThreadSanitizer: data race
../include/jemalloc/internal/edata.h:498 in edata_nfree_inc
```

```
$ ../configure --enable-ubsan --enable-debug
<...>
asan               : 0
tsan               : 0
ubsan              : 1
$ make -j`nproc` check
<...>
=== test/unit/hash ===
../test/unit/hash.c:119:16: runtime error: left shift of 176 by 24
places cannot be represented in type 'int'
<...>
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -92,6 +92,32 @@ AC_LANG_POP([C++])
 JE_CONCAT_VVV(CXXFLAGS, CONFIGURE_CXXFLAGS, SPECIFIED_CXXFLAGS)
 ])
 
+CONFIGURE_LDFLAGS=
+SPECIFIED_LDFLAGS="${LDFLAGS}"
+dnl JE_LDFLAGS_ADD(ldflag)
+dnl
+dnl LDFLAGS is the concatenation of CONFIGURE_LDFLAGS and SPECIFIED_LDFLAGS
+dnl This macro appends to CONFIGURE_LDFLAGS and regenerates LDFLAGS.
+AC_DEFUN([JE_LDFLAGS_ADD],
+[
+AC_MSG_CHECKING([whether linker supports $1])
+T_CONFIGURE_LDFLAGS="${CONFIGURE_LDFLAGS}"
+JE_APPEND_VS(CONFIGURE_LDFLAGS, $1)
+JE_CONCAT_VVV(LDFLAGS, CONFIGURE_LDFLAGS, SPECIFIED_LDFLAGS)
+AC_LINK_IFELSE([AC_LANG_PROGRAM(
+[[
+]], [[
+    return 0;
+]])],
+              [je_cv_ldflags_added=$1]
+              AC_MSG_RESULT([yes]),
+              [je_cv_ldflags_added=]
+              AC_MSG_RESULT([no])
+              [CONFIGURE_LDFLAGS="${T_CONFIGURE_LDFLAGS}"]
+)
+JE_CONCAT_VVV(LDFLAGS, CONFIGURE_LDFLAGS, SPECIFIED_LDFLAGS)
+])
+
 dnl JE_COMPILABLE(label, hcode, mcode, rvar)
 dnl
 dnl Use AC_LINK_IFELSE() rather than AC_COMPILE_IFELSE() so that linker errors
@@ -2647,6 +2673,40 @@ if test "x$enable_pageid" = "x1" ; then
   AC_DEFINE([JEMALLOC_PAGEID], [ ], [ ])
 fi
 
+AC_ARG_ENABLE([tsan],
+  [AS_HELP_STRING([--enable-tsan],
+                  [Enable thread sanitizer])],
+[if test "x$enable_tsan" = "xno" ; then
+  enable_tsan="0"
+else
+  enable_tsan="1"
+fi
+],
+[enable_tsan="0"]
+)
+if test "x$enable_tsan" = "x1" ; then
+  JE_CFLAGS_ADD([-fsanitize=thread])
+  JE_CXXFLAGS_ADD([-fsanitize=thread])
+  JE_LDFLAGS_ADD([-fsanitize=thread])
+fi
+
+AC_ARG_ENABLE([ubsan],
+  [AS_HELP_STRING([--enable-ubsan],
+                  [Enable undefined behavior sanitizer])],
+[if test "x$enable_ubsan" = "xno" ; then
+  enable_ubsan="0"
+else
+  enable_ubsan="1"
+fi
+],
+[enable_ubsan="0"]
+)
+if test "x$enable_ubsan" = "x1" ; then
+  JE_CFLAGS_ADD([-fsanitize=undefined])
+  JE_CXXFLAGS_ADD([-fsanitize=undefined])
+  JE_LDFLAGS_ADD([-fsanitize=undefined])
+fi
+
 dnl ============================================================================
 dnl Enable background threads if possible.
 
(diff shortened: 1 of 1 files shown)


guangli-dai:
This is definitely a good feature jemalloc developers can benefit from! I wonder how it may affect jemalloc users if they wrongly enable it when compiling jemalloc from source?

Also, maybe we can do some more to prevent abuse of the newly added flags? For instance, can we rename them with a prefix `dev-only`, i.e., `--enable-dev-only-tsan`, to prevent users from thinking that this is the right way to enable sanitizers on their application?

ilvokhin:
> I wonder how it may affect jemalloc users if they wrongly enable it when compiling jemalloc from source?

I think if you are going to enable a flag manually, you better know what you are doing. It is true for sanitizers flags, as well as any flags in general, also true for config options and other tunables we already have. Consequences will vary. One thing that worries me a bit is address sanitizer. It's interaction with jemalloc might be surprising for someone.

> prevent users from thinking that this is the right way to enable sanitizers on their application?

I am not sure slightly different name will prevent this to be honest, but I agree it might worth to have a better documentation for it.

ilvokhin:
Extended autoconf help string for better documentation.

guangli-dai:
> I am not sure slightly different name will prevent this to be honest

Adding a prefix to warn people is a direct way and used a lot in jemalloc like [here](https://github.com/jemalloc/jemalloc/blob/52fa9577ba8fa94f41c8c92f845a74c3fb04db80/include/jemalloc/internal/tcache_structs.h#L43). If it is for dev only instead of for users, I think we should add such a prefix in.

As I can see in the modified documentation, it says "Recommended for development use only". That seems to imply users can still use them for debugging purposes? 

> One thing that worries me a bit is address sanitizer. It's interaction with jemalloc might be surprising for someone.

For ASAN, I think even if users want to use it, enable it through jemalloc is not the right route. Maybe we can check the flag and prevent `make install` if ASAN is enabled to prevent confusion?

ilvokhin:
Could you please elaborate a bit what problem we are trying to solve by adding prefixes to configure flags? Also, what are definitions of users and developers? Maybe I misunderstood problem. Just want to make sure we are on the same page here.

Below are mine thoughts about it.

1. Flags are meant to be turned on. That's why we are having them in the first place. If we believe some flags do more harm, than good, then we should not have such flags. So, attempt to prevent usage of flags in any way is a bit odd to me to begin with.
2. Everyone who builds code and hence provides flags to configure script is a developer in my book. Maybe, not a core jemalloc developer, but developer nevertheless. As a consequences of that, they will be able to turn on any flag and even modify code in any way they would like. We should do our best to explain what this flag is supposed to do, but the rest is out of our hands. There are indefinite possibilities of misunderstanding, we can not protect against all of them, we should assume basic competence and tackle problems from this angle.

Given that, what I am trying to understand is why three letter prefix of a flag expected to do a better job in explaining what this flag is for, than sentence in plain English in help message? If sentence is doing a poor job and doesn't achieve the goal, let's fix it.

Last, but not least, why I am pushing back is because `--enable-asan` and friends seems like a [popular][1] (if not conventional) name for autoconf flag for this purpose, so I am hesitant to introduce a creative name for that reason.

[1]: https://github.com/search?q=--enable-asan&type=code

guangli-dai:
> What are definitions of users and **developers?**

From my understanding, there are lots of users compiling jemalloc from source to use jemalloc on a variety of machines they use. The users normally don't need to dig deep into jemalloc source codes unless they have to due to a blocker. For such users, I think we should try to lower the risk of any misuse for them. 

> Could you please elaborate a bit what problem we are trying to solve by adding prefixes to configure flags? 

Sure, to clarify, I am trying to discuss if users (as defined above) should use `--enable-asan` to bypass jemalloc. It seems to me that adding a config flag in the library to stop using the library logically doesn't sound right. If users want to enable ASAN, they should specify that in the higher level of the build system to avoid future confusion.  So the first question I want to ask here is do you think this flag should be used in cases where jemalloc is compiled and installed to serve other applications or it should be limited to running jemalloc's own tests like `make check`?

If we agree that users should not use the flag to bypass jemalloc, I think adding strong warnings or changing the flag name can all do the job. If the answer to the first question is yes, we can add that explanation in. Changing the flag name is more like adding one more fence at the front door.  I suggested this different naming mainly because other projects use it to use alternatives of a subcomponent but here we are using this flag to bypass ourselves. Of course if we think the answer is NO to the first question, I think the current warnings are already doing a good job.

ilvokhin:
Thanks for clarifications.

If I understand correctly most concerns are centered around Address Sanitizer, because it is hard to reason about it's interaction with memory allocator. I agree, that's a tricky thing indeed and after thinking about it for a while I am not sure what strategy should we use as well.

If Address Sanitizer is the only concern, what do you think if we'll just drop it for now?

We'll introduce flags for thread and undefined behaviour sanitizers, which are still useful and are much easier to reason about in combination with memory allocator. Meanwhile, I'll try to think if we can do something about Address Sanitizer.

guangli-dai:
> If Address Sanitizer is the only concern, what do you think if we'll just drop it for now?
> 
> We'll introduce flags for thread and undefined behaviour sanitizers, which are still useful and are much easier to reason about in combination with memory allocator. Meanwhile, I'll try to think if we can do something about Address Sanitizer.

That sounds good. We can discuss more offline.

ilvokhin:
Removed flag for Address Sanitizer.

guangli-dai (approved):

configure.ac
```diff
@@ -2869,7 +2929,8 @@ AC_MSG_RESULT([CXX                : ${CXX}])
 AC_MSG_RESULT([CONFIGURE_CXXFLAGS : ${CONFIGURE_CXXFLAGS}])
 AC_MSG_RESULT([SPECIFIED_CXXFLAGS : ${SPECIFIED_CXXFLAGS}])
 AC_MSG_RESULT([EXTRA_CXXFLAGS     : ${EXTRA_CXXFLAGS}])
-AC_MSG_RESULT([LDFLAGS            : ${LDFLAGS}])
```
interwq, line 2872: To double check, this shouldn't affect existing use cases, since LDFLAGS is included in SPECIFIED_LDFLAGS, is that correct?

If so, maybe still include the final FDFLAGS here?

ilvokhin: > To double check, this shouldn't affect existing use cases, since LDFLAGS is included in SPECIFIED_LDFLAGS, is that correct?

Yes, it shouldn't affect existing use cases. `LDFLAGS` are the ones that are used in the end. `JE_LDFLAGS_ADD` is implemented the same way as existing `JE_CFLAGS_ADD` and `JE_CXXFLAGS_ADD` functions with `AC_COMPILE_IFELSE` to `AC_LINK_IFELSE` replacement.

I removed `LDFLAGS` for consistency with `CFLAGS` and `CXXFLAGS`. If we are returning back `LDFLAGS` , should we also print `CFLAGS`  and `CXXFLAGS` there?

interwq: Thanks for explaining @ilvokhin. Everything looks good. Merging.

PR #2780, opened 2025-01-14
Extend purging algorithm with peak demand tracking

ilvokhin:
Implementation inspired by idea described in "Beyond malloc efficiency to fleet efficiency: a hugepage-aware memory allocator" [paper][1].

Primary idea is to track maximum number (peak) of active pages in use with sliding window and then use this number to decide how many dirty pages we would like to keep.

We are trying to estimate maximum amount of active memory we'll need in the near future. We do so by projecting future active memory demand(based on peak active memory usage we observed in the past within sliding window) and adding slack on top of it (an overhead is reasonable to have in exchange of higher hugepages coverage). When peak demand tracking is off, projection of future active memory is active memory we are having right now.

Estimation is essentially the same as `nactive_max * (1 + dirty_mult)`.

Peak demand purging algorithm controlled by two config options. Option `hpa_peak_demand_window_ms` controls duration of sliding window we track maximum active memory usage in and option `hpa_dirty_mult` controls
amount of slack we are allowed to have as a percent from maximum active memory usage. By default `hpa_peak_demand_window_ms == 0` now and we have same behaviour (ratio based purging) that we had before this commit.

[1]: https://storage.googleapis.com/gweb-research2023-media/pubtools/6170.pdf

### DIFF
--- a/include/jemalloc/internal/hpa_opts.h
+++ b/include/jemalloc/internal/hpa_opts.h
@@ -27,7 +27,8 @@ struct hpa_shard_opts_s {
 
 	/*
 	 * The HPA purges whenever the number of pages exceeds dirty_mult *
-	 * active_pages.  This may be set to (fxp_t)-1 to disable purging.
+	 * peak_active_pages.  This may be set to (fxp_t)-1 to disable
+	 * purging.
 	 */
 	fxp_t dirty_mult;
 
@@ -59,6 +60,13 @@ struct hpa_shard_opts_s {
 	 * Maximum number of hugepages to purge on each purging attempt.
 	 */
 	ssize_t experimental_max_purge_nhp;
+
+	/*
+	 * Sliding window duration to track active memory demand statistics.
+	 * This might be set to 0, to disable sliding window statistics
+	 * tracking and use current number of active pages for purging instead.
+	 */
+	uint64_t peak_demand_window_ms;
 };
 
 #define HPA_SHARD_OPTS_DEFAULT {					\
@@ -83,7 +91,9 @@ struct hpa_shard_opts_s {
 	/* min_purge_interval_ms */					\
 	5 * 1000,							\
 	/* experimental_max_purge_nhp */				\
-	-1								\
+	-1,								\
+	/* peak_demand_window_ms */					\
+	0								\
 }
 
 #endif /* JEMALLOC_INTERNAL_HPA_OPTS_H */
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -106,6 +106,7 @@ CTL_PROTO(opt_hpa_hugify_delay_ms)
 CTL_PROTO(opt_hpa_hugify_sync)
 CTL_PROTO(opt_hpa_min_purge_interval_ms)
 CTL_PROTO(opt_experimental_hpa_max_purge_nhp)
+CTL_PROTO(opt_hpa_peak_demand_window_ms)
 CTL_PROTO(opt_hpa_dirty_mult)
 CTL_PROTO(opt_hpa_sec_nshards)
 CTL_PROTO(opt_hpa_sec_max_alloc)
@@ -485,6 +486,8 @@ static const ctl_named_node_t opt_node[] = {
 	{NAME("hpa_min_purge_interval_ms"), CTL(opt_hpa_min_purge_interval_ms)},
 	{NAME("experimental_hpa_max_purge_nhp"),
 		CTL(opt_experimental_hpa_max_purge_nhp)},
+	{NAME("hpa_peak_demand_window_ms"),
+	    CTL(opt_hpa_peak_demand_window_ms)},
 	{NAME("hpa_dirty_mult"), CTL(opt_hpa_dirty_mult)},
 	{NAME("hpa_sec_nshards"),	CTL(opt_hpa_sec_nshards)},
 	{NAME("hpa_sec_max_alloc"),	CTL(opt_hpa_sec_max_alloc)},
@@ -2251,6 +2254,8 @@ CTL_RO_NL_GEN(opt_hpa_min_purge_interval_ms, opt_hpa_opts.min_purge_interval_ms,
     uint64_t)
 CTL_RO_NL_GEN(opt_experimental_hpa_max_purge_nhp,
     opt_hpa_opts.experimental_max_purge_nhp, ssize_t)
+CTL_RO_NL_GEN(opt_hpa_peak_demand_window_ms,
+    opt_hpa_opts.peak_demand_window_ms, uint64_t)
 
 /*
  * This will have to change before we publicly document this option; fxp_t and
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -63,6 +63,11 @@ hpa_supported(void) {
 	return true;
 }
 
+static bool
+hpa_peak_demand_tracking_enabled(hpa_shard_t *shard) {
+	return shard->opts.peak_demand_window_ms > 0;
+}
+
 static void
 hpa_do_consistency_checks(hpa_shard_t *shard) {
 	assert(shard->base != NULL);
@@ -217,6 +222,11 @@ hpa_shard_init(hpa_shard_t *shard, hpa_central_t *central, emap_t *emap,
 	shard->stats.nhugify_failures = 0;
 	shard->stats.ndehugifies = 0;
 
+	if (hpa_peak_demand_tracking_enabled(shard)) {
+		peak_demand_init(&shard->peak_demand,
+		    shard->opts.peak_demand_window_ms);
+	}
+
 	/*
 	 * Fill these in last, so that if an hpa_shard gets used despite
 	 * initialization failing, we'll at least crash instead of just
@@ -294,8 +304,37 @@ hpa_ndirty_max(tsdn_t *tsdn, hpa_shard_t *shard) {
 	if (shard->opts.dirty_mult == (fxp_t)-1) {
 		return (size_t)-1;
 	}
-	return fxp_mul_frac(psset_nactive(&shard->psset),
-	    shard->opts.dirty_mult);
+	/*
+	 * We are trying to estimate maximum amount of active memory we'll
+	 * need in the near future.  We do so by projecting future active
+	 * memory demand (based on peak active memory usage we observed in the
+	 * past within sliding window) and adding slack on top of it (an
+	 * overhead is reasonable to have in exchange of higher hugepages
+	 * coverage).  When peak demand tracking is off, projection of future
+	 * active memory is active memory we are having right now.
+	 *
+	 * Estimation is essentially the same as nactive_max * (1 +
+	 * dirty_mult), but expressed differently to factor in necessary
+	 * implementation details.
+	 */
+	size_t nactive = psset_nactive(&shard->psset);
+	size_t nactive_max = nactive;
+	if (hpa_peak_demand_tracking_enabled(shard)) {
+		/*
+		 * We release shard->mtx, when we do a syscall to purge dirty
+		 * memory, so someone might grab shard->mtx, allocate memory
+		 * from this shard and update psset's nactive counter, before
+		 * peak_demand_update(...) was called and we'll get
+		 * peak_demand_nactive_max(...) <= nactive as a result.
+		 */
+		size_t peak = peak_demand_nactive_max(&shard->peak_demand);
+		if (peak > nactive_max) {
+			nactive_max = peak;
+		}
+	}
+	size_t slack = fxp_mul_frac(nactive_max, shard->opts.dirty_mult);
+	size_t estimation = nactive_max + slack;
+	return estimation - nactive;
 }
 
 static bool
@@ -548,6 +587,16 @@ static void
 hpa_shard_maybe_do_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard,
     bool forced) {
 	malloc_mutex_assert_owner(tsdn, &shard->mtx);
+
+	/* Update active memory demand statistics. */
+	if (hpa_peak_demand_tracking_enabled(shard)) {
+		nstime_t now;
+		shard->central->hooks.curtime(&now,
+		    /* first_reading */ true);
+		peak_demand_update(&shard->peak_demand, &now,
+		    psset_nactive(&shard->psset));
+	}
+
 	if (!forced && shard->opts.deferral_allowed) {
 		return;
 	}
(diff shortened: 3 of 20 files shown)


include/jemalloc/internal/hpa_opts.h
```diff
@@ -59,6 +60,13 @@ struct hpa_shard_opts_s {
 	 * Maximum number of hugepages to purge on each purging attempt.
 	 */
 	ssize_t experimental_max_purge_nhp;
+
+	/*
+	 * Time interval to track active memory demand statistics.  This might
+	 * be set to 0, to disable sliding window statistics tracking and use
+	 * current number of active pages for purging instead.
+	 */
+	uint64_t peak_demand_interval_ms;
```
ot: nit: "interval" suggests a periodic behavior, can we call this "window", so it is more clear it is a sliding window?

src/hpa.c
```diff
@@ -295,8 +305,32 @@ hpa_ndirty_max(tsdn_t *tsdn, hpa_shard_t *shard) {
...
+		 */
+		size_t nactive_max =
+		    peak_demand_nactive_max(&shard->peak_demand);
+		assert(nactive_max >= nactive);
+		size_t demand = nactive_max - nactive;
+		size_t slack = fxp_mul_frac(nactive_max,
+		    shard->opts.dirty_mult);
+		return demand + slack;
+	}
+	return fxp_mul_frac(nactive, shard->opts.dirty_mult);
```
ot: I still think that this way of describing the algorithm doesn't make it self-evident that it is a proper extension of ratio-based purging. I would write it like this

```
size_t nactive_max;
if (hpa_peak_demand_tracking_enabled(shard)) {
  nactive_max = peak_demand_nactive_max(&shard->peak_demand);
  assert(nactive_max >= nactive);
} else {
  nactive_max = nactive;
}
return fxp_mul_frac(nactive_max, (1 + shard->opts.dirty_mult)) - nactive;
```

This makes it more obvious that we're still doing ratio-based, and just changing the baseline to which we apply the ratio, from current instantaneous active to active in the window.

src/hpa.c
```diff
@@ -295,8 +305,32 @@ hpa_ndirty_max(tsdn_t *tsdn, hpa_shard_t *shard) {
...
+		 *    exchange of higher hugepages coverage (we'll keep this
+		 *    slack and will not split up hugepages).  Slack is
+		 *    optional.
+		 *
+		 * Both components (demand and slack) are dynamic and depend on
+		 * amount of active memory we observed within sliding window.
+		 */
+		size_t nactive_max =
+		    peak_demand_nactive_max(&shard->peak_demand);
+		assert(nactive_max >= nactive);
```
ot: What guarantees this assertion? Is it always true that whatever updates psset_nactive also updates the window?

src/jemalloc.c
```diff
@@ -1546,6 +1546,13 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			    opt_hpa_opts.experimental_max_purge_nhp,
 			    "experimental_hpa_max_purge_nhp", -1, SSIZE_MAX);
 
+			CONF_HANDLE_UINT64_T(
+			    opt_hpa_opts.peak_demand_interval_ms,
+			    "hpa_peak_demand_interval_ms", 0,
+			    NSTIME_SEC_MAX * KQU(1000) < QU(UINT64_MAX) ?
+			    NSTIME_SEC_MAX * KQU(1000) : UINT64_MAX,
```
ot: What is the purpose of this? If it is to ensure that the product cannot exceed `UINT64_MAX` how does it work, wouldn't it overflow in the first place?

src/peak_demand.c
```diff
@@ -0,0 +1,79 @@
...
+#include "jemalloc/internal/jemalloc_internal_includes.h"
+
+#include "jemalloc/internal/peak_demand.h"
+
+void
+peak_demand_init(peak_demand_t *peak_demand, uint64_t interval_ms) {
+	assert(interval_ms > 0);
+	peak_demand->epoch = 0;
+	uint64_t interval_ns = interval_ms * 1000 * 1000;
+	peak_demand->epoch_interval_ns = interval_ns / PEAK_DEMAND_NBUCKETS;
```
ot, line 11: Let's do `>> PEAK_DEMAND_LG_NBUCKETS` here, so that it is clear why we're doing that.

src/peak_demand.c
```diff
@@ -0,0 +1,79 @@
...
+	memset(peak_demand->nactive_max, 0, sizeof(peak_demand->nactive_max));
+}
+
+static uint64_t
+peak_demand_epoch_ind(peak_demand_t *peak_demand) {
+	return peak_demand->epoch % PEAK_DEMAND_NBUCKETS;
```
ot, line 17: Same here.

src/peak_demand.c
```diff
@@ -0,0 +1,79 @@
...
+	}
+	peak_demand->epoch = next_epoch;
+	return;
+}
+
+void
+peak_demand_update(peak_demand_t *peak_demand, const nstime_t *now,
+    size_t nactive) {
+	peak_demand_maybe_advance_epoch(peak_demand, now);
+	uint64_t ind = peak_demand_epoch_ind(peak_demand);
```
ot: Can we have `peak_demand_maybe_advance_epoch` return `ind`? It has to know it

src/peak_demand.c
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+size_t
+peak_demand_nactive_max(peak_demand_t *peak_demand) {
+	size_t nactive_max = peak_demand->nactive_max[0];
+	for (int i = 1; i < PEAK_DEMAND_NBUCKETS; ++i) {
+		if (peak_demand->nactive_max[i] > nactive_max) {
+			nactive_max = peak_demand->nactive_max[i];
+		}
```
ot, line 71: There's no `max()` macro in the codebase?

src/peak_demand.c
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	if (delta > PEAK_DEMAND_NBUCKETS) {
+		delta = PEAK_DEMAND_NBUCKETS;
+	}
+	while (delta-- > 0) {
+		++peak_demand->epoch;
+		uint64_t ind = peak_demand_epoch_ind(peak_demand);
+		peak_demand->nactive_max[ind] = 0;
+	}
+	peak_demand->epoch = next_epoch;
```
ot: Would it be a bit clearer this way?
```
if (peak_demand->epoch + PEAK_DEMAND_NBUCKETS < next_epoch) {
  peak_demand->epoch = next_epoch - PEAK_DEMAND_NBUCKETS;
}
while (peak_demand->epoch < next_epoch) {
  ++peak_demand->epoch;
  uint64_t ind = peak_demand_epoch_ind(peak_demand);
  peak_demand->nactive_max[ind] = 0;
}
```
This way it is always true that the loop arrives at `next_epoch`, instead of having to patch it up separately. We discard the old epochs, not the new ones.

src/hpa.c
ilvokhin: > Is it always true that whatever updates psset_nactive also updates the window?

That's a good point. Before, I was thought we are holding `shard->mtx` lock and are calling `peak_demand_update` right before we call `peak_demand_nactive_max`. But in reality, we unlock `shard->mtx`, when we do `mmap` to purge dirty memory. In other words, when we purge more than one `hpdata`, then this `assert` might be false.

I'll update code and write a comment to explain why is was written this way.

src/peak_demand.c
ilvokhin: Nope, not at the moment.

src/peak_demand.c
ilvokhin: Yes, make sense, fixed.

src/peak_demand.c
ilvokhin: Are you sure we want to go that far? Any reasonable compiler will optimize it to shift anyway, but humans, will likely pause to think about what is going on here.

This case is probably fine, but modulo (`%`) replacement to `x & (PEAK_DEMAND_LG_NBUCKETS - 1)` is even more cryptic from readability point of view.

src/jemalloc.c
ilvokhin: This is a leftover from times, when `peak_demand_interval_ms` was a `ssize_t`, but it was refactored into current state without much thought.

Removed. There is still a problem: this option is in ms, but we use it internally as ns (by multiplying to 10**6), so in theory it is possible to provide value, that will overflow it. Same problem exists with `hpa_min_purge_interval_ms`.

src/peak_demand.c
ilvokhin: Yep, fixed.

src/hpa.c
ilvokhin: `frac` in `fxp_mul_frac` should be less (or equal?) than 1. So, best we can do is `nactive_max + fxp_mul_frac(nactive_max, shard->opts.dirty_mult)) - nactive`.

I rewrote code a bit, it is more close to what you suggested, but still doesn't match exactly. Please, take another look.

ilvokhin:
Factored in code review feedback. Ready for next round of review.

bmaurer:
Something that I think would be helpful is to write down some litmus tests for how this algorithm should work.

For example, IMHO one such litmus test would be:

> If a program does not allocate or deallocate memory an extended period, all dirty pages should eventually be released to the operating system.

This litmus test is important because programs should not hold on to entirely unused buffers.

I think another interesting potential litmus test:

> If a program's active memory usage drops below a low watermark value over the past N seconds, dirty memory should be freed immediately

In other words, if you have a program that allocated 20 GB of ram and has been like that for an hour, if you start freeing memory you should see an immediate drop in RSS. If you then re-allocate back to 20 GB and start freeing again, it's more reasonable to hold the memory to prevent unnecessary madvise calls.

ot:
> If a program does not allocate or deallocate memory an extended period, all dirty pages should eventually be released to the operating system.

This is actually not desirable because it would force to break up huge pages to get to zero dirty memory. This is why this formulation breaks up the estimation into forecasted demand and slack. The invariant is that if no allocation/deallocation is done for longer than the window, then the allocator holds on to at most dirty_mult * active dirty memory.

Without the time component, the dirty ratio needs to account both for forecasting future demand (and basing that on current demand is not really the right model) and the inevitable fragmentation coming from huge pages, so it needs to be a large number (say 5-10%).

By explicitly introducing the time component, we can make the slack floor much smaller, say 1%.

> If a program's active memory usage drops below a low watermark value over the past N seconds, dirty memory should be freed immediately

This is true, modulo the slack, see above. Immediately after the window is fully rotated almost all the dirty memory will be released.

bmaurer:
> > If a program does not allocate or deallocate memory an extended period, all dirty pages should eventually be released to the operating system.
> 
> This is actually not desirable because it would force to break up huge pages to get to zero dirty memory. This is why this formulation breaks up the estimation into forecasted demand and slack. The invariant is that if no allocation/deallocation is done for longer than the window, then the allocator holds on to at most dirty_mult * active dirty memory.
> 
> Without the time component, the dirty ratio needs to account both for forecasting future demand (and basing that on current demand is not really the right model) and the inevitable fragmentation coming from huge pages, so it needs to be a large number (say 5-10%).
> 
> By explicitly introducing the time component, we can make the slack floor much smaller, say 1%.

Maybe it's pedantic but it seems like the algorithm should be designed to eventually free at a minimum any contiguous huge page after an indefinite period (even if it was above the slack amount) 

> > If a program's active memory usage drops below a low watermark value over the past N seconds, dirty memory should be freed immediately
> 
> This is true, modulo the slack, see above. Immediately after the window is fully rotated almost all the dirty memory will be released.

I'm suggesting that we try to reduce the dirty memory before the window rotates. Like the diagram below (black = memory used by app; red = usage as seen by OS)

image

In other words, I'm suggesting that if nactive is below the *low watermark* of nactive over the window we should have a much lower dirty max.

Doing this enables significantly longer windows and provides a more intuative experience that applications which free long-held memory see an immediate demand drop.

ot:
> Maybe it's pedantic but it seems like the algorithm should be designed to eventually free at a minimum any contiguous huge page after an indefinite period (even if it was above the slack amount)

It can do that in a best-effort fashion, but I don't think it should be part of the contract, it is quite a hard condition to formalize.

> Like the diagram below

That diagram requires the existence of a time machine, since somehow the allocator starts getting memory from the OS before the application requests it :)

I think that you're converging to a decay-based model, which is what we would like to avoid here as it is very hard to reason about. The window can be already be made very small, we have good results with 10 seconds, which is a reasonable delay to release memory.

I can come up with many examples where a large amount of memory is freed and shortly after re-allocated, and your strategy would not guarantee full reuse (for example, a process that operates in rounds and each round is longer than the window). What I like about the peak demand window is that it is extremely easy to reason about: if the process requests as much memory as it was using at most 10 seconds ago, no memory churn should happen.

bmaurer:
We're not going to get perfect prediction for every type of application. I think it's worth looking at how stuff like this plays out in practice. I do think there's opportunity to have a bit more predictiveness without being difficult to operate. But we should see if this heutrisc ends up being close enough

ot:
Yes I should have clarified that we're prioritizing simplicity here, especially because there are plans to try out radical changes to how purging works, so we should fine-tune heuristics only after those changes settle.
However this is a very small (and natural) change which has a noticeable impact, so IMO it's worth doing it.

src/ctl.c
```diff
@@ -2246,6 +2249,8 @@ CTL_RO_NL_GEN(opt_hpa_min_purge_interval_ms, opt_hpa_opts.min_purge_interval_ms,
     uint64_t)
 CTL_RO_NL_GEN(opt_experimental_hpa_max_purge_nhp,
     opt_hpa_opts.experimental_max_purge_nhp, ssize_t)
+CTL_RO_NL_GEN(opt_hpa_peak_demand_window_ms,
```
guangli-dai, line 2257: This should definitely be another PR, but just want to ask: is there any case where it should be set on-the-fly / set for an individual arena to better fit (or to achieve different preferable trade-offs under) different workload patterns? Asking because currently pac's time-based purging allows per-arena setting through `mallctl` to be more flexible with different workload patterns.

test/unit/hpa.c
```diff
@@ -711,6 +699,7 @@ TEST_BEGIN(test_experimental_max_purge_nhp) {
 	expect_zu_eq(1, ndefer_purge_calls, "Expect purges");
 	ndefer_purge_calls = 0;
 
+
```
guangli-dai: nit: an extra blank line

src/ctl.c
ilvokhin: Yes, I think it make sense to make it more flexible in the future and allow change this value on per arena basis.

test/unit/hpa.c
ilvokhin: Thanks, fixed!

PR #2783, opened 2025-01-19
Use AVX to speedup arena_slab_reg_alloc_batch.

yetxanotherxamit:
The slab alloc path in jemalloc requires bitmap scanning to find the available slots within a page / slab. There's some opportunity to speed up the bitmap ops possibly utilizing AVX instructions.


ilvokhin:
Don't we afraid that frequency downclocking from AVX-512 instruction caused by memory allocator might have too much impact on overall application performance?

PR #2787, opened 2025-01-21
Check for 0 input when setting max_background_thread through mallctl

### LINKED ISSUE #2784: SIGFPE when max_background_threads == 0
Hi, when investigating [Debian CI failures](https://ci.debian.net/packages/r/rust-tikv-jemalloc-ctl/) of the Rust tikv-jemalloc-ctl crate (its [corresponding issue](https://github.com/tikv/jemallocator/issues/115)), I've found that, when `max_background_threads` is set to 0, this SIGFPE's due to division by zero:

https://github.com/jemalloc/jemalloc/blob/52fa9577ba8fa94f41c8c92f845a74c3fb04db80/src/background_thread.c#L527-L533

I don't really know how this knob works, maybe it's not intended to be 0, thus not considered a bug upstream (here). Or would it be a good idea to add a check and/or `max()`?

interwq:
Reported by @nc7s.

Resolves #2784

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -2181,7 +2181,8 @@ max_background_threads_ctl(tsd_t *tsd, const size_t *mib,
 			ret = 0;
 			goto label_return;
 		}
-		if (newval > opt_max_background_threads) {
+		if (newval > opt_max_background_threads ||
+		    newval == 0) {
 			ret = EINVAL;
 			goto label_return;
 		}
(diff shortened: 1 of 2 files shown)


ilvokhin:
Looks good to me, but it would be nice to add a unit-test for this as well. Looks like there is already something related in `test/unit/background_thread_enable.c`.

interwq:
> Looks good to me, but it would be nice to add a unit-test for this as well. Looks like there is already something related in `test/unit/background_thread_enable.c`.

Good point. Added in test/unit/background_thread_enable

ilvokhin (approved):

PR #2794, opened 2025-02-05
Implement process_madvise support.

interwq:
Add opt.process_madvise_max_batch which determines if process_madvise is enabled
(non-zero) and the max # of regions in each batch.  Added another limiting
factor which is the space to reserve on stack, which results in the max batch of
128.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2544,6 +2544,17 @@ if test "x${je_cv_madvise}" = "xyes" ; then
   if test "x${je_cv_madv_collapse}" = "xyes" ; then
     AC_DEFINE([JEMALLOC_HAVE_MADVISE_COLLAPSE], [ ], [ ])
   fi
+
+  dnl Check for process_madvise
+  JE_COMPILABLE([process_madvise(2)], [
+#include <sys/mman.h>
+#include <sys/syscall.h>
+], [
+	syscall(SYS_process_madvise, 0, (void *)0, 0, 0, 0);
+], [je_cv_process_madvise])
+  if test "x${je_cv_process_madvise}" = "xyes" ; then
+    AC_DEFINE([JEMALLOC_HAVE_PROCESS_MADVISE], [ ], [ ])
+  fi
 else
   dnl Check for posix_madvise.
   JE_COMPILABLE([posix_madvise], [
--- a/src/extent.c
+++ b/src/extent.c
@@ -12,6 +12,13 @@
 /* Data. */
 
 size_t opt_lg_extent_max_active_fit = LG_EXTENT_MAX_ACTIVE_FIT_DEFAULT;
+size_t opt_process_madvise_max_batch =
+#ifdef JEMALLOC_HAVE_PROCESS_MADVISE
+    PROCESS_MADVISE_MAX_BATCH_DEFAULT;
+#else
+    0
+#endif
+    ;
 
 static bool extent_commit_impl(tsdn_t *tsdn, ehooks_t *ehooks, edata_t *edata,
     size_t offset, size_t length, bool growing_retained);
@@ -1032,6 +1039,29 @@ extent_alloc_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 	return edata;
 }
 
+static void
+extent_dalloc_wrapper_finish(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
+    edata_t *edata) {
+	if (config_prof) {
+		extent_gdump_sub(tsdn, edata);
+	}
+	extent_record(tsdn, pac, ehooks, &pac->ecache_retained, edata);
+}
+
+void
+extent_dalloc_wrapper_purged(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
+    edata_t *edata) {
+	assert(edata_pai_get(edata) == EXTENT_PAI_PAC);
+	witness_assert_depth_to_rank(tsdn_witness_tsdp_get(tsdn),
+	    WITNESS_RANK_CORE, 0);
+
+	/* Verify that will not go down the dalloc / munmap route. */
+	assert(ehooks_dalloc_will_fail(ehooks));
+
+	edata_zeroed_set(edata, true);
+	extent_dalloc_wrapper_finish(tsdn, pac, ehooks, edata);
+}
+
 void
 extent_dalloc_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
     edata_t *edata) {
@@ -1077,11 +1107,7 @@ extent_dalloc_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 	}
 	edata_zeroed_set(edata, zeroed);
 
-	if (config_prof) {
-		extent_gdump_sub(tsdn, edata);
-	}
-
-	extent_record(tsdn, pac, ehooks, &pac->ecache_retained, edata);
+	extent_dalloc_wrapper_finish(tsdn, pac, ehooks, edata);
 }
 
 void
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1361,6 +1361,11 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			    "muzzy_decay_ms", -1, NSTIME_SEC_MAX * KQU(1000) <
 			    QU(SSIZE_MAX) ? NSTIME_SEC_MAX * KQU(1000) :
 			    SSIZE_MAX);
+			CONF_HANDLE_SIZE_T(opt_process_madvise_max_batch,
+			    "process_madvise_max_batch", 0,
+			    PROCESS_MADVISE_MAX_BATCH_LIMIT,
+			    CONF_DONT_CHECK_MIN, CONF_CHECK_MAX,
+			    /* clip */ true)
 			CONF_HANDLE_BOOL(opt_stats_print, "stats_print")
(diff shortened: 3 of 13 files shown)


configure.ac
```diff
@@ -2544,6 +2544,17 @@ if test "x${je_cv_madvise}" = "xyes" ; then
...
     AC_DEFINE([JEMALLOC_HAVE_MADVISE_COLLAPSE], [ ], [ ])
   fi
+
+  dnl Check for process_madvise
+  JE_COMPILABLE([process_madvise(2)], [
+#include <sys/mman.h>
+#include <sys/syscall.h>
+], [
+	syscall(SYS_process_madvise, 0, (void *)0, 0, 0, 0);
+], [je_cv_process_madvise])
```
guangli-dai, line 2554: Before kernel 6.13, only a few flags are allowed according to the [manual](https://man7.org/linux/man-pages/man2/process_madvise.2.html). We can't really know if the flag we need is permitted or not through this check, is that correct?

src/pac.c
```diff
@@ -450,6 +487,22 @@ pac_decay_stashed(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
 	bool try_muzzy = !fully_decay
 	    && pac_decay_ms_get(pac, extent_state_muzzy) != 0;
 
+	bool maximally_purge = !try_muzzy ||
+	    ecache->state == extent_state_muzzy;
+	/*
+	 * Attempt process_madvise only if 1) enabled, 2) purging to retained,
+	 * and 3) not using custom hooks.
+	 */
+	bool try_process_madvise = (opt_process_madvise_max_batch > 0) &&
```
nullptr0-0, line 497: Wouldn't batch == 1 essentially the same as madvise one by one?

src/extent.c
```diff
@@ -1032,6 +1039,29 @@ extent_alloc_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
+
+	/* Verify that will not go down the dalloc / munmap route. */
+	assert(ehooks_dalloc_will_fail(ehooks));
+
+	edata_zeroed_set(edata, true);
+	extent_dalloc_wrapper_finish(tsdn, pac, ehooks, edata);
+}
+
 void
 extent_dalloc_wrapper(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
```
nullptr0-0, line 1066: nit: add a **purged** parameter to `extent_dalloc_wrapper` instead of adding a new function?

src/jemalloc.c
```diff
@@ -1361,6 +1361,10 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			    "muzzy_decay_ms", -1, NSTIME_SEC_MAX * KQU(1000) <
 			    QU(SSIZE_MAX) ? NSTIME_SEC_MAX * KQU(1000) :
 			    SSIZE_MAX);
+			CONF_HANDLE_SIZE_T(opt_process_madvise_max_batch,
+			    "max_batched_size", 0, SIZE_T_MAX,
+			    CONF_DONT_CHECK_MIN, CONF_DONT_CHECK_MAX,
```
guangli-dai: If `max_stack` is limiting the real max batch, why don't we set it as a limit here and clip it? It is misleading for users to read a fake `max_batched_size` when it is larger than 128.

src/pac.c
```diff
@@ -435,6 +435,43 @@ pac_stash_decayed(tsdn_t *tsdn, pac_t *pac, ecache_t *ecache,
...
+decay_with_process_madvise(edata_list_inactive_t *decay_extents) {
+	cassert(have_process_madvise);
+	assert(opt_process_madvise_max_batch > 0);
+#ifndef JEMALLOC_HAVE_PROCESS_MADVISE
+	return true;
+#else
+	size_t max_stack = VARIABLE_ARRAY_SIZE_MAX / sizeof(struct iovec);
+	size_t len = max_stack < opt_process_madvise_max_batch ? max_stack :
+	    opt_process_madvise_max_batch;
+	struct iovec vec[len];
```
guangli-dai: This is VLA? Will it affect the compatibility for some C++ compilers?

src/pac.c
```diff
@@ -450,6 +487,22 @@ pac_decay_stashed(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
 	bool try_muzzy = !fully_decay
 	    && pac_decay_ms_get(pac, extent_state_muzzy) != 0;
 
+	bool maximally_purge = !try_muzzy ||
```
guangli-dai: maybe `purging_to_retained` is better?

src/pac.c
interwq: Yes. Batch == 1 and batch == 0 should be technically the same. It would be reasonable to to change the condition here to `batch > 1` so that it goes through the original madvise path. However I'd say still keep it different, to provide a testing case where `pmadvise` is always used with a single region. Otherwise batch == 1 and == 0 would be literally the same (and kind wasting a config value).

src/pac.c
```diff
@@ -435,6 +435,43 @@ pac_stash_decayed(tsdn_t *tsdn, pac_t *pac, ecache_t *ecache,
...
+#else
+	size_t max_stack = VARIABLE_ARRAY_SIZE_MAX / sizeof(struct iovec);
+	size_t len = max_stack < opt_process_madvise_max_batch ? max_stack :
+	    opt_process_madvise_max_batch;
+	struct iovec vec[len];
+	size_t cur = 0, total_bytes = 0;
+	for (edata_t *edata = edata_list_inactive_first(decay_extents);
+	     edata != NULL;
+	     edata = edata_list_inactive_next(decay_extents, edata)) {
+		size_t pages_bytes = edata_size_get(edata);
```
guangli-dai, line 454: we don't remove things away from the list. In case `pages_purge_process_madvise` failed after some success, these items will be repeatedly purged? Does this matter? Or we are doing this just to save time?

src/extent.c
interwq: Adding a new _purged function is intentional because I expect the purged variant to only be used at that single place, so that we don't have to bother other call sites of `extent_dalloc_wrapper` with another input. In other words, the _purged variant is an uncommon case; if we expect both variants to be common, adding an input arg would be the better choice.

src/pages.c
```diff
@@ -833,6 +877,7 @@ pages_boot(void) {
 		os_pages_unmap(madv_free_page, PAGE);
 	}
 #endif
+	init_proces_madvise();
```
guangli-dai: Not handling the return value here? What does it mean when this `init` fails? If the error here is not worth aborting in the opt build, is it for the debug build? Or any other ways to warn users that something went wrong?

configure.ac
interwq: Right. We only check if process_madvise is supported at compile time. If the MADV_DONTNEED flag isn't supported, the syscall will return error on the fly, and the purge will fallback to the regular madvise in that case (the `already_purged = !decay_with_process_madvise` part).

src/jemalloc.c
interwq: Good point. Added cap and clip.

src/pac.c
interwq: Good catch. Replaced with `VARIABLE_ARRAY`

src/pac.c
interwq: If pmadvise is partial success, yes some regions will be purged multiple times but there will be no correctness concern. I'm not too worried about that case, because it's not suppose to happen unless there's heap corruption.

src/pac.c
interwq: Agree. Changed.

src/pages.c
interwq: The only case I can think of when the init fails, is OOM during pidfd_open. Added return value check.

interwq (commented):
Address feedback

src/pac.c
guangli-dai: Agreed. As long as purging multiple times does not have correctness concern, we should be good. Maybe add a comment to explain this just for future references?

src/pac.c
```diff
@@ -435,6 +435,43 @@ pac_stash_decayed(tsdn_t *tsdn, pac_t *pac, ecache_t *ecache,
...
 
+static bool
+decay_with_process_madvise(edata_list_inactive_t *decay_extents) {
+	cassert(have_process_madvise);
+	assert(opt_process_madvise_max_batch > 0);
+#ifndef JEMALLOC_HAVE_PROCESS_MADVISE
+	return true;
+#else
+	size_t stack_limit = PROCESS_MADVISE_MAX_BATCH_LIMIT;
+	size_t len = stack_limit < opt_process_madvise_max_batch ? stack_limit :
```
guangli-dai: With the clip we should be safe to assert `opt_process_madvise_max_batch <= PROCESS_MADVISE_MAX_BATCH_LIMIT` instead of still using `len` here?

nullptr0-0 (approved):
code-wise lgtm, curious how was the change tested?

guangli-dai (approved):

src/pac.c
interwq: Agree. Added comments around `already_purged`

src/pac.c
interwq: The only case not covered by the config `clip` is the default value `PROCESS_MADVISE_MAX_BATCH_DEFAULT`. To be extra safe, added a manual clip in init_process_madvise and replaced here with assert.

interwq:
> curious how was the change tested?

Tested on my dev machine with an updated kernel (with support of `pmadvise(MADV_DONTNEED)`). Added logging to verify that the number of bytes covered with pmadvise and madvise are the same. We should add more unit tests once the kernel patch is merged and becomes available.

PR #2795, opened 2025-02-05
Fix arena 0 `deferral_allowed` flag init

ilvokhin:
Arena 0 have a dedicated initialization path, which differs from initialization path of other arenas. The main difference for the purpose of this change is that we initialize arena 0 before we initialize background threads. HPA shard options have `deferral_allowed` flag which should be equal to `background_thread_enabled()` return value, but it wasn't the case, before this change, because for arena 0 `background_thread_enabled()` was initialized correctly after arena 0 initialization phase already ended.

Below is initialization sequence after this commit to illustrate everything still should be initialized correctly.

* `hpa_central_init` initializes HPA Central, before we initialize every HPA shard (including arena's 0).
* `background_thread_boot1` initializes `background_thread_enabled()` return value.
* `pa_shard_enable_hpa` initializes arena 0 HPA shard.

 ```
                       malloc_init_hard -------------
                      /           /                  \   
                     /           /                    \   
                    /           /                      \   
malloc_init_hard_a0_locked  background_thread_boot1  pa_shard_enable_hpa
        /                     /                          \   
       /                     /                            \   
      /                     /                              \   
arena_boot       background_thread_enabled_seta         hpa_shard_init
     |   
     |   
pa_central_init
     |   
     |   
hpa_central_init
```

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1964,13 +1964,6 @@ malloc_init_hard_a0_locked(void) {
 		} else {
 			opt_hpa = false;
 		}
-	} else if (opt_hpa) {
-		hpa_shard_opts_t hpa_shard_opts = opt_hpa_opts;
-		hpa_shard_opts.deferral_allowed = background_thread_enabled();
-		if (pa_shard_enable_hpa(TSDN_NULL, &a0->pa_shard,
-		    &hpa_shard_opts, &opt_hpa_sec_opts)) {
-			return true;
-		}
 	}
 
 	malloc_init_state = malloc_init_a0_initialized;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/jemalloc.c [hunk 1]
```diff
@@ -2211,6 +2204,20 @@ malloc_init_hard(void) {
...
+		 * Initialization of a0->pa_shard.hpa_shard happen before
+		 * background_thread_boot1 was called, so value of
+		 * background_thread_enabled_state was false at the time of
+		 * hpa_shard->opts initialization and hence
+		 * hpa_shard->opts.deferral_allowed was set to false as well.
+		 * We need to fix it here for arena 0 to have correct options
+		 * set.
+		 */
+		arena_t *a0 = arena_get(tsd_tsdn(tsd), 0, false);
+		pa_shard_set_deferral_allowed(tsd_tsdn(tsd), &a0->pa_shard,
```
interwq: Let's move the entire a0 HPA enablement here, i.e. pa_shard_enable_hpa. Having a0's HPA partially initialized w/ a temporary deferral_allowed feels risky, e.g. if the HPA shard internal initialization makes any decision based on the deferral_allowed state. It's cleaner to make it either not enabled, or fully enabled.

ilvokhin: Make sense, fixed.

PR #2797, opened 2025-02-07
detect false failure of strerror_r

BusyJay:
See tikv/jemallocator#108.

In a summary, test on `strerror_r` can fail due to reasons other than `strerror_r` itself, so add an additional test to determine the failure is expected.


interwq:
Thanks for the patch!

PR #2801, opened 2025-02-18
Make hugify system call only when thp=madvise

ilvokhin:
When `/sys/kernel/mm/transparent_hugepage/enabled` is always or never it doesn't make much sense to call `madvise(..., MADV_HUGEPAGE)` and `madvise(..., MADV_NOHUGEPAGE)`, because it will not have desired effect anyway, but we'll waste CPU cycles on system call.

Call `madvise(..., MADV_HUGEPAGE)` and `madvise(..., MADV_NOHUGEPAGE)`, only, when `/sys/kernel/mm/transparent_hugepage/enabled` is set to `madvise`


ilvokhin:
I think we can cut even deeper and avoid `to_hugify` list operations as well.

PR #2806, opened 2025-03-03
Fix bugs on the profiling part.

guangli-dai:
1. xallocx does not retrieve the new `emap_alloc_ctx` after expanding/shrinking, which may causes problem in `prof_info_get_and_reset_recent`, fixing this;
2. Currently, no tests cover the `isfree` path for small & sampled allocations, adding that into `prof_small.c`.


interwq:
Let's polish the commit msg a bit, e.g. "Fix profiling sample metadata lookup during xallocx"

PR #2810, opened 2025-03-05
Support THP with Huge Arena in PAC

nullptr0-0:
This change aims to enable THP for the huge arena when the opt_thp setting is default.
- **no change**:  opt_thp is "always" or "disabled"
- **change**: when opt_thp is "default", we now internally enable THP for the huge arena based on the arena[0] metadata THP usage:
    - When opt_metadata_thp is "always", we will enable THP for the huge arena as well.
    - When opt_metadata_thp is "auto", we'll turn on THP when the arena[0]'s metadata THP auto switch is turned on. This is achieved by: 1) add a pending list to track the grown regions before the switch is turned on 2) triggers the switch and hugify the pending list inside **base_auto_thp_switch()**.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -46,7 +46,15 @@ size_t oversize_threshold = OVERSIZE_THRESHOLD_DEFAULT;
 
 uint32_t arena_bin_offsets[SC_NBINS];
 
-static unsigned huge_arena_ind;
+/*
+ * a0 is used to handle huge requests before malloc init completes. After
+ * that,the huge_arena_ind is updated to point to the actual huge arena,
+ * which is the last one of the auto arenas.
+ */
+unsigned huge_arena_ind = 0;
+bool opt_huge_arena_pac_thp = false;
+pac_thp_t huge_arena_pac_thp = {.thp_madvise = false,
+    .auto_thp_switched = false, .n_thp_lazy = ATOMIC_INIT(0)};
 
 const arena_config_t arena_config_default = {
 	/* .extent_hooks = */ (extent_hooks_t *)&ehooks_default_extent_hooks,
@@ -1898,6 +1906,7 @@ arena_choose_huge(tsd_t *tsd) {
 bool
 arena_init_huge(arena_t *a0) {
 	bool huge_enabled;
+	assert(huge_arena_ind == 0);
 
 	/* The threshold should be large size class. */
 	if (opt_oversize_threshold > SC_LARGE_MAXCLASS ||
@@ -1908,10 +1917,18 @@ arena_init_huge(arena_t *a0) {
 	} else {
 		/* Reserve the index for the huge arena. */
 		huge_arena_ind = narenas_total_get();
+		assert(huge_arena_ind != 0);
 		oversize_threshold = opt_oversize_threshold;
 		/* a0 init happened before malloc_conf_init. */
 		atomic_store_zu(&a0->pa_shard.pac.oversize_threshold,
 		    oversize_threshold, ATOMIC_RELAXED);
+		/* Initialize huge arena THP settings for PAC. */
+		(&huge_arena_pac_thp)->thp_madvise = opt_huge_arena_pac_thp &&
+		    metadata_thp_enabled() && (opt_thp == thp_mode_default) &&
+		    (init_system_thp_mode == thp_mode_default);
+		malloc_mutex_init(&(&huge_arena_pac_thp)->lock, "pac_thp",
+		    WITNESS_RANK_LEAF, malloc_mutex_rank_exclusive);
+		edata_list_active_init(&(&huge_arena_pac_thp)->thp_lazy_list);
 		huge_enabled = true;
 	}
 
--- a/src/base.c
+++ b/src/base.c
@@ -55,9 +55,6 @@ base_map(tsdn_t *tsdn, ehooks_t *ehooks, unsigned ind, size_t size) {
 	}
 	if (ehooks_are_default(ehooks)) {
 		addr = extent_alloc_mmap(NULL, size, alignment, &zero, &commit);
-		if (have_madvise_huge && addr) {
-			pages_set_thp_state(addr, size);
-		}
 	} else {
 		addr = ehooks_alloc(tsdn, ehooks, NULL, size, alignment, &zero,
 		    &commit);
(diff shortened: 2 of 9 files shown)


src/arena.c
```diff
@@ -46,7 +46,9 @@ size_t oversize_threshold = OVERSIZE_THRESHOLD_DEFAULT;
 
 uint32_t arena_bin_offsets[SC_NBINS];
 
-static unsigned huge_arena_ind;
+unsigned huge_arena_ind = 0;
+pac_thp_t huge_arena_pac_thp = {.thp_madvise = false,
+    .auto_thp_switched = false, .n_pending_max = ATOMIC_INIT(0)};
```
guangli-dai: The two variables here seem to have two assumptions: 1. there will be only one huge arena; 2. the huge arena index will never be 0 (unless before initialization) because there will be at least one auto arena. Is that correct? Maybe let's put these assumptions into a comment for easier future reference.

include/jemalloc/internal/witness.h
```diff
@@ -58,9 +58,9 @@ enum witness_rank_e {
 	WITNESS_RANK_HPA_CENTRAL_GROW,
 	WITNESS_RANK_HPA_CENTRAL,
 
+	WITNESS_RANK_RTREE,
```
interwq: Curious why this was needed?

src/base.c
```diff
@@ -187,6 +226,16 @@ base_auto_thp_switch(tsdn_t *tsdn, base_t *base) {
 		block = block->next;
 		assert(block == NULL || (base_ind_get(base) == 0));
 	}
+
+	/* Handle the THP auto switch for the huge area. */
```
guangli-dai: nit: arena

src/extent.c
```diff
@@ -767,6 +819,13 @@ extent_grow_retained(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 	exp_grow_size_commit(&pac->exp_grow, exp_grow_skip);
 	malloc_mutex_unlock(tsdn, &pac->grow_mtx);
 
+	/* Avoid using HUGEPAGE when the grow size is less than HUGEPAGE. */
```
guangli-dai: Will the size be smaller than HUGEPAGE if `ind == huge_arena_id`

src/base.c
```diff
@@ -153,6 +150,48 @@ base_get_num_blocks(base_t *base, bool with_new_block) {
...
+		return;
+	}
+	assert(huge_arena != NULL);
+
+	edata_list_active_t list;
+	edata_list_active_init(&list);
+
+	malloc_mutex_lock(tsdn, &pac_thp->lock);
+	edata_list_active_move(&list, &pac_thp->thp_pending_list);
+	assert(edata_list_active_empty(&pac_thp->thp_pending_list));
```
guangli-dai: Everything in `thp_pending_list` will get hugified when switched on. What would be possible states of these edatas? Do their states matter?

include/jemalloc/internal/witness.h
nullptr0-0: I'm reusing existing edata from the PAC edata_cache to store items in the thp_pending_list. However, this can lead to a reversed lock ranking order between rtree(22), and edata_cache(21) under certain conditions:

1. When extent_grow_retained is called, it invokes extent_split, which in turn calls rtree_leaf_elm_lookup. This might causes rtree_node_init to be executed, grabbing the rtree->init_lock.
2. With the rtree lock held, there's a possibility that rtree_node_alloc() will be called, which then proceeds to base_alloc.
3. In base_alloc, there's a chance that the base auto_switch will be flipped, triggering the clearance of the huge arena pending list. This returns the edata storing pending items back to the PAC edata_cache, resulting in a reversed lock ranking order between rtree and edata_cache.

I've observed this case during testing, while cannot think of any other scenarios where rtree needs to be acquired after edata_cache, thus changing the order between the two.

src/extent.c
nullptr0-0: We grow extents less than HUGEPAGE when HUGEPAGE size is for example 512M right? to avoid VM explode.

src/base.c
nullptr0-0: the state is irrelevant, merely borrowing the typed list for implementation simplicity.

src/arena.c
nullptr0-0: Right, we use arena 0 to serve huge requests in the beginning before malloc_initialized(), after that ind will be updated to the actual one, which is the last auto arena.

src/base.c
guangli-dai: I don't quite follow why it is irrelevant. The size and the address in the `edata` could be changing after appending to this list, right? There could be coalescing and splitting happening? This will affect the range of buffer the hugification applies to?

nullptr0-0: It won't, we hugify the entire grown retained region at once (see the comments in the callsite of extent_handle_huge_arena_thp() below) instead of going through extents one by one. So whenever we newly mmap() a region, we tell kernel that this address range will be using HUGEPAGE, regardless of how we use this region. We might split this region into multiple extents to serve actual requests, or merge them, the underlying address range will always be backed up by HUGEPAGE.

src/extent.c
```diff
@@ -767,6 +819,13 @@ extent_grow_retained(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
+	if (ind != 0 && ind == huge_arena_ind && ehooks_are_default(ehooks) &&
+	    likely(alloc_size >= HUGEPAGE)) {
+		extent_handle_huge_arena_thp(tsdn, &huge_arena_pac_thp,
+		    pac->edata_cache, ptr, alloc_size);
```
nullptr0-0: @guangli-dai  we pass in the `ptr` and `alloc_size` here that's returned by ehooks_alloc(), i.e. mmap, instead of the processed extent/edata address/size.

src/base.c
guangli-dai: That is indeed the case when `pac_thp->auto_thp_switched`, but before it is switched on, the extents grow go into upper level without being hugified and we keep track of them in the pending list, right? After going into the upper level and before reaching this function, those edatas will have their address and size not changed? 

My question is: comparing the address and size of the edata when they are inserted into `pending_list` and when they are popped out below, are the values guaranteed to be the same?

nullptr0-0: Yes, because the edata inserted is only updated when we insert it into the list and not touched at all before they're poped out?

guangli-dai: I think I see your point. We are using dedicated `edata` to store these information (the addr and size) of the region to be hugified so that they are safe from any other modifications, right?

nullptr0-0: >  but before it is switched on, the extents grow go into upper level without being hugified and we keep track of them in the pending list, right?

Right, but we only keep track of the "original region" once after it's mmaped, a separate edata is "created" (recycled through edata_cache) to track them, not the one that actually track the extents in prod flow.

nullptr0-0: > We are using dedicated edata to store these information (the addr and size) of the region to be hugified so that they are safe from any other modifications, right?

Correct, and to avoid the edata cost we recycle it from the PAC edata_cache.

interwq: If it can simplify things like having to access edata_cache from here, consider keeping the edata even after we hugify everything -- first the # of edata here should be low given with opt.retain, and second it would also allow us to de-hugify in the future.

nullptr0-0: Like `base_alloc_edata()` directly and stay in the list?

interwq: Yes

src/base.c
```diff
@@ -153,6 +150,40 @@ base_get_num_blocks(base_t *base, bool with_new_block) {
...
+	ql_foreach(edata, &pending_list->head, ql_link_active) {
+		assert(edata != NULL);
+		void *addr = edata_addr_get(edata);
+		size_t size = edata_size_get(edata);
+		assert(HUGEPAGE_ADDR2BASE(addr) == addr);
+		assert(HUGEPAGE_CEILING(size) == size && size != 0);
+		pages_huge(addr, size);
+		cnt++;
+	}
+	assert(cnt == atomic_load_u(&pac_thp->n_thp_lazy, ATOMIC_RELAXED));
```
interwq, line 184: Since we are not holding the lock, can we get more n_thp_lazy while doing the hugify?

nullptr0-0: No there won't be more items enqueued after we switch the `auto_thp_switched` to be true, the reason is that:
- We hold the lock in extent.c when enqueue, the lock protects the sequential operations of 1) read the `auto_thp_switched` and if it's false 2) enqueue. Thus if an item is enqueued, it's guaranteed that auto_thp_switched has not been turned on, vice versa.

nullptr0-0: oh btw, `n_thp_lazy` only tracks the number of items in the `thp_lazy_list`, it won't continue increase after the switch being turned on. (Also this is not the number of HUGE PAGRES, it's just the number of regions, counting is solely for assertion purpose)
We can indeed count the number of HUGEPAGES being requested in huge arena, but since we already have raw number of bytes mapped + retained, it's kinda duplicated for calculating them I guess? I did print the n_thp out when I was testing and removed it in the actual PR, I can add it back if it's useful.

interwq: Thanks for explaining Sherry

guangli-dai (approved):

PR #2811, opened 2025-03-07
Fix frame pointer based unwinder to handle changing stack range

guangli-dai:
There are cases where the stack range may change, i.e., folly::fibers. For these cases, the current frame pointer based unwinder may crash. This patch adds a detection to the stack range changing and will fallback to Linux's `backtrace` if changes are detected. Note that `backtrace` may get redirected to libunwind if it is built without flag `--disable-weak-backtrace`. 

Alos formatted `prof_stack_range.c` using `clang-format`.

### DIFF
--- a/scripts/gen_travis.py
+++ b/scripts/gen_travis.py
@@ -265,7 +265,13 @@ def generate_linux(arch):
     if arch != ARM64:
         exclude += [LARGE_HUGEPAGE]
 
-    return generate_jobs(os, arch, exclude, max_unusual_opts)
+    linux_configure_flags = list(configure_flag_unusuals)
+    linux_configure_flags.append(Option.as_configure_flag("--enable-prof --enable-prof-frameptr"))
+
+    linux_unusuals = (compilers_unusual + feature_unusuals
+                    + linux_configure_flags + malloc_conf_unusuals)
+
+    return generate_jobs(os, arch, exclude, max_unusual_opts, linux_unusuals)
 
 
 def generate_macos(arch):
--- a/src/prof_sys.c
+++ b/src/prof_sys.c
@@ -23,6 +23,11 @@
 #define _Unwind_Backtrace JEMALLOC_TEST_HOOK(_Unwind_Backtrace, test_hooks_libc_hook)
 #endif
 
+#ifdef JEMALLOC_PROF_FRAME_POINTER
+// execinfo backtrace() as fallback unwinder
+#include <execinfo.h>
+#endif
+
 /******************************************************************************/
 
 malloc_mutex_t prof_dump_filename_mtx;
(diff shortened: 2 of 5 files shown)


src/prof_sys.c
```diff
@@ -102,41 +107,101 @@ prof_backtrace_impl(void **vec, unsigned *len, unsigned max_len) {
...
+		}
+		void* ip = ((void **)fp)[1];
+		if (ip == 0) {
+			break;
+		}
+		vec[ii++] = ip;
+		fp = ((uintptr_t *)fp)[0];
+	}
+
+label_fallback:
```
nullptr0-0, line 186: nit: this is a bit confusing that non-fallback path also go through label_fallback? Is this label necessary?

scripts/gen_travis.py
```diff
@@ -265,7 +267,15 @@ def generate_linux(arch):
     if arch != ARM64:
         exclude += [LARGE_HUGEPAGE]
 
-    return generate_jobs(os, arch, exclude, max_unusual_opts)
+    linux_configure_flags = list(configure_flag_unusuals)
```
nullptr0-0, line 268: Is there a reason that we override the individual `"--enable-prof"` flag instead of keeping both `"--enable-prof"` and `"--enable-prof --enable-prof-frameptr"` ?

src/prof_sys.c
guangli-dai: Good point! The naming is indeed inappropriate here. I prefer to keep this label because otherwise, we will have another `if` wrapping the for loops above, jumping to the return point here makes it neater. As for the name, what about `label_return` or `label_end`?

scripts/gen_travis.py
guangli-dai: By keeping both, I think you mean adding `--enable-prof --enable-prof-frameptr` to the unusual list right? The functional difference would be: if we keep both, we will also have tests testing "prof enabled with default unwinder (libgcc)" and "prof enabled with the new unwinder". This will increase the number of tests. But I also begin to think it is worth it to cover libgcc in the test. WDYT?

When both are used, there will be a duplicated `--enable-prof` but that is totally fine.

nullptr0-0: > I think you mean adding --enable-prof --enable-prof-frameptr to the unusual list right?

Right! Though number of test will be increased, since profiling has given us headache, keeping the coverage is important?

src/prof_sys.c
nullptr0-0: Or keep the label_fallback and rearrange it a little bit? Something like below:

nullptr0-0:
```
	while (ii < max_len && fp != 0) {
		if (fp < unwind_info.stack_range.start || fp >= unwind_info.stack_range.end) {
			/*
			 * Determining the stack range from procfs can be
			 * relatively expensive especially for programs with
			 * many threads / shared libraries.  If the stack
			 * range has changed, it is likely to change again
			 * in the future (fibers or some other stack
			 * manipulation).  So fall back to backtrace for this
			 * thread.
			 */
			unwind_info.fallback = true;
			ii = 0;
			goto label_fallback;
		}
		void* ip = ((void **)fp)[1];
		if (ip == 0) {
			break;
		}
		vec[ii++] = ip;
		fp = ((uintptr_t *)fp)[0];
	}
        *len = ii;
        return;
 
label_fallback:
	/*
	 * Using the backtrace from execinfo.h here.  Note that it may get
	 * redirected to libunwind when a libunwind not built with build-time
	 * flag --disable-weak-backtrace is linked.
	 */
	assert(unwind_info.fallback)
	int nframes = backtrace(vec, max_len);
	if (nframes > 0) {
		*len = nframes;
	} else {
		*len = 0;
	}
```

scripts/gen_travis.py
guangli-dai: Modified to append this and it will add 14 travis test. @interwq Is that fine?

src/prof_sys.c
guangli-dai: This looks the best among all choices, modified.

scripts/gen_travis.py
interwq: SGTM

nullptr0-0 (approved):

PR #2812, opened 2025-03-07
background_thread: add fallback for pthread_create dlsym

rhelmot:
If jemalloc is linked into a shared library, the RTLD_NEXT dlsym call may fail since RTLD_NEXT is only specified to search all objects after the current one in the loading order, and the pthread library may be earlier in the load order. Instead of failing immediately, attempt one more time to find pthread_create via RTLD_GLOBAL.

Errors cascading from this were observed on FreeBSD 14.1.


interwq:
Thanks for the patch!

PR #2818, opened 2025-03-20
Fix assertion error in huge_arena_auto_thp_switch() when b0 is deleted and recreated

nullptr0-0:
Fix the assertion error in `huge_arena_auto_thp_switch()` when the auto switch has been turned on by b0 previously, but we delete and recreate b0 again, `causing huge_arena_auto_thp_switch()` to be called again while the switch is already turned on.
Additionally, the `thp_madvise` field was initialized at the same time as when the huge arena was initialized. However, it's possible that we could turn on the switch before the huge arena is even initialized. In that case, `thp_madvise` would return false and cause the auto switch to never be turned on.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -1904,7 +1904,7 @@ arena_choose_huge(tsd_t *tsd) {
 }
 
 bool
-arena_init_huge(arena_t *a0) {
+arena_init_huge(tsdn_t *tsdn, arena_t *a0) {
 	bool huge_enabled;
 	assert(huge_arena_ind == 0);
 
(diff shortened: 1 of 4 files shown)


src/arena.c
```diff
@@ -1915,20 +1915,26 @@ arena_init_huge(arena_t *a0) {
...
 		(&huge_arena_pac_thp)->thp_madvise = opt_huge_arena_pac_thp &&
 		    metadata_thp_enabled() && (opt_thp == thp_mode_default) &&
 		    (init_system_thp_mode == thp_mode_default);
+		(&huge_arena_pac_thp)->auto_thp_switched = b0->auto_thp_switched;
 		malloc_mutex_init(&(&huge_arena_pac_thp)->lock, "pac_thp",
 		    WITNESS_RANK_LEAF, malloc_mutex_rank_exclusive);
 		edata_list_active_init(&(&huge_arena_pac_thp)->thp_lazy_list);
+		malloc_mutex_unlock(tsdn, &b0->mtx);
+
+		oversize_threshold = opt_oversize_threshold;
```
guangli-dai: Curious why do we want to move this below? We can also init `oversize_threshold` first and then hold the lock and do the `huge_arena_pac_thp` init?

nullptr0-0: Move the huge_arena_pac_thp section to right after the huge_arena_ind assignment, so that both are protected by the lock. 
On a second thought, we can also have the b0 lock only protect huge_arena_pac_thp init and remove the assert(huge_arena_ind != 0) in the base auto switch as well.

Issue #2688, opened 2024-08-09
`dirty_decay_ms` doesn't take effect. Lack of adequate control for physical memory usage with jemalloc


zhitaoli-6:
In our environment, we do care about the process memory usage. If the RSS of our process exceeds the hard limit, OOM will occur. With the default options of jemalloc("retain:true, dirty_decay_ms:10000, muzzy_decay_ms:0"), the peak RSS exceeds 1GB while the expected is below 500MB. 

Set `dirty_decay_ms` to 1000 or 10 doesn't take effect at all. The peak RSS keeps below 500MB only when `dirty_decay_ms` is set to 0. It seems that `dirty_decay_ms` is not an always effective way to control the tradeoff between CPU and memory. **Is this as expected? If so, what is the reason?** BTW, could we limit the dirty pages according to the amount instead of the timeout? 

Following is the malloc stats when `dirty_decay_ms` is set to 1000. 
```
{"log":"Allocated: 610687176, active: 618848256, metadata: 13666544 (n_thp 0), resident: 1059119104, mapped: 1125564416, retained: 556875776\n","stream":"stderr","time":"2024-08-09T11:41:50.304038417Z"}
{"log":"decaying:  time       npages       sweeps     madvises       purged\n","stream":"stderr","time":"2024-08-09T11:41:50.304084572Z"}
{"log":"   dirty:   N/A       104766           38          121        94299\n","stream":"stderr","time":"2024-08-09T11:41:50.304086988Z"}
{"log":"   muzzy:   N/A            0            0            0            0\n","stream":"stderr","time":"2024-08-09T11:41:50.304089344Z"}
```
Following is the malloc stats when `dirty_decay_ms` is set to 10. 
```
{"log":"Allocated: 568836632, active: 578674688, metadata: 14977648 (n_thp 0), resident: 1210519552, mapped: 1276792832, retained: 695578624\n","stream":"stderr","time":"2024-08-09T12:05:22.639788751Z"}
{"log":"decaying:  time       npages       sweeps     madvises       purged\n","stream":"stderr","time":"2024-08-09T12:05:22.639878378Z"}
{"log":"   dirty:   N/A       151495           47          181       162781\n","stream":"stderr","time":"2024-08-09T12:05:22.639882258Z"}
{"log":"   muzzy:   N/A            0            0            0            0\n","stream":"stderr","time":"2024-08-09T12:05:22.639888371Z"}
```
[...]

PR #2819, opened 2025-03-26
Document the adverse effects of delayed memory return

nh2:
See #2688.

Also document the surprising behaviour that the times configured `dirty_decay_ms`/`muzzy_decay_ms` have no effect unless `background_thread:true` is given if the process is sleeping.

I was very surprised when I had a process waiting for user input sitting around at 100 GB RES memory,
and `dirty_decay_ms` not returning the memory after the default 10 seconds, but `dirty_decay_ms:0` fixing the problem immediately. I was even more confused that changing the time configured by `dirty_decay_ms` seemed to be completely ignored by jemalloc, until I got the explanation from
https://github.com/jemalloc/jemalloc/issues/2688#issuecomment-2464775694

Personally I think the default-on of `dirty_decay_ms` is questionable, as I encountered the problem documented here with
"This can lead to out-of-memory crashes ..."
where my program needed 100 GB for some processing, freed it, and then invoked a child process that also needed 100 GB RAM; the parent program continued to sit at 100 GB RES usage and thus was OOM-killed, creating a bug where none should be (as the parent process freed all memory exactly right).

This commit at least documents the impact of this questionable default.

### DIFF
--- a/TUNING.md
+++ b/TUNING.md
@@ -43,6 +43,11 @@ Runtime options can be set via
     between CPU and memory usage.  Shorter decay time purges unused pages faster
     to reduces memory usage (usually at the cost of more CPU cycles spent on
     purging), and vice versa.
+    Be aware that decay times `> 0` will not be honored until the next relevant call
+    into jemalloc, unless you also enable `background_thread:true`.
+    Without `background_thread:true`, processes that are sleeping
+    (e.g. because they call `sleep()`, block on user input, network/file activity,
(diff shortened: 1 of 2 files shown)


nh2:
Please somebody check that the resulting docs look right.

I tried `make build_doc_html` but it always output

```
Missing xsltproc.  doc/jemalloc.html not (re)built.
```

even though `xsltproc` is on `PATH`. Probably I don't have the `xsl-stylesheets`.

PR #2820, opened 2025-03-28
if process_madvise is supported, call it when purging hpa

spredolac:
Similar to common pages logic. If there is a process_madvise use it in hpa (to save few syscalls).

### DIFF
--- a/include/jemalloc/internal/hpa_hooks.h
+++ b/include/jemalloc/internal/hpa_hooks.h
@@ -13,6 +13,7 @@ struct hpa_hooks_s {
 	void (*dehugify)(void *ptr, size_t size);
 	void (*curtime)(nstime_t *r_time, bool first_reading);
 	uint64_t (*ms_since)(nstime_t *r_time);
+	bool (*vectorized_purge)(void* vec, size_t vlen, size_t nbytes);
 };
 
 extern const hpa_hooks_t hpa_hooks_default;
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -8,6 +8,16 @@
 
 #define HPA_EDEN_SIZE (128 * HUGEPAGE)
 
+#define HPA_MIN_VAR_VEC_SIZE 8
+#ifdef JEMALLOC_HAVE_PROCESS_MADVISE
+typedef struct iovec hpa_io_vector_t;
+#else
+typedef struct {
+	void *iov_base;
+	size_t iov_len;
+} hpa_io_vector_t;
+#endif
+
 static edata_t *hpa_alloc(tsdn_t *tsdn, pai_t *self, size_t size,
     size_t alignment, bool zero, bool guarded, bool frequent_reuse,
     bool *deferred_work_generated);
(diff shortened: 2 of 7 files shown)


include/jemalloc/internal/hpa_hooks.h
```diff
@@ -13,6 +13,7 @@ struct hpa_hooks_s {
 	void (*dehugify)(void *ptr, size_t size);
 	void (*curtime)(nstime_t *r_time, bool first_reading);
 	uint64_t (*ms_since)(nstime_t *r_time);
+        bool (*vectorized_purge)(struct iovec* vec, size_t vlen, size_t nbytes);
```
interwq: `struct iovec` may not be available. We need to guard the use cases with `JEMALLOC_HAVE_PROCESS_MADVISE` like in https://github.com/jemalloc/jemalloc/pull/2794/files

In the hook signature here, let's make it `void *` so that we don't have to add too many #ifdef. Then only in the supported implementation we convert it to `struct iovec`

src/hpa.c
```diff
@@ -470,14 +498,36 @@ hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
 	}
 	size_t total_purged = 0;
 	uint64_t purges_this_pass = 0;
+
+	assert(opt_process_madvise_max_batch <=
+		PROCESS_MADVISE_MAX_BATCH_LIMIT);
+	size_t len = opt_process_madvise_max_batch < HPA_MIN_VAR_VEC_SIZE ?
+		HPA_MIN_VAR_VEC_SIZE : opt_process_madvise_max_batch;
```
interwq, line 505: Question: any particular reason why the min size is enforced?

src/hpa_hooks.c
```diff
@@ -78,3 +81,14 @@ static uint64_t
 hpa_hooks_ms_since(nstime_t *past_nstime) {
 	return nstime_ms_since(past_nstime);
 }
+
+
+/* Return true if we purged all nbytes, false otherwise */
```
interwq: nit: the convention is returning true in case of error / unexpected.

src/hpa.c
```diff
@@ -470,14 +498,36 @@ hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
...
+		if (cur == len) {
+			hpa_try_vectorized_purge(shard, vec, len, total_batch_bytes);
+			cur = 0;
+			total_batch_bytes = 0;
+		}
+	}
+
+	/* Batch was not full */
+	if (cur > 0) {
+		hpa_try_vectorized_purge(shard, vec, cur, total_batch_bytes);
```
interwq, line 531: nit: add `assert(total_batch_bytes > 0);`

src/hpa.c
spredolac: Analyzer caught that it should never be zero size. This was the least disruptive way to integrate with the old logic that I could find.

interwq: That would mean we have no way to disable the feature though. Let's make the value 0 as not using the batched madvise.

PR #2822, opened 2025-04-02
Thread events - refactor thread events to make it less macro heavy and support user added events

spredolac:
There are two parts to this PR. First is refactoring existing implementation which is very macro heavy. Second is to allow registration of user's thread events. This would allow implementation of events like prof_threshold outside of jemalloc. See https://github.com/jemalloc/jemalloc/commit/73b8175f45787ef7af720f1300403275c6ea95d5 for one time test comparing these two events.


include/jemalloc/internal/prof_externs.h
```diff
@@ -104,9 +105,60 @@ void prof_prefork1(tsdn_t *tsdn);
...
+ * preparation work for some event must be done early in the allocation /
+ * deallocation calls.
+ *
+ * Currently only the profiling sampling event needs the lookahead
+ * functionality, so we don't yet define general purpose lookahead functions.
+ *
+ * Surplus is a terminology referring to the amount of bytes beyond what's
+ * needed for triggering an event, which can be a useful quantity to have in
+ * general when lookahead is being called.
+ */
```
guangli-dai, line 130: Maybe also clarify in the comments regarding what the returned value means?

include/jemalloc/internal/thread_event_registry.h
```diff
@@ -0,0 +1,98 @@
...
+struct te_alloc_data_s {
+    uint64_t a_wait; /* Every event has alloc wait */
+};
+#define TE_ALLOC_DATA_INIT                                                     \
+    { 0 }
+
+typedef struct te_ad_data_s te_ad_data_t;
+struct te_ad_data_s {
+    te_alloc_data_t a_data; /* Every event has alloc wait. Must be 1st */
+    uint64_t d_wait;
```
guangli-dai: `d_wait` for `dalloc_wait`? While `a_data`'s type explains itself, maybe call it `dalloc_wait` or add a comment to make it more readable?

include/jemalloc/internal/thread_event_table.h
```diff
@@ -0,0 +1,24 @@
+#ifndef JEMALLOC_INTERNAL_THREAD_EVENT_BASE_H
+#define JEMALLOC_INTERNAL_THREAD_EVENT_BASE_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/tsd.h"
+
+typedef struct te_base_cb_s te_base_cb_t;
+struct te_base_cb_s {
+    bool (*condition)(void);
```
guangli-dai: `enabled` sounds more appropriate here?

src/prof.c
```diff
@@ -289,31 +290,53 @@ prof_sample_new_event_wait(tsd_t *tsd) {
...
+	return TE_MAX_START_WAIT;
+#endif
+}
+
+te_base_cb_t prof_sample_te_handler = {
+	.condition = &prof_sample_condition,
+	.new_event_wait = &prof_sample_new_event_wait,
+	/*
+	 * The postponed wait time for prof sample event is computed as if we
+     * want a new wait time (i.e. as if the event were triggered).  If we
```
guangli-dai: nit: should be tab here to align

src/thread_event.c
```diff
@@ -238,18 +188,63 @@ te_adjust_thresholds_helper(tsd_t *tsd, te_ctx_t *ctx,
 	te_ctx_next_event_set(tsd, ctx, next_event);
 }
 
-static uint64_t
-te_clip_event_wait(uint64_t event_wait) {
```
guangli-dai, line 242: Why we don't want to clip the wait anymore?

guangli-dai:
By converting macros to a for loop, will the perf be affected? This is like undoing the "unroll" here. 
While I don't think the perf will be affected, at least not to the degree we care about, just want to confirm it here.

include/jemalloc/internal/thread_event_registry.h
spredolac: Ok, will change to dalloc_wait (also changing a_wait to alloc_wait for symmetry)

include/jemalloc/internal/thread_event_table.h
spredolac: Sounds good

src/thread_event.c
spredolac: That function was just noise, it did not do anything except asserting(event_wait>0) which we still do when we are calculating new value. This was the code of the clip events:
 
       ```
         assert(event_wait > 0U);
	if (TE_MIN_START_WAIT > 1U &&
	    unlikely(event_wait < TE_MIN_START_WAIT)) {
		event_wait = TE_MIN_START_WAIT;
	}
	if (TE_MAX_START_WAIT < UINT64_MAX &&
	    unlikely(event_wait > TE_MAX_START_WAIT)) {
		event_wait = TE_MAX_START_WAIT;
	}
	return event_wait;
```

Given TE_MIN_START_WAIT == ((uint64_t)1U), and TE_MAX_START_WAIT==UINT64_MAX it never clipped anything.

**Comment by spredolac** [comment]
> By converting macros to a for loop, will the perf be affected? This is like undoing the "unroll" here.
> While I don't think the perf will be affected, at least not to the degree we care about, just want to confirm it here.

Ran microbenchmark and it shows no significant move (even slightly better, but within noise). 

I can hit canaries if you think it is needed, but this path does not seem so likely to affect overall perf given that it is in the branch that is throttled, and if the path is taken overhead of event handling itself (such as profiling and statistics) is likely to dominate any potential difference in "unrolled" version (please note that this is not real unroll as we are not talking about same, but rather *similar* body of each macro. The code is more compact now, and it is hard to predict how it will behave - thus, measuring and relying on microbenchmark).

The prime goals here are to allow addition of events without touching the jemalloc repo, and make the thread events code more readable and debuggable (debugging code this macro heavy is not the most pleasant experience)

File: include/jemalloc/internal/prof_externs.h (hunk 1 above)
Comment (spredolac) [reply include/jemalloc/internal/prof_externs.h:130]: Will do. BTW, that code is just moved as it lived in thread_event before, no changes otherwise (except this addition to comment)

**Review by guangli-dai — APPROVED** [review]
Looks good to me. Similarly, let's wait for Travis' results before merging upstream. 

Regarding the perf discussion, I agree that this is less likely to be a concern here. For safety, we can canary during next release with all recent changes and check functional level to confirm no cycle increase.

**Comment by interwq** [comment]
Travis CI caught this, https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/632316411: 

> src/thread_event_table.c:52:1: error: unused function 'te_register_user_handler_locked' [-Werror,-Wunused-function]

File: src/tcache.c [hunk 6]
```diff
@@ -1912,3 +1900,15 @@ tcache_postfork_child(tsdn_t *tsdn) {
 void tcache_assert_initialized(tcache_t *tcache) {
 	assert(!cache_bin_still_zero_initialized(&tcache->bins[0]));
 }
+
+static bool
+tc_enabled(void) {
```
nullptr0-0: nit: the naming here is a bit misleading, it implies tcache enabled but indeed it's tcache_gc_enabled?

src/thread_event.c
```diff
@@ -263,46 +258,9 @@ te_event_trigger(tsd_t *tsd, te_ctx_t *ctx) {
...
-	if (is_alloc == alloc_event && condition &&			\
-	    is_##event##_triggered) {					\
-		assert(allow_event_trigger);				\
-		uint64_t elapsed = event##_fetch_elapsed(tsd);		\
-		event##_event_handler(tsd, elapsed);			\
-	}
-
-	ITERATE_OVER_ALL_EVENTS
-#undef E
+    te_trigger(tsd, ctx, accumbytes, allow_event_trigger, &wait);
```
nullptr0-0: nit: indent

src/thread_event_table.c
```diff
@@ -0,0 +1,31 @@
...
+#include "jemalloc/internal/thread_event_registry.h"
+#include "jemalloc/internal/tcache_externs.h"
+#include "jemalloc/internal/peak_event.h"
+#include "jemalloc/internal/prof_externs.h"
+#include "jemalloc/internal/prof_threshold.h"
+#include "jemalloc/internal/stats.h"
+
+
+/* Table of all the thread events.
+ *  Events share interface, but internally they will know thier
```
nullptr0-0: nit:their

src/thread_event_table.c
```diff
@@ -0,0 +1,31 @@
...
+    { .impl = &prof_sample_te_handler, .doffset = offsetof(te_data_t, prof_sample)},
+#endif
+    { .impl = &stats_interval_te_handler, .doffset = offsetof(te_data_t, stats_interval)},
+#ifdef JEMALLOC_STATS
+    { .impl = &prof_threshold_te_handler, .doffset = offsetof(te_data_t, prof_threshold) },
+#endif
+    { .impl = &tcache_gc_te_handler, .doffset = offsetof(te_data_t, tcache_gc) },
+#ifdef JEMALLOC_STATS
+    { .impl = &peak_te_handler, .doffset = offsetof(te_data_t, peak_data) },
+#endif
```
nullptr0-0: duplicated `#ifdef JEMALLOC_STATS`? could it be merged with the one above?

PR #2827, opened 2025-04-09
Hpa process madvise - perform vectorized purge across multiple huge pages

spredolac:
Support is added to purge across more than one page. Batch is limited with 

a) stack size
b) max number of ranges within one vectorized calls. Once we are above that one, we do not want to add more huge pages to the batch until we are done with this one. This is because while hooks are being called we do not allow allocations on those pages being purged.

The idea is to minimize the number of *madvise syscalls, while at the same time not keep too many huge pages unavailable for allocation.

### DIFF
--- a/include/jemalloc/internal/hpa_utils.h
+++ b/include/jemalloc/internal/hpa_utils.h
@@ -79,4 +79,38 @@ hpa_range_accum_finish(hpa_range_accum_t *ra, hpa_shard_t *shard) {
     }
 }
 
+/*
+ * For purging more than one page we use batch of these items
+ */
+typedef struct {
+	hpdata_purge_state_t state;
+	hpdata_t *hp;
+	bool dehugify;
+} hpa_purge_item_t;
+
+typedef struct hpa_purge_batch_s hpa_purge_batch_t;
+struct hpa_purge_batch_s {
+	hpa_purge_item_t *items;
+	size_t items_capacity;
+	/* Number of huge pages to purge in current batch */
+	size_t item_cnt;
+	/* Number of ranges to purge in current batch */
+	size_t nranges;
+	/* Total number of dirty pages in current batch*/
+	size_t ndirty_in_batch;
+
+	/* Max number of huge pages to purge */
+	size_t max_hp;
+	/*
+	 * Once we are above this watermark we should not add more pages
+	 * to the same batch. This is because while we want to minimize
+	 * number of madvise calls we also do not want to be preventing
+	 * allocations from too many huge pages (which we have to do
+	 * while they are being purged)
+	 */
+	size_t range_watermark;
+
+	size_t npurged_hp_total;
+};
+
 #endif /* JEMALLOC_INTERNAL_HPA_UTILS_H */
--- a/include/jemalloc/internal/hpa_utils.h
+++ b/include/jemalloc/internal/hpa_utils.h
@@ -1,0 +1,82 @@
+#ifndef JEMALLOC_INTERNAL_HPA_UTILS_H
+#define JEMALLOC_INTERNAL_HPA_UTILS_H
+
+#include "jemalloc/internal/hpa.h"
+
+#define HPA_MIN_VAR_VEC_SIZE 8
+#ifdef JEMALLOC_HAVE_PROCESS_MADVISE
+typedef struct iovec hpa_io_vector_t;
+#else
+typedef struct {
+    void *iov_base;
+    size_t iov_len;
+} hpa_io_vector_t;
+#endif
+
+/* Actually invoke hooks. If we fail vectorized, use single purges */
+static void
+hpa_try_vectorized_purge(
+  hpa_shard_t *shard, hpa_io_vector_t *vec, size_t vlen, size_t nbytes) {
+    bool success = opt_process_madvise_max_batch > 0
+      && !shard->central->hooks.vectorized_purge(vec, vlen, nbytes);
+    if (!success) {
+        /* On failure, it is safe to purge again (potential perf
+         * penalty) If kernel can tell exactly which regions
+         * failed, we could avoid that penalty.
+         */
+        for (size_t i = 0; i < vlen; ++i) {
+            shard->central->hooks.purge(vec[i].iov_base, vec[i].iov_len);
+        }
+    }
+}
+
+/*
+ * This struct accumulates the regions for process_madvise.
+ * It invokes the hook when batch limit is reached
+ */
+typedef struct {
+    hpa_io_vector_t *vp;
+    size_t cur;
+    size_t total_bytes;
+    size_t capacity;
+} hpa_range_accum_t;
+
+static inline void
+hpa_range_accum_init(hpa_range_accum_t *ra, hpa_io_vector_t *v, size_t sz) {
+    ra->vp = v;
+    ra->capacity = sz;
+    ra->total_bytes = 0;
+    ra->cur = 0;
+}
+
+static inline void
+hpa_range_accum_flush(hpa_range_accum_t *ra, hpa_shard_t *shard) {
+    assert(ra->total_bytes > 0 && ra->cur > 0);
+    hpa_try_vectorized_purge(shard, ra->vp, ra->cur, ra->total_bytes);
+    ra->cur = 0;
+    ra->total_bytes = 0;
+}
+
+static inline void
+hpa_range_accum_add(
+  hpa_range_accum_t *ra, void *addr, size_t sz, hpa_shard_t *shard) {
+    assert(ra->cur < ra->capacity);
+
+    ra->vp[ra->cur].iov_base = addr;
+    ra->vp[ra->cur].iov_len = sz;
+    ra->total_bytes += sz;
+    ra->cur++;
+
+    if (ra->cur == ra->capacity) {
+        hpa_range_accum_flush(ra, shard);
+    }
+}
+
+static inline void
+hpa_range_accum_finish(hpa_range_accum_t *ra, hpa_shard_t *shard) {
+    if (ra->cur > 0) {
+        hpa_range_accum_flush(ra, shard);
+    }
+}
+
+#endif /* JEMALLOC_INTERNAL_HPA_UTILS_H */
--- a/src/hpa.c
+++ b/src/hpa.c
@@ -2,21 +2,12 @@
 #include "jemalloc/internal/jemalloc_internal_includes.h"
 
 #include "jemalloc/internal/hpa.h"
+#include "jemalloc/internal/hpa_utils.h"
 
 #include "jemalloc/internal/fb.h"
 #include "jemalloc/internal/witness.h"
 
 #define HPA_EDEN_SIZE (128 * HUGEPAGE)
-
-#define HPA_MIN_VAR_VEC_SIZE 8
-#ifdef JEMALLOC_HAVE_PROCESS_MADVISE
-typedef struct iovec hpa_io_vector_t;
-#else
-typedef struct {
-	void *iov_base;
-	size_t iov_len;
-} hpa_io_vector_t;
-#endif
 
 static edata_t *hpa_alloc(tsdn_t *tsdn, pai_t *self, size_t size,
     size_t alignment, bool zero, bool guarded, bool frequent_reuse,
@@ -432,22 +423,12 @@ hpa_shard_has_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard) {
 	return to_hugify != NULL || hpa_should_purge(tsdn, shard);
 }
 
-/* If we fail vectorized purge, we will do single */
-static void
-hpa_try_vectorized_purge(hpa_shard_t *shard, hpa_io_vector_t *vec,
-	size_t vlen, size_t nbytes) {
-	bool success = opt_process_madvise_max_batch > 0
-		&& !shard->central->hooks.vectorized_purge(vec, vlen, nbytes);
-	if (!success) {
-		/* On failure, it is safe to purge again (potential perf
-		 * penalty) If kernel can tell exactly which regions
-		 * failed, we could avoid that penalty.
-		 */
-		for (size_t i = 0; i < vlen; ++i) {
-			shard->central->hooks.purge(vec[i].iov_base,
-				vec[i].iov_len);
-		}
-	}
+static inline size_t
+hpa_process_madvise_max_iovec_len(void) {
+	assert(opt_process_madvise_max_batch <=
+		PROCESS_MADVISE_MAX_BATCH_LIMIT);
+	return opt_process_madvise_max_batch == 0 ?
+		HPA_MIN_VAR_VEC_SIZE : opt_process_madvise_max_batch;
 }
 
 /* Returns whether or not we purged anything. */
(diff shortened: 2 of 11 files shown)


src/hpa.c
```diff
@@ -432,32 +423,72 @@ hpa_shard_has_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard) {
...
+
+	hpa_range_accum_t accum;
+	hpa_range_accum_init(&accum, vec, len);
+
+	for (size_t i = 0; i < batch_sz; ++i) {
+		hpdata_t *to_purge = batch[i].hp;
+
+		/* Actually do the purging, now that the lock is dropped. */
+		if (batch[i].dehugify) {
+			shard->central->hooks.dehugify(hpdata_addr_get(to_purge),
```
guangli-dai, line 475: just curious, can `process_madvise` also handle `dehugify`?

src/hpa.c
```diff
@@ -432,32 +423,72 @@ hpa_shard_has_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard) {
...
-/* Returns whether or not we purged anything. */
-static bool
-hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
-	malloc_mutex_assert_owner(tsdn, &shard->mtx);
-
-	hpdata_t *to_purge = psset_pick_purge(&shard->psset);
+/* Prepare purge of one page. Return num of dirty regular pages on it
+ * Return 0 if no purgable huge page is found
+ *
+ * If there was a page to purge its state is marked as purge_begin
```
guangli-dai: do you mean "at `hpdata_purge_begin`"?

src/hpa.c
```diff
@@ -480,82 +511,111 @@ hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
...
+	b->ndirty_in_batch += ndirty;
+	b->nranges += nranges;
+	return ndirty;
+}
 
-	/* Actually do the purging, now that the lock is dropped. */
-	if (dehugify) {
-		shard->central->hooks.dehugify(hpdata_addr_get(to_purge),
-		    HUGEPAGE);
+/* Finish purge of one page. */
```
guangli-dai: nit: I guess you mean "one hugepage"?

include/jemalloc/internal/hpa_utils.h
```diff
@@ -0,0 +1,122 @@
...
+
+/* Actually invoke hooks. If we fail vectorized, use single purges */
+static void
+hpa_try_vectorized_purge(
+	hpa_shard_t *shard,
+	hpa_io_vector_t *vec,
+	size_t vlen,
+	size_t nbytes) {
+	bool success = opt_process_madvise_max_batch > 0
+		&& !shard->central->hooks.vectorized_purge(vec, vlen, nbytes);
```
guangli-dai: nit: conventionally we use four space for such continuing line. The `if` below should align with line 23.

src/hpa.c
```diff
@@ -432,32 +423,72 @@ hpa_shard_has_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard) {
...
+	return old_size;
+}
+#define HPA_PURGE_BATCH_MAX hpa_purge_max_batch_size_for_test
+#endif
+
+static inline size_t
+hpa_process_madvise_max_iovec_len(void) {
+	assert(opt_process_madvise_max_batch <=
+		PROCESS_MADVISE_MAX_BATCH_LIMIT);
+	return opt_process_madvise_max_batch == 0 ?
```
guangli-dai, line 455: What does `opt_process_madvise_max_batch == 0` mean? Should batched madvise be prevented when this is true?

src/hpa.c
```diff
@@ -480,82 +511,111 @@ hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
...
-	/* Batch was not full */
-	if (cur > 0) {
-		hpa_try_vectorized_purge(shard, vec, cur, total_batch_bytes);
-	}
+static inline bool
+hpa_batch_full(hpa_purge_batch_t *b) {
+	/* It's okay for ranges to go above */
+	return b->npurged_hp_total == b->max_hp ||
+		b->item_cnt == b->items_capacity ||
+		b->nranges >= b->range_watermark;
```
guangli-dai, line 567: Currently, `opt_process_madvise_max_batch` controls `ranges_watermark`, why is that? What about making it control the number of hpdatas instead of nranges? 

I think the original motivation here is that user can tune this option to reach a balance between a) fewer madvises and b) fewer hpdatas available due to the purging in batch? The number of `hpdata` to purge is the main factor of the waiting time instead of the number of ranges to purge since in `hpa_purge_start_hp` we pick one `hpdata` to check each time? Would it be more straightforward for `opt_process_madvise_max_batch` to control number of hpdatas to purge?

nullptr0-0:
nit: there are several places where indentations are incorrect for continuation lines, please check out the [coding style](https://github.com/jemalloc/jemalloc/wiki/Coding-Style) to make sure it mimics existing jemalloc code.

src/hpa.c
```diff
@@ -654,19 +745,10 @@ hpa_shard_maybe_do_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard,
...
-				 */
-				break;
-			}
-			malloc_mutex_assert_owner(tsdn, &shard->mtx);
-			nops++;
-		}
+		malloc_mutex_assert_owner(tsdn, &shard->mtx);
+		nops += hpa_purge(tsdn, shard, max_purges);
+		malloc_mutex_assert_owner(tsdn, &shard->mtx);
+		shard->stats.npurge_passes++;
```
nullptr0-0: nit: move the stats update to the end of hpa_purge() function?

src/hpa.c
```diff
@@ -461,70 +511,111 @@ hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
...
+		.max_hp = max_hp,
+		.npurged_hp_total = 0,
+		.items = &items[0],
+		.items_capacity = HPA_PURGE_BATCH_MAX,
+		.range_watermark = hpa_process_madvise_max_iovec_len(),
+	};
+	assert(batch.range_watermark > 0);
+
+	bool done = false;
+	while (!done) {
```
nullptr0-0: Is this essentially while(1)?

src/hpa.c
```diff
@@ -498,38 +479,24 @@ hpa_try_purge(tsdn_t *tsdn, hpa_shard_t *shard) {
...
-			total_batch_bytes = 0;
-		}
-	}
-
-	/* Batch was not full */
-	if (cur > 0) {
-		hpa_try_vectorized_purge(shard, vec, cur, total_batch_bytes);
 	}
+	/* If batch was not full, flush */
+	hpa_range_accum_flush(&accum, shard);
```
nullptr0-0: nit: hpa_range_accum_finish()?

src/hpa.c
spredolac: Yes, batched madvise is prevented and we are falling into https://github.com/jemalloc/jemalloc/blob/dev/src/hpa.c#L442

src/hpa.c
spredolac: > Currently, opt_process_madvise_max_batch controls ranges_watermark, why is that?

That is proposed by @interwq. So,  `opt_process_madvise_max_batch` is maximum number of ranges you will pass in a single `process_madvise` call. Currently it is bounded by runtime option which cannot be larger than 128 (we do not want more than 2K on stack for that).

There is an option that controls maximum number of pages to do in a single purge: it is `opts.experimental_max_purge_nhp`.

We use `opt_process_madvise_max_batch` to control the `ranges_watermark` exactly as you mentioned, to balance between how long we prevent allocations from huge pages, and how many madvise calls are we making. Algorithm goes something like this:
     
        range_vector rv; 
        
        while(nhp_in_purge <   opts.experimental_max_purge_nhp) {
            mutex_lock
            get_next_hp
            prevent_allocations_from_next_hp
            add_ranges_from_next_hp_to_range_vector 
            if_we_filled_the_range
                  mutex_unlock
                  process_madvise(rv);
                  mutex_lock
                  allow_allocs_on_all_hps_in_the_batch
         }
         
 We do not stop in the middle of the hugepage, so we may be making extra madvise call to flush the leftovers above the watermark.
 
 I do not fully understand your proposal, can we discuss with @interwq ?

src/hpa.c
spredolac: Yes, planned to have that flag set inside the loop on line 603 if the batch was not empty and we are not above watermark, decided not to complicate it, and then forgot to delete it. Will change to while(1). Thanks!

src/hpa.c
guangli-dai: Thanks for the clarificaiton. Let me elaborate my proposal a bit more:
Instead of letting `opt_process_madvise_max_batch` control the maximum number of ranges to process in a single `process_madvise`, can we let it control the maximum number of hpdatas involved in a single `process_madvise`? This would be more straightforward for users to control how long we prevent allocations from hugepages?

Maybe the current design is more for the consistency with the behavior in PAC?

spredolac: I like the proposal, but would like to do it in a separate PR after discussion with Qi as there are implications for pages case as well.

spredolac: After conversation, we will add comment that `opt_process_madvise_max_batch` is really for kernel tuning not app tuning, and we will limit page batch to 16.

src/hpa.c
```diff
@@ -432,32 +423,71 @@ hpa_shard_has_deferred_work(tsdn_t *tsdn, hpa_shard_t *shard) {
...
-		&& !shard->central->hooks.vectorized_purge(vec, vlen, nbytes);
-	if (!success) {
-		/* On failure, it is safe to purge again (potential perf
-		 * penalty) If kernel can tell exactly which regions
-		 * failed, we could avoid that penalty.
-		 */
-		for (size_t i = 0; i < vlen; ++i) {
-			shard->central->hooks.purge(vec[i].iov_base,
-				vec[i].iov_len);
+#define HPA_PURGE_BATCH_MAX_DEFAULT 16
```
guangli-dai, line 436: Maybe write a comment for future reference that the intuition behind this setting and it may need tuning (from our side not user side) when we have more production traffic patterns known?

test/unit/hpa_vectorized_madvise_large_batch.c
```diff
@@ -0,0 +1,200 @@
...
+
+	enum {NALLOCS = 8 * HUGEPAGE_PAGES};
+	edata_t *edatas[NALLOCS];
+	for (int i = 0; i < NALLOCS; i++) {
+		edatas[i] = pai_alloc(tsdn, &shard->pai, PAGE, PAGE, false,
+		    false, false, &deferred_work_generated);
+		expect_ptr_not_null(edatas[i], "Unexpected null edata");
+	}
+	/* Deallocate almost 3 pages out of 8, and to force batching
+	 * leave the 2nd and 4th PAGE in the first 3 hugepages.
```
guangli-dai, line 172: I guess you mean "3 hugepages out of 8"?

test/unit/hpa_vectorized_madvise_large_batch.c
```diff
@@ -0,0 +1,200 @@
...
+
+static uint64_t
+defer_test_ms_since(nstime_t *past_time) {
+	return (nstime_ns(&defer_curtime) - nstime_ns(past_time)) / 1000 / 1000;
+}
+
+TEST_BEGIN(test_vectorized_purge) {
+	printf("opt_process_madvise_max_batch=%zu\n", opt_process_madvise_max_batch);
+	test_skip_if(!hpa_supported() ||
+		opt_process_madvise_max_batch != 64 ||
```
guangli-dai: We set `process_madvise_max_batch:64` in the corresponding `.sh` right? Do we expect it to be overrided anywhere? If not, this should be asserted.

guangli-dai (commented):
Core parts look good to me, just some nits. Somehow travis is not triggered, we should probably wait for its result just for safety. Besides this, I don't think the CI has new enough kernel to really support the `process_madvise`, which means the vectorized purge hook would always fail and fall back to normal purge in tests, is that correct?

test/unit/hpa_vectorized_madvise_large_batch.c
spredolac: Yes, we do and will do.

spredolac:
> Core parts look good to me, just some nits. Somehow travis is not triggered, we should probably wait for its result just for safety. Besides this, I don't think the CI has new enough kernel to really support the `process_madvise`, which means the vectorized purge hook would always fail and fall back to normal purge in tests, is that correct?

Sounds good. 

I am not sure what is the version of the kernel where CI runs, but I tested this on 6.11.1 (our hardened version) and without default hooks things worked fine, syscall is made and return value was as expected.  Check debugger screenshot with the test using default hooks.

Screenshot 2025-04-11 at 3 21 32 PM

guangli-dai (approved):

interwq:
@spredolac Travis CI is now fixed with  #2831. Can you please rebase the PR to the current dev to trigger the CI jobs?

interwq:
The Travis Ci failures are real, e.g. https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/632316293

spredolac:
The failure https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/632347597 could not be reproduced locally with 

`CC=clang CXX=clang++  EXTRA_CFLAGS="-Werror -Wno-array-bounds -Wno-unknown-warning-option -Wno-ignored-attributes" ./autogen.sh --enable-limit-usize-gap`
`make clean && make -j32 && make tests -32 && make check -j32`

It looks like some transient error

spredolac:
https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/632347604 same failure (psset) that is present on dev branch as well and bisected to 4 year old diff.

guangli-dai:
> The failure https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/632347597 could not be reproduced locally with
> 
> `CC=clang CXX=clang++ EXTRA_CFLAGS="-Werror -Wno-array-bounds -Wno-unknown-warning-option -Wno-ignored-attributes" ./autogen.sh --enable-limit-usize-gap` `make clean && make -j32 && make tests -32 && make check -j32`
> 
> It looks like some transient error
>
> https://app.travis-ci.com/github/jemalloc/jemalloc/jobs/632347604 same failure (psset) that is present on dev branch as well and bisected to 4 year old diff.

@spredolac Thank you for looking into this! For the context, the two arm tests have been giving dangling signals. We still cannot confirm if this means a bug in jemalloc or something wrong in the Travis. Double checked and can confirm that all other tests are good. 

I recall seeing an error previously due to misconfigured `opt_process_madvise_max_batch`. Were you able to reproduce it locally?

spredolac:
@guangli-dai ,

> I recall seeing an error previously due to misconfigured opt_process_madvise_max_batch. Were you able to reproduce it locally?

Yes, this was the fix: https://github.com/jemalloc/jemalloc/compare/45baf0dbe26e1ada0d38974389751a533b6d4ed9..b6b86c87b172cb8c7ed3a1d056faae648025685a

When process_madvise is not on at all, even though shell script attempts to set the value 64, the init code ensures it is clipped to 0 so that feature is off.

PR #2847, opened 2025-05-13
Refactor: use the cache_bin_sz_t typedef instead of direct uint16_t

xinydev:
any future changes to the underlying data type for bin sizes (such as upgrading from `uint16_t` to `uint32_t`) can be achieved by modifying only the `cache_bin_sz_t` definition.


interwq:
Thanks for the patch @xinydev !

PR #2851, opened 2025-05-20
[cache_bin] prefetch refilled pointers and add experimental fast_path_prefetch

spredolac:
This change enables two features which are off by default:

1) runtime feature which will prefetch up to `nprefetch_cache_fill` pointers when cache_bin is refilled (if the next pointer in the bin is on the same cacheline as the one prior to it, it will not be prefetched). On some architectures this may be worth cost to pay when it is expected that those pointers will be return for allocations that will come soon. Default value is 0 and this is not enabled.

2) Build time feature `--enable-experimental-fp-prefetch` is added and if enabled it will prefetch the next pointer in the cache_bin when current one is returned to allocation call. To avoid the boundary check, we will pad the cache_bin with one extra space and put the address of that space itself so that prefetch call of the last legit element of the bin still has good address (avoiding larger performance penalty on some platforms)

Beyond unit tests, ran it on internal servers and confirmed that performance when both features were disabled was neutral, while when enabled we notice movement of CPU and TLB misses depending on memory allocation traffic patterns.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1434,6 +1434,22 @@ if test "x$enable_experimental_smallocx" = "x1" ; then
 fi
 AC_SUBST([enable_experimental_smallocx])
 
+dnl Do not enable fastpath prefetch by default.
+AC_ARG_ENABLE([experimental_fp_prefetch],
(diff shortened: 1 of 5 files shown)


spredolac:
We measured this more internally and gave up on it

PR #2859, opened 2026-02-25
configure: add --with-cxx-stdlib option

yuxuanchen1997:
When C++ support is enabled, configure unconditionally probes `-lstdc++` and keeps it in LIBS if the link test succeeds. On platforms using libc++, this probe can succeed at compile time (if libstdc++ headers/libraries happen to be installed) but then cause runtime failures when configure tries to execute test binaries because `libstdc++.so.6` isn't actually available.

Add a `--with-cxx-stdlib=<libstdc++|libcxx>` option that lets the build system specify which C++ standard library to link. When given, the probe is skipped and the specified library is linked directly. When not given, the original probe behavior is preserved.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -324,6 +324,15 @@ fi
 ,
 enable_cxx="1"
 )
+AC_ARG_WITH([cxx_stdlib],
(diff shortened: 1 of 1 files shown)


guangli-dai:
Already merged.

PR #2860, opened 2026-02-28
Fix build with gcc 16.x

bkuhls:
```
src/jemalloc_cpp.cpp: In function 'void* handleOOM(std::size_t, bool)': src/jemalloc_cpp.cpp:90:22: error: '__throw_bad_alloc' is not a member
 of 'std'; did you mean '__throw_bad_cast'?
   90 |                 std::__throw_bad_alloc();
```
https://bugs.gentoo.org/967868

Patch downloaded from
https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/jemalloc/files/jemalloc-5.3.0-dont-call-libstdcxx-internals.patch?id=4b929d70d69b013ea881879d66687b461d812aec

### DIFF
--- a/src/jemalloc_cpp.cpp
+++ b/src/jemalloc_cpp.cpp
@@ -93,7 +93,7 @@ handleOOM(std::size_t size, bool nothrow) {
 	}
 
 	if (ptr == nullptr && !nothrow)
-		std::__throw_bad_alloc();
+		throw std::bad_alloc();
 	return ptr;
 }
 


lexprfuncall:
Hi @bkuhls thanks for the PR!  It looks like you found a genuine issue and I was also able to reproduce this on a FreeBSD system with gcc 16. But, I am not 100% sure that `std::__throw_bad_alloc()` is equivalent to `throw std::bad_alloc()` when compiling with `-fno-execptions`.  The workaround might be as simple as an `#ifdef` that results in calling `std::abort()` when exceptions are disabled but this is a corner of C++ that I am not especially familiar with.  Do you know?  When I am back in the office I'll ask some of the local experts so we can move forward with a fix.

lexprfuncall:
Looks like the gcc commit `62c126db6b6 — "libstdc++: Implement C++26 P3378R2 - constexpr
  exception types"` which reorganized some header files might be to blame.  A minimal fix might be to `#include <memory>`.

lexprfuncall:
Hi @bkuhls I have landed #2900 which should fully address this issue whether or not exceptions are enabled.  Please feel free to open and issue or another PR if you find other issues related to GCC 16.x.

bkuhls:
Thanks, build with gcc 16.1.0-RC-20260424 still works!

PR #2865, opened 2026-03-17
[ci] Update GitHub Actions to their latest major release

gruenich:


guangli-dai:
Thank you @gruenich for the contribution. On a side note, we need to upgrade Github actions because Github is [deprecating Node20](https://github.com/jemalloc/jemalloc/pull/2865) and the upgrade here can keep CI working properly after the deprecation.

guangli-dai (approved):

PR #2866, opened 2026-03-18
Fix typo in prof_log_rep_check: use != instead of || for alloc_count

cuiweixie:
The condition incorrectly used 'alloc_count || 0' which was likely a typo for 'alloc_count != 0'. While both evaluate similarly for the zero/non-zero case, the fix ensures consistency with bt_count and thr_count checks and uses the correct comparison operator.


guangli-dai (approved):
Good catch! Thank you for the contribution!

PR #2869, opened 2026-03-24
Documentation updates

guangli-dai:
Add documentation for the following new mallctl entries:
- opt.debug_double_free_max_scan: double-free detection scan limit
- opt.prof_bt_max: max profiling backtrace depth
- opt.disable_large_size_classes: page-align (instead of size-class align) large allocations
- opt.process_madvise_max_batch: batched purging using process_madvise
- thread.tcache.max: per-thread tcache_max control
- thread.tcache.ncached_max.read_sizeclass: query ncached_max
- thread.tcache.ncached_max.write: set ncached_max per size range
- arena.\<i\>.name: get/set arena names
- arenas.hugepage: queries the hugepage size
- approximate_stats.active: lightweight active bytes estimate

Removed config.prof_frameptr since the feature is still experimental.

### DIFF
--- a/doc/jemalloc.xml.in
+++ b/doc/jemalloc.xml.in
@@ -897,16 +897,6 @@ mallctl("arena." STRINGIFY(MALLCTL_ARENAS_ALL) ".decay",
         during build configuration.</para></listitem>
       </varlistentry>
 
-      <varlistentry id="config.prof_frameptr">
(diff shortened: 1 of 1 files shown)


gctony (approved):
LGTM!

lexprfuncall (approved):

PR #2871, opened 2026-03-26
Fix psset_pick_purge when candidate with index 0 dirtiness is ineligible

spredolac:
Fix psset_pick_purge when candidate with index 0 dirtiness is ineligible

psset_pick_purge used max_bit-- after rejecting a time-ineligible candidate, which caused unnecessary re-scanning of the same bitmap and makes assert fail in debug mode) and a size_t underflow when the lowest-index entry was rejected.  Use max_bit = ind - 1 to skip directly past the rejected index.

Added new unit test that fired the assert before the fix and passes afterwards

### DIFF
--- a/src/psset.c
+++ b/src/psset.c
@@ -417,7 +417,10 @@ psset_pick_purge(psset_t *psset, const nstime_t *now) {
 		if (nstime_compare(tm_allowed, now) <= 0) {
 			return ps;
 		}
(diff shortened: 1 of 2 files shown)


guangli-dai (approved):
The `max_bit = ind - 1` is a great optimization and the correctness is covered by test `test_purge_timing`. LGTM

PR #2873, opened 2026-03-27
Various bug fixes and comments

spredolac:
I used AI to look for bugs. Inspected each one manually and they all seem legit.  Added new unit tests that fail before and pass after the fix.

### DIFF
--- a/include/jemalloc/internal/conf.h
+++ b/include/jemalloc/internal/conf.h
@@ -9,19 +9,15 @@ void malloc_abort_invalid_conf(void);
 
 #ifdef JEMALLOC_JET
 extern bool had_conf_error;
+
 bool conf_next(char const **opts_p, char const **k_p, size_t *klen_p,
     char const **v_p, size_t *vlen_p);
-void conf_error(const char *msg, const char *k, size_t klen,
-    const char *v, size_t vlen);
+void conf_error(
+    const char *msg, const char *k, size_t klen, const char *v, size_t vlen);
 bool conf_handle_bool(const char *v, size_t vlen, bool *result);
-bool conf_handle_unsigned(const char *v, size_t vlen,
-    uintmax_t min, uintmax_t max, bool check_min, bool check_max,
-    bool clip, uintmax_t *result);
-bool conf_handle_signed(const char *v, size_t vlen,
-    intmax_t min, intmax_t max, bool check_min, bool check_max,
-    bool clip, intmax_t *result);
-bool conf_handle_char_p(const char *v, size_t vlen,
-    char *dest, size_t dest_sz);
+bool conf_handle_signed(const char *v, size_t vlen, intmax_t min, intmax_t max,
+    bool check_min, bool check_max, bool clip, intmax_t *result);
+bool conf_handle_char_p(const char *v, size_t vlen, char *dest, size_t dest_sz);
 #endif
 
 #endif /* JEMALLOC_INTERNAL_CONF_H */
--- a/include/jemalloc/internal/tcache_inlines.h
+++ b/include/jemalloc/internal/tcache_inlines.h
@@ -163,10 +163,10 @@ tcache_alloc_large(tsd_t *tsd, arena_t *arena, tcache_t *tcache, size_t size,
 			assert(usize <= tcache_max_get(tcache->tcache_slow));
 			memset(ret, 0, usize);
 		}
+	}
 
-		if (config_stats) {
-			bin->tstats.nrequests++;
-		}
+	if (config_stats) {
+		bin->tstats.nrequests++;
 	}
 
 	return ret;
--- a/src/conf.c
+++ b/src/conf.c
@@ -254,36 +254,8 @@ JEMALLOC_DIAGNOSTIC_PUSH
 JEMALLOC_DIAGNOSTIC_IGNORE("-Wunused-function")
 
 JET_EXTERN bool
-conf_handle_unsigned(const char *v, size_t vlen,
-    uintmax_t min, uintmax_t max, bool check_min, bool check_max,
-    bool clip, uintmax_t *result) {
-	char *end;
-	set_errno(0);
-	uintmax_t mv = (uintmax_t)malloc_strtoumax(v, &end, 0);
-	if (get_errno() != 0 || (uintptr_t)end - (uintptr_t)v != vlen) {
-		return true;
-	}
-	if (clip) {
-		if (check_min && mv < min) {
-			*result = min;
-		} else if (check_max && mv > max) {
-			*result = max;
-		} else {
-			*result = mv;
-		}
-	} else {
-		if ((check_min && mv < min) || (check_max && mv > max)) {
-			return true;
-		}
-		*result = mv;
-	}
-	return false;
-}
-
-JET_EXTERN bool
-conf_handle_signed(const char *v, size_t vlen,
-    intmax_t min, intmax_t max, bool check_min, bool check_max,
-    bool clip, intmax_t *result) {
+conf_handle_signed(const char *v, size_t vlen, intmax_t min, intmax_t max,
+    bool check_min, bool check_max, bool clip, intmax_t *result) {
 	char *end;
 	set_errno(0);
 	intmax_t mv = (intmax_t)malloc_strtoumax(v, &end, 0);
@@ -309,6 +281,9 @@ conf_handle_signed(const char *v, size_t vlen,
 
 JET_EXTERN bool
 conf_handle_char_p(const char *v, size_t vlen, char *dest, size_t dest_sz) {
+	if (dest_sz == 0) {
+		return false;
+	}
 	size_t cpylen = (vlen <= dest_sz - 1) ? vlen : dest_sz - 1;
 	strncpy(dest, v, cpylen);
 	dest[cpylen] = '\0';
@@ -473,11 +448,11 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			continue;
 		}
 
-		while (*opts != '\0'
-		    && !conf_next(&opts, &k, &klen, &v, &vlen)) {
+		while (
+		    *opts != '\0' && !conf_next(&opts, &k, &klen, &v, &vlen)) {
 #define CONF_ERROR(msg, k, klen, v, vlen)                                      \
 	if (!initial_call) {                                                   \
-		conf_error(msg, k, klen, v, vlen);                      \
+		conf_error(msg, k, klen, v, vlen);                             \
 		cur_opt_valid = false;                                         \
 	}
 #define CONF_CONTINUE                                                          \
--- a/src/extent.c
+++ b/src/extent.c
@@ -916,15 +916,20 @@ extent_try_coalesce_impl(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 		size_t   max_next_neighbor = max_size > edata_size_get(edata)
 		      ? max_size - edata_size_get(edata)
 		      : 0;
-		if (next != NULL && edata_size_get(next) <= max_next_neighbor) {
-			if (!extent_coalesce(
-			        tsdn, pac, ehooks, ecache, edata, next, true)) {
-				if (ecache->delay_coalesce) {
-					/* Do minimal coalescing. */
-					*coalesced = true;
-					return edata;
+		if (next != NULL) {
+			if (edata_size_get(next) > max_next_neighbor) {
+				emap_release_edata(
+				    tsdn, pac->emap, next, ecache->state);
+			} else {
+				if (!extent_coalesce(tsdn, pac, ehooks, ecache,
+				        edata, next, true)) {
+					if (ecache->delay_coalesce) {
+						/* Do minimal coalescing. */
+						*coalesced = true;
+						return edata;
+					}
+					again = true;
 				}
-				again = true;
 			}
 		}
 
@@ -934,16 +939,21 @@ extent_try_coalesce_impl(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 		size_t   max_prev_neighbor = max_size > edata_size_get(edata)
 		      ? max_size - edata_size_get(edata)
 		      : 0;
-		if (prev != NULL && edata_size_get(prev) <= max_prev_neighbor) {
-			if (!extent_coalesce(tsdn, pac, ehooks, ecache, edata,
-			        prev, false)) {
-				edata = prev;
-				if (ecache->delay_coalesce) {
-					/* Do minimal coalescing. */
-					*coalesced = true;
-					return edata;
+		if (prev != NULL) {
+			if (edata_size_get(prev) > max_prev_neighbor) {
+				emap_release_edata(
+				    tsdn, pac->emap, prev, ecache->state);
+			} else {
+				if (!extent_coalesce(tsdn, pac, ehooks, ecache,
+				        edata, prev, false)) {
+					edata = prev;
+					if (ecache->delay_coalesce) {
+						/* Do minimal coalescing. */
+						*coalesced = true;
+						return edata;
+					}
+					again = true;
 				}
-				again = true;
 			}
 		}
 	} while (again);
--- a/src/extent_dss.c
+++ b/src/extent_dss.c
@@ -153,11 +153,14 @@ extent_alloc_dss(tsdn_t *tsdn, arena_t *arena, void *new_addr, size_t size,
 			    - (uintptr_t)gap_addr_page;
 			if (gap_size_page != 0) {
 				edata_init(gap, arena_ind_get(arena),
-				    gap_addr_page, gap_size_page, false,
-				    SC_NSIZES,
+				    gap_addr_page, gap_size_page,
+				    /* slab */ false,
+				    /* szind */ SC_NSIZES,
 				    extent_sn_next(&arena->pa_shard.pac),
-				    extent_state_active, false, true,
-				    EXTENT_PAI_PAC, head_state);
+				    extent_state_active,
+				    /* zeroed */ false,
+				    /* committed */ true,
+				    /* pai */ EXTENT_PAI_PAC, head_state);
 			}
 			/*
 			 * Compute the address just past the end of the desired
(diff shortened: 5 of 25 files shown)


src/conf.c
```diff
@@ -254,9 +254,11 @@ JEMALLOC_DIAGNOSTIC_PUSH
 JEMALLOC_DIAGNOSTIC_IGNORE("-Wunused-function")
 
 JET_EXTERN bool
-conf_handle_unsigned(const char *v, size_t vlen,
-    uintmax_t min, uintmax_t max, bool check_min, bool check_max,
-    bool clip, uintmax_t *result) {
+conf_handle_unsigned(const char *v, size_t vlen, uintmax_t min, uintmax_t max,
```
guangli-dai: A question regarding these function: it seems they are not used anywhere, we have macros handling these parsings like for [char](http://github.com/jemalloc/jemalloc/blob/a87c518babfe81395a63b6b023245d8359ca1b96/src/conf.c#L567) and for [unsigned](https://github.com/jemalloc/jemalloc/blob/a87c518babfe81395a63b6b023245d8359ca1b96/src/conf.c#L555). Should we just remove the dead codes?

src/malloc_io.c
```diff
@@ -215,9 +215,12 @@ malloc_strtoumax(const char *restrict nptr, char **restrict endptr, int base) {
 			}
 			break;
 		default:
-			p++;
-			ret = 0;
-			goto label_return;
+			if (b == 0) {
```
guangli-dai: The fix looks about right, but again a question regarding the usage. When setting configs or doing mallctls on a specific arena, do users really expect an auto detection on the base? For some configs maybe yes, like `tcache_max` maybe a hex value is easier than 16384, but for specifying arena id, users may not expect 010 to be an octal? Moreover, this is not mentioned in the manual.

My take here is: don't have to solve this problem now. But if we are going to further refactor `conf.c`, i.e., instead of using the macro, use functions, this number parsing semantics should be considered and probably updated in the manual then as well.

src/conf.c
spredolac: Let's make extensive separate pass for dead code.

src/large.c
```diff
@@ -147,7 +147,7 @@ large_ralloc_no_move(tsdn_t *tsdn, edata_t *edata, size_t usize_min,
 		}
 		/* Try again, this time with usize_min. */
 		if (usize_min < usize_max && usize_min > oldusize
-		    && large_ralloc_no_move_expand(
```
guangli-dai, line 150: nit: commit message would be better if specifying this impacts both failed and successful cases for `large_ralloc_no_move` in case someone is bothered by the negative impact on perf here: On expansion success, the bug leads to falsely reported failure, causing callers to unncessarily allocate, copy, and free.

src/extent_dss.c
```diff
@@ -153,11 +153,14 @@ extent_alloc_dss(tsdn_t *tsdn, arena_t *arena, void *new_addr, size_t size,
 			    - (uintptr_t)gap_addr_page;
 			if (gap_size_page != 0) {
 				edata_init(gap, arena_ind_get(arena),
-				    gap_addr_page, gap_size_page, false,
```
guangli-dai, line 156: Not for this PR, but as a future task: the dss alloc path is barely covered by tests, which should be improved.

test/unit/hpa_stats_json.c
```diff
@@ -0,0 +1,108 @@
+#include "test/jemalloc_test.h"
```
guangli-dai: We have a json stats test added [here](https://github.com/jemalloc/jemalloc/blob/dev/test/unit/json_stats.c), are we adding another json stats test just because we need to set the MALLOC_CONF here?

src/conf.c
guangli-dai: nit: the commit message doesn't mention the fix for `unsigned`

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -163,10 +163,10 @@ tcache_alloc_large(tsd_t *tsd, arena_t *arena, tcache_t *tcache, size_t size,
 			assert(usize <= tcache_max_get(tcache->tcache_slow));
```
guangli-dai, line 163: nit: the commit message looks a bit ambiguous, I thought we were adding a new counter to know the cache misses on large allocations at the first glance. Might be better to say "Fix large alloc nrequests under-counting on cache misses"?

src/stats.c
```diff
@@ -1107,8 +1107,8 @@ stats_arena_hpa_shard_slabs_print(emitter_t *emitter, unsigned i) {
 		    &nactive_nonhuge);
 		emitter_json_kv(emitter, "ndirty_nonhuge", emitter_type_size,
 		    &ndirty_nonhuge);
-		emitter_json_kv(emitter, "nretained_nonhuge",
```
guangli-dai: I would suggest squashing this commit with the previous one like how other commits in this PR are doing formatting.

test/unit/hpa_stats_json.c
spredolac: will take a look, thx

src/conf.c
spredolac: I actually removed 'conf_handle_unsigned' as it created lot of noise in the tests. Will do it in the commit where it was fixed (instead of fixing it, will just remove it).

src/malloc_io.c
spredolac: Will skip this fix for now.

test/unit/mallctl.c
```diff
@@ -956,6 +956,52 @@ TEST_BEGIN(test_arenas_bin_constants) {
...
+	/*
+	 * Querying the lextent at index nlextents should fail because valid
+	 * indices are [0, nlextents).
+	 */
+	sz = sizeof(result);
+	malloc_snprintf(buf, sizeof(buf), "arenas.lextent.%u.size", nlextents);
+	expect_d_eq(mallctl(buf, (void *)&result, &sz, NULL, 0), ENOENT,
+	    "mallctl() should fail for out-of-bounds lextent index");
+
+	/* One below the boundary should succeed. */
```
lexprfuncall: I know it's a little redundant but I might phrase this as
```
    /* One below the boundary (nlextents - 1) should succeed. */
```
since boundary is kind of ambiguous.  You could even say 
```
    /* Querying the last enlement (nlextents - 1) should succeed. */
```
to connect it back to the previous comment.

test/unit/large_ralloc.c
```diff
@@ -0,0 +1,75 @@
+#include "test/jemalloc_test.h"
+
+/*
+ * Test that large_ralloc_no_move returns true (failure) when in-place
+ * expansion cannot succeed for either usize_max or usize_min.
```
lexprfuncall: `causes a failure (returns true) when in-place` would be better to connect this paragraph with the one below.

It would also be good to include some context so the reader knows what the "expansion" the comment is talking about refers to.  (Expansion of what?  Extents?)

test/unit/large_ralloc.c
```diff
@@ -0,0 +1,75 @@
...
+
+	int flags = MALLOCX_ARENA(arena_ind) | MALLOCX_TCACHE_NONE;
+
+	size_t large_sz = SC_LARGE_MINCLASS;
+	/* Allocate several blocks to prevent expansion of the first. */
+	void *blocks[8];
+	for (int i = 0; i < 8; i++) {
+		blocks[i] = mallocx(large_sz, flags);
+		expect_ptr_not_null(blocks[i], "Unexpected mallocx() failure");
+	}
```
lexprfuncall, line 30: We don't have an `ARRAY_SIZE` macro (or equivalent)?  If not
```
#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
```
would be a good addition to "jemalloc/internal/util.h"

Then, line 26 becomes
```
    for (int i = 0; i < ARRAY_SIZE(blocks); i++) {
```

src/malloc_io.c
```diff
@@ -692,7 +692,7 @@ malloc_vsnprintf(char *str, size_t size, const char *format, va_list ap) {
 label_out:
 	if (i < size) {
 		str[i] = '\0';
-	} else {
+	} else if (size > 0) {
```
lexprfuncall: `size` is an unsigned integer so this is just checking that `size != 0`.  That might be the better test.

src/pac.c
```diff
@@ -198,7 +198,9 @@ pac_alloc_real(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, size_t size,
 		edata = ecache_alloc_grow(tsdn, pac, ehooks,
 		    &pac->ecache_retained, NULL, size, alignment, zero,
 		    guarded);
-		newly_mapped_size = size;
+		if (edata != NULL) {
+			newly_mapped_size = size;
+		}
 	}
 
 	if (config_stats && newly_mapped_size != 0) {
```
lexprfuncall, line 206: if `edata` is null is there any way for `newly_mapped_size` to be non-zero?

src/pages.c
```diff
@@ -718,7 +718,7 @@ os_page_detect(void) {
 #else
```
lexprfuncall, line 718: Just an aside but the FreeBSD code and the sysconf below are now equivalent

https://cgit.freebsd.org/src/tree/lib/libc/gen/sysconf.c#n293

so there's really no benefit to the extra code since this commit from 2018

https://cgit.freebsd.org/src/commit/lib/libc/gen/sysconf.c?id=3a9cdc513257af5364923bd027aec0fa27a337cd

src/prof_stack_range.c
```diff
@@ -99,7 +99,7 @@ prof_mapping_containing_addr(uintptr_t addr, const char *maps_path,
 			    remaining); /* copy remaining characters to start of buf */
 			line = buf;
 
-			size_t count = malloc_read_fd(
+			ssize_t count = malloc_read_fd(
 			    fd, buf + remaining, sizeof(buf) - remaining);
 			if (count <= 0) {
```
lexprfuncall: I don't believe `errno` is set with a return value of 0 so `ret` will be garbage.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/read.html

src/extent.c
```diff
@@ -916,15 +916,20 @@ extent_try_coalesce_impl(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
+						*coalesced = true;
+						return edata;
+					}
+					again = true;
 				}
-				again = true;
+			} else {
+				emap_release_edata(tsdn, pac->emap, next,
+				    ecache->state);
 			}
```
lexprfuncall, line 933: Up to you but I find it easier to read when the shorter case is first

```
		if (next != NULL) {
			if (edata_size_get(next) > max_next_neighbor) {
				emap_release_edata(tsdn, pac->emap, next,
				    ecache->state);
            } else {
				if (!extent_coalesce(tsdn, pac, ehooks, ecache,
				        edata, next, true)) {
					if (ecache->delay_coalesce) {
						/* Do minimal coalescing. */
						*coalesced = true;
						return edata;
					}
				}
				again = true;
			}
```

src/extent.c
```diff
@@ -934,16 +939,21 @@ extent_try_coalesce_impl(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
...
+					if (ecache->delay_coalesce) {
+						/* Do minimal coalescing. */
+						*coalesced = true;
+						return edata;
+					}
+					again = true;
 				}
-				again = true;
+			} else {
+				emap_release_edata(tsdn, pac->emap, prev,
```
lexprfuncall: If you flip around the condition above, do it here as well.

src/sz.c
```diff
@@ -93,7 +93,7 @@ sz_boot_size2index_tab(const sc_data_t *sc_data) {
 	size_t dst_max = (SC_LOOKUP_MAXCLASS >> SC_LG_TINY_MIN) + 1;
 	size_t dst_ind = 0;
 	for (unsigned sc_ind = 0; sc_ind < SC_NSIZES && dst_ind < dst_max;
-	     sc_ind++) {
```
lexprfuncall, line 96: Isn't the old version correctly formatted?  It should be one space past the ( I think.

test/unit/san_bump.c
```diff
@@ -4,6 +4,51 @@
...
+static extent_hooks_t *san_bump_default_hooks;
+static extent_hooks_t  san_bump_hooks;
+static bool            fail_retained_alloc;
+static unsigned        retained_alloc_fail_calls;
+
+static void *
+san_bump_fail_alloc_hook(extent_hooks_t *extent_hooks, void *new_addr,
+    size_t size, size_t alignment, bool *zero, bool *commit,
+    unsigned arena_ind) {
+	(void)extent_hooks;
```
lexprfuncall: We have an UNUSED macro for this in jemalloc_internal_macros.h.  Much cleaner than cast to void since you'd just annotate the argument.

include/jemalloc/internal/conf.h
```diff
@@ -9,19 +9,14 @@ void malloc_abort_invalid_conf(void);
...
 #ifdef JEMALLOC_JET
 extern bool had_conf_error;
-bool conf_next(char const **opts_p, char const **k_p, size_t *klen_p,
-    char const **v_p, size_t *vlen_p);
-void conf_error(const char *msg, const char *k, size_t klen,
-    const char *v, size_t vlen);
+bool        conf_next(char const **opts_p, char const **k_p, size_t *klen_p,
+           char const **v_p, size_t *vlen_p);
+void        conf_error(
+           const char *msg, const char *k, size_t klen, const char *v, size_t vlen);
```
lexprfuncall: This looks wrong?  It's treating this like a variable not a prototype.

test/unit/conf_parse.c
```diff
@@ -25,73 +25,28 @@ TEST_BEGIN(test_conf_handle_bool_invalid) {
...
-	    1, 2048, true, true, true, &result);
-	expect_true(err, "Should fail for non-numeric input");
-}
-TEST_END
-
 TEST_BEGIN(test_conf_handle_signed_valid) {
 	intmax_t result = 0;
-	bool err = conf_handle_signed("5000", sizeof("5000") - 1,
-	    -1, INTMAX_MAX, true, false, false, &result);
+	bool     err = conf_handle_signed("5000", sizeof("5000") - 1, -1,
```
lexprfuncall, line 30: Nothing changed here we probably want to leave this alone.  Same for the stuff below.

test/unit/tcache_max.c
```diff
@@ -195,9 +195,9 @@ TEST_BEGIN(test_tcache_max) {
 
 	global_test = true;
 	for (alloc_option = alloc_option_start; alloc_option < alloc_option_end;
-	     alloc_option++) {
+	    alloc_option++) {
```
lexprfuncall, line 198: This looks like the same formatting mistake I saw in an earlier commit?

test/unit/tcache_max.c
```diff
@@ -318,9 +361,9 @@ tcache_check(void *arg) {
 	expect_zu_eq(tcache_nbins, tcache_max2nbins(new_tcache_max),
 	    "Unexpected value for tcache_nbins");
 	for (unsigned alloc_option = alloc_option_start;
-	     alloc_option < alloc_option_end; alloc_option++) {
```
lexprfuncall, line 321: Here, too.

test/unit/tcache_max.c
```diff
@@ -206,6 +206,49 @@ TEST_BEGIN(test_tcache_max) {
...
+	    mallctl("arenas.lextent.0.size", (void *)&large, &sz, NULL, 0), 0,
+	    "Unexpected mallctl() failure");
+	expect_d_eq(mallctl("thread.tcache.max", NULL, NULL, (void *)&large,
+	                sizeof(large)),
+	    0, "Unexpected mallctl() failure");
+	expect_d_eq(mallctl("thread.tcache.flush", NULL, NULL, NULL, 0), 0,
+	    "Unexpected tcache flush failure");
+
+	tsd_t    *tsd = tsd_fetch();
+	tcache_t *tcache = tcache_get(tsd);
```
lexprfuncall: I'm not sure why it wants to use a variable for tsd but since this is a test we should expect it to be non-null.

test/unit/json_stats.c
```diff
@@ -201,6 +201,22 @@ json_find_object_end(const char *object_begin) {
...
 
+static const char *
+json_find_array_end(const char *array_begin) {
+	int depth = 0;
+	for (const char *cur = array_begin; *cur != '\0'; cur++) {
+		if (*cur == '[') {
+			depth++;
+		} else if (*cur == ']') {
+			depth--;
+			if (depth == 0) {
```
lexprfuncall, line 215: Bad things will happen if you have a ']' without a '[' before it.

src/pac.c
spredolac: No. See Line 134 where variable is initialized to 0 and then mentioned on lines 202 and 193 both time guarded with if `edata != NULL`

src/pages.c
spredolac: Let's deal with that separately

src/prof_stack_range.c
spredolac: Yes, you are right. The bug was there already in both cases (0 and -1), bit this fix is not proper as it only handles negative.  Based on this doc 0 should mean EOF (for STREAMs it is

> In addition, read() shall fail if the STREAM head had processed an asynchronous error before the call. In this case, the value of errno shall not reflect the result of read(), but reflect the prior error. 
 
I will just break in the case of 0 and fix in few more spots in this function. It looks like we do not have enough to parse the line, so we cannot tell which range address belongs to, thus `ENOENT` should be fine to return.

src/sz.c
spredolac: I am doing `clang-format -i src/sz.c` and it is using the file we have in the root. Still shows me no change.  If that is the case, we need to fix https://github.com/jemalloc/jemalloc/blob/dev/.clang-format

include/jemalloc/internal/conf.h
spredolac: I see what happened. It aligns `conf_next` with the name above `had_conf_error`. If I have a blank line between then clang-format is stable. If we want it to work same we need to tweak .clang-format file.

test/unit/conf_parse.c
spredolac: Original file is not format clean.  I will make separate commit with just clang-format before this one.

test/unit/tcache_max.c
spredolac: Same stuff, clang-format did not run on original file. Will put separate commit, so that it is easy to see the change

guangli-dai (approved):

PR #2879, opened 2026-04-03
Remove hpa_sec_batch_fill_extra and calculate nallocs dynamically.

gctony:
This change includes the following improvements:
* Remove the hpa_sec_batch_fill_extra parameter.
* Refactor the hpa_alloc() code and helper functions to be able to allocate more than one extent out of a single pageslab. This way we can amortize the per-pageslab costs (active bitmap iteration, pageslab metadata updates) across multiple extents.
* Decide on a min and max number of extents that will be allocated in hpa_alloc(). The code will try to allocate at least the min and allocate up to the max as long as we can allocate additional ones from the pageslab we already have, as additional allocations are relatively cheap.
* Add extent allocation distribution stats.
* Amend hpa_sec_integration.c unit test.

### DIFF
--- a/include/jemalloc/internal/hpa.h
+++ b/include/jemalloc/internal/hpa.h
@@ -50,6 +50,60 @@ struct hpa_shard_nonderived_stats_s {
 	 * Guarded by mtx.
 	 */
 	uint64_t ndehugifies;
+
+	/*
+	 * Distribution of the min number of extents we will try to allocate
+	 * from a single hpa_alloc() call.
+	 *
+	 * Guarded by mtx.
+	 */
+	uint64_t hpa_alloc_min_extents[SEC_MAX_NALLOCS + 1];
+
+	/*
+	 * Distribution of the max number of extents we will try to allocate
+	 * from a single hpa_alloc() call.
+	 *
+	 * Guarded by mtx.
+	 */
+	uint64_t hpa_alloc_max_extents[SEC_MAX_NALLOCS + 1];
+
+	/*
+	 * Distribution of the number of extents allocated for a single
+	 * hpa_alloc() call and a single mtx lock acquisition.
+	 *
+	 * Guarded by mtx.
+	 */
+	uint64_t hpa_alloc_extents[SEC_MAX_NALLOCS + 1];
+
+	/*
+	 * Distribution of the number of ps out of which we allocated extents
+	 * for a single hpa_alloc() call and a single mtx lock acquisition.
+	 *
+	 * Guarded by mtx.
+	 */
+	uint64_t hpa_alloc_ps[SEC_MAX_NALLOCS + 1];
+
+	/*
+	 * Distribution of the number of pages allocated from a single ps.
+	 *
+	 * Guarded by mtx.
+	 */
+	uint64_t hpa_alloc_pages_per_ps[SEC_MAX_NALLOCS + 1];
+
+	/*
+	 * Distribution of the number of extents allocated from a single ps.
+	 *
+	 * Guarded by mtx.
+	 */
+	uint64_t hpa_alloc_extents_per_ps[SEC_MAX_NALLOCS + 1];
+
+	/*
+	 * Distribution of the total elapsed time (ns) for allocating extents
+	 * from a single ps.
+	 *
+	 * Guarded by mtx.
+	 */
+	uint64_t hpa_alloc_total_elapsed_ns_per_ps[SEC_MAX_NALLOCS + 1];
 };
 
 /* Completely derived; only used by CTL. */
--- a/include/jemalloc/internal/hpdata.h
+++ b/include/jemalloc/internal/hpdata.h
@@ -433,6 +433,60 @@ void *hpdata_reserve_alloc(hpdata_t *hpdata, size_t sz);
 void  hpdata_unreserve(hpdata_t *hpdata, void *addr, size_t sz);
 
 /*
+ * For buffering extent allocations we will perform out of
+ * a single ps.
+ */
+typedef struct hpdata_alloc_offset_s hpdata_alloc_offset_t;
+struct hpdata_alloc_offset_s {
+	/*
+	 * Index on the active bitmap for the extent to allocate.
+	 * It is used to know which bits we'll need to set when we perform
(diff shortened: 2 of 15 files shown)


include/jemalloc/internal/hpa.h
```diff
@@ -51,6 +51,60 @@ struct hpa_shard_nonderived_stats_s {
...
+	 */
+	uint64_t hpa_alloc_extents_per_ps[SEC_MAX_NALLOCS + 1];
+
+	/*
+	 * Distribution of the total elapsed time (ns) for allocating extents
+	 * from a single ps.
+	 *
+	 * Guarded by mtx.
+	 */
+	uint64_t hpa_alloc_total_elapsed_ns_per_ps[SEC_MAX_NALLOCS + 1];
```
guangli-dai, line 106: These are very good stats used for debugging, e.g., to identify if the current min/max setup for singel-pageslab extent retrieval is well set. However, those two are constants so these stats seem to be helpful only in terms of dev debugging. Are they worthy exposed to wider range of users? We might want to minimize the stats kept here.

include/jemalloc/internal/hpdata.h
```diff
@@ -432,6 +432,63 @@ void hpdata_init(hpdata_t *hpdata, void *addr, uint64_t age, bool is_huge);
...
+	 * [0, index + len_before), assuming the bits in the range
+	 * [index, index + npages) were set. This means that
+	 * longest_len is only valid for the index that corresponds
+	 * to the last successful extent allocation.
+	 * When we need to update the longest free range on the hpdata,
+	 * the new value is either longest_len or the longest free range
+	 * in the range [index + len_before, HUGEPAGE_PAGES),
+	 * whichever is greater.
+	 */
+	size_t longest_len;
```
guangli-dai, line 467: Can it be further simplified?

include/jemalloc/internal/util.h
```diff
@@ -6,6 +6,13 @@
 
 #define UTIL_INLINE static inline
 
+#ifndef MIN
+#	define MIN(a, b) (((a) < (b)) ? (a) : (b))
```
guangli-dai: MIN/MAX can easily pollute the namespace. The linking order could vary. I would suggest at least change the name to `JEMALLOC_MIN` and `JEMALLOC_MAX`.

Besides, for cases like `MIN(a++, b)`, the macro would potentially evaluate `a++` twice? New `__typeof__` should be able to handle that but it seems pretty new (C23). Making this a function also has problems in accepting different types. I don't have a good solution for that so maybe we can keep this part for now and just change the naming.

src/conf.c
```diff
@@ -952,9 +952,6 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			CONF_HANDLE_SIZE_T(opt_hpa_sec_opts.max_bytes,
 			    "hpa_sec_max_bytes", SEC_OPTS_MAX_BYTES_DEFAULT, 0,
 			    CONF_CHECK_MIN, CONF_DONT_CHECK_MAX, true);
-			CONF_HANDLE_SIZE_T(opt_hpa_sec_opts.batch_fill_extra,
```
guangli-dai, line 955: Can add it to the [deprecated list](https://github.com/jemalloc/jemalloc/blob/dev/src/conf.c#L228) for now as well.

include/jemalloc/internal/hpa.h
guangli-dai: Besides, for those kept, we should add unit tests for them.

src/hpa.c
```diff
@@ -715,60 +721,145 @@ hpa_try_alloc_one_no_grow(
...
+	nstime_t start;
+	nstime_init_update(&start);
+
+	hpdata_t *ps = psset_pick_alloc(&shard->psset, size);
+	if (ps == NULL) {
+		return 0;
+	}
+
+	assert(max_nallocs <= SEC_MAX_NALLOCS);
+	hpdata_alloc_offset_t alloc_offsets[SEC_MAX_NALLOCS];
```
guangli-dai, line 728: Can you provide some data to justify why `SEC_MAX_NALLOCS` is set to 4 not something higher like 8?

include/jemalloc/internal/hpa.h
gctony: Agreed. They are very helpful for debugging but maybe not helpful to most users. But it should be good to keep them for future projects.

include/jemalloc/internal/hpdata.h
gctony: I think I simplified the approach a bit and the invariant of longest_len is now a bit simpler. It now keeps track of the longest len up to index.

include/jemalloc/internal/util.h
gctony: Not sure what the right approach is. If it was C++, I'd use a template. I can use a function instead of a macro, but its name has to reflect the type. I guess there are cases of that already in the code base: `expect_u_eq`, `expect_zu_eq`, etc. I guess I can do that? But I should add that we should add something, as having min/max functions is very helpful!

src/conf.c
gctony: Damn. I had added it but I think I somehow lost that change when rebasing. Thanks for pointing that out! I'll fix that.

src/hpa.c
gctony: Good question! I'm actually rerunning with min=2 and max=8. Will report back.

gctony:
@guangli-dai Thanks for the comments. I fixed the following:

* added `hpa_sec_batch_fill_extra` to the deprecated list
* replaced the `MIN`/`MAX` macros with `min_zu`/`max_zu` functions
* simplified the invariant of the `longest_len` field

Evaluation of max of 8 instead of 4 is ongoing.

include/jemalloc/internal/util.h
guangli-dai: The current `min_zu` and `max_zu` LGTM.

guangli-dai:
The evaluation of the specific value for max is still ongoing. Merging this for now and will open another PR if we conclude another better value for the max.

PR #2885, opened 2026-04-15
Change permission modes of static libraries to 644

Integral-Tech:
Static libraries are archives and do not need the executable bit.


lexprfuncall (approved):

lexprfuncall:
@Integral-Tech thanks for the PR!  I've gone ahead and merged it as-is.

PR #2887, opened 2026-04-17
Update image used in appveyor

guangli-dai:
VS 14 no longer supported. Update appveyor to use VS 2022

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,3 +1,5 @@
+image: Visual Studio 2022
+
 version: '{build}'
 
 environment:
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -37,9 +39,11 @@ environment:
 
 install:
   - set PATH=c:\msys64\%MSYSTEM%\bin;c:\msys64\usr\bin;%PATH%
-  - if defined MSVC call "c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" %MSVC%
+  - if "%MSVC%"=="amd64" call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
+  - if "%MSVC%"=="x86"   call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars32.bat"
   - if defined MSVC pacman --noconfirm -Rsc mingw-w64-%CPU%-gcc gcc
   - pacman --noconfirm -Syuu
+  - pacman --noconfirm -Syuu
```
lexprfuncall: This line looks like a duplicate of the line above.  Does it need to be executed twice?

lexprfuncall (approved):

.appveyor.yml
guangli-dai: Although msys [doc](https://www.msys2.org/docs/ci/) says we need it, but that doesn't seem to apply here. Removed it and tests are passing.

Issue #2886, opened 2026-04-16
MSVC Windows build failed with 5.3.1


nono303:
Hi,
Quick feedback & question regarding compilation of the new release 5.3.1 under Windows MSVC _(vs18 x64)_.
  
#### 1. autoconf version report not 5.3.1
```
...
===============================================================================
jemalloc version   : 5.3.0-395-g81034ce1f1373e37dc865038e1bc8eeecf559ce8
library revision   : 2

CONFIG             : --enable-autogen --disable-doc --enable-shared --disable-static CC=cl CXX=cl
...
```

#### 2. `INSTALL.md` §"With MSVC solutions" might be updated
```patch
--- a/INSTALL-81034ce1.000.md	Thu Apr 16 12:19:56 2026 UTC
+++ b/INSTALL.md	Thu Apr 16 12:19:28 2026 UTC
@@ -419,18 +419,18 @@
    * grep
    * sed
 
-2. Install Visual Studio 2015 or 2017 with Visual C++
+2. Install Visual Studio 2015 to 2026 with Visual C++
 
 3. Add Cygwin\bin to the PATH environment variable
 
-4. Open "x64 Native Tools Command Prompt for VS 2017"
+4. Open "x64 Native Tools Command Prompt for VS"
    (note: x86/x64 doesn't matter at this point)
 
 5. Generate header files:
    sh -c "CC=cl ./autogen.sh"
 
 6. Now the project can be opened and built in Visual Studio:
-   msvc\jemalloc_vc2017.sln
+   msvc\jemalloc_vc2022.sln
 
 ### With MSYS
 This is a more involved approach that offers the same configuration flexibility

```
  - :question: Should I add `CXX=CL` or `--disable-cxx` to the autogen line?
  - :question: Are there `--disable-xxx` options to make autoconf generate only the headers _(and not documentation, tests, or anything else that takes time)_?
  
#### 3. Building with VS (or MSBuild) failed:
```
"C:\sdk\src\jemalloc\msvc\jemalloc_vc2022.sln" (Clean;jemalloc cible) (1) ->
"C:\sdk\src\jemalloc\msvc\projects\vc2022\jemalloc\jemalloc.vcxproj" (cible par défaut) (3:3) ->
(Link cible) -> 
  jemalloc.obj : error LNK2001: unresolved external symbol malloc_conf_init [C:\sdk\src\jemalloc\msvc\projects\vc2022\jemalloc\jemalloc.vcxproj]
[...]

PR #2889, opened 2026-04-17
Add conf.c into msvc and update INSTALL.md

guangli-dai:
Should resolve the MSVC error reported in #2886

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -411,7 +411,7 @@ differ in their ease of use and flexibility.
 ### With MSVC solutions
 This is the easy, but less flexible approach. It doesn't let you specify
 arguments to the `configure` script.
-  
+
 1. Install Cygwin with at least the following packages:
    * autoconf
    * autogen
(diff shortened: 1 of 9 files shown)


INSTALL.md
```diff
@@ -419,18 +419,22 @@ arguments to the `configure` script.
    * grep
    * sed
 
-2. Install Visual Studio 2015 or 2017 with Visual C++
+2. Install Visual Studio 2015, 2017, 2019, or 2022 with Visual C++
```
lexprfuncall, line 422: I wonder if we want to enumerate all of the versions of Visual Studio.  It probably only matters for the solution files.  Microsoft products are usually somewhat smart about backward compatibility and I would expect that even if a user has a later version of Visual Studio they'd be able to open the latest .sln file we provide and the IDE will offer to upgrade it.

INSTALL.md
```diff
@@ -419,18 +419,22 @@ arguments to the `configure` script.
...
 5. Generate header files:
    sh -c "CC=cl ./autogen.sh"
 
-6. Now the project can be opened and built in Visual Studio:
+6. Now the project can be opened and built in Visual Studio using the
+   corresponding solution file:
+   msvc\jemalloc_vc2015.sln
    msvc\jemalloc_vc2017.sln
+   msvc\jemalloc_vc2019.sln
+   msvc\jemalloc_vc2022.sln
```
lexprfuncall, line 437: You could even shorten this to the corresponding solution file in the msvc directory.

lexprfuncall (approved):
It might be good to split this into two commits and land them together.

INSTALL.md
nono303: Agree as 2026 is current version and not listed here

guangli-dai: Makes sense, modified to describe the current situation better without enumerating.

@nono303 After we upgrade the tests on Windows for VS 2026 and the solution file (if needed), will further change the doc here.

INSTALL.md
guangli-dai: Updated.

PR #2891, opened 2026-04-18
jemalloc: fix always_inline build failure

lxz0503:
The ph_gen macro generates a function heap_ph_cmp with static linkage, and ph.h marks internal functions with always_inline.
	JEMALLOC_ALWAYS_INLINE int a_prefix##_ph_cmp(void *a, void *b) {
		return a_cmp((a_type *)a, (a_type *)b);
	}
In ph.c, ph_gen(static, heap, node_t, link, node_cmp_magic). The called functions node_cmp_magic and node_cmp are all static int but not static inline int. When gcc is using -Og, it will fail with error like "inlining failed in call to 'always_inline' 'heap_ph_cmp': function not considered for inlining".

### DIFF
--- a/test/unit/ph.c
+++ b/test/unit/ph.c
@@ -13,7 +13,7 @@ struct node_s {
 	uint64_t    key;
 };
 
-static int
+static inline int
 node_cmp(const node_t *a, const node_t *b) {
 	int ret;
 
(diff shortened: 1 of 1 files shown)


lxz0503:
I will try another solution to fix this "always_inline build failure".

PR #2900, opened 2026-04-23
Replace std::__throw_bad_alloc call with standard C++

lexprfuncall:
Since December of 2025, std::__throw_bad_alloc is no longer visible through #include <new> causing jemalloc build failures with gcc 16. As far as I can tell, all std::__throw_bad_alloc did was arrange to raise a std::bad_alloc exception if exceptions are enabled.  I am not sure whether its usage was truly meaningful in jemalloc since the call is wrapped in a try catch and any usage of try catch is considered an error when compiling with -fno-exceptions on gcc, at least.

This change adds a check to configure.ac that determines whether exceptions are enabled by compiling a simple try catch that raises a std::bad_alloc exception.  If that test succeeds, the macro JEMALLOC_HAVE_CXX_EXCEPTIONS is defined, and jemalloc will raise an exception.  Otherwise, we call std::terminate() to abort.

### DIFF
--- a/src/jemalloc_cpp.cpp
+++ b/src/jemalloc_cpp.cpp
@@ -1,3 +1,4 @@
+#include <exception>
 #include <mutex>
 #include <new>
 // NOLINTBEGIN(misc-use-anonymous-namespace)
@@ -90,17 +91,26 @@ handleOOM(std::size_t size, bool nothrow) {
 		if (handler == nullptr)
 			break;
 
+#ifdef JEMALLOC_HAVE_CXX_EXCEPTIONS
 		try {
 			handler();
 		} catch (const std::bad_alloc &) {
 			break;
 		}
+#else
+		handler();
+#endif
 
 		ptr = je_malloc(size);
 	}
 
-	if (ptr == nullptr && !nothrow)
-		std::__throw_bad_alloc();
+	if (ptr == nullptr && !nothrow) {
+#ifdef JEMALLOC_HAVE_CXX_EXCEPTIONS
+		throw std::bad_alloc();
+#else
+		std::terminate();
+#endif
+	}
 	return ptr;
 }
 
--- a/src/jemalloc_cpp.cpp
+++ b/src/jemalloc_cpp.cpp
@@ -1,5 +1,4 @@
 #include <exception>
-#include <mutex>
 #include <new>
 // NOLINTBEGIN(misc-use-anonymous-namespace)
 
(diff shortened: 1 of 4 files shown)


src/jemalloc_cpp.cpp
```diff
@@ -90,17 +91,26 @@ handleOOM(std::size_t size, bool nothrow) {
 		if (handler == nullptr)
```
spredolac: nit: 
These are 10 years old.  Maybe we replace this with simple: 
       std::new_handler handler = std::get_new_handler();

Not sure what is some reasonable timeframe to expect compiler upgrades, but if someone uses c++ c++11 is very reasonable expectation (that was major upgrade)

spredolac (approved):
lgtm

guangli-dai (approved):
Two nits:
1. The CI does not include gcc 16, were you able to test locally that this works fine? Probably worthy adding into the summary;
2. The check in `configure.ac` should be robust enough. The first line `AC_DEFINE` is separate, probably add a `dnl` comments to explain the test a bit for easier future references?

src/jemalloc_cpp.cpp
lexprfuncall: > Not sure what is some reasonable timeframe to expect compiler upgrades, but if someone uses c++ c++11 is very reasonable expectation (that was major upgrade)

If I am not mistaken, `std::mutex` and `std::lock_guard`, both used above, already implies C++11 so it should be safe to use `std::get_new_handler()` as well.  Since that clean-up is a logically separate change, I'll add another commit to this PR to make that clean-up.

lexprfuncall:
> Two nits:
> 
>     1. The CI does not include gcc 16, were you able to test locally that this works fine? Probably worthy adding into the summary;

Yup.  I tested on FreeBSD with gcc16.  I'll add that in.

>     2. The check in `configure.ac` should be robust enough. The first line `AC_DEFINE` is separate, probably add a `dnl` comments to explain the test a bit for easier future references?

Sure.  Does a `dnl` with a comment above the `AC_LANG_PUSH` (below the `AC_DEFINE`) explaining that the code that follows checks for assertions match what you have in mind?

guangli-dai:
> Does a dnl with a comment above the AC_LANG_PUSH (below the AC_DEFINE) explaining that the code that follows checks for assertions match what you have in mind?

Yes, exactly what I am thinking.

PR #2904, opened 2026-04-27
Split jemalloc.c into cohesive modules with smaller cross-module surface

spredolac:
Split jemalloc.c into cohesive modules with minimal cross-module surface: jemalloc_init, arenas_management, and jemalloc_fork and improve test coverage

Module headers expose only functions; state is static where possible. State that must be extern (because hot inline accessors read it) stays in the module's own header rather than piling onto jemalloc_internal_externs.h. Writers go through setters for symmetry.  Improve unit tests to cover each module's public surface.

No functional changes should be present.  Code is mainly moved around and only cosmetic changes are performed (except new unit tests where they were lacking).  I did not add certain initialization test improvements that would demand adding test hooks into jemalloc_init code. Will treat that in a separate PR as this one is large and I want to keep it very mechanical


lexprfuncall:
The subject of this PR is pretty expansive but it looks like the pre-fork support and some arena management are the substance.  Can this be split into commits targeted to specific units of functionality?

spredolac:
> The subject of this PR is pretty expansive but it looks like the pre-fork support and some arena management are the substance. Can this be split into commits targeted to specific units of functionality?

Yes, I still have it split in a branch where I did not squash, I just mistakenly anticipated that reviewers will want it squashed.  Let me port few compilation fixes there and I will update. I will ping you when ready for review

madscientist:
I'm not a GitHub expert but I think if you push a PR with multiple commits then it is up to the reviewer if they want to review all the diffs as one big bunch, or look at them individually by choosing a subset.

lexprfuncall:
> Yes, I still have it split in a branch where I did not squash, I just mistakenly anticipated that reviewers will want it squashed. Let me port few compilation fixes there and I will update. I will ping you when ready for review

I just mistakenly squashed my last commit since that was the default :-(  (Was it that way on the fork?  I can't remember.)  At least for me, having smaller commits is preferable to a big squashed commit.  When bisecting to find a bug, a time when the commit history really matters, it's nicer to have the more focused commits.

include/jemalloc/internal/arena_externs.h
```diff
@@ -33,6 +33,8 @@ extern size_t oversize_threshold;
 extern bool      opt_huge_arena_pac_thp;
 extern pac_thp_t huge_arena_pac_thp;
 
+extern unsigned huge_arena_ind;
```
guangli-dai, line 36: Maybe not for this PR, but `huge_arena_ind` is not used in any inline headers. Can it be made static instead?

include/jemalloc/internal/jemalloc_init.h
```diff
@@ -0,0 +1,43 @@
...
+
+extern malloc_init_t malloc_init_state;
+
+bool malloc_is_initializer(void);
+bool malloc_initializer_is_set(void);
+void malloc_initializer_set(void);
+
+bool malloc_init_hard_a0(void);
+bool malloc_init_hard(void);
+void malloc_slow_flag_init(void);
```
guangli-dai: This looks like a helper function which can stay static inside the module without being exposed? Similarly, not requiring this in the current PR.

guangli-dai (approved):
New modules' separation is pretty reasonable. The conflict should come from the recent fix in #2905 and should be straightforward to resolve. After that we can merge.

include/jemalloc/internal/arena_externs.h
spredolac: it is used in extent.c, but could be refactored to be static (if some parts of that code move to arena management). See https://github.com/jemalloc/jemalloc/blob/dev/src/extent.c#L826 
I can handle it separately when I make an extra pass at all of this.

include/jemalloc/internal/jemalloc_init.h
spredolac: done

PR #2906, opened 2026-04-29
Preserve errno across free, free_sized, and free_aligned_sized

spredolac:
Preserve errno across free, free_sized, and free_aligned_sized
    
The slow path of free can invoke syscalls that may set errno on failure. Save and restore errno around the body of the slow-path entry points so the user-visible contract is preserved without adding any instructions to the inline fast path, which is syscall-free.
    
The inner save/restore around process_madvise in pages.c is no longer needed: the outer wrap covers user-thread purge, and background-thread purge does not require errno preservation.

Fastpath free looks the same before and after
```
/tmp/hotpath_afterhjxzA4gZBa:   file format elf64-x86-64

Disassembly of section .text:

<_Z6myfreeP5Point>:
                push    rbp
                mov     rbp, rsp
                test    rdi, rdi
                je       <L1>
                mov     rax, rdi
                shl     rax, 0x34
                je       <L0>
                mov     rax, qword ptr fs:[-0x410]
                add     rax, 0x20
                cmp     rax, qword ptr fs:[-0x408]
                jae      <L0>
                mov     rcx, qword ptr fs:[-0x3c8]
                cmp     word ptr fs:[-0x3b6], cx
                je       <L0>
                lea     rdx, [rcx - 0x8]
                mov     qword ptr fs:[-0x3c8], rdx
                mov     qword ptr [rcx - 0x8], rdi
                mov     qword ptr fs:[-0x410], rax
                pop     rbp
                ret
<L0>:
                mov     esi, 0x18
                xor     edx, edx
                call     <jemalloc_je_sdallocx_default>
<L1>:
                pop     rbp
                ret
                mov     rdi, rax
                call     <__clang_call_terminate>
```

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -2042,6 +2042,7 @@ void
 free_default(void *ptr) {
 	UTRACE(ptr, 0, 0);
 	if (likely(ptr != NULL)) {
+		int saved_errno = get_errno();
 		/*
 		 * We avoid setting up tsd fully (e.g. tcache, arena binding)
 		 * based on only free() calls -- other activities trigger the
@@ -2068,6 +2069,7 @@ free_default(void *ptr) {
 		}
 
 		check_entry_exit_locking(tsd_tsdn(tsd));
+		set_errno(saved_errno);
 	}
 }
 
@@ -2903,6 +2905,7 @@ inallocx(tsdn_t *tsdn, size_t size, int flags) {
 
 JEMALLOC_NOINLINE void
 sdallocx_default(void *ptr, size_t size, int flags) {
+	int saved_errno = get_errno();
 	assert(ptr != NULL);
 	assert(malloc_initialized() || IS_INITIALIZER);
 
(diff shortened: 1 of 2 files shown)


src/jemalloc.c
```diff
@@ -2068,6 +2069,7 @@ free_default(void *ptr) {
 		}
 
 		check_entry_exit_locking(tsd_tsdn(tsd));
+		set_errno(saved_errno);
```
guangli-dai, line 2072: We can also add a `LOG` here if the errono is reset so that users can know what went wrong by turning on LOG. LOG can be enabled by prefix, a sample usage is [here](https://github.com/jemalloc/jemalloc/blob/6e0b8e6daac28ceff382b620e11032e8fadd5c15/src/jemalloc.c#L1082). WDYT?

guangli-dai (approved):

src/jemalloc.c
spredolac: I think that if the purpose is to explain what failed it would not be helpful - user would need to know which call clobbered the value in which case we would be better adding LOG lines where system calls are.  I would leave it out of here.

src/jemalloc.c
```diff
@@ -2903,6 +2905,7 @@ inallocx(tsdn_t *tsdn, size_t size, int flags) {
 
 JEMALLOC_NOINLINE void
 sdallocx_default(void *ptr, size_t size, int flags) {
+	int saved_errno = get_errno();
```
lexprfuncall, line 2908: Can we scope this a little more tightly?  I'm not sure the compiler is smart enough to move this call around.

PR #2907, opened 2026-04-30
Use SEC in PAC to reduce lock contention on the ecaches

binliu19:
The first commit refactors SEC as a preparation: use atomic SEC bin counters so stats merging does not take sec_bin locks. Also split shard picking from SEC cache operations.
The second commit introduces SEC to PAC, similar to SEC in HPA.

### DIFF
--- a/src/pac.c
+++ b/src/pac.c
@@ -7,6 +7,18 @@
 #include "jemalloc/internal/san.h"
 #include "jemalloc/internal/witness.h"
 
+static inline uint8_t
+pac_sec_shard_pick(tsdn_t *tsdn, sec_t *sec) {
+	if (sec->opts.nshards <= 1) {
+		return 0;
+	}
+	if (tsdn_null(tsdn)) {
+		return 0;
+	}
+	tsd_t *tsd = tsdn_tsd(tsdn);
+	return sec_shard_pick(tsd, sec, tsd_pac_sec_shardp_get(tsd));
+}
+
 static inline void
 pac_decay_data_get(pac_t *pac, extent_state_t state, decay_t **r_decay,
     pac_decay_stats_t **r_decay_stats, ecache_t **r_ecache) {
@@ -95,6 +107,16 @@ pac_init(tsdn_t *tsdn, pac_t *pac, base_t *base, emap_t *emap,
 	pac->stats_mtx = stats_mtx;
 	atomic_store_zu(&pac->extent_sn_next, 0, ATOMIC_RELAXED);
 
+	if (sec_init(tsdn, &pac->sec, base, &opt_pac_sec_opts)) {
+		/* sec_init already zeroed nshards and max_alloc. */
+	}
+	if (!sec_is_used(&pac->sec) || dirty_decay_ms == 0) {
+		atomic_store_zu(&pac->sec_max_alloc, 0, ATOMIC_RELAXED);
+	} else {
+		atomic_store_zu(&pac->sec_max_alloc,
+		    pac->sec.opts.max_alloc, ATOMIC_RELAXED);
+	}
+
 	return false;
 }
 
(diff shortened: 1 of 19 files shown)


src/stats.c
```diff
@@ -835,6 +835,37 @@ stats_arena_hpa_shard_sec_print(emitter_t *emitter, unsigned i) {
...
 }
 
+static void
+stats_arena_pac_sec_print(emitter_t *emitter, unsigned i) {
+	size_t sec_bytes;
+	size_t sec_hits;
+	size_t sec_misses;
+	size_t sec_dalloc_flush;
+	size_t sec_dalloc_noflush;
+	CTL_M2_GET("stats.arenas.0.pac_sec_bytes", i, &sec_bytes, size_t);
```
guangli-dai, line 848: When SEC is enabled, we should probably also add bytes in SEC into the dirty section of stats. Please also check the resident estimate is also correct there. Otherwise, malloc stats parsing would view SEC in PAC as a free-of-charge contention relieving tool while it actually consumes memory.

guangli-dai: Similarly for pinned bytes calculation.

src/pac.c
```diff
@@ -133,6 +156,15 @@ pac_alloc_real(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, size_t size,
 
 	edata_t *edata = NULL;
 
+	if (!guarded && !zero && alignment <= PAGE
+	    && size <= atomic_load_zu(&pac->sec_max_alloc, ATOMIC_RELAXED)) {
+		edata = sec_alloc(tsdn, &pac->sec, size,
```
guangli-dai, line 167: nit: I agree that we do not need to call `sec_fill` here when there is nothing cached in SEC because allocating multiple extents from PAC is strictly one after another and cannot easily optimize like HPA does now. Maybe add a comment explain the motivation of not doing `sec_fill` here?

src/pac.c
```diff
@@ -133,6 +156,15 @@ pac_alloc_real(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, size_t size,
 
 	edata_t *edata = NULL;
 
+	if (!guarded && !zero && alignment <= PAGE
+	    && size <= atomic_load_zu(&pac->sec_max_alloc, ATOMIC_RELAXED)) {
```
guangli-dai, line 162: This seems to overlap with the [check](https://github.com/jemalloc/jemalloc/blob/4de3a4c3d1bb4520acdc856ddab3e57a28eb7795/src/sec.c#L182) in `sec_alloc` itself, do we need to keep both?

test/unit/pac_sec_integration.c
```diff
@@ -0,0 +1,368 @@
+#include "test/jemalloc_test.h"
+
+/*
+ * Use 1 shard for deterministic stat assertions and a small max_bytes so
+ * overflow triggers quickly.  Background threads are disabled to prevent
+ * asynchronous decay from interfering with precise stat checks.
+ */
+const char *malloc_conf =
+    "experimental_pac_sec_nshards:1,background_thread:false";
```
guangli-dai: Should be a better coverage if we test with multi-shard setup.

src/pac.c
```diff
@@ -732,6 +797,11 @@ pac_decay_ms_set(tsdn_t *tsdn, pac_t *pac, extent_state_t state,
 	pac_maybe_decay_purge(tsdn, pac, decay, decay_stats, ecache, eagerness);
 	malloc_mutex_unlock(tsdn, &decay->mtx);
 
+	if (update_pac_sec && decay_ms != 0) {
+		atomic_store_zu(&pac->sec_max_alloc,
```
guangli-dai, line 817: Since SEC only as a per-bin lock, a potential racy condition here is the deallocation enters the `sec_dalloc` path and finishes after the sec flushes. Not a critical racy problem, that deallocated region will likely stay there as a leak until SEC is activated again or the process ends. Considering how rare this situation is, probably not worthy a fix, but should add comments (probably at dalloc path) to note this down for future reference.

src/stats.c
binliu19: will do. SEC bytes will be counted toward active, dirty, and pinned.

src/pac.c
binliu19: will add a comment.

src/pac.c
binliu19: Note that `pac->sec_max_alloc` can be 0 when dirty decay is set to 0, in order to offer a fast path; otherwise it is equal to `sec->opts.max_alloc`.  So yes, in the current version, the PAC-sec path can have an redundant comparison.   I will add the size check in HPA so we don't have to check inside `sec_alloc` (and instead asserts).

test/unit/pac_sec_integration.c
binliu19: I can do that. However, with a single thread, it doesn't give us much better coverage. And with multiple threads, randomness can happen and we may have trouble making the process determinsitic.

guangli-dai (approved):

PR #2909, opened 2026-04-30
Make huge_arena_ind static within the arena module (Follow up for #2904)

spredolac:
EASY - Make huge_arena_ind static within the arena module (Follow up for #2904)

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -47,7 +47,7 @@ static unsigned nbins_total;
  * that, the huge_arena_ind is updated to point to the actual huge arena,
  * which is the last one of the auto arenas.
  */
-unsigned  huge_arena_ind = 0;
+static unsigned huge_arena_ind = 0;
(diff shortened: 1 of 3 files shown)


src/arena.c
```diff
@@ -1978,6 +1978,11 @@ arena_init_huge(tsdn_t *tsdn, arena_t *a0) {
 	return huge_enabled;
 }
 
+bool
+arena_ind_is_huge(unsigned ind) {
+	return ind != 0 && ind == huge_arena_ind;
```
guangli-dai: IIUC, it should be `huge_arena_ind != 0 && ind == huge_arena_ind`? `huge_arena_ind != 0` means there is a huge arena and then we judge whether the given arena is the huge one. Using `ind != 0` is not technically wrong but I think `huge_arena_ind != 0` is better for reading

spredolac: sure.

PR #2910, opened 2026-04-30
Improve const correctness in the repo

spredolac:
Improve const correctness in the repo

I noticed few of these while reading the code and figured this is something that is easy to delegate to LLM. The tests should be pretty much CI.


guangli-dai (approved):

guangli-dai:
The appveyor error looks like irrelevant. LGTM

PR #2913, opened 2026-05-02
Fix extra semicolon

gruenich:


include/jemalloc/internal/ph.h
```diff
@@ -510,7 +510,7 @@ ph_enumerate_next(
 	typedef struct {                                                       \
 		void               *bfs_queue[a_max_queue_size];               \
 		ph_enumerate_vars_t vars;                                      \
-	} a_prefix##_enumerate_helper_t;
+	} a_prefix##_enumerate_helper_t
```
guangli-dai: This actually requires an extra semicolon when calling `ph_structs`. Looks more reasonable to keep the semicolon here and remove the one when calling it like [here](https://github.com/jemalloc/jemalloc/blob/1ffd5e96d6ea882b5d5a85d57562c9b777a5669a/include/jemalloc/internal/edata.h#L98)? Same for other `ph_` macros.

gruenich: Done

guangli-dai (approved):

guangli-dai:
@gruenich Thank you  for the contribution!

PR #2914, opened 2026-05-06
[EASY] Hide methods from jemalloc_init.h that are not used in other units

spredolac:
Two methods in this header could be private to translation unit.  They are anyway executed and tested in every test because we must init malloc, so keeping them public is more noise than actual value.


guangli-dai:
The mac CI failures look real?

spredolac:
yeah, will fix.

guangli-dai (approved):

PR #2915, opened 2026-05-08
Cleanup interface between tcache,cache_bin,bin,arena and improve test coverage

spredolac:
Cleanup interface between `tcache,cache_bin,bin,arena` and improve test coverage

We are addressing some technical debt in this PR that cleans up the interfaces between `arena, tcache, bin, and cache_bin` in the spirit of the design comment in `cache_bin.h`. Arena owns the per-arena state, `tcache` owns the per-thread cache, and they are not reaching into each other's internals.

`tcache` no longer references `bin_t` or any of its primitives `arena.c `has no non-routing calls into `tcache` (so the only things left are member variables in arena stats that refer to `tcache` which is fine, and `tcache` as an forwarded argument in allocation dispatching). Allocation dispatching may be extracted as a separate translation unit, but that is out of scope of this PR. 

There is no behavior change and no hot-path code touched: identical locks, identical operations, identical order.

Coverage is improved with test additions. There are only two functional changes:
a) mutex that protects `cache_bin_array_descriptor_ql` in the arena is initilaized close to the list it protects after postfork (no need to lock the list anyway as that is only surviving thread, but it is cleaner to see them together)
b) `tcache_ql` is removed from arena and not used anymore. Communication goes completely via cache bin array descriptor and there are no circles `arena->tcache->arena`


include/jemalloc/internal/jemalloc_internal_inlines_b.h
```diff
@@ -7,6 +7,15 @@
 #include "jemalloc/internal/extent.h"
 #include "jemalloc/internal/jemalloc_internal_inlines_a.h"
 
+static inline void
+thread_migrate_arena(tsd_t *tsd, arena_t *oldarena, arena_t *newarena) {
+	arena_migrate(tsd, oldarena, newarena);
+	if (tcache_available(tsd)) {
+		tcache_arena_reassociate(tsd_tsdn(tsd),
```
guangli-dai, line 17: nit: we can move the assertions to this level as well? Specifically I mean `newarena != NULL && oldarena != NULL` at the beginning of `thread_migrate_arena`). Now the two assertions already exist but one in the caller and another buried in the callee of this funciton.

test/unit/fork.c
```diff
@@ -213,8 +213,81 @@ TEST_BEGIN(test_fork_multithreaded) {
...
+	expect_d_eq(mallctl(ctl, &parent_bytes, &bsz, NULL, 0),
+	    0, "read tcache_bytes pre-fork");
+	expect_zu_gt(parent_bytes, 0,
+	    "Parent should see cached bytes from cache_bin descriptor");
+
+	pid_t pid = fork();
+	if (pid == -1) {
+		test_fail("Unexpected fork() failure");
+	} else if (pid == 0) {
+		/*
```
guangli-dai, line 261: Maybe also do an atual allocation here to make sure it is functioning properly.

test/unit/fork.c
```diff
@@ -213,8 +213,81 @@ TEST_BEGIN(test_fork_multithreaded) {
 }
 TEST_END
 
+TEST_BEGIN(test_fork_postfork_descriptor_relink) {
+#ifndef _WIN32
+	test_skip_if(!config_stats);
+	test_skip_if(!tcache_available(tsd_fetch()));
+
+	/* Set up a manually managed arena bound to this thread. */
```
guangli-dai, line 221: Two things potentially missing from the unit tests:
1. when multiple arenas are involved, will the relink be done properly.
2. When there are multiple threads, only the surviving threads' cache bins will be added in. No assertion regarding this? 

These two are probably out of the scope of this unit test and new tests need to be added because `test_fork_multithreaded` and `test_fork_child_usability` both don't cover these cases yet. Could be a follow-up PR instead of this one.

guangli-dai (approved):
Just a nit in the assertions and some comments on the tests.

test/unit/fork.c
spredolac: Will leave this as a follow up

PR #2916, opened 2026-05-09
arena_s: Replace bin_t all_bins[0] by [] for C99 or newer

gruenich:
Zero-sized arrays are not allowed by ISO C.
Type-checking fails, because all_bins is asigned malloced storage of length > 0.
Found by GCC and Clang (-Wpedantic).

### DIFF
--- a/include/jemalloc/internal/arena_structs.h
+++ b/include/jemalloc/internal/arena_structs.h
@@ -104,7 +104,11 @@ struct arena_s {
 	    "Do not use this field directly. "
 	    "Use `arena_get_bin` instead.")
 	JEMALLOC_ALIGNED(CACHELINE)
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
+	bin_t all_bins[];
+#else
 	bin_t all_bins[0];
+#endif
 };
 
 #endif /* JEMALLOC_INTERNAL_ARENA_STRUCTS_H */


include/jemalloc/internal/arena_structs.h
```diff
@@ -105,7 +105,7 @@ struct arena_s {
 	    "Do not use this field directly. "
 	    "Use `arena_get_bin` instead.")
 	JEMALLOC_ALIGNED(CACHELINE)
-	bin_t all_bins[0];
+	bin_t *all_bins;
```
spredolac: This is not the same thing.  One was the old trick to say I have some array at the end of the structs, but the layout will be cache aligned after the filed `name`. The other means it is a pointer and you will allocate it elsewhere. If you look at `arena_new` code you will see

```
size_t arena_size = ALIGNMENT_CEILING(sizeof(arena_t), CACHELINE)
            + sizeof(bin_t) * nbins_total;
arena = (arena_t *)base_alloc(tsdn, base, arena_size, CACHELINE);
```
which would be broken with your change.

The c99 way would be `all_bins[]`; See the section 6.7.2.1, paragraph 16 (Structure and union specifiers with example that matches arena_new code

spredolac (changes requested):

include/jemalloc/internal/arena_structs.h
gruenich: Thanks for the explanation! What about using `all_bins[]` in case of C99 or newer? I adjusted my patch, what do you think?

include/jemalloc/internal/arena_structs.h
```diff
@@ -105,7 +105,12 @@ struct arena_s {
 	    "Do not use this field directly. "
 	    "Use `arena_get_bin` instead.")
 	JEMALLOC_ALIGNED(CACHELINE)
+#if __STDC_VERSION__ >= 199901L
```
spredolac: If you do
 `#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L`
it will be more robust to some compilers emitting warnings when `__STDC_VERSION__` is not defined.

include/jemalloc/internal/arena_structs.h
```diff
@@ -105,7 +105,12 @@ struct arena_s {
...
+	// Only C99 offers a ISO compliant way
```
spredolac: nit: No need for comment, it is clear from the guard.

spredolac (commented):
Thanks. Please make these small changes and I will merge it in.

gruenich:
I adjusted the code, thanks for the review!

PR #2918, opened 2026-05-11
Fix psset_enumerate_search pages-vs-bytes comparison

binliu19:


spredolac:
Mind adding a test that catches that fails before the fix and passes afterwards? Also, create a follow up to evaluate the effects. on perf and fragmentation internally if this enumerate search is added/removed (technically we lived as if it was not there).

spredolac:
I checked some results and this fix improves fragmentation with HPA. So, just add a test that is broken before the fix and works after the fix and we are good to go.

spredolac (approved):
Just add a test before landing.

guangli-dai (approved):

test/unit/hpa_pageslab_packing.c
```diff
@@ -0,0 +1,204 @@
+#include "test/jemalloc_test.h"
+
+/*
+ * Tests that HPA cleanly packs allocations into a single pageslab when they
+ * collectively fit, instead of growing a new one.  See block comment on the
+ * malloc_conf options below for why each is required.
+ */
+const char *malloc_conf =
```
guangli-dai, line 7: Although that shouldn't be done when running tests, but with environment variable set, the configs here could be overriden. We should add more `test_skip_if` on necessary setups in the tests to prevent those false alarming failures.

test/unit/hpa_pageslab_packing.c
```diff
@@ -0,0 +1,204 @@
...
+TEST_END
+
+TEST_BEGIN(test_hpa_pageslab_packing_three_allocs) {
+	test_skip_if(!config_stats);
+	test_skip_if(!hpa_supported());
+	setup_arena();
+
+	/*
+	 * Sweep both c (the third allocation) and a (the first allocation,
+	 * varied from LARGE_MIN_PAGES up to half of the remaining pages so
```
guangli-dai: The upper bound of a seems to be `rest - LARGE_MIN_PAGES` instead of `rest/2`? Is the comment stale or the implementation needs update?

guangli-dai (commented):
The fix itself is correct. Thanks for catching that! Two nits on the tests added.

test/unit/hpa_pageslab_packing.c
binliu19: you are right. this comment was for a prior version. the latest version tries to test more combination of sizes.

PR #2919, opened 2026-05-11
Refactor/remove pai vtable

guangli-dai:


include/jemalloc/internal/hpa.h
```diff
@@ -154,6 +154,23 @@ bool hpa_shard_init(tsdn_t *tsdn, hpa_shard_t *shard, hpa_central_t *central,
     emap_t *emap, base_t *base, edata_cache_t *edata_cache, unsigned ind,
     const hpa_shard_opts_t *opts, const sec_opts_t *sec_opts);
 
+/*
+ * Implementations of the page allocator interface for the HPA.  PA calls these
+ * directly, with no vtable indirection.  expand and shrink are stubs (always
```
spredolac: The part of the comment 'PA calls these directly, with no vtable indirection' is unnecessary as it only makes sense given prior state.  I would delete it.

include/jemalloc/internal/pac.h
```diff
@@ -165,6 +165,21 @@ bool pac_init(tsdn_t *tsdn, pac_t *pac, base_t *base, emap_t *emap,
     ssize_t dirty_decay_ms, ssize_t muzzy_decay_ms, pac_stats_t *pac_stats,
     malloc_mutex_t *stats_mtx);
 
+/*
+ * Implementations of the page allocator interface for the PAC.  PA calls these
+ * directly, with no vtable indirection.
```
spredolac: Same - please delete "PA calls these directly, with no vtable indirection."

src/pa.c
```diff
@@ -216,8 +221,12 @@ pa_dalloc(tsdn_t *tsdn, pa_shard_t *shard, edata_t *edata,
 	edata_addr_set(edata, edata_base_get(edata));
 	edata_szind_set(edata, SC_NSIZES);
 	pa_nactive_sub(shard, edata_size_get(edata) >> LG_PAGE);
-	pai_t *pai = pa_get_pai(shard, edata);
-	pai_dalloc(tsdn, pai, edata, deferred_work_generated);
+	if (edata_pai_get(edata) == EXTENT_PAI_HPA) {
+		hpa_dalloc(tsdn, &shard->hpa_shard, edata,
```
spredolac: nit: shard->hpa_shard could be renamed into shard->hpa to have symmetry with pac and make this line shorter.

src/pac.c
```diff
@@ -292,10 +274,9 @@ pac_alloc_impl(tsdn_t *tsdn, pai_t *self, size_t size, size_t alignment,
 	return edata;
 }
 
-static bool
-pac_expand_impl(tsdn_t *tsdn, pai_t *self, edata_t *edata, size_t old_size,
+bool
+pac_expand(tsdn_t *tsdn, pac_t *pac, edata_t *edata, size_t old_size,
     size_t new_size, bool zero, bool *deferred_work_generated) {
```
spredolac, line 279: nit: I know you are guarding the call by returning if extent is hpa, but you may want to add assert here that edata is pac.

src/pac.c
```diff
@@ -360,10 +341,9 @@ pac_expand_impl(tsdn_t *tsdn, pai_t *self, edata_t *edata, size_t old_size,
 	return false;
 }
 
-static bool
-pac_shrink_impl(tsdn_t *tsdn, pai_t *self, edata_t *edata, size_t old_size,
+bool
+pac_shrink(tsdn_t *tsdn, pac_t *pac, edata_t *edata, size_t old_size,
```
spredolac, line 347: nit: assert, same as pac_extent

spredolac (approved):
Code looks pretty straightforward.  Please fix the comments (it puts context that was relevant to the llm task at hand, but it is not valuable to the destination - just creates noise) before merging.

src/pa.c
guangli-dai: Good idea, renamed.

src/pac.c
guangli-dai: Added, similar assertions added to `pac_expand`, `pac_shrink`, and `pac_dalloc` where requests come in with a previously used page-level allocator specified.

PR #2922, opened 2026-05-14
Add follow-up test for post fork multithread

spredolac:
Add follow-up test for post fork multithread case and fix FreeBsd bug

See [comment](https://github.com/jemalloc/jemalloc/pull/2915#discussion_r3229673979)

The test actually discovered a bug on FreeBsd where _malloc_postfork is called.  On FreeBsd it seems that it is called both in parent and the child ([here if I am not mistaken](https://github.com/freebsd/freebsd-src/blob/9306d0449/lib/libthr/thread/thr_fork.c#L264-L288C4)), thus, we need to forward the call to child version if we detect that we are invoked in child.  

reproed with small driver on VM FreeBsd 15 and confirmed that fix  is working.

The consequences of this bug are that on freebsd system after fork we may unnecessary retain memory in child processes.

### DIFF
--- a/src/jemalloc_fork.c
+++ b/src/jemalloc_fork.c
@@ -12,6 +12,19 @@
  * malloc during fork().
  */
 
+#ifdef JEMALLOC_MUTEX_INIT_CB
+/*
+ * When JEMALLOC_MUTEX_INIT_CB is defined, pthread_atfork registration is
+ * skipped and the platform calls _malloc_prefork/_malloc_postfork directly.
+ * FreeBSD's libthr calls _malloc_postfork in both parent and child.  Detect
+ * the child by pid change so we route to jemalloc_postfork_child, which resets
+ * per-arena state the parent handler does not touch (nthreads, descriptor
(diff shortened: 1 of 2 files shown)


spredolac:
failure looks unrelated

guangli-dai:
> failure looks unrelated

Yes, unrelated, somehow the appveyor is failing on signature sporadically and it doesn't reproduce stable.

guangli-dai (approved):

guangli-dai:
I think a question to answer here is: 1. should we further apply the tweak in these tests in all jemalloc's postfork functions so that we distinguish parent and child process ourselves proactively and execute corresponding codes or 2. should we escalate this to the FreeBSD community and require/submit a patch there?

spredolac:
Well, doing this is never wrong (recording the pid pre-fork and making sure it is the same in parent and forwarding to a child if not.  I added one more test for sequential forking.   If we want to this unconditionally/on all platforms, I think we do a follow up. This should be already nice fix for freebsd folks.

PR #2923, opened 2026-05-18
Set SEC_MAX_NALLOCS to 8.

gctony:
Increase SEC_MAX_NALLOCS from 4 to 8. From our testing it looks as if 8 (compared to 4) increases hugepage-backed memory further and decreases hpa locking ops/sec further.


guangli-dai:
Discussed offline, to conclude the discussion for future reference:
1. conducted canaries on internal services. Experimental results show ~neutral impact from this change but overall a stable increase in hugepage coverage, which is a good sign;
2. intuitionally, the default value of optiojn `sec_extra_fill` is 3, so refilling 4 items is the default behavior. Setting max to 4 is a bit too constrained.

Issue #2925, opened 2026-05-26
Veracode Flaw (static): Numeric Errors


brookheather:
We would like to move to the latest jemalloc 5.3.1 but our code has to pass a Veracode scan and with the latest jemalloc there are a number of issues raised by Veracode - the most serious rated is at line sz.h:323 for Integer Overflow or Wraparound.  Having examined the code we think this would only be an issue when allocating exabytes of memory but nonetheless we are requesting that this line of code is fixed - we are not allowed to make local changes to FOSS source code - we have to pull the release version directly from Github in order to use the code.  The full Veracode error description is:

_This arithmetic operation results in an integer overflow error. Because the result of this calculation is larger than the maximum possible value for this data type, the result may wrap to become a very small, or negative number, therefore providing an unintended value. Integer overflows can often trigger buffer overflows, which can be exploited to execute arbitrary code._

Veracode also identified these issues:

 jemalloc.c:1411: This arithmetic operation results in an integer underflow error. Because the result of this calculation is less than the minimum allowable integer value for this data type, the result may wrap to become a very large, positive number, thereby providing an unintended value. Integer underflows do not generally result in arbitrary code execution but can trigger unexpected application behavior.

sz.h:406: This arithmetic operation results in an integer underflow error. Because the result of this calculation is less than the minimum allowable integer value for this data type, the result may wrap to become a very large, positive number, thereby providing an unintended value. Integer underflows do not generally result in arbitrary code execution but can trigger unexpected application behavior.

spredolac:
Hi @brookheather , thanks for reporting.  I do not have veracode, so I was wondering if you have a way to run it on the [PR 2926](https://github.com/jemalloc/jemalloc/pull/2926)?

spredolac:
@brookheather , merged onto dev branch, when you get a chance please test so that we can close.  Thanks!

brookheather:
Thanks - it will take a few days for us to deploy and retest so will get back to you next week hopefully.

brookheather:
@spredolac I have updated the PR with the scan results.

spredolac:
> [@spredolac](https://github.com/spredolac) I have updated the PR with the scan results.

[...]

PR #2926, opened 2026-05-26
Fix numeric overflow checks in size classes

spredolac:
This should address issue #2925 

I do not have veracode to run locally, but based on description this should fit. 
This change does not affect the ordinary malloc() fast path. `sz_sa2u()` logic is only reached for aligned allocations.

`SC_LARGE_MAXCLASS` is expressed as pointer-width constant instead of macro arithmetic. `configure.ac` only supports `LG_SIZEOF_PTR == 2` or `3` (32-bit or 64-bit pointers, with the same MSVC mapping), so the constants are equivalent while avoiding arithmetic expansion at `jemalloc.c` use sites (one of them most likely reported by veracode).

For `sz_sa2u()`, zero alignment now returns `0` before computing `alignment - 1`. Power-of-two alignment is internal caller contract enforced by the existing assert in debug build (all current callers respect it by either validating alignment, deriving it from`MALLOCX_ALIGN_GET()`, or passing known power-of-two constants).

New tests fail before the change.

### DIFF
--- a/include/jemalloc/internal/sc.h
+++ b/include/jemalloc/internal/sc.h
@@ -268,12 +268,17 @@
 #define SC_LARGE_MINCLASS ((size_t)1ULL << (LG_PAGE + SC_LG_NGROUP))
 #define SC_LG_LARGE_MINCLASS (LG_PAGE + SC_LG_NGROUP)
 
-/* Internal; only used for the definition of SC_LARGE_MAXCLASS. */
-#define SC_MAX_BASE ((size_t)1 << (SC_PTR_BITS - 2))
-#define SC_MAX_DELTA ((size_t)1 << (SC_PTR_BITS - 2 - SC_LG_NGROUP))
-
-/* The largest size class supported. */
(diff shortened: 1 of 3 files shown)


guangli-dai (approved):
LGTM, since CI is green, we can merge once @brookheather confirms that it passes Veracode checks.

brookheather:
Thanks for the quick response however I will need a branch or tag with this change in it to download it on our side (internally I need to provide just a URL) and then I can get it scanned with Veracode.

spredolac:
I will merge into `dev` then as changes are okay in any case.  We will not close your issue until you confirm.

brookheather:
We have run Veracode on the latest dev branch and unfortunately the Integer Overflow error is still present on sz.h:331 and there is now an additional Integer Overflow error on sz.h:361 so the latest code is actually worse it seems.  I think sz.h:331 will need to be rewritten.

Lines sz.h:331 and sz.h:421 also have an Integer Underflow error.

image

PR #2927, opened 2026-05-27
Test coverage improvement in tcache, pac, hpa_central

spredolac:
With the help of LLM generated several unit tests that were missing.  nit: Cleaned up few assert only writes in a separate commit

### DIFF
--- a/include/jemalloc/internal/cache_bin.h
+++ b/include/jemalloc/internal/cache_bin.h
@@ -523,14 +523,13 @@ cache_bin_stash(cache_bin_t *bin, void *ptr) {
 /* Get the number of stashed pointers. */
 JEMALLOC_ALWAYS_INLINE cache_bin_sz_t
 cache_bin_nstashed_get_internal(cache_bin_t *bin) {
-	cache_bin_sz_t ncached_max = cache_bin_ncached_max_get(bin);
 	cache_bin_sz_t low_bits_low_bound = cache_bin_low_bits_low_bound_get(
 	    bin);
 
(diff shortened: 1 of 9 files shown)


include/jemalloc/internal/cache_bin.h
```diff
@@ -523,14 +523,13 @@ cache_bin_stash(cache_bin_t *bin, void *ptr) {
 /* Get the number of stashed pointers. */
 JEMALLOC_ALWAYS_INLINE cache_bin_sz_t
 cache_bin_nstashed_get_internal(cache_bin_t *bin) {
-	cache_bin_sz_t ncached_max = cache_bin_ncached_max_get(bin);
```
gctony, line 526: I assume the assert-related changes were to avoid potentially doing calculations that are only needed in asserts?

gctony (approved):
Loooks good. I have to admit I only did a high level scan of each new test and understood what it tried to achieve without looking super carefully at all the details.

guangli-dai (approved):
At a glance, I though some public functions (called by other modules) were not covered by the test cases, i.e., `emap_alloc_ctx.*` and `pac_expand/shrink`. But double checked and found they are actually covered by other existing tests. LGTM.

include/jemalloc/internal/cache_bin.h
spredolac: yes, some tools would report it as "dead write", so it is mostly noise reduction (compiler would optimize it away anyway in non-debug build)

PR #2928, opened 2026-05-28
Ctl refactor and removal of some interface

spredolac:
This PR deduplicates some code in CTL, extracts method for some logic checks and removes several interfaces:
* `malloc_conf_2_conf_harder` is private variable and no need to have interface for it. Only dumped in stats
*  `safety_check_abort` was only used in integration tests and we can use test hooks for it
* `pactivep` - could not find usage internally or externally
*  utilization queries are something that is used externally (for example in [valkey](https://github.com/valkey-io/valkey)) for defragmentation and they use a bit cleaner part of the interface that at least does not deal with current slab.  The part we are removing was used and removed in this [project](https://github.com/memkind/memkind/commit/5cbb16a281af6954b433497d0f2b15b19f8512c8)

`experimental_infallible_new`  will be removed separately


src/ctl.c
```diff
@@ -2693,6 +2688,40 @@ arena_i_initialized_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+
+	for (i = 0; i < narenas; i++) {
+		tarenas[i] = arena_get(tsdn, i, false);
+	}
+
+	/*
+	 * No further need to hold ctl_mtx, since narenas and tarenas contain
+	 * everything needed below.
+	 */
+	malloc_mutex_unlock(tsdn, &ctl_mtx);
```
gctony: So, it's assumed that you're holding the mutex when the function is called but not when the function returns? Is there a way to do this in a more symmetrical / natural way? You should at least assert that you're holding the mutex at the start of the function.

src/ctl.c
```diff
@@ -2693,6 +2688,40 @@ arena_i_initialized_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
...
+		}
+	}
+}
+
+static void
+arena_i_decay_one(tsdn_t *tsdn, unsigned arena_ind, bool all) {
+	arena_t *tarena = arena_get(tsdn, arena_ind, false);
+
+	/* No further need to hold ctl_mtx. */
+	malloc_mutex_unlock(tsdn, &ctl_mtx);
```
gctony, line 2733: Ditto.

src/ctl.c
```diff
@@ -3295,24 +3295,35 @@ arenas_lextent_i_index(
 	return super_arenas_lextent_i_node;
 }
 
+static int
+ctl_arena_create(tsd_t *tsd, void *oldp, size_t *oldlenp,
+    const arena_config_t *config) {
+	int ret;
```
gctony, line 3301: You don't actually need ret, right? just return EAGAIN instead of the goto and return 0 at the end?

src/ctl.c
```diff
@@ -3295,24 +3295,35 @@ arenas_lextent_i_index(
...
 	WRITE(config.extent_hooks, extent_hooks_t *);
-	if ((arena_ind = ctl_arena_init(tsd, &config)) == UINT_MAX) {
-		ret = EAGAIN;
-		goto label_return;
-	}
-	READ(arena_ind, unsigned);
 
-	ret = 0;
+	ret = ctl_arena_create(tsd, oldp, oldlenp, &config);
 label_return:
```
gctony, line 3327: is the label still needed?

src/ctl.c
```diff
@@ -3321,21 +3332,15 @@ arenas_create_ctl(tsd_t *tsd, const size_t *mib, size_t miblen, void *oldp,
...
 	WRITE(config, arena_config_t);
 
-	if ((arena_ind = ctl_arena_init(tsd, &config)) == UINT_MAX) {
-		ret = EAGAIN;
-		goto label_return;
-	}
-	READ(arena_ind, unsigned);
-	ret = 0;
+	ret = ctl_arena_create(tsd, oldp, oldlenp, &config);
 label_return:
```
gctony, line 3344: is the label still needed?

src/ctl.c
```diff
@@ -3584,118 +3589,65 @@ prof_log_stop_ctl(tsd_t *tsd, const size_t *mib, size_t miblen, void *oldp,
 	return 0;
 }
 
+#define PROF_HOOK_CTL_BODY(hook_type, hook_get, hook_set, allow_null)          \
```
gctony: is there any way to turn this into a function? Not sure I love something this elaborate being a macro

gctony (commented):
A few comments, mostly stylistic.

src/ctl.c
spredolac: yes, these macros are using it.  It is too much to change in one pass, that is worth separate PR which only changes the logic there.

src/ctl.c
spredolac: Too much to do in same PR - these macros are using ret.  I think that is worth rewriting to lose the macros and goto labels for most of these things.

src/ctl.c
spredolac: yeah, great point, this was automatic extraction of the method and starting code was asymmetrical.  I would add assert_owner for this PR and assert !owner on exit at least.  I will examine the calling places to see what needs to be protected (first one is easy, second looks more complex) and if relatively simple will make separate commit on top of this stack (so that we can review correctness)

src/ctl.c
spredolac: yes, macros use it

src/ctl.c
spredolac: No, since types are different and without C++ you cannot paramterize the types.  I think your instinct is good though and I should probably revert this part and keep it inlined.  There will be a PR to split CTL code and I think this will end up in ctl_prof.c so number of lines will be more manageable.

src/ctl.c
```diff
@@ -981,6 +981,21 @@ ctl_accum_atomic_zu(atomic_zu_t *dst, atomic_zu_t *src) {
...
+ctl_arena_ind_is_deprecated_all(size_t i, unsigned narenas) {
+	/*
+	 * Historical compatibility for treating arena.<narenas> as the merged
+	 * all-arenas entry.  New code should use MALLCTL_ARENAS_ALL.
+	 */
+	return i == narenas;
+}
+
+static bool
+ctl_arena_ind_is_all(size_t i, unsigned narenas) {
```
guangli-dai, line 983: There are similar use cases that could be replaced by the new functioni: [1](https://github.com/jemalloc/jemalloc/blob/dev/src/ctl.c#L1013) and [2](https://github.com/jemalloc/jemalloc/blob/dev/src/ctl.c#L3144). Please double check if any similar usage can be replaced by the new functions.

src/ctl.c
```diff
@@ -981,6 +981,21 @@ ctl_accum_atomic_zu(atomic_zu_t *dst, atomic_zu_t *src) {
 
 /******************************************************************************/
 
+static bool
+ctl_arena_ind_is_deprecated_all(size_t i, unsigned narenas) {
+	/*
+	 * Historical compatibility for treating arena.<narenas> as the merged
+	 * all-arenas entry.  New code should use MALLCTL_ARENAS_ALL.
+	 */
+	return i == narenas;
```
guangli-dai, line 975: `narenas` could be subject to racings so we need `ctl_mtx` to protect it, is that correct? If so, we should mention that in comment and also add assertions to make sure we have the lock when reading the `narenas` value.

test/unit/prof_hook.c
```diff
@@ -334,9 +334,23 @@ TEST_BEGIN(test_prof_sample_hooks) {
...
 
+TEST_BEGIN(test_prof_hook_noop) {
+	test_skip_if(!config_prof);
+
+	const char *hooks[] = {"experimental.hooks.prof_backtrace",
+	    "experimental.hooks.prof_dump", "experimental.hooks.prof_sample",
+	    "experimental.hooks.prof_sample_free"};
+
+	for (unsigned i = 0; i < sizeof(hooks) / sizeof(hooks[0]); i++) {
+		expect_d_eq(mallctl(hooks[i], NULL, NULL, NULL, 0), EINVAL,
```
guangli-dai, line 346: Besides passing a NULL in, passing a function pointer pointing to a NULL is also critical. It seems `prof_backtrace` has this test already [here](https://github.com/jemalloc/jemalloc/blob/dev/test/unit/prof_hook.c#L92) but `prof_dump` doesn't.

guangli-dai (commented):
First batch of comments, reviewing the rest.

spredolac:
* rewrote the logic around arena helpers so that it is easy to see mutex locks/unlocks (instead of having asymmetry)
* added comments near checkers around deprecated all
* `ctl_narenas_get()` will check that it is owner of the mutex and value will be obtained like that from all the reasonable places  
* revert PROF macros and added PROF test for ptr to NULL target hook
* For places where i == narenas is called instead of deprecated all, that is because semantic is different, we are checking if the slot is free for creation, not using deprecated "all arenas" meaning.

test/unit/prof_hook.c
```diff
@@ -334,9 +334,50 @@ TEST_BEGIN(test_prof_sample_hooks) {
 }
 TEST_END
 
+TEST_BEGIN(test_prof_hook_noop) {
```
guangli-dai, line 337: nit: The commit message should be updated since the dedpulicate part is gone.

guangli-dai (approved):
One nit. Latest changes and other commits LGTM.
One question, why revert the prof hooks deduplicate changes?

gctony (approved):

spredolac:
Will change commit message.  

I was not very happy/unsure with deduplication of prof hooks to begin with and when @gctony brought the same during the code review it just flipped me to the side of do not do it now.   I will have PR where we are splitting things into components. It will be more manageable to think about the code then.  Certain things tripped my code reading few times, mostly usage of ret/goto in macros and all of that being so macro heavy (then we do not have C++ templates for this).   I will make another pass at CTL after the split to see if we can rewrite something to make it less macro heavy.

PR #2929, opened 2026-05-29
Ctl split experiment

spredolac:
Split ctl handlers by mallctl namespace
    
    Move mallctl handler implementations out of src/ctl.c into namespace-oriented ctl_* modules, while keeping the mallctl tree and dispatch machinery centralized in ctl.c.
    
    Add shared ctl mallctl helper macros and thin ctl arena/stat interfaces needed by the split modules. Wire the new sources into Makefile.in and MSVC project files.
    
    Keep behavior unchanged; this is intended as a readability and navigation refactor.


gctony (commented):
Splitting a 4,000 line file is clearly a helpful change! One nit though: is there a reason why `ctl` is used as a suffix (e.g., `tcache_create_ctl()`) instead of a prefix?

spredolac:
I am not happy with exporting more symbols from these files. ctl is large file, but it is pretty much boiler plate.  I am proposing just [commenting and adding TOC](https://github.com/jemalloc/jemalloc/pull/2937) instead.

spredolac:
We will not be doing this.  Conceptually ctl is relatively simple. With LLMs it may not be that hard to maintain it.  In the future we may attempt to use code generator to describe the mallctl tree, but this is not the highest priority at the moment.

PR #2930, opened 2026-05-29
Create malloc_dispatch. Get rid of jemalloc_internal_includes.h

spredolac:
This PR cleans up jemalloc’s internal header dependency structure by removing the jemalloc_internal_includes.h umbrella header and replacing it with explicit module includes. It does it by consolidating several split internal headers (arena_*, prof_*, tcache_*, bin_*, etc.) and breaking existing include cycles.
  
It also moves malloc routing logic into a new malloc_dispatch module.

It should result in a clearer ownership between arena, tcache, and profiling.

[Add header dep graph script for cleanup verification](https://github.com/jemalloc/jemalloc/pull/2930/commits/0336dee3d09b90d7759ec5ebefee8d240e4a37d5) can be removed as it was only used during development, but we may as well keep it.


include/jemalloc/internal/background_thread.h
```diff
@@ -66,4 +65,33 @@ struct background_thread_stats_s {
 };
 typedef struct background_thread_stats_s background_thread_stats_t;
 
-#endif /* JEMALLOC_INTERNAL_BACKGROUND_THREAD_STRUCTS_H */
+extern bool                      opt_background_thread;
```
guangli-dai, line 68: Not for this PR for sure (since it's already super large), some of these `extern` variables might be unnecessary and should be refactored later.

guangli-dai (approved):
A great PR! Things I am concerned seems to be well handled:
1. no changes of inline hints, so should be perf neutral;
2. Compile-time header fan-out is well taken care of and minimal headers (like `arena_decay_constants.h`) were created to avoid heavy includes for small things;
3. The mitigation of splitting modules into types/structs/externs and inlines into a/b/c are gone together with its root cause the circular dependency.

I think there is no harm to keep the script generating the dependency graph. We can even consider improving that into the unit test to prevent future cyclic dependency to show up again in the future.

PR #2931, opened 2026-05-31
void function should not return anything

gruenich:
Found by GCC 16 (-Wpedantic).


guangli-dai (approved):
LGTM, thank you for the patch!

PR #2932, opened 2026-06-03
Replace ctl mallctl macros with helper functions

spredolac:
The `READ/WRITE/READONLY/WRITEONLY/VERIFY_READ/ASSURED_WRITE/MIB_UNSIGNED/ NEITHER_READ_NOR_WRITE/READ_XOR_WRITE `macros hid control flow (`goto label_return`) inside the mallctl handlers as well as dependency on the context (`oldp,newp`,...).  Convert them to memcpy-based helper functions.

Add error-path tests for the converted handlers and direct unit tests for the helpers.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -1860,99 +1860,99 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
 /******************************************************************************/
 /* *_ctl() functions. */
 
-#define READONLY()                                                             \
-	do {                                                                   \
-		if (newp != NULL || newlen != 0) {                             \
-			ret = EPERM;                                           \
-			goto label_return;                                     \
-		}                                                              \
-	} while (0)
+static inline int
+ctl_writeonly(void *oldp, size_t *oldlenp) {
+	if (oldp != NULL || oldlenp != NULL) {
+		return EPERM;
+	}
+	return 0;
+}
 
-#define WRITEONLY()                                                            \
-	do {                                                                   \
-		if (oldp != NULL || oldlenp != NULL) {                         \
-			ret = EPERM;                                           \
-			goto label_return;                                     \
-		}                                                              \
-	} while (0)
+static inline int
+ctl_assured_write(void *dst, size_t dst_size, const void *newp,
+    size_t newlen) {
+	if (newp == NULL || newlen != dst_size) {
+		return EINVAL;
+	}
+	memcpy(dst, newp, dst_size);
+	return 0;
+}
 
-/* Can read or write, but not both. */
-#define READ_XOR_WRITE()                                                       \
-	do {                                                                   \
-		if ((oldp != NULL && oldlenp != NULL)                          \
-		    && (newp != NULL || newlen != 0)) {                        \
-			ret = EPERM;                                           \
-			goto label_return;                                     \
-		}                                                              \
-	} while (0)
+static inline int
+ctl_read(void *oldp, size_t *oldlenp, const void *src, size_t src_size) {
+	if (oldp == NULL || oldlenp == NULL) {
+		return 0;
+	}
+	const size_t oldlen = *oldlenp;
+	if (oldlen != src_size) {
+		size_t copylen = (src_size <= oldlen) ? src_size : oldlen;
+		memcpy(oldp, src, copylen);
+		*oldlenp = copylen;
+		return EINVAL;
+	}
+	memcpy(oldp, src, src_size);
+	return 0;
+}
 
-/* Can neither read nor write. */
-#define NEITHER_READ_NOR_WRITE()                                               \
-	do {                                                                   \
-		if (oldp != NULL || oldlenp != NULL || newp != NULL            \
-		    || newlen != 0) {                                          \
-			ret = EPERM;                                           \
-			goto label_return;                                     \
-		}                                                              \
-	} while (0)
+static inline int
+ctl_write(void *dst, size_t dst_size, const void *newp, size_t newlen) {
+	if (newp == NULL) {
+		return 0;
+	}
+	if (newlen != dst_size) {
+		return EINVAL;
+	}
+	memcpy(dst, newp, dst_size);
+	return 0;
+}
 
-/* Verify that the space provided is enough. */
-#define VERIFY_READ(t)                                                         \
-	do {                                                                   \
-		if (oldp == NULL || oldlenp == NULL                            \
-		    || *oldlenp != sizeof(t)) {                                \
-			if (oldlenp != NULL) {                                 \
-				*oldlenp = 0;                                  \
-			}                                                      \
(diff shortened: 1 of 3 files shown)


src/ctl.c
```diff
@@ -1852,99 +1852,97 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
...
-			goto label_return;                                     \
-		}                                                              \
-		v = (unsigned)mib[i];                                          \
-	} while (0)
+static inline int
+ctl_read(void *oldp, size_t *oldlenp, const void *src, size_t src_size) {
+	if (oldp == NULL || oldlenp == NULL) {
+		return 0;
+	}
+	if (*oldlenp != src_size) {
```
gctony: Worth assigning:
`const size_t oldlen = *oldlenp;`
and using it in these two lines?

src/ctl.c
```diff
@@ -1852,99 +1852,97 @@ ctl_mtx_assert_held(tsdn_t *tsdn) {
...
+			*oldlenp = 0;
+		}
+		return EINVAL;
+	}
+	return 0;
+}
+
+JET_EXTERN int
+ctl_mib_unsigned(unsigned *dst, const size_t *mib, size_t mib_index) {
+	if (mib[mib_index] > UINT_MAX) {
```
gctony: Again, maybe:
`const size_t value = mib[mib_index];`
and use `value`?

src/ctl.c
```diff
@@ -2126,45 +2096,40 @@ max_background_threads_ctl(tsd_t *tsd, const size_t *mib, size_t miblen,
 
 	malloc_mutex_lock(tsd_tsdn(tsd), &ctl_mtx);
 	malloc_mutex_lock(tsd_tsdn(tsd), &background_thread_lock);
-	if (newp == NULL) {
```
gctony, line 2137: In the old version, `if (newp == NULL) { ... }` is the first check after taking the locks and `READ()` will be reached after that single NULL check. In the new version, `ctl_read()` will be reached after multiple checks. Is that important in any way? Is `newp` more likely to be NULL or not NULL here? The same pattern appears in other places too.

src/ctl.c
```diff
@@ -3209,45 +3144,31 @@ arena_i_index(tsdn_t *tsdn, const size_t *mib, size_t miblen, size_t i) {
...
-		}
-		if (dirty
-		        ? arena_dirty_decay_ms_default_set(*(ssize_t *)newp)
-		        : arena_muzzy_decay_ms_default_set(*(ssize_t *)newp)) {
+	ssize_t oldval = dirty ? arena_dirty_decay_ms_default_get()
+	                       : arena_muzzy_decay_ms_default_get();
+	int ret = ctl_read(oldp, oldlenp, &oldval, sizeof(oldval));
+	if (ret == 0 && newp != NULL) {
+		ssize_t newval;
+		ret = ctl_write(&newval, sizeof(newval), newp, newlen);
```
gctony, line 3186: The previous version doesn't do a write. Is this correct?

gctony (approved):
This is a great change and I think the new version is a lot more readable and maintainable.

One suggestion (maybe for a follow-up change): A lot of the values that are being read/written are <= 8 bytes. If you're worried about the memcpy overhead, would a wrapper like this work?

```
if (src_size == 1) {
  *((uint8_t) dst) = (uint8_t) val;
} else if (src_size == 2) {
  *((uint16_t) dst) = (uint16_t) val;
  ...
} else {
  memcpy(...);
}
```

With inlining, maybe `src_size` will be constant and the if-statement (or switch-statement?) elided.

src/ctl.c
```diff
@@ -2083,38 +2060,32 @@ background_thread_ctl(tsd_t *tsd, const size_t *mib, size_t miblen, void *oldp,
...
-			}
-		} else {
-			if (background_threads_disable(tsd)) {
-				ret = EFAULT;
-				goto label_return;
+	int ret = 0;
+	if (newp != NULL && newlen != sizeof(bool)) {
+		ret = EINVAL;
+	}
+	if (ret == 0) {
```
guangli-dai: One key decision point here is whether we should keep the `label_return`. Previously, all those `label_return` gotos were hidden in the macros so we should definitely remove the macros. Yet with the new functions done here, perhaps it's OK to keep the `label_return` to skip a chain of `if (ret == 0)`? Such `goto`s would be very clear in serving early returns and would make the primary logic cleaner. I would vote for it in functions like this one.

src/ctl.c
guangli-dai: I don't think this matters as long as the semantic stays the same for users calling `mallctl`s.

spredolac: it's not significant here. mallctl is a control-plane API, not on any allocation hot path, so the extra one or two comparisons before ctl_read() are negligible and branch predicts trivially.  I kept the uniform structure for readability

src/ctl.c
spredolac: Correct! (and behaviorally identical). 
The old version did what WRITE was doing here instead of using the WRITE() macro.  See
 ```
if (newp != NULL) {
   if (newlen != sizeof(ssize_t)) {
       ret = EINVAL; goto ...;
   }
   arena_dirty_decay_ms_default_set( *(ssize_t *)newp);
 
```  
 while `ctl_write(&newval, sizeof(newval), newp, newlen) `does exactly the same thing: validates `newlen == sizeof(ssize_t)` (EINVAL otherwise) and copies into the local `newval`, which is passed to arena_dirty_decay_ms_default_set().

src/ctl.c
spredolac: Yeah, let me look at several of these. I prefer without goto unless there is too much nesting. I will make a pass and use some judgement. This one would be flatter I think.

spredolac:
> This is a great change and I think the new version is a lot more readable and maintainable.
> 
> One suggestion (maybe for a follow-up change): A lot of the values that are being read/written are <= 8 bytes. If you're worried about the memcpy overhead, would a wrapper like this work?
> 
> ```
> if (src_size == 1) {
>   *((uint8_t) dst) = (uint8_t) val;
> } else if (src_size == 2) {
>   *((uint16_t) dst) = (uint16_t) val;
>   ...
> } else {
>   memcpy(...);
> }
> ```
> 
> With inlining, maybe `src_size` will be constant and the if-statement (or switch-statement?) elided.

On the common path (sizes match), the compiler already lowers `memcpy(oldp, src, 8) `to a single 
  store — `str x4, [x3] (gcc) / str x9, [x0] (clang)`. No `memcpy` call is emitted. A real call only survives on the truncation/EINVAL path, where the
  length is runtime-variable.

src/ctl.c
spredolac: done. I did functions that had more than 3 levels of nesting.

PR #2933, opened 2026-06-04
Infallible new

spredolac:
This PR replaces runtime option `experimental_infallible_new` with compile time option.  At the same time it fixes the `new (nothrow)` contract when option is used.  Compile time option will allow for certain LTO.


gctony (approved):

src/jemalloc_cpp.cpp
```diff
@@ -66,20 +66,20 @@ void operator delete[](
...
-
+	const char *huge_warning = (size >= ((std::size_t)1 << 30))
+	    ? "This may be caused by heap corruption, if the large size "
+	      "is unexpected (suggest building with sanitizers for "
+	      "debugging). "
+	    : "";
+	safety_check_fail(
+	    "<jemalloc>: Allocation of size %zu failed. %sAborting.\n",
+	    size, huge_warning);
+	return nullptr;
```
swkrueger, line 81: I was initially confused by this `return null`, but then I noticed that safety_check_fail will call abort. This return should be unreachable, right?

swkrueger (approved):

src/jemalloc_cpp.cpp
spredolac: Yes in prod, but not in tests. Hook is allowed instead of abort so that it is easier to test it (as opposed to capturing abort).

src/stats.c
```diff
@@ -1774,7 +1774,6 @@ stats_general_print(emitter_t *emitter) {
 	OPT_WRITE_BOOL("zero")
 	OPT_WRITE_BOOL("utrace")
 	OPT_WRITE_BOOL("xmalloc")
-	OPT_WRITE_BOOL("experimental_infallible_new")
```
guangli-dai, line 1777: We also have a section in stats showing some critical compile-time options [here](https://github.com/jemalloc/jemalloc/blob/dev/src/stats.c#L1623), i.e., `config\..*`, we should add the infallible_new there? Proposing this because it is something useful to know especially when debugging.

guangli-dai (approved):

PR #2936, opened 2026-06-09
Surface cleanup

spredolac:
This PR is cleaning up few things mechanically: 
* removes some dead-code
* Makes some function static if they are only called from single TU (not counting tests). This makes the surface of symbols smaller
* `ehooks_default_merge`  was once the only extern hook, but commit [add63659](https://github.com/jemalloc/jemalloc/commit/add636596afecb87e220d31ae75a9ba0b4601fbc) is the commit that rendered the export dead ([6342da09](https://github.com/jemalloc/jemalloc/commit/6342da0970257187f5fcc9504301eba75f92ccca) is the one that had introduced it)


spredolac:
Unrelated failure.

guangli-dai (approved):

PR #2937, opened 2026-06-09
Add a table of contents and section banners to ctl.c

spredolac:
Instead of [splitting](https://github.com/jemalloc/jemalloc/pull/2929) which would require making certain symbols external to translation units this may be cheap compromise to help navigation.

### DIFF
--- a/src/ctl.c
+++ b/src/ctl.c
@@ -27,6 +27,29 @@
 #include "jemalloc/internal/witness.h"
 
 /******************************************************************************/
+/*
+ * File layout.  Each section below is introduced by a matching banner comment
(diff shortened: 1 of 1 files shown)


guangli-dai (approved):
Discussed offline, compared to splitting the large `ctl.c` into multiple modules, adding comments to make it more readable should be a better solution to reduce future addition/removal trouble. However, the available mallctls are still not clear and adding/removing mallctls has to go through the primitive manual tree builds. We will follow up to see if there is anything we can do to make the mallctl name tree system more automated and available mallctls clearer.

PR #2938, opened 2026-06-12
Handle jemalloc calls after TSD teardown (tsd_generic)

spredolac:
### Summary:
This only changes behavior for the tsd_generic path, where pthread_getspecific() can report no TSD after the pthread key destructor has finished.  Other TSD backends keep tsd_teardown_done() as a constant false, so the added checks compile out there.

Avoid recreating TSD for late deallocations.  Preserve existing reincarnation behavior for late allocations and nonzero reallocations.

Add Linux CI coverage for force_tls=0, with and without --enable-debug, to exercise the Android-equivalent generic TSD path.

### Report:
I received several stacks from Android where we configure with force_tls=0. These were the stacks that point to TSD re-created during a teardown free() 
 ```
- A1: a0ialloc ← je_free ← thread_data_t::trampoline ← __pthread_start ← __start_thread                                                                                                                           
 - A2: je_tsd_init_check_recursion ← je_free ← thread_data_t::trampoline ← __pthread_start ← __start_thread                                                                                                        
 - A3: je_extent_heap_remove_first ← arena_bin_malloc_hard ← je_arena_malloc_hard ← a0ialloc ← je_free ← thread_data_t::trampoline ← __pthread_start ← __start_thread                                              
 - A4: je_extent_heap_first ← extent_recycle ← je_extent_alloc_wrapper ← arena_bin_malloc_hard ← je_arena_malloc_hard ← a0ialloc ← je_free ← thread_data_t::trampoline ← __pthread_start ← __start_thread          
 - A5: je_malloc_usable_size ← LimitFree ← thread_data_t::trampoline ← __pthread_start ← __start_thread 
```

### Testing:
Created new tests to run on added CI configurations (on Linux, but simulate force_tls).  Tests fail before the fix and pass afterwards.

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1025,12 +1025,50 @@ isfree(tsd_t *tsd, void *ptr, size_t usize, tcache_t *tcache, bool slow_path) {
 	thread_dalloc_event(tsd, usize);
 }
 
+JEMALLOC_ALWAYS_INLINE bool
+dealloc_no_tsd(void *ptr) {
+	/*
+	 * On the generic pthread_getspecific() TSD path, a deallocation can run
+	 * after the thread's TSD has been torn down.  Using tsd_fetch_min() here
+	 * would allocate and publish a fresh TSD wrapper mid-teardown,
+	 * reincarnating the state teardown just released (the crashes this change
+	 * fixes).  Instead we hand the object straight back to its arena via
+	 * idalloctm() with a NULL tsdn.
+	 *
+	 * This deliberately bypasses the per-thread bookkeeping that the normal
+	 * ifree()/isfree() slow path performs, none of which is available or
+	 * meaningful without a live TSD.  The trade-offs, all acceptable for
+	 * these rare teardown-time frees (the memory itself is still correctly
+	 * returned to the arena):
+	 *   - Profiling: prof_free() is skipped, so a sampled object is not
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: src/jemalloc.c [hunk 1]
```diff
@@ -1022,12 +1022,32 @@ isfree(tsd_t *tsd, void *ptr, size_t usize, tcache_t *tcache, bool slow_path) {
...
+	/*
+	 * On the generic pthread_getspecific() TSD path, a deallocation may run
+	 * with no TSD wrapper.  Avoid tsd_fetch_min(); it would allocate and
+	 * publish a fresh wrapper during thread teardown.
+	 */
+	if (!tsd_teardown_done()) {
+		return false;
+	}
+
+	idalloctm(TSDN_NULL, ptr, /* tcache */ NULL, /* alloc_ctx */ NULL,
```
guangli-dai, line 1057: The profiled allocations are not handled. This causes profiler-side stats stale and maybe the profile at the end will show such objects still alive (not fully confirmed). I prefer not to complicate the logic here for this extreme case, but probably worthy noting down in the comments.

guangli-dai: Besides profiling, other behaviors like sized free and junk fill are also changed. But same as profiling, I tend to think it's fine but needs to be well documented in the comments.

guangli-dai (approved):

PR #2942, opened 2026-06-22
Accept NULL in free_sized() and free_aligned_sized()

bigbruno:
free_sized() and free_aligned_sized() forward straight to sdallocx(), which expects a non-NULL pointer and asserts on it in debug builds. But C23 says both should accept NULL and do nothing, like free(NULL), so a NULL argument either trips the assert or feeds NULL into the dealloc path in release builds.

It's not hard to hit: glibc 2.41 ships free_sized()/free_aligned_sized(), and a C++ sized delete of a null pointer compiles down to free_sized(). Once jemalloc is preloaded its versions win, and that NULL call takes down the process. I ran into it with GTK4/GLib apps under LD_PRELOAD.

The fix just checks for NULL first, the way free() already does, leaving the non-NULL path alone. Added test/integration/free_sized.c covering both functions including the NULL case (it trips the assert in sdallocx_default() with the old code under --enable-debug).

All the GTK4 programs I tested were affected, for example: https://github.com/manjaro/pamac/issues/562

### DIFF
--- a/src/jemalloc.c
+++ b/src/jemalloc.c
@@ -1069,14 +1069,35 @@ JEMALLOC_EXPORT void JEMALLOC_NOTHROW
 je_free_sized(void *ptr, size_t size) {
 	LOG("core.free_sized.entry", "ptr: %p, size: %zu", ptr, size);
 
-	je_sdallocx_noflags(ptr, size);
+	/*
+	 * free_sized(NULL, size) is a no-op, like free(NULL).  Do the NULL
+	 * check here rather than deferring to the je_free_impl() fallback:
+	 * without a size hint, free_fastpath() does an emap lookup that NULL
+	 * fails, so it reaches free_default(), which handles NULL.  With a
+	 * size hint, the fast path skips that lookup and falls through to
+	 * sdallocx_default(), which expects a non-NULL pointer.
+	 */
+	if (likely(ptr != NULL)) {
+		je_sdallocx_noflags(ptr, size);
+	}
 
 	LOG("core.free_sized.exit", "");
 }
 
 JEMALLOC_EXPORT void JEMALLOC_NOTHROW
 je_free_aligned_sized(void *ptr, size_t alignment, size_t size) {
-	je_sdallocx(ptr, size, /* flags */ MALLOCX_ALIGN(alignment));
+	LOG("core.free_aligned_sized.entry",
+	    "ptr: %p, alignment: %zu, size: %zu", ptr, alignment, size);
+
+	/*
+	 * Same as je_free_sized() above: the sdallocx path expects a non-NULL
+	 * pointer, so handle the C23 free_aligned_sized(NULL, ...) no-op here.
+	 */
+	if (likely(ptr != NULL)) {
+		je_sdallocx_impl(ptr, size, /* flags */ MALLOCX_ALIGN(alignment));
+	}
+
+	LOG("core.free_aligned_sized.exit", "");
 }
 
 /*
--- a/Makefile.in
+++ b/Makefile.in
@@ -330,6 +330,7 @@ endif
 TESTS_INTEGRATION := $(srcroot)test/integration/aligned_alloc.c \
 	$(srcroot)test/integration/allocated.c \
 	$(srcroot)test/integration/extent.c \
+	$(srcroot)test/integration/free_sized.c \
 	$(srcroot)test/integration/malloc.c \
 	$(srcroot)test/integration/mallocx.c \
 	$(srcroot)test/integration/MALLOCX_ARENA.c \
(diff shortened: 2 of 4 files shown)


src/jemalloc.c
```diff
@@ -1069,14 +1069,18 @@ JEMALLOC_EXPORT void JEMALLOC_NOTHROW
 je_free_sized(void *ptr, size_t size) {
 	LOG("core.free_sized.entry", "ptr: %p, size: %zu", ptr, size);
 
-	je_sdallocx_noflags(ptr, size);
+	if (likely(ptr != NULL)) {
```
spredolac, line 1080: nit: Mind adding a comment to explain that you cannot check this only after you fail in `free_fastpath` like `je_free_impl` does because in `je_free_impl` we rely on failure in emap lookup.

src/jemalloc.c
```diff
@@ -1069,14 +1069,18 @@ JEMALLOC_EXPORT void JEMALLOC_NOTHROW
...
+		je_sdallocx_noflags(ptr, size);
+	}
 
 	LOG("core.free_sized.exit", "");
 }
 
 JEMALLOC_EXPORT void JEMALLOC_NOTHROW
 je_free_aligned_sized(void *ptr, size_t alignment, size_t size) {
-	je_sdallocx(ptr, size, /* flags */ MALLOCX_ALIGN(alignment));
+	if (likely(ptr != NULL)) {
```
spredolac, line 1096: I think it would be better to call je_sdallocx_impl directly (instead of je_sdallocx which logs its own thing)
```
LOG("core.free_aligned_sized.entry",
            "ptr: %p, alignment: %zu, size: %zu", ptr, alignment, size);

if (likely(ptr != NULL)) {
                je_sdallocx_impl(ptr, size, /* flags */ MALLOCX_ALIGN(alignment));
}

LOG("core.free_aligned_sized.exit", "");
```

spredolac (commented):
Thanks for doing this it is great!  I can merge if you just add comment and change the logging in free_aligned_sized and call directly the impl version.

Optionally (you do not have to do it in this PR, but you may want to make change in cpp as well to make it consistent - logging ptr!=NULL as well and using same idiom to check)

```
JEMALLOC_ALWAYS_INLINE
  void
  sizedDeleteImpl(void *ptr, std::size_t size) noexcept {
        LOG("core.operator_delete.entry", "ptr: %p, size: %zu", ptr, size);

        if (likely(ptr != nullptr)) {
                je_sdallocx_noflags(ptr, size);
        }

        LOG("core.operator_delete.exit", "");
  }
```

and

```
JEMALLOC_ALWAYS_INLINE
  void
  alignedSizedDeleteImpl(
      void *ptr, std::size_t size, std::align_val_t alignment) noexcept {
        if (config_debug) {
                assert(((size_t)alignment & ((size_t)alignment - 1)) == 0);
        }
        LOG("core.operator_delete.entry", "ptr: %p, size: %zu, alignment: %zu",
            ptr, size, alignment);
  
        if (likely(ptr != nullptr)) {
                je_sdallocx_impl(ptr, size, MALLOCX_ALIGN(alignment));
        }
  
        LOG("core.operator_delete.exit", "");
  }
```

Thanks, again!

bigbruno:
Added the comments, and free_aligned_sized() now emits its own entry/exit logging and calls je_sdallocx_impl() directly instead of going through the je_sdallocx() wrapper. I also applied the same idiom to the C++ sizedDeleteImpl/alignedSizedDeleteImpl paths for consistency. Thanks for the review!

src/jemalloc.c
spredolac: Sorry, I was not clear enough.  The comment here was to explain the difference in the implementation between `je_free` and `je_free_sized`.   The first question I had when I was reviewing your code was "Why not do it the same way `je_free` [did it](https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/jemalloc_internal_inlines_c.h#L575)?  `je_free` does not branch first, but goes into `free_fastpath` and if that fails fallsback into `free_default`, and free_default` then does the [NULL ptr check](https://github.com/jemalloc/jemalloc/blob/dev/src/jemalloc.c#L1032).  so, je_free is able to avoid checking on the fast path because without size_hint we nicely[ fail in emap lookup ](https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/jemalloc_internal_inlines_c.h#L484C14-L484C44)for NULL, thus we do not hit the asserts and then can check on slow path.


Comment should be something like this:
_We check here rather than deferring to the slow path like je_free_impl() because size-hinted fastpath skips the emap lookup that lets free() reject NULL and thus prevents us from checking on slow path only_

spredolac:
Looks good, only comment is not stating what I intended, sorry I was not clear enough.
CI is green, all is well, please update the comment and I will merge.  Thanks @bigbruno!

bigbruno:
The comment now includes a more complete explanation.
I hope it meets expectations now. Thanks for the review.

PR #2944, opened 2026-06-22
Fix tcache corruption when fibers migrate across threads under whole-program LTO

### LINKED ISSUE #2890: Crashes in tcache when jemalloc bundled as static library with LTO into result binary
We noticed that once we integrated jemalloc with LTO into result binary (ClickHouse), it sometimes crashes in random places.

Looks like the issue is related to `tcache`, disabling it helps

After adding [some debugging](https://github.com/ClickHouse/jemalloc/pull/6) (sorry for the mess) it looks like tcache contains same pointer twice - https://pastila.nl/?004b9459/5c787018b0d6d9a32312c7e2a18f3dc1#aWEcjy8XsIOSFv0WrPKUhg==GCM (note that I cannot get stack when those pointers added because it hides the bug)

I will continue investigation, but maybe you will have some suggestion/ideas.

Refs: https://github.com/ClickHouse/ClickHouse/issues/102460#issuecomment-4270861892

azat:
The problem is `tsd_get()` returns `&tsd_tls`. Under jemalloc's default initial-exec TLS model the compiler forms this as `thread_pointer + const_offset` and treats it as loop-invariant, so under whole-program LTO it hoists the read out of the inlined `malloc`/`free` and keeps it in a callee-saved register across a user-space context switch (`swapcontext`, `ucontext`). If the fiber resumes on a different OS thread, two threads race on one `tcache` - silent heap corruption, reproducible only under LTO and invisible to sanitizers.

Two patches:

- `JEMALLOC_TLS_ADDR(tsd_tls)` — a `noinline` accessor that takes the address behind a **memory barrier**. It's an opaque call the optimizer cannot hoist or CSE across the switch. Because it's opaque, it's correct regardless of how the address is computed - including the dynamic TLS models, where `&tsd_tls` is a `__tls_get_addr` call whose result is just as cacheable. MSVC has no inline asm and keeps the plain address (a Windows build with fibers
under `/GL` needs `/GT`, fiber-safe TLS).
- Inline fast path on top, for the initial-exec model only (the default). The opaque call costs ~30–40% on the `malloc`/`free` fastpath (table), so for the static model the `tpoff` constant is captured once at runtime into a global by a `noinline` helper, and `tsd_get` then computes `volatile_asm_thread_pointer + offset`. Computing `&tsd_tls - __builtin_thread_pointer()` inline is not equivalent: clang hoists the stale `&tsd_tls` while re-materializing the tp terms, and `fresh_tp + (stale_addr - fresh_tp)` cancels the volatile read back to the stale address (tested - still corrupts).

`test/stress/microbench` (`malloc(1)`/`free` pairs, pinned core, clang):

|                  | unpatched | accessor only | + fast path |
|------------------|-----------|---------------|-------------|
| malloc, no LTO   | 7.0 ns    | 9.8 (+40%)    | 7.0 (±0%)   |
| free, no LTO     | 6.5 ns    | 9.2 (+40%)    | 6.5 (±0%)   |
| malloc, ThinLTO  | 7.0 ns    | 9.0 (+28%)    | 7.6 (~+8%)  |
| free, ThinLTO    | 6.5 ns    | 8.4 (+28%)    | 7.0 (~+8%)  |

Fixes: https://github.com/jemalloc/jemalloc/issues/2890
Refs:
- https://learn.microsoft.com/en-us/cpp/build/reference/gt-support-fiber-safe-thread-local-storage?view=msvc-170
- https://github.com/llvm/llvm-project/issues/98479
- https://lkml.iu.edu/hypermail/linux/kernel/0711.2/1245.html

Thanks to @xinydev

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -707,4 +707,23 @@ jobs:
         make check
 
 
+  test-linux-lto:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install clang, lld and llvm
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y clang lld llvm
+
+    - name: Build and test (whole-program ThinLTO, je_ prefix)
+      run: |
+        autoconf
+        CC=clang AR=llvm-ar NM=llvm-nm RANLIB=llvm-ranlib \
+          ./configure --with-jemalloc-prefix=je_ EXTRA_CFLAGS=-flto=thin
+        make -j3 EXTRA_LDFLAGS="-flto=thin -fuse-ld=lld"
+        make -j3 tests EXTRA_LDFLAGS="-flto=thin -fuse-ld=lld"
+        make check
+
 
--- a/Makefile.in
+++ b/Makefile.in
@@ -59,6 +59,7 @@ enable_autogen := @enable_autogen@
 enable_doc := @enable_doc@
 enable_shared := @enable_shared@
 enable_static := @enable_static@
+have_ucontext := @have_ucontext@
 enable_prof := @enable_prof@
 enable_zone_allocator := @enable_zone_allocator@
 enable_experimental_smallocx := @enable_experimental_smallocx@
@@ -345,6 +346,24 @@ ifeq (@enable_experimental_smallocx@, 1)
 TESTS_INTEGRATION += \
   $(srcroot)test/integration/smallocx.c
 endif
+# tcache_fiber_migration is a standalone LTO reproducer (issue #2890): the
+# dedicated rule below links it against the static archive (--whole-archive) so
+# the allocator fastpath inlines next to the swapcontext.  It needs the static
+# archive, GNU-ld --whole-archive semantics (the ELF gate), and working
+# ucontext fibers -- have_ucontext is a configure link test because musl
+# declares getcontext/makecontext/swapcontext but does not implement them, so
+# the ELF gate alone would break `make check` on Alpine/musl.  Built only when
+# the build uses LTO (-flto in CFLAGS): without inlining across the swapcontext
+# the bug cannot reproduce.
+ifeq (elf, $(ABI))
+ifeq ($(enable_static), 1)
+ifeq (1, $(have_ucontext))
+ifneq (,$(findstring -flto,$(CFLAGS)))
+TESTS_INTEGRATION += $(srcroot)test/integration/tcache_fiber_migration.c
+endif
+endif
+endif
+endif
 ifeq (@enable_cxx@, 1)
 CPP_SRCS := $(srcroot)src/jemalloc_cpp.cpp
 TESTS_INTEGRATION_CPP := $(srcroot)test/integration/cpp/basic.cpp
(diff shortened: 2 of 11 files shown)


spredolac:
@azat,  the way we use fibers now, [as Jay explained](https://github.com/jemalloc/jemalloc/issues/2890#issuecomment-5335512351), is such that we did not experience this issue.  Obviously we would be vulnerable to the same problem if we migrate fibers that touch TLS.  In the Boost doc you can see a little footnote that says "[Of course it would be problematic to migrate a fiber that relies on thread-local storage.](https://www.boost.org/doc/libs/latest/libs/fiber/doc/html/fiber/migration.html#fiber.migration.f1) 

Our position: **this isn't the right place/layer to fix it, and carrying it here has costs — but we want to help you and other not (silently) corrupt heaps.** 

`&tsd_tls` is well-defined C and storing that computation into a callee-saved register is a legal optimization, because within the C abstract machine the current thread does not change mid-function. Which thread is current is invisible to everyone involved. Boost.Context is following the ABI, jemalloc is writing correct C, and the compiler is applying a legal transform.  This is a situation when everyone is acting legally, but you end up with some damage unfortunately. 

Imho, MSVC solved this well with `/GT`, and clang/gcc have no equivalent as you pointed out.  We measured 3 versions of the fix: yours from this PR, version that is similar to yours but instead of runtime offset relies on link-time TLS relocation (`@tpoff`/`@gottpoff`), and version that is the simplest (no inline, asm volatile) with no arch specific asm that could be seen in [this qemu macro ](https://github.com/qemu/qemu/blob/master/include/qemu/coroutine-tls.h#L161-L163).  In terms of performance: link-time-TLS-relo < your_PR < qemu, and it is all reverse in terms of simplicity/maintainability.   Given all of the above, we would prefer the solution that is opt-in (so that we do not penalize majority of users that do not need to worry about this),  documents the limitation (access to tls will only be safe via jemalloc, but errno, or other libs in the app may still fail), adds CI config with the new option, and preferably has no per arch asm to keep our maintenance cost lower.  

Would you be willing to:
1. Add `--enable-experimental-fiber-safe-tls` option, default off, and byte-for-byte unchanged codegen when disabled, iow, the macro compiles to plain `&tlsvar`. we want programs that never migrate fiber to pay nothing and get that as default behavior.
2. Document scope clearly as a partial mitigation, not a fix for whole program, (for example `errno` still-exposed, other libs).
3. Add CI lane with the flag on, to prevent rot.
4. Contain in separate header for readability. Put `JEMALLOC_TLS_ADDR()` in a dedicated header (`tsd_tls_addr.h`) included from the `tsd_*.h` files, so those keep reading as `return JEMALLOC_TLS_ADDR(tsd_tls);`.
5. We would prefer no-arch dependent approach like QEMU's approach from `include/qemu/coroutine-tls.h`. In our production on x86 arch it was costing +0.86 .. +1.33%, while your version was +0.59 .. +1.11%, so we would prefer to have simpler code.  Please measure in your prod on real apps (as opposed to tight-loop malloc-free benchmarks)
 
To be explicit about what `experimental` means here: we reserve the right to remove it if it becomes a maintenance burden, or hopefully once a compiler-level mechanism makes it unnecessary.   Thank you for your work and patience - it took us a while to get opinion from several people.  Let us know what you think, please.

azat:
@spredolac thanks for taking a look and for such a detailed feedback!

There is no rush here indeed, better to have a patch everyone will be happy with. And thanks for pointing to the link-time-TLS-relo!

I will simplify patches and re-measure everything (and will address all other comments)

And a small note, AFAIR 40% was only in that micro bench, ClickHouse (the application in this case) does not have such huge degradation (allocations are not that frequent) - likely it was close to nothing.

The reason I used this "micrbench" for benchmark numbers to show real overhead not inside some specific workload.

guangli-dai:
@azat Thank you for the efforts here!

> And a small note, AFAIR 40% was only in that micro bench, ClickHouse (the application in this case) does not have such huge degradation (allocations are not that frequent) - likely it was close to nothing.

Regarding the perf measurement, we highly agree the benchmarks cannot be very accurate. We do not find a microbenchmark that can be very accurate towards production usage of jemalloc yet. That to say, we highly appreciate results from your real production traffic and we can conduct canaries on Meta's internal services once the update on the patch is done. When sharing production results, appreciate if you can share the malloc stats (or some metrics like nmalloc/s and nrequests/s) there as well: then we know the jemalloc usage pattern there for better understanding of this patch's impact.

azat:
@guangli-dai here is a simpler version https://github.com/jemalloc/jemalloc/pull/2990

I decided to submit it separately since this one is way faster, but this is probably not a big deal for a real world programs, at first glance I don't see any noticeable overhead for ClickHouse (I will test this change more deeply and will get back with the results if there will be something noticeable), so we can switch to a simpler version if it will be upstreamed

azat:
Closed in favor of https://github.com/jemalloc/jemalloc/pull/2990

PR #2949, opened 2026-07-01
Fix `--enable-prof-frameptr` when cross-compiling

ArniDagur:
Gate `--enable-prof-frameptr` behind `${host_os}` instead of `uname -s`. The latter reflects only the build machine, so cross-compiling for Linux from a non-Linux host (for example Darwin -> x86_64-unknown-linux-gnu) didn't work.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1511,8 +1511,8 @@ if test "x$backtrace_method" = "x" -a "x$enable_prof_libunwind" = "x1" ; then
   fi
 fi
 
-if test `uname -s` = "Linux"
-then
+case "${host_os}" in
+  linux*)
   AC_ARG_ENABLE([prof-frameptr],
     [AS_HELP_STRING([--enable-prof-frameptr], [Use optimized frame pointer unwinder for backtracing (Linux only)])],
   [if test "x$enable_prof_frameptr" = "xno" ; then
@@ -1534,9 +1534,11 @@ then
   else
     enable_prof_frameptr="0"
   fi
-else
+  ;;
+  *)
   enable_prof_frameptr="0"
-fi
+  ;;
+esac
 
 AC_ARG_ENABLE([prof-libgcc],
   [AS_HELP_STRING([--disable-prof-libgcc],


ArniDagur:
I have verified this fixes the issues I was having with Jemalloc + Rust + `cargo-zigbuild`

guangli-dai:
Thank you for the patch. However, we are going to remove the frame pointer based unwinder because previous bugs in libunwind have been fixed in https://github.com/libunwind/libunwind/pull/965 and https://github.com/libunwind/libunwind/pull/947. Also the current frame pointer unwinder needs more fix to work properly for some cases where thread boundaries are moved, e.g., fibers. Overall, we feel frame pointer unwinder is adding more complexity than we need it for now. Based on this consideration, we deliberately remove it from the manual and did not mention it in the release change log.

Do you have any special use case where frame pointer based unwinder is especially helpful and cannot be replaced by libunwind?

ArniDagur:
> Do you have any special use case where frame pointer based unwinder is especially helpful and cannot be replaced by libunwind?

Yes. The gcc and `libunwind` based unwinders crash for my application, while the frame pointer based unwinder doesn't. Essentially, as soon as we set `prof.active` to true, it segfaults as follows:
```
warning: 236    src/prof_sys.c: No such file or directory

Thread 14 "fdb-7.3.69-0" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffea5ff640 (LWP 152)]
0x00005555564f390b in prof_backtrace_impl (vec=0x7fffd952c4c0, len=len@entry=0x7fffea5fa498, max_len=128) at src/prof_sys.c:236
"
=== CRASH DETECTED ===
Thread 14 (Thread 0x7fffea5ff640 (LWP 152) "fdb-7.3.69-0"):
#0  0x00005555564f390b in prof_backtrace_impl (vec=0x7fffd952c4c0, len=len@entry=0x7fffea5fa498, max_len=128) at src/prof_sys.c:236
        p = <optimized out>
#1  0x00005555564f3773 in _rjem_je_prof_backtrace (tsd=tsd@entry=0x7fffea5fe810, bt=bt@entry=0x7fffea5fa490) at src/prof_sys.c:530
        prof_backtrace_hook = 0x5555564f37d0 <prof_backtrace_impl>
#2  0x00005555564e87d0 in _rjem_je_prof_tctx_create (tsd=0x7fffea5fe810) at src/prof.c:232
        bt = {vec = 0x7fffd952c4c0, len = 5}
        tdata = <optimized out>
#3  0x00005555564af3cf in prof_alloc_prep (prof_active=true, tsd=<optimized out>, sample_event=<optimized out>) at include/jemalloc/internal/prof_inlines.h:171
        ret = <optimized out>
#4  imalloc_body (sopts=<optimized out>, dopts=<optimized out>, tsd=<optimized out>) at src/jemalloc.c:1613
        prof_active = true
        tctx = <optimized out>
        alloc_ctx = <optimized out>
        sample_event = <optimized out>
        size = 16712
        ind = 37
        usize = 20480
        allocation = <optimized out>
        reentrancy_level = <optimized out>
        label_invalid_alignment = <optimized out>
#5  imalloc (sopts=<optimized out>, dopts=<optimized out>) at src/jemalloc.c:1750
        tsd = <optimized out>
#6  _rjem_je_malloc_default (size=16712) at src/jemalloc.c:1785
        sopts = {may_overflow = false, bump_empty_aligned_alloc = false, assert_nonempty_alloc = false, null_out_result_on_error = true, set_errno_on_error = true, min_alignment = 0, oom_string = <optimized out>, invalid_alignment_string = <optimized out>, slow = false, usize = false}
        dopts = {result = <optimized out>, usize = 20480, num_items = 1, item_size = <optimized out>, alignment = 0, zero = false, tcache_ind = 4294967294, arena_ind = 4294967295}
        ret = <optimized out>
#7  0x00007ffff0787433 in ssl3_setup_read_buffer () from /tmp/libfdb_c.so-JRpbwV
No symbol table info available.
#8  0x00007fffea5fa5a0 in ?? ()
No symbol table info available.
#9  0x0000000000000000 in ?? ()
No symbol table info available.
```
The crash seems to happen inside the FoundationDB shared library.

ArniDagur:
Correction: It also seems like `libunwind` avoids the crash. Previously, I think even though I was enabling `libunwind`, it silently wasn't being compiled in (since `libunwind` was unavailable) and thus falling back to gcc. Ideally the compilation should fail instead of silently performing a fallback.

guangli-dai:
> Correction: It also seems like `libunwind` avoids the crash. Previously, I think even though I was enabling `libunwind`, it silently wasn't being compiled in (since `libunwind` was unavailable) and thus falling back to gcc. Ideally the compilation should fail instead of silently performing a fallback.

Great to know, we will then continue to remove the frame pointer unwinder as planned. Regarding the failure in GCC, we could probably also take a look since we do plan to evaluate unwinders there to determine if we should still keep it in as an option. Appreciate if you can provide hardware type, gcc version you are using when the gcc unwinder is crashing. cc: @gctony

guangli-dai:
As mentioned above, closing this PR because we plan to remove the whole frame pointer unwinder soon. Still thanks to @ArniDagur for the patch and please feel free to re-open the PR or open a new issue if any other issues/questions relevant.

PR #2950, opened 2026-07-01
Use CLOCK_MONOTONIC for background thread sleep to prevent clock rollback stalls

antonio2368:
`background_thread_sleep()` computed its `pthread_cond_timedwait` deadline from `gettimeofday()` (`CLOCK_REALTIME`), so a backward wall-clock jump (eg NTP correction) could push the deadline far into the future and stall dirty-page purging for hours or days, which we observed in production as unbounded RSS growth. This switches the deadline computation to `nstime_init_update()` and sets `CLOCK_MONOTONIC` on the condvar via `pthread_condattr_setclock()` at both init sites (`background_thread_boot1`, `background_thread_postfork_child`), falling back to `CLOCK_REALTIME` where the monotonic clock or condattr setup is unavailable.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2224,6 +2224,43 @@ dnl Check if we have dlsym support.
   if test "x${je_cv_pthread_get_name_np}" = "xyes" ; then
     AC_DEFINE([JEMALLOC_HAVE_PTHREAD_GET_NAME_NP], [ ], [ ])
   fi
+  dnl Runtime-check pthread_condattr_setclock and clock_gettime with
+  dnl CLOCK_MONOTONIC to confirm the kernel actually supports the
+  dnl monotonic clock.
+  AC_CACHE_CHECK([whether pthread_condattr_setclock with CLOCK_MONOTONIC is usable],
+                 [je_cv_pthread_cond_timedwait_monotonic],
+                 AC_RUN_IFELSE([AC_LANG_PROGRAM(
+[[
+#include <pthread.h>
+#include <time.h>
+]], [[
+  pthread_condattr_t attr;
+  pthread_cond_t cond;
+  struct timespec ts;
+  int ret = pthread_condattr_init(&attr);
+  if (ret != 0) {
+    return 1;
+  }
+  ret = pthread_condattr_setclock(&attr, CLOCK_MONOTONIC);
+  if (ret != 0) {
+    pthread_condattr_destroy(&attr);
+    return 1;
+  }
+  ret = pthread_cond_init(&cond, &attr);
+  pthread_condattr_destroy(&attr);
+  if (ret != 0) {
+    return 1;
+  }
+  ret = clock_gettime(CLOCK_MONOTONIC, &ts);
+  pthread_cond_destroy(&cond);
+  return ret != 0;
+]])],
+                 [je_cv_pthread_cond_timedwait_monotonic=yes],
+                 [je_cv_pthread_cond_timedwait_monotonic=no],
+                 [je_cv_pthread_cond_timedwait_monotonic=no]))
+  if test "x${je_cv_pthread_cond_timedwait_monotonic}" = "xyes" ; then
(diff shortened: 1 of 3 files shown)


src/background_thread.c
```diff
@@ -751,7 +747,27 @@ background_thread_postfork_child(tsdn_t *tsdn) {
...
+		if (pthread_condattr_init(&cond_attr)) {
+			ret = pthread_cond_init(&info->cond, NULL);
+		} else if (pthread_condattr_setclock(&cond_attr,
+		    CLOCK_MONOTONIC)) {
+			/*
+			 * Fall back to default (CLOCK_REALTIME)
+			 * attributes if setclock fails.
+			 */
+			pthread_condattr_destroy(&cond_attr);
+			ret = pthread_cond_init(&info->cond, NULL);
```
guangli-dai: When this happens, there will be a mismatch in the clock. The deadline is using monotonic clock while the pthread is using a realtime clock. To solve the problem, it's probably better to add compile-time detection (through `configure.ac`) to make sure `pthread_condattr_setclock(&cond_attr, CLOCK_MONOTONIC)` won't fail (although from what I can find failure should be a very rare case). Then the deadline and the pthread can be determined using exactly the same logic (among different macros) to make sure there is no mismatch.

antonio2368: Yes, I went with this route is I understood the failure of these functions is extremely rare.
I applied the recommended changes, hopefully I did it correctly.

configure.ac
```diff
@@ -2224,6 +2224,19 @@ dnl Check if we have dlsym support.
   if test "x${je_cv_pthread_get_name_np}" = "xyes" ; then
     AC_DEFINE([JEMALLOC_HAVE_PTHREAD_GET_NAME_NP], [ ], [ ])
   fi
+  dnl Check if pthread_condattr_setclock with CLOCK_MONOTONIC is available.
+  JE_COMPILABLE([pthread_condattr_setclock(..., CLOCK_MONOTONIC)], [
```
guangli-dai: JE_COMPILABLE here only checks if the codes below compile and link properly, which means it checks for the existence of 1. CLOCK_MONOTONIC macro and 2. pthread_condattr_.* functions. It does not guarantee setting up `pthread_condattr_setclock` would succeed.

We can instead use `AC_RUN_IFELSE` here to confirm `pthread_condattr_.*` functions would succeed (although we still want to keep runtime check when setting, I like the current setup where we return true if the initialization fails in `background_thread_cond_init`). Can refer to [here](https://github.com/jemalloc/jemalloc/blob/dev/configure.ac#L558C22-L558C35) for `AC_RUN_IFELSE` usage.

I would also recommend checking `clock_gettime` there:
```
  ...
  int ret = pthread_condattr_init(&attr);
  if (ret != 0) {
    return 1;
  }
  ret = pthread_condattr_setclock(&attr, CLOCK_MONOTONIC);
 if (ret != 0) {
    pthread_condattr_destroy(&attr);
    return 1;
  }
  ret = pthread_cond_init(&cond, &attr);
  pthread_condattr_destroy(&attr);
  if (ret != 0) {
    return 1;
  }
  ret = clock_gettime(CLOCK_MONOTONIC, &ts);
  pthread_cond_destroy(&cond);
  return ret != 0;
```

antonio2368: Done. I just assumed that if the macro compiles, `pthread_condattr_setclock` cannot fail as unknown clock_id is the only reported failure. This seems safer nevertheless. 
What to about the cross-compile case? We just fallback to `no`?

guangli-dai: Yes, we can leave the fallback to `no` now. Cross-compiling users can always enable monotonic clocks by passing in extra compilation flags defining `JEMALLOC_HAVE_PTHREAD_COND_TIMEDWAIT_MONOTONIC`. 

Will merge once all signals are green.

guangli-dai:
Thank you @antonio2368 for the patch!

PR #2951, opened 2026-07-02
Initialize TSD tcache before enabling it

fzakaria:
During TSD initialization, tcache setup can briefly make the TSD appear nominal before the tcache bin data is fully initialized. If allocation reenters jemalloc during that window, for example from an asynchronous signal handler, the recursive allocation can take the fast path and dereference an uninitialized cache bin stack pointer.

At Meta, we experienced this in our LLVM ThinLTO workload, where worker threads could be asynchronously interrupted while performing their first allocation. The interrupt path performed an allocation before the interrupted thread had finished TSD/tcache bootstrap. That exposed a window where jemalloc had updated TSD state enough for a recursive allocation to take the fast path, but had not yet initialized the tcache bin backing storage.

* Initialize the automatic tcache data before setting `tcache_enabled` during TSD bootstrap. This
  matches the runtime `tcache_enabled_set()` ordering: allocate and initialize the cache bins first,
  then publish the enabled state and recompute TSD slow state. As a result, recursive allocation during
  bootstrap can no longer observe `tcache_enabled == true` while the cache bins are still unbacked.
* Adds a signal-based stress repro and a deterministic unit test for recursive allocation during TSD/tcache bootstrap.

Without the fix the stress repro fails accessing the invalid bin stack.

```console
> lldb --batch \
      -o 'process handle SIGUSR1 --stop false --pass true --notify false' \
      -o run \
      -k 'thread backtrace all' \
      -- ./test/stress/tsd_reentrant
(lldb) target create "./test/stress/tsd_reentrant"
Current executable set to '/home/fmzakari/code/github.com/jemalloc/jemalloc/test/stress/tsd_reentrant' (x86_64).
(lldb) process handle SIGUSR1 --stop false --pass true --notify false
NAME         PASS     STOP     NOTIFY
===========  =======  =======  =======
SIGUSR1      true     false    false
(lldb) run
Process 2704907 launched: '/home/fmzakari/code/github.com/jemalloc/jemalloc/test/stress/tsd_reentrant' (x86_64)
Process 2704907 stopped
* thread #2, name = 'tsd_reentrant', stop reason = signal SIGSEGV: address not mapped to object (fault address=0x0)
       frame #0: 0x00007ffff7c21b90 libjemalloc.so.2`cache_bin_alloc_impl(adjust_low_water=false, success=<unavailable>, bin=<unavailable>) at cache_bin.h:395:17 [inlined]
   392           * This may read from the empty position; however the loaded value won't
   393           * be used.  It's safe because the stack has one more slot reserved.
   394           */
-> 395          void          *ret = *bin->stack_head;
   396          cache_bin_sz_t low_bits = (cache_bin_sz_t)(uintptr_t)bin->stack_head;
   397          void         **new_head = bin->stack_head + 1;
   398
(lldb) thread backtrace all
  thread #1, name = 'tsd_reentrant'
       frame #0: 0x00007ffff74883da libc.so.6`__futex_abstimed_wait_common + 202
       frame #1: 0x00007ffff748cfe4 libc.so.6`__pthread_clockjoin_ex + 324
       frame #2: 0x00000000004140f0 tsd_reentrant`test_tsd_bootstrap_signal_reentrant at tsd_reentrant.c:121:2
       frame #3: 0x00000000004c5add tsd_reentrant`p_test_impl(do_malloc_init=true, do_reentrant=false, t=(tsd_reentrant`test_tsd_bootstrap_signal_reentrant at tsd_reentrant.c:110:1), ap=0x00007fffffffd278) at test.c:154:3
       frame #4: 0x00000000004c6299 tsd_reentrant`p_test_no_reentrancy(t=<unavailable>) at test.c:214:8
       frame #5: 0x00007ffff742a610 libc.so.6`__libc_start_call_main + 128
       frame #6: 0x00007ffff742a6c0 libc.so.6`__libc_start_main@@GLIBC_2.34 + 128
       frame #7: 0x0000000000413cc5 tsd_reentrant`_start + 37
* thread #2, name = 'tsd_reentrant', stop reason = signal SIGSEGV: address not mapped to object (fault address=0x0)
   * frame #0: 0x00007ffff7c21b90 libjemalloc.so.2`cache_bin_alloc_impl(adjust_low_water=false, success=<unavailable>, bin=<unavailable>) at cache_bin.h:395:17 [inlined]
     frame #1: 0x00007ffff7c21b75 libjemalloc.so.2`cache_bin_alloc_easy(success=<unavailable>, bin=<unavailable>) at cache_bin.h:433:9 [inlined]
     frame #2: 0x00007ffff7c21b75 libjemalloc.so.2`imalloc_fastpath(fallback_alloc=0x000000000001f610, size=16) at jemalloc_internal_inlines_c.h:356:8
     frame #3: 0x00007ffff7c21b30 libjemalloc.so.2`malloc(size=16) at jemalloc.c:811:14
     frame #4: 0x0000000000413fb8 tsd_reentrant`on_signal(signo=<unavailable>) at tsd_reentrant.c:57:12
     frame #5: 0x00007ffff743fc60 libc.so.6`__restore_rt
     frame #6: 0x00007ffff748d21c libc.so.6`__pthread_kill_implementation + 284
     frame #7: 0x00007ffff743fbb6 libc.so.6`raise + 22
     frame #8: 0x00007ffff7cc851f libjemalloc.so.2`je_tsd_tcache_enabled_data_init(tsd=0x00007ffff63fe710) at tcache.c:1055:3
     frame #9: 0x00007ffff7ccae4a libjemalloc.so.2`tsd_data_init(tsd=0x00007ffff63fe710) at tsd.c:136:9 [inlined]
     frame #10: 0x00007ffff7ccae1d libjemalloc.so.2`je_tsd_fetch_slow(tsd=0x00007ffff63fe710, minimal=<unavailable>) at tsd.c:215:4
     frame #11: 0x00007ffff7c1f96a libjemalloc.so.2`tsd_fetch_impl(minimal=false, init=true) at tsd.h:121:10
     frame #12: 0x0000000000413da2 tsd_reentrant`worker_start(arg=<unavailable>) at tsd_reentrant.c:79:12
     frame #13: 0x00007ffff748b3f9 libc.so.6`start_thread + 825
     frame #14: 0x00007ffff75105d0 libc.so.6`__clone3 + 48
```

Assisted-By: OpenAI Codex


spredolac:
Hey, thanks for working on this.  It looks to me like real issue and your stress test shows the problem well.  

* Your fix is okay, but I am wondering if this fix is what we should go with.  I'd just suggest a slightly smaller one that
  targets the root cause. 
  The problem/window exists because bootstrap flips tcache_enabled on before the cache bins are allocated. The runtime path, `tcache_enabled_set()`, already does it the safe way (allocate bins, then enable); if `tsd_tcache_enabled_data_init()` matches that order, the "enabled but unbacked" state never exists and no
  reentrancy handling is needed.  You can have `tsd_tcache_enabled_data_init` do something like this:
```
     tcache_default_settings_init(tsd_tcache_slowp_get(tsd));
     bool err = false;
       if (opt_tcache) {
               /* Initialize the bins before advertising the tcache. */
               err = tsd_tcache_data_init(
                   tsd, NULL, tcache_get_default_ncached_max());
       }
       /* Enable last: a recursive alloc must never see enabled with unbacked bins. */
       tsd_tcache_enabled_set(tsd, opt_tcache && !err);
       tsd_slow_update(tsd);
       return err;
```


With this the tsd.c changes can be dropped — the reentrancy guard isn't needed once the invariant holds, and keeping it also has the side effect of binding the bootstrap tcache to arena 0 (then reassociating) instead of the thread's real arena. (Re-initialization isn't a concern either: `tsd_fetch_slow` sets the state to nominal before calling `tsd_data_init`, so a reentrant fetch just takes the nominal_slow "no work" path.)

 On the tests: test/stress isn't run by make check, so the deterministic unit test is really what guards against regressions — nice to have. Given that, you could drop the stress test and wrap the hook in `#if defined(JEMALLOC_JET) || defined(JEMALLOC_UNIT_TEST)` (the guard `test_hooks.h` already uses for the other hooks), so it adds no symbol to the production library.   Your `test_tsd_reentrant_bootstrap` covers the `uninitialized ->nominal path`,  would be nice to add case that exercises `minimal_initialized -> nominal promotion` (a thread that deallocates before it ever allocates — e.g. a reclaim thread). I confirmed that path crashes without the fix and passes with it too, so it's worth a second case:
 
   <details><summary>minimal→nominal case</summary>

  ```
static bool tsd_bootstrap_min_promote_hook_ran;

  static void
  tsd_bootstrap_min_promote_hook(void) {
        tsd_bootstrap_min_promote_hook_ran = true;
        test_hooks_tsd_bootstrap_hook = NULL;

        void *p = malloc(16);
        expect_ptr_not_null(p, "Unexpected recursive malloc() failure");
        free(p);
  }

  static void *
  thd_start_min_promote_bootstrap(void *arg) {
        (void)arg;

        /*
         * Force the thread into the minimal_initialized state (as a
         * dealloc-only thread would), so the following malloc() promotes it to
         * nominal and runs tcache bootstrap through the minimal->nominal path.
         */
        tsd_t *tsd = tsd_fetch_min();
        if (tsd_state_get(tsd) != tsd_state_minimal_initialized) {
                /* Allocation happened implicitly; can't exercise this path. */
                return NULL;
        }
        expect_ptr_not_null(p, "Unexpected malloc() failure");
        free(p);
        test_hooks_tsd_bootstrap_hook = NULL;

        expect_true(tsd_bootstrap_min_promote_hook_ran,
            "TSD bootstrap hook should have executed on promotion");
        return NULL;
  }

  TEST_BEGIN(test_tsd_reentrant_bootstrap_min_promote) {
        thd_t thd;

        thd_create(&thd, thd_start_min_promote_bootstrap, NULL);
        thd_join(thd, NULL);
  }
  TEST_END

```  
</details>

spredolac (commented):
Also CI must be green.

guangli-dai:
Agree with what @spredolac says, swapping the order during tcache initialization would be better (neater and directly eliminates the gap) and interrupting allocation would also fall back to arena level without causing crashes. 

Besides, I assume the efficiency here is not a big deal? In the worst case, if the initialization is interrupted for a long time and there are lots of allocations made during the interrupt, with no tcache, the perf would be bad and we should prevent such cases if we care about CPU perf there.

spredolac:
Final version looks good. Can you just squash the commits so that we do not have this hostory of adding and removing reentrancy lines?  Also, that would make it easier to deal with reverts if needed.

spredolac:
I messed up merging non-squashed and repo does not allow force push to fix it.

fzakaria:
> I messed up merging non-squashed and repo does not allow force push to fix it.

Sorry about that since I had asked you to squash on approval.
On the bright side, I have extra commits in jemalloc which is a personal badge of honor.

stanhu:
Is it possible to cut a release with this fix? I think we are seeing seg faults in our applications due to this bug after updating to 5.3.1.

guangli-dai:
> Is it possible to cut a release with this fix? I think we are seeing seg faults in our applications due to this bug after updating to 5.3.1.

We have a release planned this half. Before that, please feel free to use the dev branch.

PR #2952, opened 2026-07-06
Background thread module refactor

guangli-dai:
Due to historical reasons, background thread, pac decay, and arena are coupled tightly while HPA has a relatively clean interaction with the background thread module. This PR:
1. symmetrizes PAC's deferred-work execution; 
2. move constants and 
3. functions out of arenas into pa module (a thin dispatch in the page allocator level), PAC, and HPA correspondingly;
4. Move the background lifecycle control into the background thread module
5. Remove arena module's deferral relevant functions and change them to call pa directly.

This enables:
1. A clean and symmetric interface for page allocator's deferral operations;
2. Easier future rerouting and arena-less exploration (targeting HPA better locality) in the page allocator level;

Tests:
Improve the test coverage for PAC and background thread relevant modules.


include/jemalloc/internal/pac.h
```diff
@@ -231,6 +231,32 @@ bool pac_maybe_decay_purge(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
...
+/*
+ * Result of a pac_do_deferred_work() pass.  Reported per decay state so the
+ * caller (on the user-inline path) can decide, per state, whether to notify the
+ * background thread.  npages_new is the per-state epoch backlog delta, only
+ * meaningful when the corresponding *_epoch_advanced is true.
+ */
+typedef struct pac_deferred_work_result_s pac_deferred_work_result_t;
+struct pac_deferred_work_result_s {
+	bool   dirty_epoch_advanced;
+	size_t dirty_npages_new;
```
spredolac, line 250: nit: reorder fields  for less padding

include/jemalloc/internal/pac.h
```diff
@@ -247,6 +273,13 @@ bool    pac_decay_ms_set(tsdn_t *tsdn, pac_t *pac, extent_state_t state,
        ssize_t decay_ms, pac_purge_eagerness_t eagerness);
 ssize_t pac_decay_ms_get(pac_t *pac, extent_state_t state);
 
+/* Whether muzzy decay should be skipped entirely (no muzzy pages, no decay). */
+static inline bool
+pac_dont_decay_muzzy(pac_t *pac) {
+	return ecache_npages_get(&pac->ecache_muzzy) == 0
```
spredolac: Easier to avoid negation and then double negation down in the code `!dont `. Can you invert this into  `pac_should_decay_muzzy` instead?

src/arena.c
```diff
@@ -339,7 +337,7 @@ arena_handle_deferred_work(tsdn_t *tsdn, arena_t *arena) {
 	    tsdn_witness_tsdp_get(tsdn), WITNESS_RANK_CORE, 0);
 
 	if (decay_immediately(&arena->pa_shard.pac.decay_dirty)) {
```
spredolac: not for this PR, but this should be hidden in `pac.h` which already includes `decay.h` as inline so that we at least avoid last `'.'`

   `if(pac_decay_immediately(&arena->pa_shard.pac))`

src/arena.c
```diff
@@ -506,74 +504,67 @@ arena_decay_ms_get(arena_t *arena, extent_state_t state) {
...
+		 * like thread death, or manual purge calls).  This blocking
+		 * flush-and-fully-decay path is kept separate from the deferred
+		 * (all=false) path below.
 		 */
 		pa_shard_flush(tsdn, &arena->pa_shard);
-	}
-	if (arena_decay_dirty(tsdn, arena, is_background_thread, all)) {
+		pac_decay_all_now(
+		    tsdn, &arena->pa_shard.pac, extent_state_dirty);
+		if (!pac_dont_decay_muzzy(&arena->pa_shard.pac)) {
```
spredolac: This is where inverted question will help in readability

src/arena.c
```diff
@@ -1585,17 +1376,16 @@ arena_get_ehooks(const arena_t *arena) {
...
+	 * than reaching into info->mtx directly.
+	 */
+	background_thread_serialize_lock(tsd, arena);
 	/* No using the HPA now that we have the custom hooks. */
 	pa_shard_disable_hpa(tsd_tsdn(tsd), &arena->pa_shard);
 	extent_hooks_t *ret = base_extent_hooks_set(arena->base, extent_hooks);
-	if (have_background_thread) {
-		malloc_mutex_unlock(tsd_tsdn(tsd), &info->mtx);
-	}
+	background_thread_serialize_unlock(tsd, arena);
```
spredolac: You can do `arena_ind_get` here.

src/pac.c
```diff
@@ -779,9 +781,219 @@ pac_maybe_decay_purge(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
...
+/*
+ * Notify the background thread that a decay epoch advanced: if it sleeps
+ * indefinitely, wake it now; otherwise wake it early when the projected backlog
+ * crosses the purge threshold before its next scheduled wakeup.  Non-blocking
+ * (trylocks info->mtx) to keep the application free path cheap.
+ */
+static void
+pac_maybe_wake_bg(tsdn_t *tsdn, pac_t *pac, decay_t *decay, size_t npages_new) {
+	background_thread_info_t *info =
+	    background_thread_info_get(base_ind_get(pac->base));
```
spredolac: This is repeated few times - perhaps, you want to make a function (and document there that  `base_ind_get(pac->base) == arena_ind`)?

include/jemalloc/internal/pac.h
```diff
@@ -231,6 +239,56 @@ bool pac_maybe_decay_purge(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
...
+/*
+ * Fully decay the extents of the given state, acquiring decay->mtx internally.
+ */
+void pac_decay_all_now(tsdn_t *tsdn, pac_t *pac, extent_state_t state);
+
+/*
+ * Decide the unforced decay-purge eagerness from the caller context, i.e.,
+ * background-thread state.
+ */
+pac_purge_eagerness_t pac_decide_purge_eagerness(bool is_background_thread);
```
spredolac: Make it static, only used in pac

include/jemalloc/internal/pac.h
```diff
@@ -231,6 +239,56 @@ bool pac_maybe_decay_purge(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
...
+	size_t muzzy_npages_new;
+};
+
+/*
+ * Non-forced deferred decay-purge for both the dirty and muzzy states, at the
+ * given eagerness; corresponding decay->mtx are acquired internally.  Reports
+ * per-state epoch-advance in *result.  Exposed for deterministic testing;
+ * production callers use pac_do_deferred_work.
+ */
+void pac_decay_deferred(tsdn_t *tsdn, pac_t *pac,
```
spredolac: If this is exposed for testing only (do not see it yet) it should be JET_EXTERN to avoid adding symbol as public

src/pa.c
```diff
@@ -249,9 +257,33 @@ pa_shard_set_deferral_allowed(
 	}
 }
 
+/*
+ * Reactive hook: runs on an APPLICATION thread after a pa op reported
+ * deferred_work_generated.  Must NEVER be called on the background thread.
+ */
```
spredolac: nit: We have comment in header and here, probably merge content?

spredolac (approved):
Looks good.  I gave you  few nits, see if it makes sense

include/jemalloc/internal/pac.h
guangli-dai: Reordered to reduce the struct size from 32B to 24B. Putting the two `size_t`s first to leave the pads in the trailing position although it should not matter here.

include/jemalloc/internal/pac.h
guangli-dai: Yes, test only, modified.

src/pa.c
guangli-dai: Merged this into the header.

src/pac.c
guangli-dai: Good idea, considering the mapping might change in the future, also highlighted `base_ind_get(pac->base) == arena_ind` is true under the current contract but might be changed.

src/arena.c
guangli-dai: Added in this PR.

include/jemalloc/internal/background_thread.h
```diff
@@ -5,6 +5,9 @@
 #include "jemalloc/internal/base.h"
 #include "jemalloc/internal/mutex.h"
 
+/* Forward decl (used by the serialize-lock API); avoid depending on arena.h. */
+typedef struct arena_s arena_t;
```
spredolac: You can remove this now when we addressed lock interface.

src/pa.c
```diff
@@ -250,8 +258,28 @@ pa_shard_set_deferral_allowed(
...
+void
+pa_shard_do_deferred_work(
+    tsdn_t *tsdn, pa_shard_t *shard, bool is_background_thread) {
+	pac_do_deferred_work(tsdn, &shard->pac, is_background_thread);
+	/*
+	 * Application threads self-throttle HPA deferred work inline from their
+	 * own alloc/dalloc path (hpa_shard_maybe_do_deferred_work, capped), so
+	 * only drive it (forced, uncapped) from here on the background thread.
+	 */
+	if (is_background_thread && pa_shard_uses_hpa(shard)) {
```
spredolac, line 282: I asked  claude to generate the test for this in `test/unit/pa.c` since comment states that it should be forced.  It could do it with hap_hooks like hpa tests, but perhaps we can just rely on purge_passes to avoid boiler plate just for it.  Feel free to ignore this if you wnat to land what you have so far. 

```
TEST_BEGIN(test_do_deferred_work_hpa_gate) {
        test_skip_if(!hpa_supported());

        hpa_shard_opts_t opts = {
            /* slab_max_alloc */ HUGEPAGE,
            /* hugification_threshold */ HUGEPAGE,
            /* dirty_mult */ FXP_INIT_PERCENT(25),
            /* deferral_allowed */ true,
            /* hugify_delay_ms */ 10 * 1000,
            /* hugify_sync */ false,
            /* min_purge_interval_ms */ 0,
            /* purge_threshold */ 1,
            /* min_purge_delay_ms */ 0,
            /* hugify_style */ hpa_hugify_style_lazy};

        /* init_test_data gains a `bool hpa` passed to pa_central_init. */
        test_data_t *data = init_test_data(-1, -1, /* hpa */ true);
        pa_shard_t  *shard = &data->shard;
        sec_opts_t   sec_opts = SEC_OPTS_DEFAULT;
        sec_opts.nshards = 0;
        assert_false(pa_shard_enable_hpa(TSDN_NULL, shard, &opts, &sec_opts),
            "Unexpected pa_shard_enable_hpa() failure");
        expect_true(atomic_load_b(&shard->use_hpa, ATOMIC_RELAXED),
            "HPA should be enabled");

        tsdn_t *tsdn = tsd_tsdn(tsd_fetch());
        bool    dwg = false;

        /* Fill a hugepage, then free half -> dirty pages eligible for purge. */
        edata_t *edatas[HUGEPAGE_PAGES];
      for (int i = 0; i < (int)HUGEPAGE_PAGES; i++) {
                edatas[i] = pa_alloc(tsdn, shard, PAGE, PAGE, /* slab */ false,
                    /* szind */ 0, /* zero */ false, /* guarded */ false, &dwg);
                expect_ptr_not_null(edatas[i], "Unexpected pa_alloc() failure");
        }
        for (int i = 0; i < (int)HUGEPAGE_PAGES / 2; i++) {
                pa_dalloc(tsdn, shard, edatas[i], &dwg);
        }

        /* Application path must NOT force HPA deferred purge. */
        uint64_t npurge_before = shard->hpa.stats.npurge_passes;
        pa_shard_do_deferred_work(tsdn, shard, /* is_background_thread */ false);
        expect_u64_eq(shard->hpa.stats.npurge_passes, npurge_before,
            "Application path must not force HPA deferred purge");

        /* Background-thread path MUST force HPA deferred purge. */
        pa_shard_do_deferred_work(tsdn, shard, /* is_background_thread */ true);
        expect_u64_gt(shard->hpa.stats.npurge_passes, npurge_before,
            "Background-thread path must force HPA deferred purge");

        for (int i = (int)HUGEPAGE_PAGES / 2; i < (int)HUGEPAGE_PAGES; i++) {
                pa_dalloc(tsdn, shard, edatas[i], &dwg);
        }
        destroy_test_data(data);
}
TEST_END
```

spredolac (approved):
still accepted

include/jemalloc/internal/hpa.h
```diff
@@ -13,6 +13,14 @@
 #include "jemalloc/internal/psset.h"
 #include "jemalloc/internal/sec.h"
 
+/*
+ * HPA-specific deferral interval bounds (currently unused).  Deferral tuning
```
swkrueger, line 17: nit: Could this be removed since it is unused?

src/pac.c
```diff
@@ -779,9 +781,238 @@ pac_maybe_decay_purge(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
...
+ *       Splitting it would require copying these values across a new API while
+ *       still holding both locks, adding interface surface for no behavioral
+ *       gain.
+ *   (b) It runs on the application FREE path (pac_do_deferred_work ->
+ *       pac_maybe_wake_bg), where perf parity matters most.
+ *
+ * The info->mtx-outer / decay->mtx-inner trylock nesting is load-bearing for
+ * lock ordering; keep it.
+ *
+ * N1 (parity trap): when npages_new == 0 (e.g. an epoch advanced with no new
```
swkrueger: Nit: What does N1 mean?

src/pac.c
```diff
@@ -779,9 +781,238 @@ pac_maybe_decay_purge(tsdn_t *tsdn, pac_t *pac, decay_t *decay,
...
+	    > PAC_DECAY_PURGE_NPAGES_THRESHOLD;
+}
+
+/*
+ * The PAC's base index is, under the current PA/PAC construction contract, the
+ * owning arena index (pa_shard_init asserts base_ind_get(base) == ind).  This
+ * is a current construction invariant, not a permanent PAC guarantee.
+ */
+static unsigned
+pac_ind_get(const pac_t *pac) {
```
swkrueger, line 896: nit: could and should this be inlined?

include/jemalloc/internal/hpa.h
guangli-dai: We can remove it. But we are going to re-evaluate and improve HPA purging soon with the HPA global hugepage pool introduced. I will leave this here for now for future reference.

src/pa.c
guangli-dai: Yes, we should add a test for this. I will do this in a follow-up PR.

src/pac.c
guangli-dai: Good catch, it's a phase I used during development and should not appear in the comment. Removed.

src/pac.c
guangli-dai: We could but shouldn't in my opinion. Reasons:
1. Such small static functions are anyway inlined under `-O2` compiler options, which is what production should use, so it's not a perf-critical choice anyway.
2. non-inlined static function (in .c files), when not used, will be report with flag [`-Wunused-function`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wunused-function), so for better maintainability, we should use `static` instead. There are some `static inline` in this .c file and we probably should clean then although not high priority.

PR #2953, opened 2026-07-07
Fix malloc_getcpu on macOS to read the current CPU number correctly

Algunenano:
On macOS `malloc_getcpu` reads the CPU number from the low 3 bits of
`tpidrro_el0`. That layout no longer holds on Apple Silicon: those bits read
back as 0, so `malloc_getcpu` returns 0 for every thread. With
`percpu_arena:percpu` this either funnels all allocations into arena 0, or the
arena is disabled at init via the `malloc_getcpu() < 0` guard.

### Fix

Read the CPU number the same way as libplatform's `_os_cpu_number`:
- arm64: low 12 bits of `tpidr_el0`
- x86: low 12 bits of the IDT base, read with `sidt`

The `__APPLE__` branch is also moved ahead of the `rdtscp` branch, so Apple x86
uses `sidt` (what `_os_cpu_number` relies on) rather than `rdtscp`.

Reference: https://github.com/apple-oss-distributions/xnu/blob/main/libsyscall/os/tsd.h

### Verification

On a 12-CPU Apple M2 Pro, with `MALLOC_CONF="percpu_arena:percpu"`:

| | before | after |
|---|---|---|
| `opt.percpu_arena` | `disabled` | `percpu` |
| distinct `thread.arena` | `{0}` | `{0,1,…,11}` |

Before the fix every thread mapped to arena 0; after it, allocations spread
across all 12 per-CPU arenas.

### DIFF
--- a/include/jemalloc/internal/jemalloc_internal_inlines_a.h
+++ b/include/jemalloc/internal/jemalloc_internal_inlines_a.h
@@ -19,9 +19,17 @@ malloc_getcpu(void) {
 #elif defined(JEMALLOC_HAVE_SCHED_GETCPU)
 	return (malloc_cpuid_t)sched_getcpu();
 #elif defined(__APPLE__)
-	/* No sched_getcpu() on macOS; read the CPU number like _os_cpu_number(), see
+	/*
+	 * No sched_getcpu() on macOS; read the CPU number like _os_cpu_number()
+	 * does, from the low 12 bits of tpidr_el0 (arm64) or the IDT base (x86).
+	 * The 0xfff mask is xnu's __TPIDR_CPU_NUM_MASK, kept in sync with
+	 * _os_cpu_number in libsyscall/os/tsd.h (the kernel counterpart is
+	 * MACHDEP_TPIDR_CPUNUM_MASK in osfmk/arm64/machine_machdep.h):
 	 * https://github.com/apple-oss-distributions/xnu/blob/main/libsyscall/os/tsd.h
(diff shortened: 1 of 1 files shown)


include/jemalloc/internal/jemalloc_internal_inlines_a.h
```diff
@@ -18,15 +18,26 @@ malloc_getcpu(void) {
 	return GetCurrentProcessorNumber();
 #elif defined(JEMALLOC_HAVE_SCHED_GETCPU)
 	return (malloc_cpuid_t)sched_getcpu();
+#elif defined(__APPLE__)
+	/* No sched_getcpu() on macOS; read the CPU number like _os_cpu_number(), see
```
guangli-dai: If I understand correctly, looking into the past version of xnu, the current detection would fail on macOS prior to 12? Those old versions are officially retired by apple so I am ok with that, but maybe add that into the comments for future references?

Besides a nit: multi-line comment formats are:
```
/*
 * multi-line comments
 * like this
 */
```

Algunenano: Yes, you are right. This changed in macOS 12 (and has been stable since then). I'll adjust the comment

guangli-dai:
Merged, @Algunenano thank you so much for your contribution!

mrpippy:
Instead of this fix, the public API `pthread_cpu_number_np()` (added in macOS 11) should be used to discover the CPU number

PR #2954, opened 2026-07-10
test: handle profiled xallocx no-grow in extent test

booxter:
With profiling active, xallocx may decline to grow an allocation that is not page-aligned.  The profiling path can require sampled xallocx growth to use page-aligned pointers; otherwise it can return the old usable size without committing the purged tail.

Observed on Darwin, where decommit/commit are real VM state changes. The shrink/purge half succeeds and records a decommit, then the profiling-limited grow skips the matching commit.

This is not Darwin-specific.  Linux no-overcommit should expose the same mismatch.  Usual Linux overcommit behavior papers it over by making commit/decommit hooks report unsupported, so both sides of the old equality check stay false.

Relax the default-path assertion only while profiling is active and we can prove xallocx hasn't reached hooks. Add a page-aligned case that always requires xallocx shrink and grow to succeed, preserving commit/merge coverage.

### DIFF
--- a/test/integration/extent.c
+++ b/test/integration/extent.c
@@ -5,15 +5,77 @@
 #include "jemalloc/internal/arena.h"
 
 static void
+test_decommit_commit(unsigned arena_ind, int flags, size_t large0,
+    size_t *purge_mib, size_t purge_miblen, bool expect_xallocx_success,
+    bool allow_prof_no_grow) {
+	void  *p;
+	bool   prof_no_grow;
+	bool   xallocx_success_b, xallocx_success_c;
+
+	try_dalloc = false;
+	try_decommit = true;
+	p = mallocx(large0 * 2, flags);
+	expect_ptr_not_null(p, "Unexpected mallocx() error");
+	did_decommit = false;
+	did_commit = false;
+	called_split = false;
+	did_split = false;
+	did_merge = false;
+	xallocx_success_b = (xallocx(p, large0, 0, flags) == large0);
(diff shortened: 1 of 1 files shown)


swkrueger (approved):

test/integration/extent.c
```diff
@@ -4,16 +4,71 @@
...
+	if (expect_xallocx_success) {
+		expect_true(xallocx_success_b, "Expected xallocx shrink");
+		expect_true(xallocx_success_c, "Expected xallocx grow");
+	}
+	if (did_split && xallocx_success_c) {
+		expect_b_eq(
+		    did_decommit, did_commit, "Expected decommit/commit match");
+	}
+	if (xallocx_success_b && xallocx_success_c) {
+		expect_true(did_merge, "Expected merge");
```
guangli-dai, line 45: We should also try to distinguish two cases: 1. xallocx failure due to profiling enablement; 2. xacllocx failure due to other reasons which should still clean up and have `did_decommit == did_commit`. The current version seems to regress the coverage by skipping entirely at failure.

booxter: If I understand you correctly, you mean that if we can prove that hooks were never reached, only then we can skip the assertion. I pushed an updated version of the patch, should hopefully cover this concern. Let me know.

booxter: @guangli-dai if you could revisit this, thanks. The latest version should address the concern.

guangli-dai (approved):

guangli-dai:
Merged. Thanks @booxter a lot for the patch!

PR #2955, opened 2026-07-10
Remove frameptr-bsed unwinder.

gctony:
frameptr-based unwinder was an interim solution while issues with libunwind were resolved. It's better if we remove it, so we don't have to support it moving forward.


.github/workflows/linux-ci.yml
```diff
@@ -60,11 +60,6 @@ jobs:
               CXX: g++
               CONFIGURE_FLAGS: --with-lg-page=16
               EXTRA_CFLAGS: "-Werror -Wno-array-bounds"
-          - env:
-              CC: gcc
-              CXX: g++
-              CONFIGURE_FLAGS: "--enable-prof --enable-prof-frameptr"
```
guangli-dai, line 66: Not for this PR, but I realize we don't have libunwind tests in CI, we should add that as a part of the unwinder sort out as well (should be a separate PR).

.github/workflows/linux-ci.yml
```diff
@@ -60,11 +60,6 @@ jobs:
               CXX: g++
```
guangli-dai, line 60: There are some other dead functions like [this](https://github.com/jemalloc/jemalloc/blob/fb5499aa9c13aa7c0e11041009ae894c584b64bd/include/jemalloc/internal/malloc_io.h#L93) introduced during the development of this unwinder. Let's remove them together.

guangli-dai (approved):
Overall LGTM. Two more changes suggested. The CI change should definitely be another PR, feel free to do the dead function removal in this PR or a new one.

gctony:
> Overall LGTM. Two more changes suggested. The CI change should definitely be another PR, feel free to do the dead function removal in this PR or a new one.

"the CI change" - you mean removing the tests that set --enable-prof-frameptr from linux-ci.yml and .travix.yml? I can definitely land that first. See https://github.com/jemalloc/jemalloc/pull/2961

.github/workflows/linux-ci.yml
gctony: I'll remove `malloc_lseek()`. Looks as if the rest of the functions in that file are still used elsewhere (at least, for now).

.github/workflows/linux-ci.yml
gctony: Agreed re: separate PR.

guangli-dai (approved):

PR #2957, opened 2026-07-15
Allow resuming per-CPU arena selection via thread.arena

### LINKED ISSUE #1236: Not possible to bind threads back to auto arenas when using percpu_arena.
Currently thread.arena control does not allow binding threads to auto arena range (i.e. per cpu arenas). If we bind a thread to an manual arena, then we have no way to bind them back to auto arenas.

Algunenano:
## Motivation

`percpu_arena` binds each thread to the arena for the CPU it currently runs on, keeping allocations CPU-local and avoiding cross-CPU contention. Some workloads want to combine this with a small number of manually created, long-lived arenas: general allocation stays on the per-CPU arenas, but a bounded region of work routes its allocations into a dedicated arena. This isolates that memory (typically longer-lived and with a different allocation pattern) from the general per-CPU churn, so it can be purged or accounted for as a unit and does not fragment the per-CPU arenas. After that region, the thread should return to normal per-CPU allocation.

Individual allocations can already be directed to a specific arena with `mallocx(..., MALLOCX_ARENA(a))`. That is enough when every allocation call is under your control. It does not cover a whole region of code that performs many allocations, especially when they go through plain `malloc` / `operator new` or through existing library code that you do not want to (or cannot) modify to pass `MALLOCX_ARENA`. For that, the natural tool is to switch the calling thread's arena for the duration of the region, so every allocation in it, including in unmodified code, lands in the dedicated arena.

`thread.arena` is that tool, but under `percpu_arena` it is one-directional. A thread can bind to a manually created arena (an index at or above the per-CPU auto range), but there is no way back:

- `thread_arena_ctl` returns `EPERM` for any index within the per-CPU auto range when `percpu_arena` is enabled.
- `arena_choose_impl` only re-selects a per-CPU arena for threads whose current arena is already within the auto range, so a thread bound to a manual arena is never reclaimed by percpu.

Once a thread touches a manual arena it stays pinned to it for its whole life, and every later allocation lands in that arena instead of following the CPU. On a thread pool this is especially harmful: any worker that ever runs the dedicated-arena region is permanently removed from per-CPU management.

## Change

When `percpu_arena` is enabled, treat setting `thread.arena` to an index within the per-CPU auto range as a request to resume automatic per-CPU selection, instead of returning `EPERM`. The thread is handed back to percpu management (rebound to its current CPU's arena via `percpu_arena_update`) and the call returns 0.

Binding to a manual arena (index at or above the auto range) is unchanged. The requested index is advisory: under `percpu` the thread is governed by the CPU it runs on, so it resumes on the current CPU's arena regardless of the value passed. This is migration safe. If the thread moved CPUs between switching to the manual arena and resuming, it lands on the arena for the CPU it is actually running on.

## Tests

- New `test/unit/percpu_arena_resume` (with a `.sh` setting `percpu_arena:percpu`) exercises the round trip: bind to a manual arena, resume, and assert the thread is back in the per-CPU range. It skips when percpu is disabled.
- `test_thread_arena` in `test/unit/mallctl.c` asserted the old `EPERM`; it now asserts the resume succeeds.

Closes #1236

### DIFF
--- a/test/unit/percpu_arena_resume.c
+++ b/test/unit/percpu_arena_resume.c
@@ -1,0 +1,74 @@
+#include "test/jemalloc_test.h"
+
+/*
+ * When percpu_arena is enabled a thread is bound to a manually created arena
+ * (an index at or above the per-CPU auto range) to route a bounded region of
+ * work to a dedicated arena, then handed back to automatic per-CPU selection.
+ *
+ * Setting thread.arena to an index within the per-CPU auto range is the signal
+ * to resume per-CPU management. Without it a thread bound to a manual arena is
+ * never reclaimed by percpu (see arena_choose_impl), so it would stay pinned to
+ * the manual arena forever and its later allocations would land there instead
+ * of following the CPU.
+ */
+TEST_BEGIN(test_thread_arena_resume_percpu) {
+	test_skip_if(!have_percpu_arena
+	    || !PERCPU_ARENA_ENABLED(opt_percpu_arena));
+
+	unsigned limit = percpu_arena_ind_limit(opt_percpu_arena);
+
+	/* Force the thread to be bound to its current CPU's arena. */
+	void *p = mallocx(1, 0);
+	expect_ptr_not_null(p, "Unexpected mallocx() failure");
+
+	unsigned cur;
+	size_t sz = sizeof(cur);
+	expect_d_eq(mallctl("thread.arena", (void *)&cur, &sz, NULL, 0), 0,
+	    "Unexpected mallctl() failure");
+	expect_u_lt(cur, limit,
+	    "Thread should start bound to a per-CPU (auto) arena");
+
+	/* Create a dedicated arena, which is always outside the auto range. */
+	unsigned manual;
+	sz = sizeof(manual);
+	expect_d_eq(mallctl("arenas.create", (void *)&manual, &sz, NULL, 0), 0,
+	    "Unexpected arenas.create() failure");
+	expect_u_ge(manual, limit,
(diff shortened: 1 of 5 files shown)


test/unit/percpu_arena_resume.c
```diff
@@ -0,0 +1,74 @@
...
+	sz = sizeof(manual);
+	expect_d_eq(mallctl("arenas.create", (void *)&manual, &sz, NULL, 0), 0,
+	    "Unexpected arenas.create() failure");
+	expect_u_ge(manual, limit,
+	    "A manually created arena must be outside the per-CPU range");
+
+	/* Binding to a manual arena is allowed and takes effect. */
+	unsigned old;
+	sz = sizeof(old);
+	expect_d_eq(mallctl("thread.arena", (void *)&old, &sz, (void *)&manual,
```
guangli-dai, line 35: After each arena binding, can also try to allocate and deallocate (bypass tcache) to make sure arena-level allocation and deallocations are working properly. I would also suggest add an extreme case where we allocate a region in one arena and after the new arena binding, free that region to see if it goes through well.

guangli-dai (commented):
Thank you for the patch. Checked the history, we have issue #1236 documenting this gap already. I think this PR solves the problem well. Commented on the test to improve the coverage.

test/unit/percpu_arena_resume.c
Algunenano: Thanks for the review. I've improved the test with your comments

guangli-dai (approved):

guangli-dai:
Thank you @Algunenano for the contribution! Resolves #1236.

PR #2959, opened 2026-07-16
[tech debt] Stats.c refactor

spredolac:
Primary focus of this PR is to improve readability of stats.c file and provide better separation between reading/gathering the stats and printing them.  Several bugs were fixed, and free form HPA output is replaced with simpler key:values.  

`src/stats.c` renders jemalloc's statistics in two formats — the human-readable table and JSON (the `"J"` option). Historically only the table existed; JSON was added later, and both were produced by a few very large, bottom-up functions
that mixed `mallctl` reads with per-value emit macros, with the public `stats_print()` at the very bottom of the file.  The intent is to refactor the code in such a way that it is easier to navigate to functions responsible for each section, as well as get some intuition about the structure of the output while looking at the code. 

JSON - output is not changed in this PR. 

Each part is a separate commit; the refactor is squashed into one.

## Commits

| commit | kind | note |
|---|---|---|
| Fix duplicate opt.stats_print / opt.stats_print_opts in malloc_stats_print | bugfix | output not changed |
| Refactor stats.c: separate gathering from emission; read top-down | refactor | output not changed |
| Format change: HPA shard counters → structured rows | table/human readable format change | JSON unchanged |
| Format change: fold full/empty pageslabs into the nonfull table | able/human readable format change | JSON unchanged |
| Format change: label size-class ranges in the per-class stats tables | table format change | JSON unchanged |
| Deduplicate mutex stats reader: merge identical global/arena helpers | refactor |  output not changed |
| Fix `col_ind` to write `unsigned_val`, matching its column type | bug fix | pre-existing; output not changed |

## The refactor (output-preserving) had following intent

- **Separate gathering from emission where it is worth it.** Most sections become a small `stats_gather_*` (the `mallctl` reads into a typed struct) plus an emit that renders it; the plumbing macros (`CTL_*`, `COL*`) and the gather structs
moved to `stats_internal.h`, out of the reading path. This is a balance, not a rule: the per-size-class tables stay
  O(1) streaming (per-row gather+emit — no large stack buffers, and avoiding heap/base allocator), the mutex rows keep reading straight into emitter columns, and the already-readable `config`/`opt` probes are left as they are.
- **Reads like the output.** The file opens with a table-of-contents comment and the public `stats_print()`, then the section functions top-down, with the interval/boot/fork plumbing at the bottom. Each fixed section is a predictable gather-then-emit pair; the per-size-class tables stay O(1) streaming (per-row gather+emit — a full-table buffer would be ~15 KB of stack).
- **One emission engine.** The `emitter` remains the single dual-format renderer, so each section has one rendering path rather than parallel renderers. (The two formats are not field-identical — some rows are table-only, e.g. the `*_ps` rates, and the per-size-class metadata is JSON-only — but they share one engine.) Some table-only concerns moved into it: gap-collapsing (the `"---"` run separator) and a dual columnar row op (`emitter_row`), so section code carries less table-format detail.
- **JSON discoverability.** The ToC names the function that prints each section and its JSON keys where useful.

## Table format changes (human table only; JSON byte-identical)

- **HPA shard counters:** the free-form (`HPA shard stats:\n  Pageslabs: N (H huge, NH nonhuge)\n  Purge passes: N (R / sec)\n …`) becomes aligned `key: value` rows keyed by the JSON field names (`npageslabs`, `nactive`, …).
- **full/empty pageslabs:** the two `In full slabs:` / `In empty slabs:` blocks become two symbolic rows (`size = full` / `empty`) at the top of the per-size-class table, which is then retitled from `In nonfull slabs:` to `pageslabs:`.
- **size-class ranges:** size classes above USIZE_GROW_SLOW_THRESHOLD cover a range of sizes, so the `extents` / `lextents` / `hpa-nonfull` tables label those rows `(prev_size,size]` (e.g. `(32768, 40960]`) instead of a single size. No JSON change: the size-class metadata arrays are ordered, so the range is already implicit there and any machine processing can handle it easily.

## Before / after (human table; JSON is byte-identical throughout)

Real captures from a fixed workload (values match; only the format differs).

**HPA shard counters** — free-form → structured rows:

```
# before                                    # after
HPA shard stats:                            npageslabs: 10
  Pageslabs: 10 (8 huge, 2 nonhuge)         nactive: 4247
  Active pages: 4247 (3884 huge, 363 non..) ndirty: 272
  Dirty pages: 272 (212 huge, 60 nonhuge)   npurge_passes: 429 (per_sec: 429)
  Retained pages: 601                       npurges: 7 (per_sec: 7)
  Purge passes: 429 (429 / sec)             nhugifies: 8 (per_sec: 8)
  Purges: 7 (7 / sec)                       nhugify_failures: 0 (per_sec: 0)
  Hugeifies: 8 (8 / sec)                    ndehugifies: 0 (per_sec: 0)
  Hugify failures: 0 (0 / sec)              slabs
  Dehugifies: 0 (0 / sec)                     npageslabs_nonhuge: 2
                                              nactive_nonhuge: 363
                                              ndirty_nonhuge: 60
                                              nretained_nonhuge: 601
                                              npageslabs_huge: 8
                                              nactive_huge: 3884
                                              ndirty_huge: 212
```

**full / empty / nonfull slabs** → one `pageslabs:` table (nonhuge columns
trimmed here for width):

```
# before                                         # after
  In full slabs:                                 pageslabs:
      npageslabs: 2 huge, 0 nonhuge                          size ind npageslabs_huge nactive_huge ndirty_huge
      nactive: 1024 huge, 0 nonhuge                          full   -               2         1024           0
      ndirty: 0 huge, 0 nonhuge                             empty   -               0            0           0
      nretained: 0 huge, 0 nonhuge                           4096   0               4         1918         130
  In empty slabs:                                            8192   1               2          942          82
      npageslabs: 0 huge, 0 nonhuge                          ...
      ... (all zero)
  In nonfull slabs:
                size ind npageslabs_huge ...
                  4096   0               4 ...
                8192   1               2 ...
```

**Size-class ranges** (classes above `8*PAGE`, in `extents`/`lextents`/`hpa-nonfull`):

```
# before          # after
        40960  8          (32768, 40960] 8
        98304 13          (81920, 98304] 13
```

## How it was verified

- **Before/after output diff:** captured the full `malloc_stats_print` output(text and JSON) for many option combinations (`"" g a b l x e h m d abehlx  gmdablxeh`) under a fixed `MALLOC_CONF` with ASLR disabled, and diffed before
  vs after — masking only genuinely time-volatile fields (uptime, purge durations). The refactor commits produce byte-identical output; the format commits change only the text (JSON stays identical). A separate `hpa:true` capture covers the HPA sections.
- **Unit tests:** `test/unit/{stats,stats_print,json_stats,emitter}` pass, new emitter cases added for new functions.

## Out of scope 

- **Bytes everywhere (HPA):** also showing HPA page/pageslab counts in bytes. Deferred — as real JSON fields it roughly doubles the count fields and would push the pageslabs table to ~14 columns. Let's do that when we support larger sizes in HPA

Potential follow-up which I decided not to do here: 
Migrate bins/extents/HPA to emitter_row() and reconcile the order first — either reorder the table columns to match JSON, or accept reordering the JSON keys. Reordering json keys is probably fine, but this could be another one of "make stats.c prettier changes"

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -702,6 +702,26 @@ mutex_stats_emit(emitter_t *emitter, emitter_row_t *row,
 #undef EMITTER_TYPE_uint64_t
 }
 
+/*
+ * Set a per-size-class row's "size" column.  Above the slow-growth threshold,
+ * display the bucket bounds as "(prev_size,size]".  This notation does not
+ * imply that every integer in the interval is a valid usable size.  Smaller
+ * classes are displayed as a single size.  buf must outlive row emission.
+ * (This is table labeling only; the ordered JSON arrays imply the bounds.)
+ */
+static void
+stats_size_col_set(emitter_col_t *col, size_t size, size_t prev_size, char *buf,
+    size_t bufsz) {
+	if (size > USIZE_GROW_SLOW_THRESHOLD) {
+		malloc_snprintf(buf, bufsz, "(%zu,%zu]", prev_size, size);
+		col->type = emitter_type_title;
+		col->str_val = buf;
+	} else {
+		col->type = emitter_type_size;
+		col->size_val = size;
+	}
+}
+
 JEMALLOC_COLD
 static void
 stats_arena_bins_print(
@@ -1012,6 +1032,15 @@ stats_arena_lextents_print(emitter_t *emitter, unsigned i, uint64_t uptime) {
 		CTL_LEAF_PREPARE(prof_stats_mib, 0, "prof.stats.lextents");
 	}
 
+	/*
+	 * Carry the previous size class's size forward for the "(prev,size]"
+	 * range label instead of looking it up by a large size-class index:
+	 * with large size classes disabled (the default) that index may be out
+	 * of range, and indexing sz_index2size_tab there would be unsafe.  The
+	 * class before the first large one is the last bin; sz_index2size_compute
+	 * derives it arithmetically (no table access).
+	 */
+	size_t prev_size = sz_index2size_compute(nbins - 1);
 	emitter_table_sparse_begin(emitter);
 	for (j = 0; j < nlextents; j++) {
 		stats_arena_lextent_t lext;
@@ -1020,7 +1049,10 @@ stats_arena_lextents_print(emitter_t *emitter, unsigned i, uint64_t uptime) {
 
 		bool is_gap = (lext.nrequests == 0);
 
-		col_size.size_val = lext.lextent_size;
+		char size_buf[48];
+		stats_size_col_set(&col_size, lext.lextent_size, prev_size,
+		    size_buf, sizeof(size_buf));
+		prev_size = lext.lextent_size;
 		col_ind.unsigned_val = nbins + j;
 		col_allocated.size_val = lext.curlextents * lext.lextent_size;
 		col_nmalloc.uint64_val = lext.nmalloc;
@@ -1106,7 +1138,9 @@ stats_arena_extents_print(emitter_t *emitter, unsigned i) {
 		    &e.pinned_bytes);
 		emitter_json_object_end(emitter);
 
-		col_size.size_val = sz_pind2sz(j);
+		char size_buf[48];
+		stats_size_col_set(&col_size, sz_pind2sz(j),
+		    j > 0 ? sz_pind2sz(j - 1) : 0, size_buf, sizeof(size_buf));
 		col_ind.size_val = j;
 		col_ndirty.size_val = e.ndirty;
 		col_dirty.size_val = e.dirty_bytes;
@@ -1413,7 +1447,9 @@ stats_arena_hpa_shard_slabs_print(emitter_t *emitter, unsigned i) {
 
 		bool is_gap = (s.npageslabs_huge == 0 && s.npageslabs_nonhuge == 0);
 
-		col_size.size_val = sz_pind2sz(j);
+		char size_buf[48];
+		stats_size_col_set(&col_size, sz_pind2sz(j),
+		    j > 0 ? sz_pind2sz(j - 1) : 0, size_buf, sizeof(size_buf));
 		col_ind.size_val = j;
 		col_npageslabs_huge.size_val = s.npageslabs_huge;
 		col_nactive_huge.size_val = s.nactive_huge;
--- a/src/stats.c
+++ b/src/stats.c
@@ -1313,18 +1313,8 @@ stats_arena_hpa_shard_counters_print(
 }
 
 static void
-stats_emit_arena_hpa_slab(emitter_t *emitter, const stats_arena_hpa_slab_t *s,
-    const char *label, const char *json_key) {
-	emitter_table_printf(emitter,
-	    "  In %s slabs:\n"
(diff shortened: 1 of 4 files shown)


src/stats.c
```diff
@@ -1011,6 +1031,15 @@ stats_arena_lextents_print(emitter_t *emitter, unsigned i, uint64_t uptime) {
 		CTL_LEAF_PREPARE(prof_stats_mib, 0, "prof.stats.lextents");
 	}
 
+	/*
+	 * Carry the previous size class's size forward for the "prev+1..size"
```
guangli-dai: usize calculation enforces PAGE-aligned for sizes here so it can be more accurate `prev + PAGE..size`. Previously I was thinking about this presentation `(prev, size]`. But no strong preference as long as it's clear and not misleading to readers.

include/jemalloc/internal/stats_private.h
```diff
@@ -0,0 +1,333 @@
+#ifndef JEMALLOC_INTERNAL_STATS_PRIVATE_H
```
guangli-dai: nit: stats_internal.h sounds like a better name fitting current conventions.

src/stats.c
```diff
@@ -74,6 +38,471 @@ static counter_accum_t stats_interval_accumulated;
...
+ *   ___ Begin jemalloc statistics ___
+ *   { "jemalloc":
+ *     general [g]               -> stats_general_print
+ *       version                 -> stats_general_version
+ *       config                  -> stats_general_config
+ *       system                  -> stats_general_system
+ *       opt.*                   -> stats_general_opts
+ *       prof                    -> stats_general_prof
+ *       arenas metadata         -> stats_general_arenas_print   (arenas.*)
+ *     stats  (config_stats)     -> stats_print_helper           (stats.*)
```
guangli-dai: It seems `stats_print_helper` is a name inherited but it doesn't really describe what it does well. What about `stats_print_runtime_stats`?

guangli-dai (commented):
Mostly good, left some comments regarding naming and comments mainly. 

Also to share my opinions on the json format, which is not in the scope of this PR: Reordering and adding fields to the json format should be acceptable. Currently, we have more fields in human-readable than the json format. It should be changed and we should prepare json format to be the default output format, which make automated malloc stats parsing and agent-based analysis much easier.

src/stats.c
spredolac: Sounds good. I would use `(prev, size]`

Allocating `32769` bytes with `disable_large_size_classes:true`, the usable size is `36864`; with it disabled, the usable size is `40960`. In both cases, the allocation falls in the lextent bucket whose size is `40960`. If we write `(32768,40960]` it should be good.

guangli-dai (approved):
LGTM. Can merge after resolving the conflicts.

One suggestion on tests that can be done in the follow-up PR. The human-readable and json stats matching is already much better. The current [json stats test](https://github.com/jemalloc/jemalloc/blob/dev/test/unit/json_stats.c) is still very primitive. We should ad more tests in so that the whole json stats is covered by this test properly.

PR #2960, opened 2026-07-17
[tech debt] Stats refactor - emit row by row using data descriptors

spredolac:
Stats refactor - emit row by row using data descriptors

This PR builds on top of [PR 2959](https://github.com/jemalloc/jemalloc/pull/2959)  by separating per-size-class row emission and gathering. 

* First commit adds a descriptor-driven emitter for column tables and converts the bin, lextent, extent, and HPA pageslab tables to use shared per-row emit helpers. Rows remain streamed with O(1) memory, while explicit JSON column ordering is preserved. The emitter supports generic conditional columns and includes unit coverage for table rendering, JSON selection and ordering, and conditional fields. There are no intended output changes.

* Second commit adds those columns to JSON output in descriptor order and simplifies the code. Removed all four JSON-order arrays and most column enums. Text output remains unchanged.  Old JSON key/value remain, superset verification confirms only new fields were added.

### DIFF
--- a/include/jemalloc/internal/emitter.h
+++ b/include/jemalloc/internal/emitter.h
@@ -590,6 +590,99 @@ emitter_sparse_row(emitter_t *emitter, emitter_row_t *row, bool is_gap) {
 	}
 }
 
+/*
+ * Descriptor-driven column tables.  A descriptor array defines table order;
+ * callers provide a separate index array when JSON requires a different
+ * subset or ordering.  The getter lets one engine serve different row types
+ * while keeping derived values (rates, utilization, and so on) with the
+ * table-specific code.  A column is active when all its required_flags are
+ * present in the active_flags supplied by the caller; the emitter assigns no
+ * domain-specific meaning to those bits.
+ */
+typedef struct emitter_col_desc_s emitter_col_desc_t;
+struct emitter_col_desc_s {
+	const char       *json_key;
+	const char       *table_label;
+	emitter_justify_t justify;
+	int               width;
+	emitter_type_t    type;
+	unsigned          required_flags;
+	void            (*get)(const void *row, emitter_col_t *col);
+};
+
+static inline bool
+emitter_col_desc_active(const emitter_col_desc_t *desc,
+    unsigned active_flags) {
+	return (desc->required_flags & ~active_flags) == 0;
+}
+
+static inline void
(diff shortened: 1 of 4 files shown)


include/jemalloc/internal/emitter.h
```diff
@@ -666,17 +665,18 @@ emitter_col_table_fill(const emitter_col_desc_t *descs, size_t ndescs,
 static inline void
 emitter_col_table_emit_json(emitter_t *emitter,
     const emitter_col_desc_t *descs, size_t ndescs, unsigned active_flags,
-    emitter_col_t *cols, const unsigned *json_order, size_t json_order_len) {
-	for (size_t i = 0; i < json_order_len; i++) {
-		unsigned col_ind = json_order[i];
-		assert(col_ind < ndescs);
-		const emitter_col_desc_t *desc = &descs[col_ind];
+    emitter_col_t *cols) {
+	for (size_t i = 0; i < ndescs; i++) {
```
guangli-dai, line 672: Would an early return if json mode is not enabled make this more efficient?

guangli-dai (commented):
The refactoring part looks reasonable. However, the `json_stats.c` coverage is still not complete. I would suggest:
1. before the first commit, insert a commit to improve `json_stats.c` coverage so that unit tests can help prove the refactoring is no-op
2. In the commit that "Expose stats table columns ...", cover all newly added metrics. The current addition does not seem to cover all new metrics exposed?

test/unit/json_stats.c
```diff
@@ -502,7 +1487,9 @@ TEST_END
 
 int
 main(void) {
-	return test_no_reentrancy(test_json_stats_mutexes,
+	return test_no_reentrancy(test_json_stats_global_and_arena,
+	    test_json_stats_arena_tables, test_json_stats_hpa,
+	    test_json_stats_mutexes,
 	    test_hpa_shard_json_ndirty_huge,
```
guangli-dai: I think the three old tests are now well included in the newly added tests? We can remove these old ones?

guangli-dai (approved):
Overall LGTM, but I can still see some gaps in the current stats test:
1. per-arena stats not covered (auto, manual, destroyed);
2. many stats print options not tested;

The gaps are non-blocking for this PR. We should though bridge the gap before turning json mode to be the default.

PR #2961, opened 2026-07-17
change tests to set --enable-prof-libunwind instead of --enable-prof-frameptr

gctony:
Change tests that use --enable-prof-frameptr to instead use --enable-prof-libunwind. This is a follow up to https://github.com/jemalloc/jemalloc/pull/2955

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -63,7 +63,7 @@ jobs:
           - env:
               CC: gcc
               CXX: g++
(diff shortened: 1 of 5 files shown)


guangli-dai:
Instead of simply moving the `--enable-prof-frameptr`, we should probably replace it with `--enable-prof-libunwind` so that we also have CI coverage for libunwind? Libunwind needs some extra installation, so it is more than just replacing a flag, probably need to tweak the test environment preparation steps as well.

guangli-dai (approved):

PR #2963, opened 2026-07-21
Os layer 

guangli-dai:
## Motivation

The OS-specific pieces jemalloc depends on — mutexes/condvars, file I/O, clocks, virtual memory,
/proc reads, CPU/thread queries, fork handling, DSS — currently live inline across many .c/.h
files, each behind its own #ifdef ladder. OS-touching logic ends up scattered and duplicated:
there is no single place to see what the allocator actually requires from the OS, and porting to a
new platform means hunting #ifdefs throughout the tree.

## Design

`os.h` is the single header portable code includes. Each facility gets a dispatcher
`os/<module>.h` that picks `os/posix/<module>.h` (default) or `os/<os>/<module>.h` (override, e.g.
Windows) based on `JEMALLOC_OS_POSIX` (`os/detect.h`). Adding a new module means dropping in a
dispatcher + posix/override files and wiring one include into `os.h`; adding OS support for an
existing module means dropping in one override file.

## Behavior

Purely mechanical — no logic changes. Verified line-by-line against the pre-refactor
code (syscall sequences, casts, EINTR retry loop, short-read/write accumulation, EOF handling all
preserved exactly) and independently confirmed by two adversarial reviews (Codex, Gemini) with no
findings.

### DIFF
--- a/include/jemalloc/internal/os/file.h
+++ b/include/jemalloc/internal/os/file.h
@@ -1,0 +1,36 @@
+#ifndef JEMALLOC_INTERNAL_OS_FILE_H
+#define JEMALLOC_INTERNAL_OS_FILE_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/os/detect.h"
+
+/*
+ * File I/O interface.
+ * Default: posix/.  Override: Windows (CRT <io.h>).
+ */
+
+/* Functions required for implementation in each backend. */
+JEMALLOC_ALWAYS_INLINE ssize_t os_file_write_once(int fd, const void *buf,
+    size_t bytes);
+JEMALLOC_ALWAYS_INLINE ssize_t os_file_read_once(int fd, void *buf,
+    size_t bytes);
+/*
+ * Full retry-until-bytes-or-error versions of the above: loop over the
+ * _once primitive until target bytes are transferred or a non-retryable
+ * error occurs. Retrying on EINTR is a POSIX-only concept (Windows has no
+ * such signal-interruption semantics for this call), so that decision lives
+ * entirely in each backend rather than leaking into callers.
+ */
+JEMALLOC_ALWAYS_INLINE ssize_t os_file_write(int fd, const void *buf,
+    size_t bytes);
+JEMALLOC_ALWAYS_INLINE ssize_t os_file_read(int fd, void *buf, size_t bytes);
+
+#if defined(_WIN32)
+#  include "jemalloc/internal/os/windows/file.h"
+#elif defined(JEMALLOC_OS_POSIX)
+#  include "jemalloc/internal/os/posix/file.h"
+#else
+#  error "OS layer: no file backend for this platform; add os/<os>/file.h"
+#endif
+
+#endif /* JEMALLOC_INTERNAL_OS_FILE_H */
--- a/include/jemalloc/internal/os/windows/file.h
+++ b/include/jemalloc/internal/os/windows/file.h
@@ -1,0 +1,52 @@
+#ifndef JEMALLOC_INTERNAL_OS_WINDOWS_FILE_H
+#define JEMALLOC_INTERNAL_OS_WINDOWS_FILE_H
+
+/*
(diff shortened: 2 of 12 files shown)


include/jemalloc/internal/os/file.h
```diff
@@ -0,0 +1,26 @@
...
+#include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/os/detect.h"
+
+/*
+ * File I/O interface.
+ * Default: posix/.  Override: Windows (CRT <io.h>).
+ */
+
+/* Functions required for implementation in each backend. */
+JEMALLOC_ALWAYS_INLINE ssize_t os_file_write_once(int fd, const void *buf,
```
guangli-dai, line 13: Adding `JEMALLOC_ALWAYS_INLINE` here is actually a no-op, but adding it just as a part of the template to remind people what inline hints in the backend should be.

include/jemalloc/internal/os/windows/file.h
```diff
@@ -0,0 +1,21 @@
...
+
+/*
+ * Windows file-I/O backend, via the C runtime io.h (_read/_write).
+ */
+#include "jemalloc/internal/jemalloc_preamble.h"
+
+#include <io.h>
+
+JEMALLOC_ALWAYS_INLINE ssize_t
+os_file_write_once(int fd, const void *buf, size_t count) {
```
gctony: Why use the `_once` suffix for the read and write functions?

test/unit/malloc_io.c
```diff
@@ -269,9 +273,145 @@ TEST_BEGIN(test_malloc_snprintf_zero_size) {
...
+ * latter is never called with O_CREAT in production (only O_RDONLY, on
+ * files that already exist), and its 2-arg signature has no mode_t
+ * parameter to pass along if it were -- calling the underlying variadic
+ * open() with O_CREAT but no mode gives the new file garbage permissions.
+ */
+static const char *test_io_filename = "malloc_io_test_file.tmp";
+
+static void
+create_empty_test_io_file(void) {
+	FILE *fp = fopen(test_io_filename, "wb");
```
gctony, line 296: Should you use `malloc_open()` / `malloc_close()` here?

gctony (approved):
Overall, a great and helpful refactoring that will make the code more readable and maintainable. Looks good, minus two very minor questions (see inline).

include/jemalloc/internal/os/windows/file.h
guangli-dai: The `malloc_write_fd` calling it has a loop calling this multiple times so that naming wants to highlight the fact of writing only once (need an external loop to handle it).

However, this reminds me a previous change I had that makes sure the `malloc_write_fd` won't include `#ifdef _WIN32` to handle the error code, which I forgot to include in this version. Including it so that the whole read/write loop goes into the os layer.

test/unit/malloc_io.c
guangli-dai: `malloc_open()` does not take the `mode` as the input but to create a file with `open()` syscall, it is required to [set the mode bit](https://man7.org/linux/man-pages/man2/open.2.html#:~:text=The%20mode%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20argument-,must,-be%20supplied%20if). That's why we don't use it.

For more context, we don't use `fopen` in jemalloc because it may recursively end up in jemalloc, but this is a test that is safe to exempt from the convention.

src/malloc_io.c
```diff
@@ -758,79 +759,14 @@ malloc_printf(const char *format, ...) {
 	va_end(ap);
 }
 
-static ssize_t
-malloc_write_fd_syscall(int fd, const void *buf, size_t count) {
-#if defined(JEMALLOC_USE_SYSCALL) && defined(SYS_write)
-	/*
```
swkrueger, line 764: nit: Would it be worth copying the comment as well to retain the context?

gctony (approved):

src/malloc_io.c
guangli-dai: Good point, will add that in next PR.

PR #2964, opened 2026-07-21
Avoid background-thread reset shutdown deadlock

spredolac:
A worker paused during arena reset could still be waiting for the global background-thread lock when shutdown reacquired it and began joining workers.

Wait on the worker condition variable instead, signal it when reset finishes, and add a bounded regression test.

@guangli-dai , will send you handoff

### DIFF
--- a/include/jemalloc/internal/background_thread.h
+++ b/include/jemalloc/internal/background_thread.h
@@ -16,7 +16,7 @@
 typedef enum {
 	background_thread_stopped,
 	background_thread_started,
-	/* Thread waits on the global lock when paused (for arena_reset). */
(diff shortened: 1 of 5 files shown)


guangli-dai:
Picked up and merged in #2966. Since this is relevant to thread scheduling, two approaches if we want to add a test:
1. use hook to enforce deadlock to happen during the tests, making this test deeply coupled with the implementation and the protection provides is not worthy the complication;
2. use stress tests assuming a certain pattern will show up after enough repeats, which could make tests flaky

Therefore, not adding a test for now without a good enough solution since flaky tests brought quite some trouble in the past. Open to solutions if any ideas on how to add tests for this case gracefully.

PR #2965, opened 2026-07-22
[RFC] Adapt tcache fill and retention

spredolac:
Try to simplify tcache fill and flush logic.  I tested this extensively on variety of services and what is consistent is the win on slow_path rate (taken on average 20% less frequently).  Other than that the cpu and memory looked like mostly noise where CPU trended a bit positive and memory a bit negative.  In general, if we believe the code is simpler then the fact that we are hitting slow-path rate less frequently and the rest is flat-ish may be like an insurance where spending more cycles in slow path may be less costly.


src/tcache.c
```diff
@@ -583,14 +543,9 @@ tcache_alloc_small_hard(tsdn_t *tsdn, arena_t *arena, tcache_t *tcache,
...
-	if (nfill == 0) {
-		nfill = 1;
-	}
-	cache_bin_sz_t nfill_min = opt_experimental_tcache_gc
-	    ? ((nfill >> 1) + 1)
-	    : nfill;
+	cache_bin_sz_t ncached_max = cache_bin_ncached_max_get(cache_bin);
+	cache_bin_sz_t nfill = tcache_nfill_small_get(tcache_slow, binind);
+	cache_bin_sz_t nfill_min = nfill;
 	cache_bin_sz_t nfill_max = nfill;
```
guangli-dai: I would really doubt whether we should make nfill_min and nfill_max the same, this takes away the flexibility arena bins used to have in filling. Although production canaries show no measurable impacts, I still think the flexibility is helpful.

From implementation, if we can prove indeed it's good (or at least not bad) to have nfill_min and nfill_max the same, we should change `arena_ptr_array_fill_small` to take in only nfill instead of both `nfill_min` and `nfill_max`?

include/jemalloc/internal/tcache.h
```diff
@@ -124,7 +124,6 @@ extern unsigned opt_tcache_nslots_small_max;
 extern unsigned opt_tcache_nslots_large;
 extern ssize_t  opt_lg_tcache_shift;
 extern size_t   opt_tcache_gc_incr_bytes;
-extern size_t   opt_tcache_gc_delay_bytes;
 extern unsigned opt_lg_tcache_flush_small_div;
```
guangli-dai: We can also remove this option? If we can make large adaptive in the same way, we can also remove `opt_lg_tcache_flush_large_div` below.

guangli-dai (commented):
The partial refill is probably the part that needs more discussions, but I think the overall direction is surely correct: tcache currently has too many options, more adaptive policies should be the target.

In the long term, besides the flush options, we should also consider removing `tcache_nslots_small` and `tcache_nslots_large`, they are used to calculate default each bin's `ncached_max` and now that users can directly set `ncached_max` for each bin, these two options are just adding confusions.

spredolac:
I managed to remove more options and internal canaries preserved the results.  One key element for your arena being given a range - we can give it a range instead of using fixed, but the whole effect of having more hits (aka fewer slow-paths) comes from using the target as the bottom not the top - thus, I use it as the beginning of the range and still let it stop if it hits full slab. Let me know what you think overall. 

 For large bins there is nothing to size it from as they are not batch filled. The target is halved on every full-bin flush, and large bins are small (ncached_max defaults to 20): a single burst of frees creates more flushes as it drives target down quickly. Small bins never hit this as ncached_max is up to 200 and refills replenish in batches. we can still remove teh option and flush to tcache_bin_nslots_max(ncached_max), i.e. ncached_max >> 1 — exactly what ncached_max >> lg_tcache_flush_large_div computed at the option's default. Removing it is a no-behavior-change simplification not policy change.

src/conf.c
```diff
@@ -732,8 +732,6 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				CONF_HANDLE_BOOL(opt_xmalloc, "xmalloc")
 			}
 
-			CONF_HANDLE_BOOL(opt_experimental_tcache_gc,
```
guangli-dai, line 735: For all the options removed here, maybe we should also put them into the "deprecated" to prevent abortion until next external release?

include/jemalloc/internal/tcache.h
```diff
@@ -47,6 +47,26 @@ typedef struct tcaches_s     tcaches_t;
...
+ * TCACHE_NSLOTS_SMALL_MAX]: a refill can then take a whole slab's worth at a
+ * time, and the cache holds enough to keep the bin busy between refills.
+ *
+ * Large allocations have no slab to scale against -- each one is its own
+ * extent -- so there is no natural proportion to derive a count from, and large
+ * bins take the flat TCACHE_NSLOTS_LARGE instead.  It has to stay small in
+ * absolute terms regardless: the objects are orders of magnitude bigger, so the
+ * same slot count would cost far more memory per thread.
+ *
+ * Any individual bin can be given a different count with tcache_ncached_max.
```
guangli-dai: The previous comment in `tcache_ncached_max_compute` should be preserved here and translated to requirements to the constant values like `TCACHE_NSLOT_SMALL_MIN/MAX` should be even, min should be smaller than max. This is to prevent future unexpected changes to these constant values.

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -225,8 +304,15 @@ tcache_dalloc_large(
...
-		    >> opt_lg_tcache_flush_large_div;
+		/*
+		 * Large bins use a fixed flush target rather than the adaptive
+		 * one small bins use.  Halving a retention target here compounds
+		 * within a single burst of frees -- a burst of twice ncached_max
+		 * overflows three times, walking the target down to its floor --
+		 * which leaves the bin empty for the burst that follows.
+		 */
+		unsigned remain =
+		    tcache_bin_nslots_max(cache_bin_ncached_max_get(bin));
```
guangli-dai: When `ncached_max` is 1, the remain would also be 1, it flushes nothing and then the next insertion would fail?

src/tcache.c
```diff
@@ -122,96 +108,108 @@ tcache_gc_postponed_event_wait(tsd_t *tsd) {
...
-		item_delay = delay_max - 1;
-	}
-	return (uint8_t)item_delay;
+	assert(low_water <= ncached);
+	assert(ncached <= ncached_max);
+
+	cache_bin_sz_t used_since_gc = (cache_bin_sz_t)(ncached - low_water);
+	cache_bin_sz_t nretain;
+	if (used_since_gc == 0) {
+		nretain = tcache_bin_nslots_min(ncached_max);
```
guangli-dai: For an allocation-only thread (after last gc), `used_since_gc == 0` and therefore `nretain` will be minimized. This looks like a large harm to such allocation-only threads: even we fill aggressively, in each gc triggered, we will be moving most of the cached items away?

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -179,6 +179,47 @@ tcache_alloc_large(tsd_t *tsd, arena_t *arena, tcache_t *tcache, size_t size,
...
+JEMALLOC_ALWAYS_INLINE cache_bin_sz_t
+tcache_bin_nslots_min(cache_bin_sz_t ncached_max) {
+	assert(ncached_max > 0);
+	return (ncached_max < 2) ? ncached_max : 2;
+}
+
+JEMALLOC_ALWAYS_INLINE cache_bin_sz_t
+tcache_nretain_small_reduce(tcache_slow_t *tcache_slow, szind_t binind,
+    cache_bin_sz_t ncached_max) {
+	cache_bin_sz_t nretain = tcache_nretain_small_get(tcache_slow, binind);
```
guangli-dai: When `ncached_max == 2`, the max would be 1 and min would be 2, this `tcache_nretain_small_reduce` will actually increase the number of nretain from 1 to 2?

guangli-dai (commented):
1. The windows CI failure looks real: `test_approximate_stats_active:test/unit/stats.c:545: Failed assertion: (approximate_active_after) > (approximate_active_before) --> 192512 <= 217088: approximate_stats.active should increase after the allocation`;
2. While production canary results look good, extreme threads that only allocate or only deallocate may trigger boundary cases. Worthy develop some local benchmarks to know the impact on them.
3. It's great that we remove lots of options since those configurability is not strongly needed. In the commit message, let's list the changes in configurability as we remove some options. For instance, `opt_tcache_nslots_small_min`, `opt_tcache_nslots_small_max`, `opt_tcache_nslots_large`, and `opt_lg_tcache_nslots_mul` were used to control the default thread cache ncached_max and they can now be configured directly through `tcache_ncached_max` in a more direct and easy-to-understand way.

spredolac:
@gdai, I hope that I addressed the latest feedback and shaped this a bit better (I think we have simpler algorithm now then several we had before some based on divisor/offset, but it was bugging me that it is not easier to see it after some patching.  I attempted to refactor to make it more readable, hopefully it is close now. 

Preserved partial-refill flexibility.  Benchmarked the current policy against a simpler alternative using production-like and allocation/deallocation-focused workloads, comparing fills, flushes, and tcache_bytes. The alternative did not consistently improve the tradeoff, so I kept the existing policy.
 
 - added/fixed target and flush invariants for all bin capacities.
 - Refactored the implementation to separate policy from mechanism: t`cache_ncached_target.h` contains only pure calculations of desired cached counts; `tcache.c` handles GC/refill events and updates state; and `tcache_inlines.h` applies the calculation on the dalloc hot path

include/jemalloc/internal/tcache_inlines.h
spredolac: `tcache_ncached_retain_after_overflow`() should guarantee that the retained count is strictly below `ncached_max`

include/jemalloc/internal/tcache.h
spredolac: translated the old assumptions into compile-time checks

include/jemalloc/internal/tcache_inlines.h
```diff
@@ -207,7 +213,10 @@ tcache_dalloc_small(
 			return;
 		}
 		cache_bin_sz_t max = cache_bin_ncached_max_get(bin);
-		unsigned       remain = max >> opt_lg_tcache_flush_small_div;
+		cache_bin_sz_t *nretain =
+		    &tcache->tcache_slow->bin_nretain[binind];
+		*nretain = tcache_ncached_retain_after_overflow(*nretain, max);
```
guangli-dai, line 218: This may set `*nretain` to 0 (when `ncached_max == 1`, which may trigger the assertion added above. Two choices, we can change the assertion, or we can change the implementation here so that `*nretain` stays > 0, but for the immediately used remain, keep it 0.

A more appropriate question here is probably what does `bin_nretain[ind] == 0` supposed to mean here? With the assertion above, is this state really invalid? Also, it seems this problem is not reflected in current tests, we should consider improving the new tests to cover the extreme state here.

guangli-dai (commented):
Other changes all look good, just one more concern regarding small capacity case specifically in `tcache_ncached_retain_after_overflow`.

include/jemalloc/internal/tcache_inlines.h
spredolac: Good catch. bin_nretain is persistent state, so zero should mean "disabled bins". I changed the overflow calculation to keep bin_nretain positive, and compute the flush remainder separately, always freeing at least one slot for the incoming object. For ncached_max == 1, the target remains 1 while the flush temporarily retains 0.

guangli-dai (approved):

PR #2967, opened 2026-07-27
Fix SEC shard count integer overflow (#2892)

### LINKED ISSUE #2892: LLM audit
This is an audit of jemalloc using GPT 5.4. A similar [audit in mimalloc](https://github.com/microsoft/mimalloc/issues/1271) turned up quite a few low severity bugs with some false positives.

<details>
<summary>List of unacknowledged bugs</summary>

### 1. SEC shard count accepts values the implementation cannot represent

The HPA SEC shard-count setting accepts any `size_t`, but shard selection and cached shard identifiers use `uint8_t`, and SEC metadata sizing multiplies the shard count without overflow checks.

**Steps to reproduce**
1. Enable HPA SEC support and configure the shard count above `255`, or to a value near integer-overflow limits.
2. Exercise SEC-backed allocation and deallocation paths across multiple threads.
3. Observe wrapped shard identifiers for moderate values, or undersized SEC metadata arrays for extreme values.

**Relevant source locations and functions**
- `include/jemalloc/internal/sec_opts.h:13-20`
- `include/jemalloc/internal/tsd_internals.h:81,102`
- `src/conf.c:945-947`
- `src/sec.c:22-52` `sec_init`
- `src/sec.c:56-83` `sec_shard_pick`
- `src/sec.c:85-90` `sec_bin_pick`
- `src/sec.c:110-149` `sec_multishard_trylock_alloc`
- `src/sec.c:212-237` `sec_multishard_trylock_dalloc`

**Expected impact**
Moderate misconfiguration can silently break sharding, and extreme values can underallocate SEC metadata and lead to out-of-bounds access or crashes.

---

### 2. Generic TSD recursion guard can return an uninitialized wrapper during reentrant initialization

The generic TSD recursion guard adds a stack record to its recursion list before storing the wrapper pointer in that record. Same-thread reentry during that window can read an uninitialized pointer and treat it as a valid TSD wrapper.

**Steps to reproduce**
1. Use a build that takes the generic TSD path.
2. Trigger first-time TSD initialization.
3. Cause same-thread reentry before the wrapper pointer is stored in the recursion block.

**Relevant source locations and functions**
- `include/jemalloc/internal/tsd_generic.h:78-100` `tsd_wrapper_get`
- `include/jemalloc/internal/tsd_generic.h:165-173` `tsd_get`
- `src/tsd.c:524-541` `tsd_init_check_recursion`

**Expected impact**
Rare reentrant initialization can crash the allocator early in thread startup or corrupt internal allocator state.

---

### 3. `b0_alloc_tcache_stack()` reinserts `edata_avail` without `base->mtx`

`b0_alloc_tcache_stack()` removes an `edata_t` from `base->edata_avail` under the base mutex, but one failure path reinserts it without taking that same lock.

**Steps to reproduce**
1. Trigger the path where `base_alloc_base_edata()` succeeds but the later metadata allocation fails.
2. Run another thread that also allocates or frees base metadata at the same time.
3. Observe the unlocked reinsertion race with other `edata_avail` mutations.

**Relevant source locations and functions**
- `src/base.c:280-295` `base_alloc_base_edata`
- `src/base.c:297-305` `base_extent_bump_alloc_post`
- `src/base.c:653-676` `b0_alloc_tcache_stack`

**Expected impact**
Concurrent failure handling can corrupt base-allocator metadata and destabilize future allocator operations.

---

### 4. Logging buffer append logic can write past the end of the stack buffer

`log_impl_varargs()` treats `malloc_snprintf()` and `malloc_vsnprintf()` return values as bounded append lengths. Once truncation happens, the computed write offset can move past the fixed log buffer and the next append writes out of bounds.

**Steps to reproduce**
1. Reach a logging path that calls `log_impl_varargs()`.
2. Pass a log name or formatted payload long enough to exceed the fixed log buffer.
3. Let a follow-up append or newline write use the oversized offset.

**Relevant source locations and functions**
- `include/jemalloc/internal/log.h:91-105` `log_impl_varargs`
- `src/malloc_io.c:693-703` `malloc_vsnprintf`, `malloc_snprintf`

**Expected impact**
Stack out-of-bounds write in the logging path, usually ending in a crash or corrupted diagnostic output.

---

### 5. Bootstrap calloc can underallocate after multiplication overflow

`bootstrap_calloc()` multiplies the element count and element size without overflow checking. A wrapped product can produce a too-small allocation instead of failing.

**Steps to reproduce**
1. Reach a bootstrap path that uses `bootstrap_calloc()`.
2. Pass values whose product wraps but does not become zero.
3. Use the returned allocation as if it covered the full requested element count.

**Relevant source locations and functions**
- `src/jemalloc.c:373-382` `bootstrap_calloc`

**Expected impact**
Underallocation on bootstrap-only paths can lead to out-of-bounds access if the caller trusts the logical allocation size.

---

### 6. Explicit tcache IDs are dereferenced before table and index validation

Explicit tcache lookups dereference the `tcaches` table before verifying that the table exists or that the requested index is valid.

**Steps to reproduce**
1. Use an explicit tcache ID before any explicit tcache has been created, or supply an invalid large tcache ID through the control interface.
2. Reach the lookup, flush, or destroy path for that ID.
3. Observe a null dereference or out-of-bounds table access before the invalid-ID check rejects the request.

**Relevant source locations and functions**
- `include/jemalloc/internal/tcache_inlines.h:229-239` `tcaches_get`
- `include/jemalloc/internal/jemalloc_internal_inlines_c.h:389-416` `tcache_get_from_ind`
- `include/jemalloc/internal/jemalloc_internal_types.h:51-64`
- `src/tcache.c:1385-1406` `tcaches_flush`, `tcaches_destroy`
- `src/ctl.c:2615-2639` `tcache_flush_ctl`, `tcache_destroy_ctl`

**Expected impact**
Malformed or premature explicit tcache use can crash the process and may permit out-of-bounds in-process memory access through control paths.

---

### 7. Profiling log node allocation failures can turn into null dereferences

Several profiling log paths allocate bookkeeping nodes with `iallocztm()` and dereference the result immediately, without checking for allocation failure.

**Steps to reproduce**
1. Enable profiling log support.
2. Force internal allocation failure while the profiler is trying to create a new log node.
3. Trigger a sampled allocation, sampled free, new thread log record, or new backtrace log record.

**Relevant source locations and functions**
- `src/prof_log.c:123-164` `prof_log_bt_index`
- `src/prof_log.c:166-201` `prof_log_thr_index`
- `src/prof_log.c:241-275` `prof_try_log`

**Expected impact**
Memory pressure in profiling log mode can turn allocator bookkeeping failure into an immediate process crash.

---

### 8. Empty-batch `mpsc_queue_push_batch()` dereferences a null first element

The batched push helper assumes the source list is non-empty, but the API does not document that precondition. Passing an empty list yields a null `first` pointer that is dereferenced immediately.

**Steps to reproduce**
1. Instantiate an MPSC queue and an empty source list.
2. Call the generated batch-push helper with the empty list.
3. Observe the first list element dereference even though the batch is empty.

**Relevant source locations and functions**
- `include/jemalloc/internal/mpsc_queue.h:61-82` `push_batch` macro expansion
- `include/jemalloc/internal/ql.h:44,74-75`

**Expected impact**
Immediate crash on an empty batch push.

---

### 9. Pairing-heap enumeration can crash on an empty heap in release builds

Pairing-heap enumeration always queues the root pointer, even when the heap is empty. In release builds that leads to a null dereference during the first enumeration step.

**Steps to reproduce**
1. Prepare enumeration for an empty pairing heap.
2. Advance the enumeration.
3. Observe traversal dereference the queued null root pointer.

**Relevant source locations and functions**
- `include/jemalloc/internal/ph.h:460-474` `ph_enumerate_prepare`
- `include/jemalloc/internal/ph.h:477-499` `ph_enumerate_next`

**Expected impact**
Release builds can crash when a caller enumerates an empty pairing heap.

---

### 10. `MADV_DONTNEED` self-test continues after `mmap()` failure and dereferences `MAP_FAILED`

The `MADV_DONTNEED` zeroing self-test logs `mmap()` failure, but if abort-on-error is not enabled it still proceeds to touch and unmap the failed address.

**Steps to reproduce**
1. Reach the page-boot self-test path that probes `MADV_DONTNEED` behavior.
2. Make the temporary mapping fail.
3. Let execution continue instead of aborting immediately.

**Relevant source locations and functions**
- `src/pages.c:68-105` `madvise_MADV_DONTNEED_zeroes_pages`
- `src/pages.c:884-888` `pages_boot`

**Expected impact**
Allocator initialization can crash on a failed temporary mapping.

---

### 11. Frame-pointer unwinding can read past the validated stack range

The frame-pointer backtrace path checks only that the frame pointer itself lies inside the recorded stack range, then reads two machine words from it without confirming that the full frame record also fits inside the range.

**Steps to reproduce**
1. Use a build that takes the frame-pointer profiling unwinder path.
2. Trigger a sampled allocation with a corrupted or boundary-adjacent frame chain.
3. Let unwinding reach a frame pointer close to the top of the validated stack range.

**Relevant source locations and functions**
- `src/prof_sys.c:130-202` `prof_backtrace_impl`
- `src/prof_sys.c:165-185` `prof_backtrace_impl`

**Expected impact**
Profiling backtrace capture can perform an out-of-bounds stack read and crash.

---

### 12. `malloc_vsnprintf()` can read past short strings when `%.*s` precision is large

When formatting `%s` with an explicit precision, `malloc_vsnprintf()` copies exactly the precision length instead of clamping to the actual string length.

**Steps to reproduce**
1. Format a short NUL-terminated string with `%.*s`.
2. Supply a precision larger than the string length.
3. Let the formatter copy the requested precision bytes from the short source buffer.

**Relevant source locations and functions**
- `src/malloc_io.c:374-381` `APPEND_S`
- `src/malloc_io.c:662-668` `malloc_vsnprintf`

**Expected impact**
Out-of-bounds read, potentially leaking adjacent bytes into formatted output or crashing near an unmapped boundary.

---

### 13. `malloc_strtoumax()` has undefined behavior on large negative boundary values

The negative-number path casts the parsed unsigned magnitude to `intmax_t` and negates it. Boundary values can make that cast implementation-defined and the negation undefined.

**Steps to reproduce**
1. Parse a large negative value at or beyond the signed minimum boundary.
2. Reach the negative-result conversion path.
3. Observe undefined behavior or inconsistent result handling around the signed boundary.

**Relevant source locations and functions**
- `src/malloc_io.c:117-245` `malloc_strtoumax`

**Expected impact**
Signed-boundary parsing can produce undefined behavior and inconsistent numeric results.

---

### 14. Signed formatting of `INTMAX_MIN` can trigger undefined behavior

The integer formatting helper negates negative values directly. Negating `INTMAX_MIN` is undefined because the positive value is not representable.

**Steps to reproduce**
1. Format the minimum signed integer value through the signed decimal formatting path.
2. Reach the helper that flips the sign before decimal conversion.
3. Observe undefined behavior on the negation.

**Relevant source locations and functions**
- `src/malloc_io.c:309-336` `d2s`
- `src/malloc_io.c:592-616` `malloc_vsnprintf`

**Expected impact**
Boundary-value formatting can trigger undefined behavior, sanitizer failures, or wrong output.

---

### 15. `arena->last_thd` is a cross-thread data race in arena selection

The per-arena `last_thd` heuristic is a plain pointer that is read and written on the arena-selection fast path without atomics or locking.

**Steps to reproduce**
1. Enable per-CPU arena mode.
2. Run multiple threads that share an auto arena and allocate concurrently.
3. Observe simultaneous reads and writes of `last_thd` in the arena-choose fast path.

**Relevant source locations and functions**
- `include/jemalloc/internal/arena_structs.h:41-46`
- `include/jemalloc/internal/jemalloc_internal_inlines_b.h:74-84` `arena_choose_impl`
- `src/arena.c:1795` arena initialization path

**Expected impact**
The race can misroute the arena-selection heuristic and introduces undefined behavior on a hot path.

---

### 16. Huge-arena initialization ignores mutex creation failure

Huge-arena setup initializes `huge_arena_pac_thp.lock` but does not check whether mutex creation succeeded before later using that mutex.

**Steps to reproduce**
1. Enable the huge-arena THP path.
2. Force mutex initialization for `huge_arena_pac_thp.lock` to fail during setup.
3. Trigger an oversized huge-arena allocation that later takes the lock.

**Relevant source locations and functions**
- `src/arena.c:1960-1973` `arena_init_huge`
- `src/extent.c:673-699` `extent_handle_huge_arena_thp`
- `src/extent.c:824-830` `extent_alloc_wrapper_hard`
- `src/mutex.c:141-177` `malloc_mutex_init`

**Expected impact**
Rare initialization failure can leave the allocator using an uninitialized mutex and crashing later in oversized allocation paths.

---

### 17. Partial commit verification checks uncommitted pages in debug builds

`ehooks_commit()` validates zero-filled memory for the whole extent even when the commit hook was asked to commit only a subrange.

**Steps to reproduce**
1. Use a debug build with custom extent hooks.
2. Implement a valid partial-commit hook that commits and zeroes only the requested subrange.
3. Trigger a partial commit on an extent whose untouched pages still contain nonzero data.

**Relevant source locations and functions**
- `include/jemalloc/internal/ehooks.h:164-190` `ehooks_debug_zero_check`
- `include/jemalloc/internal/ehooks.h:249-265` `ehooks_commit`

**Expected impact**
Debug builds can abort even when a partial-commit hook obeys the documented contract.

---

### 18. Failed trailing split leaks the salvaged lead extent

On the two-step interior split path, if the first split succeeds and the second fails, the leading piece is deregistered but not returned to an extent cache or destroyed.

**Steps to reproduce**
1. Recycle an extent that needs both a leading and trailing split.
2. Let the first split succeed.
3. Force the trailing split to fail through metadata exhaustion or split-hook failure.

**Relevant source locations and functions**
- `src/extent.c:372-389` `extent_deregister`
- `src/extent.c:488-536` `extent_split_interior`
- `src/extent.c:557-600` `extent_recycle_split`

**Expected impact**
Rare split-failure handling can leak retained virtual memory.

---

### 19. Sanitizer bump allocation can leak the previous current region after split failure

When the sanitizer bump allocator grows successfully and then fails while splitting the new region, it loses the old current region instead of restoring or destroying it.

**Steps to reproduce**
1. Start with a non-empty sanitizer bump current region.
2. Trigger a larger guarded allocation that grows the region successfully.
3. Force the split of the new region to fail.

**Relevant source locations and functions**
- `src/san_bump.c:23-35` `san_bump_alloc`
- `src/san_bump.c:43-66` `san_bump_alloc`
- `src/san_bump.c:87-103` `san_bump_grow_locked`
- `src/extent.c:1305-1362` `extent_split_impl`

**Expected impact**
Rare sanitizer allocation failure can leak retained address space.

---

### 20. `pages_huge_impl()` flips success and error in the `memcntl()` path

The `memcntl()` branch returns success in the place where the rest of the pages API uses `true` to mean error.

**Steps to reproduce**
1. Use a platform that takes the `memcntl()` huge-page path.
2. Trigger a hugify attempt that succeeds.
3. Observe the caller treat the success result as an error.

**Relevant source locations and functions**
- `src/pages.c:527-539` `pages_huge_impl`
- `src/hpa_hooks.c:47-67` `hpa_hooks_hugify`
- `src/hpa.c:547-566` `hpa_try_hugify`

**Expected impact**
Huge-page advice can be misreported and misaccounted on affected platforms.

---

### 21. Time-query failures are ignored, and the Mach path can divide by zero

`clock_gettime()`, `gettimeofday()`, and `mach_timebase_info()` failures are ignored. That can leave timestamp structures uninitialized, and on Mach can produce a divide-by-zero when the timebase denominator is never populated.

**Steps to reproduce**
1. Force one of the time queries used by `nstime_get()` or the realtime helper to fail.
2. Trigger code that updates allocator timestamps or profiling timestamps.
3. On Mach, allow the failed timebase query to leave a zero denominator.

**Relevant source locations and functions**
- `src/nstime.c:208-247` `nstime_get`
- `src/nstime.c:266-271` `nstime_get_realtime`
- `src/nstime.c:278-303` `nstime_prof_update_impl`, `nstime_update_impl`

**Expected impact**
Broken timekeeping can distort profiling and scheduling behavior, and Mach builds can crash on divide by zero.

---

### 22. NetBSD background-thread affinity uses an empty CPU set

The NetBSD affinity helper allocates a CPU set but never inserts the target CPU before calling `pthread_setaffinity_np()`.

**Steps to reproduce**
1. Use NetBSD with background threads and per-CPU arena mode enabled.
2. Start a background thread.
3. Let the thread attempt to pin itself to its target CPU.

**Relevant source locations and functions**
- `src/background_thread.c:121-158` `set_current_thread_affinity`
- `src/background_thread.c:516-518` `background_thread_entry`

**Expected impact**
Background threads can fail to pin correctly, degrading or breaking per-CPU behavior.

---

### 23. CPU-count helpers can feed uninitialized affinity masks to `CPU_COUNT()`

The CPU-count helpers ignore failures from affinity queries and then count bits in an uninitialized CPU set.

**Steps to reproduce**
1. Force the affinity query used during CPU counting to fail.
2. Trigger the code that computes CPU count or CPU-count determinism.
3. Let `CPU_COUNT()` consume the uninitialized mask contents.

**Relevant source locations and functions**
- `src/jemalloc.c:736-772` `malloc_ncpus`
- `src/jemalloc.c:781-808` `malloc_cpu_count_is_deterministic`

**Expected impact**
Initialization can use garbage CPU counts, leading to bad arena sizing or other wrong CPU-topology decisions.

---

### 24. Registering a new user thread-event handler leaves existing threads stale

Registering a user thread-event handler recalculates event thresholds only for the registering thread. Other live threads keep stale wait counters and thresholds until they later hit an unrelated recalculation point.

**Steps to reproduce**
1. Start worker threads and let them enter a steady allocation or free workload.
2. Register a new user thread-event handler from another thread.
3. Observe the existing workers delay or miss the new handler compared with the registering thread.

**Relevant source locations and functions**
- `src/thread_event_registry.c:195-213` `te_recalculate_current_thread_data`
- `src/thread_event_registry.c:215-238` `te_register_user_handler`
- `src/thread_event.c:198-215` `te_init_waits`
- `src/thread_event.c:295-347` `te_update_alloc_events`, `te_update_dalloc_events`

**Expected impact**
Thread-event delivery becomes inconsistent across already-running threads.

---

### 25. Recent-allocation dumping races concurrent deallocation updates

The recent-allocation dump path iterates dumped records after dropping the recent-allocation mutex, while the deallocation-reset path can still update fields inside those same records.

**Steps to reproduce**
1. Enable recent-allocation tracking.
2. Dump recent allocations from one thread.
3. Free recently recorded sampled allocations concurrently from another thread.

**Relevant source locations and functions**
- `include/jemalloc/internal/prof_structs.h:210-220`
- `src/prof_recent.c:145-185` recent-allocation accessors
- `src/prof_recent.c:195-239` `prof_recent_alloc_reset`
- `src/prof_recent.c:487-531` `prof_recent_alloc_dump_node`
- `src/prof_recent.c:536-582` `prof_recent_alloc_dump`

**Expected impact**
Profiling diagnostics can observe torn record state, emit inconsistent output, or crash under unlucky races.

---

### 26. `prof_log_stop()` can leave profiling stuck in the dumping state

If profiling log shutdown cannot create its output file, it returns before restoring the global log state from `dumping` to `stopped`.

**Steps to reproduce**
1. Start profiling log collection.
2. Point the log output at an unwritable or invalid file path.
3. Stop logging and let file creation fail.

**Relevant source locations and functions**
- `src/prof_log.c:448-460` `prof_log_start`
- `src/prof_log.c:621-654` `prof_log_stop`

**Expected impact**
Profiling log control can wedge permanently for the lifetime of the process.

---

### 27. Long profile dump prefixes can truncate uniqueness suffixes and overwrite dumps

Automatically generated profile dump filenames reserve only `PATH_MAX + 1` bytes total, even though the configurable prefix itself can consume the same budget. Long prefixes can therefore truncate the PID or sequence suffixes that make dump filenames unique.

**Steps to reproduce**
1. Configure a profile dump prefix near the path-length limit.
2. Trigger multiple automatic or explicit profile dumps.
3. Observe later dumps reuse the same truncated filename.

**Relevant source locations and functions**
- `include/jemalloc/internal/prof_types.h:71-75`
- `src/prof_sys.c:678-690` `prof_dump_open_file_impl`
- `src/prof_sys.c:917-954` `prof_dump_filename`
- `src/prof_sys.c:957-969` `prof_get_default_filename`
- `src/prof_sys.c:983-1005` `prof_prefix_set`

**Expected impact**
Profile dumps can overwrite each other and silently lose diagnostic data.

---

### 28. Profiling thread-name validation passes negative `char` values to ctype APIs

The profiling thread-name setter passes plain `char` values directly to `isgraph()` and `isblank()`. On platforms with signed `char`, non-ASCII bytes become negative and invoke undefined behavior.

**Steps to reproduce**
1. Use a platform where `char` is signed.
2. Set a profiling thread name containing bytes above `0x7f`.
3. Let the validation loop call the ctype predicates on those bytes.

**Relevant source locations and functions**
- `src/prof_data.c:453-460` `prof_thread_name_set_impl`

**Expected impact**
Thread-name validation can behave unpredictably or crash on affected libc implementations.

---

### 29. Shared fallback state in `buf_writer_pipe()` is unsynchronized across threads

The OOM fallback path in `buf_writer_pipe()` uses one static buffer and one static writer object for all callers.

**Steps to reproduce**
1. Force buffered writer initialization to fall back to the static backup path.
2. Call `buf_writer_pipe()` concurrently from two threads.
3. Use different callbacks or callback contexts in the two callers.

**Relevant source locations and functions**
- `src/buf_writer.c:41-68` `buf_writer_init`
- `src/buf_writer.c:115-145` `buf_writer_pipe`

**Expected impact**
Concurrent fallback logging can race, corrupt output, or send data through the wrong callback state.

---

### 30. Background-thread control paths can report failure after leaving global state changed

The `background_thread` and `max_background_threads` control handlers flip the global enabled-state flag before calling fallible enable or disable helpers, and do not reliably roll that flag back on failure.

**Steps to reproduce**
1. Trigger a background-thread enable, disable, or resize operation through the control interface.
2. Force the underlying thread creation or shutdown helper to fail.
3. Read the background-thread state after the control call reports failure.

**Relevant source locations and functions**
- `include/jemalloc/internal/background_thread_inlines.h:20-23` `background_thread_enabled_set`
- `src/ctl.c:2012-2059` `background_thread_ctl`
- `src/ctl.c:2063-2117` `max_background_threads_ctl`
- `src/background_thread.c:539-597` `background_thread_create_locked`
- `src/background_thread.c:613-680` `background_threads_enable`, `background_threads_disable`

**Expected impact**
Control-plane state can become misleading or internally inconsistent after a failed background-thread operation.

---

### 31. Prefix-based config parsing accepts malformed keys and enum values

Several configuration branches accept prefix matches instead of exact matches for option names and enumerated values.

**Steps to reproduce**
1. Provide configuration input with truncated or extended prefixes for one of the affected options.
2. Initialize jemalloc or reload the configuration path.
3. Observe the malformed token accepted as if it were a valid key or value.

**Relevant source locations and functions**
- `src/conf.c:469-470` `CONF_MATCH`
- `src/conf.c:569-585` `malloc_conf_init_helper`
- `src/conf.c:588-610` `malloc_conf_init_helper`
- `src/conf.c:813-835` `malloc_conf_init_helper`
- `src/conf.c:910-927` `malloc_conf_init_helper`
- `src/conf.c:1061-1081` `malloc_conf_init_helper`

**Expected impact**
Malformed configuration can silently change allocator behavior instead of being rejected.

---

### 32. `ctl_lookup()` accepts partially numeric path components as valid indexes

Indexed control-path elements are parsed as numbers without verifying that the entire path component was numeric.

**Steps to reproduce**
1. Call a control API with a path component that begins with digits but also contains trailing junk.
2. Reach an indexed control node.
3. Observe the partially numeric component resolve as a valid index.

**Relevant source locations and functions**
- `src/ctl.c:1491-1547` `ctl_lookup`

**Expected impact**
Malformed control names can target the wrong indexed node instead of failing cleanly.

---

### 33. JSON emission does not escape keys or string values

The JSON emitter wraps keys and string values in quotes but does not escape embedded quotes, backslashes, or control characters.

**Steps to reproduce**
1. Supply runtime state that contributes a string containing a quote, backslash, or newline to JSON output.
2. Trigger stats, recent-allocation, or profiler JSON emission.
3. Inspect the generated JSON.

**Relevant source locations and functions**
- `include/jemalloc/internal/emitter.h:121-149` `emitter_emit_str`
- `include/jemalloc/internal/emitter.h:283-289` `emitter_json_key`
- `src/stats.c:1185-1200`, `src/stats.c:1520-1537`, `src/stats.c:1613-1630`
- `src/prof_recent.c:500-519`
- `src/prof_log.c:502-505`, `src/prof_log.c:593-601`

**Expected impact**
Downstream tooling can receive invalid JSON or attacker-controlled JSON structure in emitted reports.

---

### 34. Retained-growth preparation rejects the exact maximum legal size class

The retained-growth helper rejects a grow size exactly equal to `SC_LARGE_MAXCLASS`, even though that size class is otherwise legal and used as the configured upper limit.

**Steps to reproduce**
1. Reach retained growth near the maximum large size class.
2. Make the first acceptable grow size exactly equal to the configured maximum class.
3. Observe the helper reject it instead of selecting it.

**Relevant source locations and functions**
- `include/jemalloc/internal/exp_grow.h:23-37` `exp_grow_size_prepare`
- `src/exp_grow.c:5-13` `exp_grow_init`
- `src/extent.c:708-728` `extent_grow_retained`

**Expected impact**
Boundary-case retained-growth can fail or fragment unexpectedly at the largest supported size.

---

### 35. HPA retained-page stats can underflow because related fields are read without a single snapshot

HPA retained-page reporting computes `total - active - dirty` from separate stats reads that are not tied to one consistent snapshot.

**Steps to reproduce**
1. Run HPA activity that changes page-slab counts while stats are being refreshed.
2. Print allocator stats repeatedly.
3. Let the retained-page formula mix fields from different snapshots.

**Relevant source locations and functions**
- `src/stats.c:852-859` `stats_arena_hpa_shard_counters_print`
- `src/stats.c:960-967`, `src/stats.c:1003-1010`, `src/stats.c:1068-1075` `stats_arena_hpa_shard_slabs_print`
- `src/ctl.c:1899-1917` stats reads
- `src/ctl.c:4053-4066` HPA stats exposure

**Expected impact**
Printed retained-page counts can wrap to large bogus values and mislead diagnostics.

---

### 36. `KQD` expands through an undefined macro

`KQD` expands through `QI`, but the header defines `QD` and does not define `QI`.

**Steps to reproduce**
1. Use the `KQD` macro in a translation unit.
2. Compile the file.
3. Observe preprocessing or compilation fail because the expansion chain is broken.

**Relevant source locations and functions**
- `include/jemalloc/internal/jemalloc_internal_macros.h:28` `KQD`

**Expected impact**
Any path that relies on `KQD` will fail to build.

---

### 37. Hash block readers perform typed reads from arbitrary byte buffers

The hash helpers accept arbitrary byte buffers but cast aligned inputs to `uint32_t *` or `uint64_t *` and dereference them directly, which can violate strict-aliasing or effective-type rules.

**Steps to reproduce**
1. Hash an aligned buffer whose declared type is not `uint32_t[]` or `uint64_t[]`.
2. Compile with aggressive optimization and strict aliasing enabled.
3. Let the aligned fast path perform typed block reads from the byte buffer.

**Relevant source locations and functions**
- `include/jemalloc/internal/hash.h:26-48` `hash_get_block_32`, `hash_get_block_64`
- `include/jemalloc/internal/hash.h:85-89` `hash_x86_32`
- `include/jemalloc/internal/hash.h:148-156` `hash_x86_128`
- `include/jemalloc/internal/hash.h:312-317` `hash_x64_128`
- `include/jemalloc/internal/hash.h:422-434` `hash`

**Expected impact**
Stricter toolchains can miscompile hashing paths or produce incorrect hash results.

---

### 38. Decay smoothstep arithmetic can overflow on very large page backlogs

Several decay calculations multiply page counts by fixed-point smoothstep weights and accumulate the results in `uint64_t` without overflow checks.

**Steps to reproduce**
1. Populate decay state with very large synthetic or real page counts.
2. Trigger backlog-limit or purge-estimation calculations.
3. Let the weighted products and sums exceed `uint64_t`.

**Relevant source locations and functions**
- `include/jemalloc/internal/smoothstep.h:27-30`, `include/jemalloc/internal/smoothstep.h:130`, `include/jemalloc/internal/smoothstep.h:230`
- `src/decay.c:97-109` `decay_backlog_npages_limit`
- `src/decay.c:159-173` `decay_npages_purge_in`
- `src/decay.c:229-239` `decay_npurge_after_interval`

**Expected impact**
Very large heaps can drive wrapped purge calculations and distort decay behavior.
</details>

NITIN9181:
Fixes #2892

**Description:**
The Small Extent Cache (SEC) configuration allowed the `nshards` to be configured with any `size_t` value. However, the internal functions that select and cache shard identifiers (like `sec_shard_pick`) use a `uint8_t`. If `nshards` exceeded 255, the shard IDs silently wrapped around. Furthermore, an unbounded `nshards` triggered a multiplication integer overflow when allocating the internal SEC metadata arrays.

This PR introduces two safety nets:
1. `hpa_sec_nshards` and `experimental_pac_sec_nshards` in `src/conf.c` now use `CONF_CHECK_MAX` with a maximum limit of 255. Any configuration values greater than 255 will automatically be clipped safely.
2. `sec_init()` in `src/sec.c` now asserts `opts->nshards <= 255` to guarantee no oversized configurations can trigger a buffer overflow or wrap-around programmatically.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -937,8 +937,8 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 				CONF_CONTINUE;
 			}
 			CONF_HANDLE_SIZE_T(opt_hpa_sec_opts.nshards,
-			    "hpa_sec_nshards", 0, 0, CONF_CHECK_MIN,
-			    CONF_DONT_CHECK_MAX, true);
+			    "hpa_sec_nshards", 0, 255, CONF_CHECK_MIN,
+			    CONF_CHECK_MAX, true);
 			CONF_HANDLE_SIZE_T(opt_hpa_sec_opts.max_alloc,
 			    "hpa_sec_max_alloc", PAGE,
 			    USIZE_GROW_SLOW_THRESHOLD, CONF_CHECK_MIN,
@@ -947,8 +947,8 @@ malloc_conf_init_helper(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
 			    "hpa_sec_max_bytes", SEC_OPTS_MAX_BYTES_DEFAULT, 0,
 			    CONF_CHECK_MIN, CONF_DONT_CHECK_MAX, true);
 			CONF_HANDLE_SIZE_T(opt_pac_sec_opts.nshards,
-			    "experimental_pac_sec_nshards", 0, 0,
-			    CONF_CHECK_MIN, CONF_DONT_CHECK_MAX, true);
+			    "experimental_pac_sec_nshards", 0, 255,
+			    CONF_CHECK_MIN, CONF_CHECK_MAX, true);
 			CONF_HANDLE_SIZE_T(opt_pac_sec_opts.max_alloc,
 			    "experimental_pac_sec_max_alloc", PAGE,
 			    USIZE_GROW_SLOW_THRESHOLD, CONF_CHECK_MIN,
--- a/src/sec.c
+++ b/src/sec.c
@@ -36,6 +36,7 @@ sec_init(tsdn_t *tsdn, sec_t *sec, base_t *base, const sec_opts_t *opts) {
 		sec->opts.max_alloc = 0;
 		return false;
 	}
+	assert(opts->nshards <= 255);
 	assert(opts->max_alloc >= PAGE);
 
 	/*


guangli-dai (approved):

guangli-dai:
Thank you for the patch! @NITIN9181

PR #2970, opened 2026-07-29
Background flaky test on FreeBSD 32-bit CI fix

guangli-dai:
1. Fix the test by reducing narenas used there to avoid virtual memory address exhaustion on 32-bit platforms;
2. Add comments in to retain the context as a follow-up of #2963


gctony (approved):
LGTM

PR #2971, opened 2026-07-30
Fix memory leak in extent_hooks split failure path (#2815)

### LINKED ISSUE #2815: Memory leak: `arena_extent_dalloc()` is not called for extents for which `arena_extent_split()` failed
There is a memory leak: `arena_extent_dalloc()` is not called for extents for which `arena_extent_split()` failed

An example "trace" from gdb (breakpoints were set in all jemalloc hooks):

```
arena_extent_alloc (extent_hooks=0x7ffff7da6640 <arena_extent_hooks>, new_addr=0x0, size=2097152, alignment=4096, zero=0x7fffffffbf1e, commit=0x7fffffffbf1f, arena_ind=49)
arena_extent_split (extent_hooks=0x7ffff7da6640 <arena_extent_hooks>, addr=0x7ffff67ff000, size=2097152, size_a=4096, size_b=2093056, committed=true, arena_ind=49) - FAILED
arena_extent_alloc (extent_hooks=0x7ffff7da6640 <arena_extent_hooks>, new_addr=0x0, size=4096, alignment=4096, zero=0x7fffffffbe6c, commit=0x7fffffffbf1d, arena_ind=49)
arena_extent_dalloc (extent_hooks=0x7ffff7da6640 <arena_extent_hooks>, addr=0x7ffff7702000, size=4096, committed=true, arena_ind=49)
```

As you can see above `arena_extent_dalloc()` was not called for the first allocated extent.

jemalloc commit: 81f35e0b55c52cb0c3e1171afd477e1cb66fafaf

Ref: https://github.com/oneapi-src/unified-memory-framework/pull/902#issuecomment-2721334161

NITIN9181:
Fixes #2815

## Description
This PR addresses a memory leak that occurs when custom extent hooks are used, specifically when the custom `split` hook returns an error (failure to split). 

Previously, if `extent_recycle_split` or `extent_grow_retained` requested a split via the `split` hook and it failed, the code would handle the failure by calling `extents_abandon_vm` on the un-splittable region. While `extents_abandon_vm` properly cleans up internal extent metadata, it *leaks the underlying virtual memory mapping*. The user's custom `dalloc` hook was never invoked to clean up the original allocation.

This PR modifies the failure paths in `extent_recycle_split` and `extent_grow_retained` to explicitly call `extent_dalloc_wrapper` instead of `extents_abandon_vm`. This ensures that `jemalloc` appropriately delegates the un-splittable region back to the custom deallocation hook (`dalloc`), completely preventing the virtual memory leak.

## Changes Made
- **`src/extent.c: extent_recycle_split`**: Replaced calls to `extents_abandon_vm` with `extent_dalloc_wrapper` for the `to_salvage` and `to_leak` segments when a split operation fails.
- **`src/extent.c: extent_grow_retained`**: Replaced the call to `extents_abandon_vm` with `extent_dalloc_wrapper` on the split failure path.
- **`src/extent.c: extent_grow_retained`**: Added `extent_gdump_add(tsdn, to_salvage)` and `extent_gdump_add(tsdn, to_leak)` right before their respective deallocations to ensure heap profiling metadata remains balanced during the failure path cleanup.

## Verification
- Verified by creating a test program where custom `alloc`, `split`, and `dalloc` hooks were registered.
- Hardcoded the `split` hook to always return `true` (failure).
- Verified that previously this would leak memory, but with this PR applied, `custom_dalloc` is successfully invoked to clean up the allocation after the split failure.
- Compiled and verified on an Ubuntu Linux Docker environment.

### DIFF
--- a/src/extent.c
+++ b/src/extent.c
@@ -606,20 +606,14 @@ extent_recycle_split(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks,
 		 * fulfill our allocation request.
 		 */
 		assert(result == extent_split_interior_error);
+		malloc_mutex_unlock(tsdn, &ecache->mtx);
 		if (to_salvage != NULL) {
-			extent_deregister(tsdn, pac, to_salvage);
+			extent_dalloc_wrapper(tsdn, pac, ehooks, to_salvage);
 		}
 		if (to_leak != NULL) {
-			extent_deregister_no_gdump_sub(tsdn, pac, to_leak);
-			/*
-			 * May go down the purge path (which assume no ecache
-			 * locks).  Only happens with OOM caused split failures.
-			 */
-			malloc_mutex_unlock(tsdn, &ecache->mtx);
-			extents_abandon_vm(tsdn, pac, ehooks, ecache, to_leak,
-			    growing_retained);
-			malloc_mutex_lock(tsdn, &ecache->mtx);
+			extent_dalloc_wrapper(tsdn, pac, ehooks, to_leak);
 		}
+		malloc_mutex_lock(tsdn, &ecache->mtx);
 		return NULL;
 	}
 	unreachable();
@@ -811,9 +805,10 @@ extent_grow_retained(tsdn_t *tsdn, pac_t *pac, ehooks_t *ehooks, size_t size,
 			pac_record_grown(tsdn, pac, ehooks, to_salvage);
 		}
 		if (to_leak != NULL) {
-			extent_deregister_no_gdump_sub(tsdn, pac, to_leak);
-			extents_abandon_vm(tsdn, pac, ehooks,
-			    &pac->ecache_retained, to_leak, true);
+			if (config_prof) {
+				extent_gdump_add(tsdn, to_leak);
+			}
+			extent_dalloc_wrapper(tsdn, pac, ehooks, to_leak);
 		}
 		goto label_err;
 	}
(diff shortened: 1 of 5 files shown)


NITIN9181:
I've pushed a fix for the AppVeyor test failure (<jemalloc>: Should own 0 locks of rank >= 14: extent_grow(17)). The error paths in extent_grow_retained and extent_recycle_split were calling extent_dalloc_wrapper while holding pac->grow_mtx (rank 17), which caused a lock rank violation since extent_dalloc_wrapper requires no core locks to be held. I've updated the code to drop the lock before deallocating.

NITIN9181:
@guangli-dai please review

guangli-dai:
Some CIs are failing including [windows](https://ci.appveyor.com/project/jasone/jemalloc/builds/54463280/job/v9m1g6372b3xim8n), [mac](https://github.com/jemalloc/jemalloc/actions/runs/30603225674/job/91828359149?pr=2971), and more. Can you fix? Also, please rebase to dev and [this commit](https://github.com/jemalloc/jemalloc/pull/2971/changes/ae5d7e1a149620a6766b0277d714fe9086abc006) can be removed.

src/extent.c
nikhilkrishnan-ai, line 1141: `return NULL);``

nikhilkrishnan-ai (commented):
Files changed

PR #2974, opened 2026-08-03
Os layer 3

guangli-dai:
Following #2963  and #2972, this PR is the last of the OS layer migration, it covers modules process, VM management, boot-time setup for overcommit, error code handling, and FMT printf macros.

Some special cases in this PR:
1. Commit "Move boot and overcommit setup" is able to remove some macro symbols which were used to record the OS detection results found in `configure.ac`. Instead, per-OS dispatching is now done in `os.h`;
2. Error code handling and FMT module migration are required in some utilities and included before `os.h` is included. Still put them in the list of `os.h` dispatching list just for observability, they do not rely on `os.h` dispatching.

### DIFF
--- a/include/jemalloc/internal/os/posix/overcommit.h
+++ b/include/jemalloc/internal/os/posix/overcommit.h
@@ -1,0 +1,22 @@
+#ifndef JEMALLOC_INTERNAL_OS_POSIX_OVERCOMMIT_H
+#define JEMALLOC_INTERNAL_OS_POSIX_OVERCOMMIT_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+
+#include <sys/mman.h>
+
+/* Defined in src/pages.c. */
+extern int mmap_flags;
+
+JEMALLOC_ALWAYS_INLINE bool
+os_overcommit_boot(void) {
+	mmap_flags = MAP_PRIVATE | MAP_ANON;
+#ifdef __NetBSD__
+	os_overcommits = true;
+#else
+	os_overcommits = false;
+#endif
+	return false;
+}
+
+#endif /* JEMALLOC_INTERNAL_OS_POSIX_OVERCOMMIT_H */
--- a/include/jemalloc/internal/os/windows/vm.h
+++ b/include/jemalloc/internal/os/windows/vm.h
@@ -2,6 +2,7 @@
 #define JEMALLOC_INTERNAL_OS_WINDOWS_VM_H
 
 #include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/os/overcommit.h"
 
 JEMALLOC_ALWAYS_INLINE void *
 os_vm_reserve(void *hint, size_t size, size_t alignment, bool *commit) {
@@ -110,6 +111,13 @@ JEMALLOC_ALWAYS_INLINE bool
 os_vm_purge_lazy(void *addr, size_t size) {
 	VirtualAlloc(addr, size, MEM_RESET, PAGE_READWRITE);
 	return false;
+}
+
+JEMALLOC_ALWAYS_INLINE size_t
+os_vm_page_size(void) {
+	SYSTEM_INFO si;
+	GetSystemInfo(&si);
+	return (size_t)si.dwPageSize;
 }
 
 #endif /* JEMALLOC_INTERNAL_OS_WINDOWS_VM_H */
--- a/include/jemalloc/internal/os/windows/vm.h
+++ b/include/jemalloc/internal/os/windows/vm.h
@@ -1,0 +1,115 @@
+#ifndef JEMALLOC_INTERNAL_OS_WINDOWS_VM_H
+#define JEMALLOC_INTERNAL_OS_WINDOWS_VM_H
+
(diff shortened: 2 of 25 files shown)


src/jemalloc_init.c
```diff
@@ -347,13 +347,18 @@ malloc_init_hard_recursible(void) {
 		}
 	}
 
-#if (defined(JEMALLOC_HAVE_PTHREAD_ATFORK) && !defined(JEMALLOC_MUTEX_INIT_CB) \
-    && !defined(JEMALLOC_ZONE) && !defined(_WIN32)                             \
-    && !defined(__native_client__))
-	/* LinuxThreads' pthread_atfork() allocates. */
```
gctony, line 353: Is this comment still relevant?

include/jemalloc/internal/os/windows/vm.h
```diff
@@ -0,0 +1,123 @@
...
+	        : (!VirtualFree(addr, size, MEM_DECOMMIT)));
+}
+
+JEMALLOC_ALWAYS_INLINE bool
+os_vm_commit(void *addr, size_t size) {
+	return os_vm_commit_impl(addr, size, true);
+}
+
+JEMALLOC_ALWAYS_INLINE bool
+os_vm_decommit(void *addr, size_t size) {
```
gctony, line 79: Should this be `uncommit` (instead of `decommit`)?

include/jemalloc/internal/os/windows/vm.h
```diff
@@ -0,0 +1,123 @@
...
+	return os_vm_commit_impl(addr, size, false);
```
gctony, line 80: Also, why have you introduced the `_impl()` functions here (same for the posix file too)?

include/jemalloc/internal/os/windows/vm.h
guangli-dai: `decommit` is the one used all the time like [here](https://github.com/jemalloc/jemalloc/blob/85599c8b1213ffd410195885663920b463732681/src/pages.c#L400)

include/jemalloc/internal/os/windows/vm.h
guangli-dai: To avoid redundancy in codes because commit and decommit basically share the same common path and just differ in the flag passed in. Before migration, the previous function [`os_pages_commit`](https://github.com/jemalloc/jemalloc/blob/85599c8b1213ffd410195885663920b463732681/src/pages.c#L356) is designed the same way although its name it a bit misleading (called commit but can actually be used for decommitting by passing `commit=false`).

src/jemalloc_init.c
guangli-dai: That comment was added [quite a while ago in 2013](https://github.com/jemalloc/jemalloc/commit/ac4403ca) and current LinuxThreads are replaced by [NPTL](https://embeddedpathashala.com/linuxthreads-vs-nptl-linux-pthreads/). So the comment does not seem necessary. However, we have the protection mechanism remained which makes jemalloc able to handle potential reentrancy here (codes are byte-to-byte identical).

include/jemalloc/internal/os/windows/vm.h
gctony: I see, OK!

gctony (approved):
LGTM

include/jemalloc/internal/os/posix/overcommit.h
```diff
@@ -0,0 +1,21 @@
+#ifndef JEMALLOC_INTERNAL_OS_POSIX_OVERCOMMIT_H
+#define JEMALLOC_INTERNAL_OS_POSIX_OVERCOMMIT_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+
+#include <sys/mman.h>
+
+extern int mmap_flags;
```
swkrueger, line 9: Would it be worth adding a note that this is declared in src/pages.c?

src/pages.c
```diff
@@ -8,40 +8,16 @@
...
 size_t os_page;
 
-#ifndef _WIN32
-#	define PAGES_PROT_COMMIT (PROT_READ | PROT_WRITE)
-#	define PAGES_PROT_DECOMMIT (PROT_NONE)
-static int mmap_flags;
-#endif
-static bool os_overcommits;
+/* Set here. Consumed by os_vm_reserve/os_vm_commit_impl in os/posix/vm.h. */
+int mmap_flags;
```
swkrueger, line 18: Could there be any consequences or risks from changing these variables from static to global? Do we have protection against these internal symbol leaking into the application's namespace?

guangli-dai: I don't see any, there are quite some existing conventions there already and these global variables will be in the generated `private_namespace.h` per current [build automation](https://github.com/jemalloc/jemalloc/blob/dev/include/jemalloc/internal/private_namespace.sh). They are internal only and won't be exposed to application's namespace.

include/jemalloc/internal/os/posix/overcommit.h
guangli-dai: Current convention doesn't do that, but taking another look, no harm to specify in the comments. Added that in.

PR #2979, opened 2026-08-12
Fix background thread state after creation failure

xinydev:
The background_thread mallctl enables workers dynamically, and changing max_background_threads while enabled also restarts them through background_threads_enable(). Before creating thread 0, this function marks the required nonzero worker slots as started.

If thread 0 creation fails, only its own slot is reset. The ctl returns EFAULT, but the enabled flag, pre-marked slots, and n_background_threads remain stale. A repeated enable is ignored because the flag is already true. Foreground decay may also keep deferring purging despite no background thread being available.

Roll back all marked state and disable background threads so the failed request leaves the allocator consistent. A later explicit enable can retry cleanly.

### DIFF
--- a/src/background_thread.c
+++ b/src/background_thread.c
@@ -703,9 +703,11 @@ background_threads_enable(tsd_t *tsd) {
 	assert(background_thread_enabled());
 	malloc_mutex_assert_owner(tsd_tsdn(tsd), &background_thread_lock);
 
-	VARIABLE_ARRAY(bool, marked, max_background_threads);
+	const size_t const_max_background_threads = max_background_threads;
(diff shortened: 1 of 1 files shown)


xinydev:
The error in CI appears to be related to the test environment.

guangli-dai:
Some CIs were failing irrelevantly, but the static analysis warning looks real: https://github.com/jemalloc/jemalloc/actions/runs/31561192205/artifacts/9153345575. Please fix that.

xinydev:
@guangli-dai I've updated the patch. Could you approve CI and take another look?

guangli-dai (approved):
Thanks @xinydev for the patch!

PR #2980, opened 2026-08-20
OS layer 4

guangli-dai:
Remove some unnecessary OS detection conditions and migrate some more modules to the OS layer:
1. process maps and PID namespace;
2. CPU affinity setup;
3. Spin's thread yield;
4. symlink reading.

### DIFF
--- a/include/jemalloc/internal/os/darwin/proc_maps.h
+++ b/include/jemalloc/internal/os/darwin/proc_maps.h
@@ -22,7 +22,7 @@ typedef struct segment_command os_segment_command_t;
 #	define OS_LC_SEGMENT_VALUE LC_SEGMENT
 #endif
 
-JEMALLOC_ALWAYS_INLINE long
+JEMALLOC_ALWAYS_INLINE uint64_t
 os_prof_pid_namespace(void) {
 	/* Not supported on Darwin. */
 	return 0;
--- a/include/jemalloc/internal/os/darwin/proc_maps.h
+++ b/include/jemalloc/internal/os/darwin/proc_maps.h
@@ -1,0 +1,87 @@
+#ifndef JEMALLOC_INTERNAL_OS_DARWIN_PROC_MAPS_H
+#define JEMALLOC_INTERNAL_OS_DARWIN_PROC_MAPS_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/malloc_io.h"
+
+#include <mach-o/dyld.h>
+
+#define JEMALLOC_OS_PROF_NO_OPEN_MAPS
+
+#ifdef __LP64__
+typedef struct mach_header_64     os_mach_header_t;
+typedef struct segment_command_64 os_segment_command_t;
+#	define OS_MH_MAGIC_VALUE MH_MAGIC_64
+#	define OS_MH_CIGAM_VALUE MH_CIGAM_64
+#	define OS_LC_SEGMENT_VALUE LC_SEGMENT_64
+#else
+typedef struct mach_header     os_mach_header_t;
+typedef struct segment_command os_segment_command_t;
+#	define OS_MH_MAGIC_VALUE MH_MAGIC
+#	define OS_MH_CIGAM_VALUE MH_CIGAM
+#	define OS_LC_SEGMENT_VALUE LC_SEGMENT
+#endif
+
+JEMALLOC_ALWAYS_INLINE long
+os_prof_pid_namespace(void) {
+	/* Not supported on Darwin. */
+	return 0;
+}
+
+JEMALLOC_ALWAYS_INLINE int
+os_prof_open_maps(void) {
+	/* Darwin dumps dyld images instead of reading an fd-based maps file. */
+	return -1;
+}
+
+JEMALLOC_ALWAYS_INLINE void
+os_prof_dump_dyld_image_vmaddr(buf_writer_t *buf_writer, uint32_t image_index) {
+	const os_mach_header_t *header = (const os_mach_header_t *)
+	    _dyld_get_image_header(image_index);
+	if (header == NULL
+	    || (header->magic != OS_MH_MAGIC_VALUE
+	        && header->magic != OS_MH_CIGAM_VALUE)) {
+		/* Invalid header. */
+		return;
+	}
+
+	intptr_t             slide = _dyld_get_image_vmaddr_slide(image_index);
+	const char          *name = _dyld_get_image_name(image_index);
+	struct load_command *load_cmd = (struct load_command *)((char *)header
+	    + sizeof(os_mach_header_t));
+	for (uint32_t i = 0; load_cmd && (i < header->ncmds); i++) {
+		if (load_cmd->cmd == OS_LC_SEGMENT_VALUE) {
+			const os_segment_command_t *segment_cmd =
+			    (const os_segment_command_t *)load_cmd;
+			if (!strcmp(segment_cmd->segname, "__TEXT")) {
+				char buffer[PATH_MAX + 1];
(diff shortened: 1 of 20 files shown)


gctony (approved):
LGTM

include/jemalloc/internal/os/freebsd/proc_maps.h
```diff
@@ -0,0 +1,67 @@
...
+#include "jemalloc/internal/malloc_io.h"
+
+#define JEMALLOC_OS_PROF_OPEN_MAPS_SUPPORTED
+
+JEMALLOC_ALWAYS_INLINE long
+os_prof_pid_namespace(void) {
+	long ret = 0;
+	char buf[PATH_MAX];
+	ssize_t linklen =
+#ifndef JEMALLOC_READLINKAT
```
spredolac: `  ssize_t linklen = os_readlink("/proc/curproc/ns/pid", buf, PATH_MAX);`

include/jemalloc/internal/os/posix/proc_maps.h
```diff
@@ -0,0 +1,84 @@
...
+#include "jemalloc/internal/malloc_io.h"
+#include "jemalloc/internal/os/process.h"
+
+#define JEMALLOC_OS_PROF_OPEN_MAPS_SUPPORTED
+
+JEMALLOC_ALWAYS_INLINE long
+os_prof_pid_namespace(void) {
+	long ret = 0;
+	char buf[PATH_MAX];
+	ssize_t linklen =
```
spredolac: ` ssize_t linklen = os_readlink("/proc/self/ns/pid", buf, PATH_MAX);`

include/jemalloc/internal/os/windows/proc_maps.h
```diff
@@ -0,0 +1,40 @@
+#ifndef JEMALLOC_INTERNAL_OS_WINDOWS_PROC_MAPS_H
+#define JEMALLOC_INTERNAL_OS_WINDOWS_PROC_MAPS_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/malloc_io.h"
+
+#define JEMALLOC_OS_PROF_OPEN_MAPS_SUPPORTED
```
spredolac: a bit confusing. Name is `JEMALLOC_OS_PROF_OPEN_MAPS_SUPPORTED` yet below `os_prof_open_maps()`  returns -1; as in "not supported" ?

  You do not define it in Darwin only, a new backend that forgets the `#define` would silently disable maps dumping and skips the unit test `test_mdump_maps_error` . 

  nit: maybe declare os_prof_open_maps in os/proc_maps.h "required for implementation" block, and invert the macro to `JEMALLOC_OS_PROF_NO_OPEN_MAPS` defined by Darwin only where the other function is not using the hook?

include/jemalloc/internal/os/windows/proc_maps.h
```diff
@@ -0,0 +1,40 @@
...
+	int mfd = *(int *)read_cbopaque;
+	assert(mfd != -1);
+	return malloc_read_fd(mfd, buf, limit);
+}
+
+JEMALLOC_ALWAYS_INLINE void
+os_prof_dump_maps(buf_writer_t *buf_writer, int (*open_maps)(void)) {
+	int mfd = open_maps();
+	if (mfd == -1) {
+		return;
```
spredolac, line 34: Maybe add a comment here stating that because of `os_prof_open_maps` returning -1 the rest is unreachable?

include/jemalloc/internal/pages.h
```diff
@@ -61,12 +61,12 @@ extern size_t os_page;
  * re-commit, but doing so is racy, and if re-commit fails it's a pain to
  * propagate the "poisoned" memory state.  Since we typically decommit as the
  * next step after purging on Windows anyway, there's no point in adding such
- * complexity.
+ * complexity.  Both conditions cannot be true on windows and therefore no need
```
spredolac: nit: Neither condition can be true on Windows

include/jemalloc/internal/os/darwin/proc_maps.h
```diff
@@ -0,0 +1,79 @@
+#ifndef JEMALLOC_INTERNAL_OS_DARWIN_PROC_MAPS_H
+#define JEMALLOC_INTERNAL_OS_DARWIN_PROC_MAPS_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/malloc_io.h"
+
+#include <mach-o/dyld.h>
+
+#ifdef __LP64__
+typedef struct mach_header_64     mach_header_t;
```
spredolac: Add `os_` in front of these for consistency, but especially `mach_header_t` may conflict

```
typedef struct mach_header_64     os_mach_header_t;
typedef struct segment_command_64 os_segment_command_t;
#    define OS_MH_MAGIC_VALUE MH_MAGIC_64
#    define OS_MH_CIGAM_VALUE MH_CIGAM_64
#    define OS_LC_SEGMENT_VALUE LC_SEGMENT_64
#else
typedef struct mach_header     os_mach_header_t;
typedef struct segment_command os_segment_command_t;
#    define OS_MH_MAGIC_VALUE MH_MAGIC
#    define OS_MH_CIGAM_VALUE MH_CIGAM
#    define OS_LC_SEGMENT_VALUE LC_SEGMENT

```

include/jemalloc/internal/os/freebsd/proc_maps.h
```diff
@@ -0,0 +1,67 @@
...
+#else
+	    readlinkat(AT_FDCWD, "/proc/curproc/ns/pid", buf, PATH_MAX)
+#endif
+	    ;
+
+	/* namespace string is expected to be like pid:[4026531836] */
+	if (linklen > 0) {
+		/* Trim the trailing "]" */
+		buf[linklen - 1] = '\0';
+		char *index = strtok(buf, "pid:[");
```
spredolac: So, this will sue static buffer behind the scene (we serialize, but it is for the whole process not just jemalloc part). Maybe use 
```c
                char *saveptr;
                char *index = strtok_r(buf, "pid:[", &saveptr);
                if (index != NULL) {
                        ret = atol(index);
                }
        }
```
 or actually even better to parse yourself as it will be more robust to certain forms (code below generated with AI thats suggests it could get those forms in some unusual cases). 

```
        static const char prefix[] = "pid:[";
        const size_t      prefix_len = sizeof(prefix) - 1;
        if (linklen > (ssize_t)prefix_len && (size_t)linklen <= sizeof(buf)
            && buf[linklen - 1] == ']'
            && memcmp(buf, prefix, prefix_len) == 0) {
                buf[linklen - 1] = '\0'; /* Trim the trailing "]". */
                char      *end;
                uintmax_t  ns = malloc_strtoumax(&buf[prefix_len], &end, 10);
                if (end != &buf[prefix_len] && *end == '\0') {
                        ret = (long)ns;
                }
        }
```

spredolac (commented):
Few small things and something that you did not introduce, but should be fixed (`strtok`)

include/jemalloc/internal/os/windows/proc_maps.h
guangli-dai: Changed to `JEMALLOC_OS_PROF_NO_OPEN_MAPS` instead and mac has an implementation returning -1 but will never be used.

include/jemalloc/internal/os/freebsd/proc_maps.h
guangli-dai: Added a separate commit at the end for this. Also changed `long` to `uint64_t`

spredolac (approved):

PR #2983, opened 2026-08-24
dynamic page size

gctony:
PR that introduces the dynamic page size feature. This PR will contain a series of commits, each updating a different module of jemalloc, to make the change more easy to code review:

commit1:

* Add --enable-dyanamic-page-size, --with-min-lg-page=, --with-max-lg-page= configure args.
* Add necessary macros.
* Move page size initialization to a separate function that's called earlier in the jemalloc initialization sequence.
* Split conf initialization to address a tricky ordering issue: Page size initialization needs opt_abort and maybe a user-specified lg page size (to be introduced later). But it also needs to happen before the main bulk of options are processed, as some need the page size to calculated defaults. So, we added an early initialization step for a small number of options.

commit2:

* Update hpdata and HUGEPAGE_PAGES macro to work with the dynamic page size.
* A lot of unit tests needed to be updated to malloc instead of stack-allocating some arrays, as they were getting very large and causing SEGVs on arm64.

commit3:

* Update sc and SC_NPSIZES macro to work with the dynamic page size.

commit4:

* Update sc and  SC_NBINS / SC_LG_LARGE_MINCLASS macros to work with the dynamic page size.
* Update tcache and TCACHE_BINS macro to work with the dynamic page size.
* There was already a TCACHE_NBINS_MAX macro and I needed to add a max macro on top of that. So I renamed TCACHE_NBINS_MAX as TCACHE_NBINS and introduced TCACHE_NBINS_MAX as the max value for TCACHE_NBINS. Let me know if I should use TCACHE_NBINS_MAX and TCACHE_NBINS_MAX_MAX instead.

commit5:

* Update rtree relater macros to work with the dynamic page size.
* Given that rtree_levels are now not a constant, they are initialized in rtree_new() when dynamic page size is enabled. But they are compiled in as a constant when it is disabled. I introduced some macros for the rtree level values so that I don't replicate the code in these two cases.

commit6:

* Change static initializers that were called on some HPA/PAC globals to be called at initialization, after page_size has been set.

commit7:

* Replace static initializers that were called on some unit tests with a function that is called after page_size has been set.

commit8:

* Update slab/edata to use a conservative static size for `e_slab_data` in `edata_t`. This can add an unreasonable footprint overhead on jemalloc's metadata due to an unreasonable increase in edata (8x-9x). We will fix this after the first version of this change lads.
* I removed `SC_SLAB_MAXREGS` as it was not being used anywhere.

commit9:

* Renamed `lg_page` as `lg_page_size`, as `lg_page` is used as a function argument in some places.
* When dynamic page size is enabled, `PAGE` and `LG_PAGE` will now pick up the dynamic value (`page_size` and `lg_page_size`).
* Remove `DYNAMIC_PAGE` and DYNAMIC_LG_PAGE` and replace them with `PAGE` and `LG_PAGE`.
* Small numbers of tweaks to avoid compilation warnings.
* Finally set `lg_page_size` from `os_page`.


configure.ac
```diff
@@ -2095,6 +2146,13 @@ if test "x${LG_PAGE}" != "xundefined" -a \
         "${je_cv_lg_hugepage}" -lt "${LG_PAGE}" ; then
   AC_MSG_ERROR([Huge page size (2^${je_cv_lg_hugepage}) must be at least page size (2^${LG_PAGE})])
 fi
+dnl With a dynamic page size the runtime page size can grow to 2^MAX_LG_PAGE,
+dnl so that, rather than LG_PAGE, is what has to fit inside a huge page.
+dnl Otherwise HUGEPAGE_PAGES becomes 0 and the HPA bitmaps degenerate.
```
guangli-dai, line 2152: Please also update the configuration outputs [here](https://github.com/jemalloc/jemalloc/blob/dev/configure.ac#L3139) so that users can clearly understand `LG_PAGE` input is no longer respected, instead the page size read dynamically at the runtime will be used.

src/pages.c
```diff
@@ -454,15 +459,86 @@ init_thp_state(void) {
...
+
+	/*
+	 * As we incrementally migrate the code to use a dynamic page size,
+	 * we'll fix the dynamic lg page to be same as LG_PAGE. This way,
+	 * we can migrate parts of the code to the dynamic version without
+	 * breaking anything. When all the changes are done, we'll update this to:
+	 *   lg_page = lg_floor(os_page)
+	 * or read it from the config.
+	 */
+	lg_page = LG_PAGE;
```
guangli-dai: Since in the transitional stage, we still use `LG_PAGE` should we still assert that `os_page <= LG_PAGE`? Then we can remove the assertion when the dynamic page size feature is completely done in the last commit.

src/hpa.c
```diff
@@ -56,7 +56,7 @@ hpa_supported(void) {
 	 * If we couldn't detect the value of HUGEPAGE, HUGEPAGE_PAGES becomes
 	 * this sentinel value -- see the comment in pages.h.
 	 */
-	if (HUGEPAGE_PAGES == 1) {
```
guangli-dai, line 59: I would still view `HUGEPAGE_PAGES == 1` as an invalid case: in this case, there is no need to introduce HPA shard and psset management: it's a completely waste of cycles in managing things that can easily be done within the per-PAGE management. Even when using dynamic page sizes, users should make sure their OS has a valid setup to use HPA, i.e., HUGEPAGE > PAGE.

configure.ac
gctony: Great suggestion! I added:

```
if test "x${enable_dynamic_page_size}" = "x0" ; then
  AC_MSG_RESULT([LG_PAGE            : ${LG_PAGE}])
else
  AC_MSG_RESULT([MIN_LG_PAGE        : ${MIN_LG_PAGE}])
  AC_MSG_RESULT([MAX_LG_PAGE        : ${MAX_LG_PAGE}])
fi
```

src/pages.c
gctony: Good point. And, given that we'll eventually allow the user to set lg_page in the conf, I think I'll do a runtime check that os_page <= page_size.

gctony: I moved the `if (os_page > DYNAMIC_PAGE) {` check outside the `#ifdef DYNAMIC_PAGE_SIZE ...`

src/hpa.c
gctony: The diff is:

```
-       if (HUGEPAGE_PAGES == 1) {
+       if (LG_HUGEPAGE == 0) {
```

and `LG_HUGEPAGE == 0` will be a compile-time expression, as `LG_HUGEPAGE` is a constant. I guess I'm not sure what else I can do here. Also, why is `HUGEPAGE_PAGES` set to `1` and not `0`?

Re: `HUGEPAGE > PAGE`: the configure script checks that `LG_HUGEPAGE > MAX_LG_PAGE`.

configure.ac
```diff
@@ -2133,6 +2147,13 @@ if test "x${LG_PAGE}" != "xundefined" -a \
...
+if test "x${enable_dynamic_page_size}" = "x1" -a \
+        "${je_cv_lg_hugepage}" -lt "${MAX_LG_PAGE}" ; then
```
guangli-dai: We should be careful with hugepage == page case, it should be viewed as invalid making hpa unsupported. Specifically, on dev all such HPA tests will be skipped:
```
$ ./autoconf.sh --with-lg-page=12 --with-lg-hugepage=12 && make -j64 && make tests -j64 && sh test/test.sh test/unit/hpa
...
test_alloc_max:test/unit/hpa.c:130: Test skipped: (!hpa_supported())
...
```
Under this change, several unit test will fail there:
```
$ ./autoconf.sh--with-lg-page=12 --with-lg-hugepage=12 && make -j64 && make tests -j64 && sh test/test.sh test/unit/hpa
...
test_starts_huge:test/unit/hpa.c:742: Failed assertion: (stat->full_slabs[1].npageslabs) == (0) --> 1 != 0:
test_starts_huge:test/unit/hpa.c:759: Failed assertion: (stat->full_slabs[0].npageslabs) == (0) --> 1 != 0:
```
With dynamic page size enabled, it is even worse, executing 
```
./autogen.sh --enable-dynamic-page-size --with-min-lg-page=12 --with-max-lg-page=12 --with-lg-hugepage=12 && make -j64 && make tests -j64 && sh test/test.sh test/unit/hpa
```
would get it stuck. At a glance, it should be stuck in the `prng_range_zu(...)` call. 

To conclude, we should 1. stop the invalid inputs where hugepage = page here by changing it to `-le` and 2. add `HUGEPAGE_PAGES == 1` detection back to `hpa_supported()` to prevent HPA from being launched there so that even if someone overrides HUGEPAGE value through compiler options to an invalid one, the harm can be minimized.

configure.ac
```diff
@@ -1938,6 +1901,57 @@ if test "x$enable_uaf_detection" = "x1" ; then
...
+  [Enable dynamic page size])],
+[if test "x$enable_dynamic_page_size" = "xno" ; then
+  enable_dynamic_page_size="0"
+else
+  enable_dynamic_page_size="1"
+fi
+],
+[enable_dynamic_page_size="0"]
+)
+if test "x$enable_dynamic_page_size" = "x1" ; then
```
guangli-dai, line 1953: What will happen if I set `--enable-dynamic-page-size` and also try to set `--with-lg-page=xxx`? It seems the lg-page will just be ignored starting from the compile time all the way to runtime, is that correct? Maybe we should just reject setting lg_page up in the configuration when dynamic page size is enabled? WDYT?

include/jemalloc/internal/conf.h
```diff
@@ -3,8 +3,13 @@
 
 #include "jemalloc/internal/sc.h"
 
-void malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
+/* Number of sources for initializing malloc_conf, should be used for opts_cache */
```
guangli-dai: nit: this line exceeds 80 chars.

include/jemalloc/internal/conf.h
```diff
@@ -3,8 +3,13 @@
...
+#define MALLOC_CONF_NSOURCES 5
+
+void malloc_conf_init_pre(const char *opts_cache[MALLOC_CONF_NSOURCES],
     char readlink_buf[PATH_MAX + 1]);
+void malloc_conf_init(sc_data_t *sc_data, unsigned bin_shard_sizes[SC_NBINS],
+    const char *opts_cache[MALLOC_CONF_NSOURCES]);
```
guangli-dai, line 12: Are these changes mandatory here?

include/jemalloc/internal/pages.h
```diff
@@ -39,8 +39,11 @@ extern unsigned lg_page;
 /* page size based on lg_page, only used when dynamic page size is enabled */
 extern size_t page_size;
 
+/* number of pages per hugepage, only used when dynamic page size is enabled */
```
guangli-dai, line 54: nit: same >80 chars line

src/pages.c
```diff
@@ -521,6 +522,12 @@ pages_pre_boot(void) {
 		}
 		return true;
 	}
+
+#	if LG_HUGEPAGE != 0
```
guangli-dai: Good to be defensive here. Would like to check if my understanding is correct, `LG_HUGEPAGE == 0` is a very rare case right? It should be refused if configured through `configure.ac` and would only happen if someone overrides it through compiler overrides?

include/jemalloc/internal/tcache.h
```diff
@@ -37,12 +37,21 @@ typedef struct tcaches_s     tcaches_t;
...
-#define TCACHE_GC_NEIGHBOR_LIMIT ((uintptr_t)1 << 21)       /* 2M */
+#define TCACHE_NBINS_FOR(_nbins, _maxclass_limit, _large_minclass)             \
+	((_nbins) + SC_NGROUP * ((_maxclass_limit) - (_large_minclass)) + 1)
+#define TCACHE_NBINS                                                           \
+	TCACHE_NBINS_FOR(                                                      \
+	    SC_NBINS, TCACHE_LG_MAXCLASS_LIMIT, SC_LG_LARGE_MINCLASS)
+#ifdef DYNAMIC_PAGE_SIZE
+#	define TCACHE_NBINS_MAX                                               \
+		TCACHE_NBINS_FOR(SC_NBINS_MAX, TCACHE_LG_MAXCLASS_LIMIT_MAX,   \
+		    SC_LG_LARGE_MINCLASS_MIN)
```
guangli-dai, line 50: IIUC, `TCACHE_LG_MAXCLASS_LIMIT_MAX` and `SC_LG_LARGE_MINCLASS_MIN` cannot be reached at the same time based on current calculation because before this commit, both can be reduced to the for k*PAGE and the difference between them is always the same. However, it's critical to consider what will happen in the future and what is used here is a safer version but with a small gap that will never be used in the current implementation.

An optimal way is to use `SC_LG_LARGE_MINCLASS_MAX` instead and leave a comment on `TCACHE_LG_MAXCLASS_LIMIT_MAX`, i.e., `USIZE_GROW_SLOW_THRESHOLD`, to make sure when that is changed, this calculation needs to be changed accordingly as well.

guangli-dai (commented):
First batch of comments, will finish reviewing the last two commits soon. One question I don't have a good answer yet: considering how much changes it is and the criticality in compatibility, I think we need to find a way to test this thoroughly in CI. One way is to set a test-only MALLOC_CONF override so we can do something like `TEST_MALLOC_CONF:lg_page:12/13/14/15/16 make check` and that lg_page will override the read from the system just for testing purposes. But it sounds too hacky. Any  thoughts?

gctony:
> First batch of comments, will finish reviewing the last two commits soon. One question I don't have a good answer yet: considering how much changes it is and the criticality in compatibility, I think we need to find a way to test this thoroughly in CI. One way is to set a test-only MALLOC_CONF override so we can do something like `TEST_MALLOC_CONF:lg_page:12/13/14/15/16 make check` and that lg_page will override the read from the system just for testing purposes. But it sounds too hacky. Any thoughts?

I don't have the ci changes yet. I'm planning a follow-up PR to expand the ci test matrix to also test the dynamic page size option thoroughly. I also do have a follow-up change that introduces the `lg_page` conf option (this is how I have been testing locally, FWIW). So, yes, we'll be able to set, say, `MALLOC_CONF=lg_page:14` and test with 16K pages. And we can try all page options (min -> max).

configure.ac
gctony: Good catch re: using `-le`. I made the change, I also changed the `hpa_supported()` check back to `HUGEPAGE_PAGES == 1`. However, note that it will now be a dynamic check. Not sure whether that will matter.

configure.ac
gctony: Yes, lg-page will be ignored. I can add an extra check to disallow it.

configure.ac
gctony: Yeah, in fact `HUGEPAGE_PAGES` is used before `hugepage_pages` is initialized. What's the objection to checking `LG_HUGEPAGE == 0` instead? It's a constant and, if it's 0, HUGEPAGE_PAGES will be 1. I think the two checks are equivalent?

include/jemalloc/internal/conf.h
gctony: fixed

include/jemalloc/internal/conf.h
gctony: I need to do two passes to initialize malloc conf, the first to initialize `opt_abort` (and in the future `lg_page`) so that we know how to abort if we have issues determining the page size. And the page size needs to be determined before the rest of the conf options are initialized (as some of the defaults depend on it). So, I had to expose the two separate functions, which are called from `jemalloc_init.c`.

include/jemalloc/internal/pages.h
gctony: I think this one is actually 79 chars?

include/jemalloc/internal/tcache.h
gctony: Yeah, this will slightly oversize `TCACHE_NBINS_MAX`. I do have a change that addresses that. Can I do that on a follow-up PR?

src/pages.c
gctony: Actually, I looked into it a bit more closely. I think that `LG_HUGEPAGE == 0` cannot happen (even though there have been explicit checks for that). If autodetect fails, it will default to `21`. And an explicit `--with-lg-hugepage=0` is rejected. Shall I clean that up on a separate PR (I think it's a self-contained change)?

configure.ac
gctony: See comment below, I don't think the `HUGEPAGE_PAGES == 1` is actually needed.

guangli-dai:
> I don't have the ci changes yet. I'm planning a follow-up PR to expand the ci test matrix to also test the dynamic page size option thoroughly. I also do have a follow-up change that introduces the `lg_page` conf option (this is how I have been testing locally, FWIW). So, yes, we'll be able to set, say, `MALLOC_CONF=lg_page:14` and test with 16K pages. And we can try all page options (min -> max).

I don't think simply setting MALLOC_CONF works because tests override it (some tests have .sh, which will surely not override that setup). On one hand, it looks critical for us to have CI coverage of different page size cases. On the other way, adding a new layer of malloc_conf just for experimental results, specifically just for this feature also sounds too overwhelming. Don't have a better idea here.

src/pages.c
gctony: How about this cleanup? https://github.com/jemalloc/jemalloc/pull/2998

gctony:
> > I don't have the ci changes yet. I'm planning a follow-up PR to expand the ci test matrix to also test the dynamic page size option thoroughly. I also do have a follow-up change that introduces the `lg_page` conf option (this is how I have been testing locally, FWIW). So, yes, we'll be able to set, say, `MALLOC_CONF=lg_page:14` and test with 16K pages. And we can try all page options (min -> max).
> 
> I don't think simply setting MALLOC_CONF works because tests override it (some tests have .sh, which will surely not override that setup). On one hand, it looks critical for us to have CI coverage of different page size cases. On the other way, adding a new layer of malloc_conf just for experimental results, specifically just for this feature also sounds too overwhelming. Don't have a better idea here.

Oh, I see what you mean. You're recommending an additional mechanism to set conf options that will be picked up by all tests. Sure. But, I'll do that as a follow-up when I work in enabling ci tests.

src/hpa.c
gctony: PR https://github.com/jemalloc/jemalloc/pull/2998 eliminated this check.

src/pages.c
gctony: This is a non-issue now after PR https://github.com/jemalloc/jemalloc/pull/2998 was merged.

configure.ac
gctony: added

include/jemalloc/internal/tcache.h
gctony: FWIW, I think the difference is 384 bytes per thread.

gctony:
> > > I don't have the ci changes yet. I'm planning a follow-up PR to expand the ci test matrix to also test the dynamic page size option thoroughly. I also do have a follow-up change that introduces the `lg_page` conf option (this is how I have been testing locally, FWIW). So, yes, we'll be able to set, say, `MALLOC_CONF=lg_page:14` and test with 16K pages. And we can try all page options (min -> max).
> > 
> > 
> > I don't think simply setting MALLOC_CONF works because tests override it (some tests have .sh, which will surely not override that setup). On one hand, it looks critical for us to have CI coverage of different page size cases. On the other way, adding a new layer of malloc_conf just for experimental results, specifically just for this feature also sounds too overwhelming. Don't have a better idea here.
> 
> Oh, I see what you mean. You're recommending an additional mechanism to set conf options that will be picked up by all tests. Sure. But, I'll do that as a follow-up when I work in enabling ci tests.

@guangli-dai  Can we use `--with-malloc-conf=lg_page:14` to avoid having to add an additional mechanism? Also, it looks as if `test.sh` concatenates external `MALLOC_CONF` to what the test sets?

```
# Make a copy of the MALLOC_CONF passed in to this script, so
# it can be repeatedly concatenated with per test settings.
export MALLOC_CONF_ALL=${MALLOC_CONF}
# Concatenate the individual test's MALLOC_CONF and MALLOC_CONF_ALL.
export_malloc_conf() {
  if [ "x${MALLOC_CONF}" != "x" -a "x${MALLOC_CONF_ALL}" != "x" ] ; then
    export MALLOC_CONF="${MALLOC_CONF},${MALLOC_CONF_ALL}"
  else
    export MALLOC_CONF="${MALLOC_CONF}${MALLOC_CONF_ALL}"
  fi
}
```

guangli-dai:
> @guangli-dai Can we use `--with-malloc-conf=lg_page:14` to avoid having to add an additional mechanism? Also, it looks as if `test.sh` concatenates external `MALLOC_CONF` to what the test sets?
> 
> ```
> # Make a copy of the MALLOC_CONF passed in to this script, so
> # it can be repeatedly concatenated with per test settings.
> export MALLOC_CONF_ALL=${MALLOC_CONF}
> # Concatenate the individual test's MALLOC_CONF and MALLOC_CONF_ALL.
> export_malloc_conf() {
>   if [ "x${MALLOC_CONF}" != "x" -a "x${MALLOC_CONF_ALL}" != "x" ] ; then
>     export MALLOC_CONF="${MALLOC_CONF},${MALLOC_CONF_ALL}"
>   else
>     export MALLOC_CONF="${MALLOC_CONF}${MALLOC_CONF_ALL}"
>   fi
> }
> ```

`--with-malloc-conf` works, that would be adding a separate set of flags, say we add 16KB and 64KB page sizes in, they will combine with other options and create a bunch of new tests if we add them in [here](https://github.com/jemalloc/jemalloc/blob/dev/scripts/gen_gh_actions.py#L102). If external `MALLOC_CONF` works, that would be even better, that means we normally extend each existing CI tests with different page size used, it is a more thorough test.

include/jemalloc/internal/tcache.h
guangli-dai: Sure another PR works, current overhead is acceptable. Maybe even after the CI added.

include/jemalloc/internal/pages.h
guangli-dai: Meant to say the line above (line 51's comment). Auto-formatting cannot handle comments well? But LLM should be able to check and fix all these places well enough.

include/jemalloc/internal/edata.h
```diff
@@ -203,7 +203,7 @@ struct edata_s {
 #define EDATA_BITS_SZIND_MASK                                                  \
 	MASK(EDATA_BITS_SZIND_WIDTH, EDATA_BITS_SZIND_SHIFT)
 
-#define EDATA_BITS_NFREE_WIDTH (SC_LG_SLAB_MAXREGS + 1)
+#define EDATA_BITS_NFREE_WIDTH (SC_LG_SLAB_MAXREGS_MAX + 1)
```
guangli-dai, line 206: edata is one of the most complicated and well-compressed data structure in jemalloc, so these changes of macros are not as straightforward as other changes. Can you add a comment somewhere concluding all the changes in edata due to the dynamic page sizes?

I am thinking about something like this (just a draft, correct and rephrase):
Dynamic page sizes adds the runtime/max split
for `SC_LG_SLAB_MAXREGS` in `sc.h`, uses the maximum in
`LG_BITMAP_MAXBITS` in `bitmap.h`, and widens `EDATA_BITS_NFREE_WIDTH` in
`edata.h`.  The `nfree` change only expands its field from 10 to 14 bits inside
the existing `uint64_t e_bits`, so it adds no bytes to `edata_t`; the size
increase comes from the bitmap capacity, which enables `BITMAP_USE_TREE` and
changes `bitmap_info_t` from 16 to 64 bytes and `bin_info_t` from 40 to 88
bytes, while the inline `slab_data_t.bitmap[]` grows from 64 to 1048 bytes and
consequently makes `edata_t` grow from 128 to 1112 bytes.  At this commit,
`base_alloc_edata()` rounds that to a 1152-byte charge, versus 128 bytes in a
fixed 4 KiB build.  With the runtime-sized allocation in `6fe1036b`, a 4 KiB
process instead needs a 136-byte descriptor (64-byte prefix plus a 72-byte,
nine-word runtime tree bitmap), which 128-byte edata alignment rounds to 256
bytes: this explains the observed 2x `stats.metadata_edata`, although compiled
`sizeof(edata_t)` remains 1112.

include/jemalloc/internal/tsd_win.h
```diff
@@ -106,7 +106,8 @@ tsd_boot0(void) {
 #	if JEMALLOC_WIN32_TLSGETVALUE2
 	tgv2_mod = LoadLibraryA("api-ms-win-core-processthreads-l1-1-8.dll");
 	if (tgv2_mod != NULL) {
-		tls_get_value2 = (TGV2)GetProcAddress(tgv2_mod, "TlsGetValue2");
+		tls_get_value2 = (TGV2)(void (*)(void))GetProcAddress(
```
guangli-dai: Why is this relevant?

gctony:
> > @guangli-dai Can we use `--with-malloc-conf=lg_page:14` to avoid having to add an additional mechanism? Also, it looks as if `test.sh` concatenates external `MALLOC_CONF` to what the test sets?
> > ```
> > # Make a copy of the MALLOC_CONF passed in to this script, so
> > # it can be repeatedly concatenated with per test settings.
> > export MALLOC_CONF_ALL=${MALLOC_CONF}
> > # Concatenate the individual test's MALLOC_CONF and MALLOC_CONF_ALL.
> > export_malloc_conf() {
> >   if [ "x${MALLOC_CONF}" != "x" -a "x${MALLOC_CONF_ALL}" != "x" ] ; then
> >     export MALLOC_CONF="${MALLOC_CONF},${MALLOC_CONF_ALL}"
> >   else
> >     export MALLOC_CONF="${MALLOC_CONF}${MALLOC_CONF_ALL}"
> >   fi
> > }
> > ```
> 
> `--with-malloc-conf` works, that would be adding a separate set of flags, say we add 16KB and 64KB page sizes in, they will combine with other options and create a bunch of new tests if we add them in [here](https://github.com/jemalloc/jemalloc/blob/dev/scripts/gen_gh_actions.py#L102). If external `MALLOC_CONF` works, that would be even better, that means we normally extend each existing CI tests with different page size used, it is a more thorough test.

So, external `MALLOC_CONF` seems to work if I use `MALLOC_CONF=... test/test.sh test/unit/...` but not if I use `make -j check` (it fails with some weird parsing error, I can look into it). But, we do seem to have a couple of options we can look at when we work on this change.

src/pages.c
```diff
@@ -454,16 +459,87 @@ init_thp_state(void) {
 }
 
 bool
-pages_boot(void) {
+pages_pre_boot(void) {
 	os_page = os_vm_page_size();
```
guangli-dai, line 464: Question: this function falls back to `PAGE`. When dynamic page size is enabled, `PAGE` may go uninitialized? Can you double check: is the fallback in different OS still valid under dynamic page size. Specifically, a valid fallback here means under dynamic page size mode, it should halt the initialization or abort appropriately.

include/jemalloc/internal/pages.h
gctony: So, no, auto-formatting hasn't helped with long comment lines. But maybe I have the wrong VS Code settings. I'll use an LLM to try to find any long lines.

include/jemalloc/internal/tsd_win.h
gctony: I was getting compilation failures on windows (on the github ci) and this seems to have fixed them. However, I'm also realizing that the last PR I just merged doesn't have this fix and it didn't have any ci failures on windows either. I'll remove these changes and let's see what happens...

src/pages.c
gctony: Interesting question that led into an interesting investigation!

First, the debug build has an assert that catches any uses of `PAGE` before it is initialized in `pages_pre_boot()`. This is helpful to avoid any incorrect ordering (and it caught several errors while I was working on this change).

And yes `PAGE` will not be initialized if `os_vm_page_size()` returns 0 or an invalid value. If this happens, `pages_pre_boot()` will either abort (if `opt_abort == true`) or return `true`. In the latter case, it will be handled like any other init function called from `malloc_init_hard_a0_locked()` that returns true.

So, I thought we were good and this failure will be handled like all others during initialization. However, there's a fly-in-the-ointment! Here's what the code in the Posix `os_vm_page_size()` looks like (for non FreeBSD):

```
	long result = sysconf(_SC_PAGESIZE);
	if (result == -1) {
		return PAGE;  <<<< oops!
	}
```

In a non-debug build, this will actually work as expected: `PAGE` will evaluate to `0` (as `page_size` is not initialized) and `pages_pre_boot()` will reject that value and exit with the standard error path.

In a debug build, however, `return PAGE;` should fail with an assertion failure that `page_size` is not initialized.

Maybe `os_vm_page_size()` should not return `PAGE`, return 0, and let the caller handle the error? And what should `pages_pre_boot()` do in this case?

gctony: Thinking about it a bit more: If `os_vm_page_size()` returns 0, in the non-dps case we set `os_page` to `PAGE` and in the dps case we just fail (but use a user-provided `lg_page` when that's introduced?).

gctony: I made that change, let me know if you'd like an alternative approach.

include/jemalloc/internal/pages.h
gctony: I think I fixed all the long comments.

include/jemalloc/internal/edata.h
gctony: How about this?

```
/*
 * The dynamic page size feature sets SC_LG_SLAB_MAXREGS_MAX and
 * LG_BITMAP_MAXBITS to the largest possible values for the given min/max page
 * size. Further, the bitmap size also depends on whether the bitmap uses a tree
 * or not, which is a compilation-time decision (see BITMAP_USE_TREE), as the
 * tree requires extra bits in the bitmap. So, EDATA_BITS_NFREE_WIDTH is set to
 * the largest possible value. However, when we allocate edata, we can
 * dynamically size it based on the current page size. For min/max page size of
 * 4K/64K, the bitmap tree will be used (due to the 64K max page), so the bitmap
 * for the 4K page case will be 72 bytes, instead of 64 bytes (for when the
 * bitmap tree is not used). This makes the edata size 136 bytes, which will be
 * aligned to 256 bytes (instead of 128 bytes for when the bitmap tree is not
 * used). We can improve on this further by making the decision on whether the
 * bitmap tree is used dynamically.
 */
```

include/jemalloc/internal/tsd_win.h
gctony: Warning seems to have disappeared, not sure why I was getting it before. Removed this change.

gctony:
> > > @guangli-dai Can we use `--with-malloc-conf=lg_page:14` to avoid having to add an additional mechanism? Also, it looks as if `test.sh` concatenates external `MALLOC_CONF` to what the test sets?
> > > ```
> > > # Make a copy of the MALLOC_CONF passed in to this script, so
> > > # it can be repeatedly concatenated with per test settings.
> > > export MALLOC_CONF_ALL=${MALLOC_CONF}
> > > # Concatenate the individual test's MALLOC_CONF and MALLOC_CONF_ALL.
> > > export_malloc_conf() {
> > >   if [ "x${MALLOC_CONF}" != "x" -a "x${MALLOC_CONF_ALL}" != "x" ] ; then
> > >     export MALLOC_CONF="${MALLOC_CONF},${MALLOC_CONF_ALL}"
> > >   else
> > >     export MALLOC_CONF="${MALLOC_CONF}${MALLOC_CONF_ALL}"
> > >   fi
> > > }
> > > ```
> > 
> > 
> > `--with-malloc-conf` works, that would be adding a separate set of flags, say we add 16KB and 64KB page sizes in, they will combine with other options and create a bunch of new tests if we add them in [here](https://github.com/jemalloc/jemalloc/blob/dev/scripts/gen_gh_actions.py#L102). If external `MALLOC_CONF` works, that would be even better, that means we normally extend each existing CI tests with different page size used, it is a more thorough test.
> 
> So, external `MALLOC_CONF` seems to work if I use `MALLOC_CONF=... test/test.sh test/unit/...` but not if I use `make -j check` (it fails with some weird parsing error, I can look into it). But, we do seem to have a couple of options we can look at when we work on this change.

Created PR https://github.com/jemalloc/jemalloc/pull/2999 that introduces `MALLOC_CONF_TEST` that works with both `make check` and `bash test/test.sh`.

src/pages.c
guangli-dai: I think this resolves the concern well.

include/jemalloc/internal/edata.h
guangli-dai: Clear but one nit, I think the "we can dynamically size it..." part is not true until this commit (https://github.com/jemalloc/jemalloc/pull/3004/changes/940845846c890b8483e93be4b0e3557ea486cb57#diff-f1e3a699381f48a6076d43a39b1e8a8f2d5bc2e6d7838bd0ca04ee1d650792da) lands and it is not in this PR. So we should avoid saying so here?

guangli-dai (approved):
One nit on the comment in edata.h and after conflicts solved, we are good to merge.

include/jemalloc/internal/edata.h
gctony: You're totally right. I was being a bit lazy! I'll update the comment and make sure I update it back when I add the dynamic edata sizing.

PR #2985, opened 2026-08-25
Add Illumos support

notpeter:
An initial attempt to add (restore?) illumos support to jemalloc.
It has not been extensively tested, but I validated it compiles and passes the test suite on:
```
OmniOS r151046 with GCC 11.3.0, Clang 17
OmniOS r151054 with GCC 14.2.0, Clang 20
Oxide Helios 3 with GCC 14.2.0, Clang 15 and Clang 20.
```

Includes no allocator implementation changes.
Only configure fixes, test changes and better CI coverage.

## Changes

- Gate `-herror_on_warning` on `je_cv_cray`
Introduced in https://github.com/jemalloc/jemalloc/commit/3dee73faf2c9c1008a4e3281cd7dd0123d16a8d3 as the Cray compiler’s `-Werror` equivalent.
Probing this Cray-specific flag with GCC and Clang interferes with configure on illumos. 
https://github.com/jemalloc/jemalloc/pull/423

- Improve cleanup in the HPA unit test to invalidate cached rtree leaf pointers after destroying a private emap.
Previously, a later test could follow stale thread-local rtree leaf pointers after the destroyed emap’s address range was reused, causing a crash on illumos.
Follows the PAC teardown pattern. See:
https://github.com/jemalloc/jemalloc/pull/2927

- Only configure tests with `percpu_arena:percpu` when `JEMALLOC_PERCPU_ARENA` is enabled.  
illumos supports background threads but lacks `sched_getcpu` required for per-CPU arenas.
Move the configuration into the compile-time-gated C test and remove the now-unneeded shell script.

- Remove a nondeterministic assertion that assumed allocations would be adjacent.
This assumption is false on Illumos. Happy to restore and `#ifndef __sun` if preferred.
Retain deterministic checks for destroyed bytes and cleared emap entries.

- Fix quoting on `T_APPEND_V` to support multiple flags (illumos uses `-lposix4 -lsocket -lnsl`)

- Add Illumos to CI test matrix. Tests LLVM and GCC on OmniOS using [vmactions/omnios-vm](https://github.com/vmactions/omnios-vm).
Running on OmniOS r151054 (Current LTS releases)


scripts/gen_gh_actions.py
```diff
@@ -711,6 +711,74 @@ def generate_freebsd_job(arch):
...
+            cc: /opt/ooce/bin/clang-20
+            cxx: /opt/ooce/bin/clang++-20
+            packages: ooce/developer/clang-20
+          - omnios: r151046
+            release: r151046
+            compiler: GCC
+            cc: gcc
+            cxx: g++
+            packages: build-essential
+          - omnios: r151046
```
guangli-dai: This version seems to reach its end of life already? I am referring to https://omnios.org/schedule.

test/unit/extent_alloc_flags.c
```diff
@@ -473,10 +473,6 @@ TEST_BEGIN(test_pinned_destroy) {
 	expect_zu_ge(pinned_destroy_bytes - destroy_bytes_before, total,
 	    "destroy hook should be called with at least the allocated total "
 	    "(coalesced fragments returned to OS)");
-	if (maps_coalesce) {
```
guangli-dai, line 476: This is a real assertion touched in Unix systems. So removing it reduces the coverage. However, at a second look, it seems this assertion is indeed not deterministic since coalesce cannot be guaranteed to happen. I would recommend separate this change into another commit (before the commit adding illumos CI) for better future record. It is removed not just because it fails on illumos but because it is not deterministic.

guangli-dai (commented):
Thank you for adding CI for OmniOS, we appreciate more OS's CI coverage here! Can you rebase and address the comments?

scripts/gen_gh_actions.py
notpeter: You are correct. During dev I intentionally wanted to test against an older OmniOS to make sure things also compiled correctly there, but didn't find anything notable. I don't think it's worth testing there -- the currently supported OmniOS LTS release (`r151054`) is plenty going forward.

I've removed the `r151046` from the matrix leaving `r151054` for gcc and llvm.

notpeter:
I've rebased current dev, switched CI so it only runs on OmniOS LTS `r15054` and split the non-determistic test removal into a separate commit.

Thanks for taking the time to review!

guangli-dai (approved):
The new CI is working fine. Thanks @notpeter for the patch! 

One follow-up though: the `nm` on OmniOS seems to reject option `-a` used by jemalloc which may lead to private namespace not working as expected. Public APIs should be just fine, which is why the CI is still green with the `nm` issue. It can be fixed in a separate PR.

PR #2987, opened 2026-08-27
Silently tolerate invalid experimental_ and deprecated options without stderr

guangli-dai:
Silently tolerate invalid experimental_ and deprecated options without stderr. This is because we don't abort for these cases even `abort_conf:true`. In that case, printing to stderr can easily trigger false alarming, which is not expected.

### DIFF
--- a/test/unit/conf.c
+++ b/test/unit/conf.c
@@ -1,6 +1,33 @@
 #include "test/jemalloc_test.h"
 
 #include "jemalloc/internal/conf.h"
+#include "jemalloc/internal/malloc_io.h"
+
+static unsigned conf_error_message_count;
+
+static void
+conf_error_message(void *opaque, const char *s) {
+	(void)opaque;
+	(void)s;
+	conf_error_message_count++;
+}
+
+static void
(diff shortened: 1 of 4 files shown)


test/unit/conf.c
```diff
@@ -105,9 +132,18 @@ TEST_BEGIN(test_conf_next_trailing_comma) {
 }
 TEST_END
 
+TEST_BEGIN(test_conf_error) {
+	expect_conf_error("experimental_removed", false, false);
+	expect_conf_error("hpa_sec_bytes_after_flush", false, false);
```
gctony: Is it worth moving the deprecated options to a global constant and iterate over them here instead of replicating them?

gctony (approved):
LGTM. See one optional minor suggestion.

spredolac (approved):

test/unit/conf.c
guangli-dai: Yes, that's a good idea. Also #2965 introduces `test_conf_error_deprecated` already, so I just reuse this test.

gctony: Yes, I like the new version. Thanks!

PR #2988, opened 2026-08-30
[RFC] Fragmentation Profiler

azat:
At ClickHouse we have few issues with jemalloc fragmentation and the usual approach was to "guess" based on the live allocations which are old enough, but it is tricky.

This is an RFC for fragmentation profiling, it will be nice if someone can take a look and tell is it the right approach for this problem, and then I will polish this patch set

Track live sampled allocations on a per-`gctx` list under the `gctx` lock (this will increase contention, I can make fragmentation profiler under option)

Dumps emit `f:` and `frag_util:` lines

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -283,6 +283,7 @@ TESTS_UNIT := \
 	$(srcroot)test/unit/prof_hook.c \
 	$(srcroot)test/unit/prof_idump.c \
 	$(srcroot)test/unit/prof_log.c \
+	$(srcroot)test/unit/prof_frag.c \
 	$(srcroot)test/unit/prof_mdump.c \
 	$(srcroot)test/unit/prof_recent.c \
 	$(srcroot)test/unit/prof_reset.c \
--- a/bin/jefrag.py
+++ b/bin/jefrag.py
@@ -1,0 +1,344 @@
+#!/usr/bin/env python3
+"""Fragmentation report from a jemalloc heap dump.
+
+Consumes a heap profile produced by prof.dump (heap_v2 format) that contains
+the fragmentation records emitted by jemalloc:
+
+  f: <age_ns> <request_size> <usize> <szind> <arena_ind> <thr_uid>
+      -- one per live sampled allocation, listed under its "@ <pc>..." stack
+  frag_util: <arena_ind> <binind> <reg_size> <slab_size> <nregs> <n_shards>
+      <curslabs> <curregs> <nonfull_slabs>
+      -- one per (arena, small size class)
+
+Ground-truth slab waste per (arena, size class) comes from frag_util; the
+sampled records statistically attribute that waste to the call stacks that
+keep old allocations alive in the class.  Sampling is byte-weighted, so each
+record is unbiased into an estimated object count with the standard Poisson
+factor 1 / (1 - exp(-usize / sample_interval)).
+
+Note: curregs counts regions cached in tcaches as allocated, so waste is a
+lower bound estimate ("waste" regions may partially be tcache-cached).
+
+Outputs:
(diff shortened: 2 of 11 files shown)


spredolac:
Hi @azat, 

I am wondering if different approach would allow you to get results here without being this intrusive.  We would like to avoid getting these locks and lists under gctx as profiling is already expensive.  I would also prefer not to extend the heap profile dump format unless we first prove that existing mechanisms are insufficient.  

One possible layered approach:

**1)  No Allocator Change**
Offline tool: heap profile dump + malloc_stats JSON -> fragmentation attribution report

You create heap profile’s per-stack and you get `curobjs`, `curbytes`. `jemalloc` records `curbytes += usize` for sampled allocations, `curbytes / curobjs` gives a usable-size average for that stack. If that average maps cleanly to one bin, attribute that stack to that bin. Your stacks may allocate similar size objects. Then join with bin stats:
     `waste[bin] = curslabs * slab_size - curregs * reg_size`
  
 And distribute each bin’s waste across stacks mapped to that bin by live object count. This is approximate, but may be good enough for the common case where a stack allocates one object shape.

**2)  Better, Still No Jemalloc Change**
  Existing experimental.prof_recent.alloc_dump is interesting. It already dumps sampled allocations as JSON with:  size, usize, released, alloc_trace, alloc_time, dalloc_trace, if released

See [src/prof_recent.c:494](https://github.com/jemalloc/jemalloc/blob/dev/src/prof_recent.c#L494). With a large enough existing `prof_recent_alloc_max`, you can get a sampled stack-to-size-class distribution directly. That helps with mixed-size stacks without changing heap_v2. Then the tool can use: `P(bin | stack)` from prof_recent, and apply that distribution to the normal heap profile’s live counts. This is more powerful than average-size guessing at 1), still uses existing mechanisms, and avoids a dump format change. Prof recent already gives you usize and you can compute the bin index offline. 

**3) Use experimental hooks**
  Could this be prototyped entirely with the existing `prof_sample` / `prof_sample_free` hooks? They already expose ptr, requested size, usize, and backtrace for sampled allocations. Your library can maintain a live sampled-allocation table, derive szind from usize, optionally read arena via arenas.lookup while the pointer is live, and then join that with existing bin stats from mallctl/malloc_stats JSON.
  
we can discuss further if none of this works, but I hope it does. Cheers, and thanks for contributing to jemalloc!

EDIT: If you design a tool using hooks, or this statistical offline processing, we would consider adding some  `tools/` (cc: @guangli-dai ) directory (prefer that than adding complexity to the core library).  This would fit some longer term plans that we have.

azat:
Hi @spredolac thanks for taking a look!

>1) No Allocator Change

Interesting idea, though I'm worrying about approximated results

>2) Better, Still No Jemalloc Change

About, `experimental.prof_recent`, it seems that it has one global lock for each allocation - https://github.com/jemalloc/jemalloc/blob/f0386eb307c4226e8b28809d12eb3883e3f6633a/src/prof_recent.c#L292

Plus the API is not great, seems that it dumps always all allocations, likely there will be no old allocations and implementing deduplication won't be hard, but still.

>3) Use experimental hooks

I actually considered this option, but decided that it will be useful to have all information in one place, that way others may benefit from it too, the gctx lock issue probably can be fixed or/and we can make fragmentation profiler under it's own option.

But if you think it does not worth to change the profiler format and extend it, this sounds the best option and I can take a look at this one later.

(it may require new hooks to get arena index at least, we can extend old one and it will be ABI compatible for an app linked with the older jemalloc, but not vise versa)

spredolac:
@azat let me know how it goes. For the hooks - if you make a general tool using hooks (if something needs to be extended there let's consider it) we can make it available to others as well.

PR #2989, opened 2026-08-31
fix spurious failures in test/unit/hpa_background_thread

gctony:
I've been seeing the error "Should have seen a background purge" when running `test/unit/hpa_background_thread` under heavy cpu load. The reason is that the background thread starves and it doesn't purge in the required time window. The fix is to wake up the background thread early so that it can make some progress.

Testing under heavy load, in 275,000 iterations the bug was reproduced 28 times without the fix and 0 times with the fix.

### DIFF
--- a/test/unit/hpa_background_thread.c
+++ b/test/unit/hpa_background_thread.c
@@ -164,13 +164,21 @@ expect_deferred_purging(unsigned arena_ind) {
 	expect_true(observed_dirty_page, "");
 
 	/*
-	 * Under high concurrency / heavy test load (e.g. using run_test.sh),
-	 * the background thread may not get scheduled for a longer period of
-	 * time.  Retry 100 times max before bailing out.
+	 * Now wait for the background thread to purge the deferred dirty page.
+	 * An idle background thread sleeps indefinitely until signaled, so a
+	 * missed wakeup on the deallocation path could stall us under heavy test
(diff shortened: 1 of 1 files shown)


test/unit/hpa_background_thread.c
```diff
@@ -164,13 +164,26 @@ expect_deferred_purging(unsigned arena_ind) {
...
+	 * Under high concurrency / heavy test load (e.g. using run_test.sh) the
+	 * background thread may not get scheduled for a longer period of time.
+	 * Worse, once it has nothing to do it enters an indefinite sleep and
+	 * only runs again when signaled, so a single missed or suppressed wakeup
+	 * on the deallocation path would otherwise stall us until the deadline.
+	 * Rather than waiting passively, explicitly nudge the background thread
+	 * awake on every iteration: that drives its purge forward at our polling
+	 * rate instead of relying on its (possibly indefinite) sleep interval.
+	 * A background thread that never purges still fails the assertion below
+	 * once we give up after 100 retries.
```
spredolac: Can you make this concise?

guangli-dai (approved):

test/unit/hpa_background_thread.c
gctony: @spredolac Is the new version better?

spredolac: Yes

spredolac (approved):

PR #2990, opened 2026-08-31
Fix possible tcache corruption on fiber migration (under --enable-experimental-fiber-safe-tls)

### LINKED ISSUE #2890: Crashes in tcache when jemalloc bundled as static library with LTO into result binary

azat:
_Note, this the second, more simpler version of https://github.com/jemalloc/jemalloc/pull/2944/ without per-arch asm_

Under LTO `&tsd_tls` is hoisted out of the inlined malloc/free and survives swapcontext in a callee-saved register: a fiber resumed on another OS thread races on the old thread's tcache.

Fix it by taking the address in a noinline volatile-asm barrier accessor.

It is done under option, because it make the fastpath slower, on a microbench (that is very heavy depends on the most optimal case) the overhead is 35%, but this is unlikely the real world scenario.

Anyway here are some numbers for `microbench.c` compiled with LTO (AMD 5975WX):

type    |vanilla (ns)|`--enable-experimental-fiber-safe-tls` (ns)|optimized (ns)
--------|------------|-------------------------------------------|--------------
malloc  |6.51        |8.77 **(+34%)**                            |6.79 (4%)
mallocx |8.48        |10.62 **(25%)**                            |8.76 (3%)
free    |6.46        |8.68 **(34%)**                             |6.76 (5%)
dallocx |8.190       |10.97 **(33%)**                            |8.74 (7%)

(optimized version is a version that submitted in another PR with per-arch asm to reduce overhead)

Note though that **difference is less on a server CPUs** and even way less on ARM64:
- [AMD EPYC 9R14](https://pastila.nl/?00e097f0/ff3bc749c303ca64bce3860d74ab572e#7U7BcERPaGtfm9luNpexjw==GCM) - **23%**
- [Neoverse-V2 (**arm64**)](https://pastila.nl/?00e01df2/94c98976b67d5fa2eb2c955145407f60#g7hoK8X4LQZk+rocOcro4g==GCM) - **4%**

Fixes: https://github.com/jemalloc/jemalloc/issues/2890
Thanks to @xinydev

### DIFF
--- a/include/jemalloc/internal/tsd_malloc_thread_cleanup.h
+++ b/include/jemalloc/internal/tsd_malloc_thread_cleanup.h
@@ -5,6 +5,7 @@
 
 #include "jemalloc/internal/jemalloc_preamble.h"
 #include "jemalloc/internal/tsd_internals.h"
+#include "jemalloc/internal/tsd_tls_addr.h"
 #include "jemalloc/internal/tsd_types.h"
 
 #define JEMALLOC_TSD_TYPE_ATTR(type) __thread type JEMALLOC_TLS_MODEL
@@ -12,6 +13,9 @@
 extern JEMALLOC_TSD_TYPE_ATTR(tsd_t) tsd_tls;
 extern JEMALLOC_TSD_TYPE_ATTR(bool) tsd_initialized;
 extern bool tsd_booted;
+
+JEMALLOC_TLS_ADDR_DECLARE(tsd_tls)
+JEMALLOC_TLS_ADDR_DECLARE(tsd_initialized)
 
 /* Initialization/cleanup. */
 JEMALLOC_ALWAYS_INLINE bool
@@ -53,13 +57,15 @@ tsd_get_allocates(void) {
 /* Get/set. */
 JEMALLOC_ALWAYS_INLINE tsd_t *
 tsd_get(bool init) {
-	return &tsd_tls;
+	return JEMALLOC_TLS_ADDR(tsd_tls);
 }
 JEMALLOC_ALWAYS_INLINE void
 tsd_set(tsd_t *val) {
+	tsd_t *tsd = JEMALLOC_TLS_ADDR(tsd_tls);
+
 	assert(tsd_booted);
-	if (likely(&tsd_tls != val)) {
-		tsd_tls = (*val);
+	if (likely(tsd != val)) {
+		*tsd = (*val);
 	}
-	tsd_initialized = true;
+	*JEMALLOC_TLS_ADDR(tsd_initialized) = true;
 }
--- a/src/tsd.c
+++ b/src/tsd.c
@@ -23,9 +23,12 @@ JEMALLOC_DIAGNOSTIC_IGNORE_MISSING_STRUCT_FIELD_INITIALIZERS
 #ifdef JEMALLOC_MALLOC_THREAD_CLEANUP
 JEMALLOC_TSD_TYPE_ATTR(tsd_t) tsd_tls = TSD_INITIALIZER;
 JEMALLOC_TSD_TYPE_ATTR(bool) JEMALLOC_TLS_MODEL tsd_initialized = false;
+JEMALLOC_TLS_ADDR_DEFINE(tsd_tls)
+JEMALLOC_TLS_ADDR_DEFINE(tsd_initialized)
 bool tsd_booted = false;
 #elif (defined(JEMALLOC_TLS))
 JEMALLOC_TSD_TYPE_ATTR(tsd_t) tsd_tls = TSD_INITIALIZER;
+JEMALLOC_TLS_ADDR_DEFINE(tsd_tls)
 pthread_key_t tsd_tsd;
 bool          tsd_booted = false;
 #elif (defined(_WIN32))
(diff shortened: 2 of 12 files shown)


src/tsd.c
```diff
@@ -1,3 +1,6 @@
+/* JEMALLOC_TLS_ADDR definitions here. */
+#define JEMALLOC_TSD_C_
```
spredolac: Let's get rid of this `JEMALLOC_TSD_C_` translation-unit marker.

`JEMALLOC_TLS_ADDR_DEFINE() `is already gated by `JEMALLOC_EXPERIMENTAL_FIBER_SAFE_TLS`, so the definitions could be placed directly beside their corresponding TLS variables in this file.

This keeps declarations in the headers, definitions in the owning .c file, and makes the linkage easier to understand without changing behavior. For example:

```
JEMALLOC_TSD_TYPE_ATTR(tsd_t) tsd_tls = TSD_INITIALIZER;
JEMALLOC_TLS_ADDR_DEFINE(tsd_tls)

```  
The platform branches in src/tsd.c can similarly define accessors for tsd_initialized and tsd_wrapper_tls. I tested this arrangement with Clang ThinLTO, and the fiber-migration test still passes.

src/tsd.c
spredolac, line 25: You do it here
```
JEMALLOC_TLS_ADDR_DEFINE(tsd_tls)
JEMALLOC_TLS_ADDR_DEFINE(tsd_initialized)
```

spredolac, line 30: and here
`JEMALLOC_TLS_ADDR_DEFINE(tsd_tls)`

spredolac, line 40: JEMALLOC_TLS_ADDR_DEFINE(tsd_wrapper_tls)

include/jemalloc/internal/tsd_malloc_thread_cleanup.h
```diff
@@ -13,6 +14,13 @@ extern JEMALLOC_TSD_TYPE_ATTR(tsd_t) tsd_tls;
 extern JEMALLOC_TSD_TYPE_ATTR(bool) tsd_initialized;
 extern bool tsd_booted;
 
+JEMALLOC_TLS_ADDR_DECLARE(tsd_tls)
+JEMALLOC_TLS_ADDR_DECLARE(tsd_initialized)
+#ifdef JEMALLOC_TSD_C_
+JEMALLOC_TLS_ADDR_DEFINE(tsd_tls)
+JEMALLOC_TLS_ADDR_DEFINE(tsd_initialized)
+#endif
```
spredolac: and this goes away and similar in other headers.

test/integration/tcache_fiber_migration.c
```diff
@@ -0,0 +1,136 @@
...
+	}
+	for (i = 0; i < num_workers; i++) {
+		pthread_join(threads[i], NULL);
+	}
+	for (i = 0; i < num_fibers; i++) {
+		free(stacks[i]);
+	}
+
+	/* Completing without a tcache-corruption crash is success. */
+	return live_fibers == 0 ? 0 : 1;
```
spredolac: In [test/test.sh.in lines 41–44](https://github.com/jemalloc/jemalloc/blob/2acb25e80b87ca0d5c95ba49dc781f8b29dfec1e/test/test.sh.in#L41-L44):
```

  pass_code=0
  skip_code=1
  fail_code=2
```

  [Lines 73–81](https://github.com/jemalloc/jemalloc/blob/2acb25e80b87ca0d5c95ba49dc781f8b29dfec1e/test/test.sh.in#L73-L81) classify the result, and [100–103](https://github.com/jemalloc/jemalloc/blob/2acb25e80b87ca0d5c95ba49dc781f8b29dfec1e/test/test.sh.in#L100-L103) return success if no test was classified as failed.

This returns 1 when `live_fibers != 0`, but jemalloc’s test harness defines exit code 1 as skip_code and 2 as fail_code. If this is a failure path, it should be reported as failed, because if it skipped and CI would remain green. I think this should return 2 instead?

test/integration/tcache_fiber_migration.c
```diff
@@ -0,0 +1,136 @@
...
+	(void)arg;
+	for (;;) {
+		int id = pop_ready_fiber();
+		if (id == -2) {
+			break;
+		}
+		if (id < 0) {
+			continue;
+		}
+		return_context[id] = &scheduler_context;
```
spredolac, line 106: If this could enqueue fiber on the same worker then we are not really testing what we want, right? You may need to route to different worker to make sure you migrated.

spredolac (commented):
Looks better to me than the last one.  Have few suggestions and some things about the test.

Could you please squash these three commits? The implementation, regression test, and dedicated CI coverage together form one logical change, and it makes it easier later to follow and revert if needed.

test/integration/tcache_fiber_migration.c
azat: Thanks, indeed. Fixed. Though this example will crash w/o the fix

test/integration/tcache_fiber_migration.c
azat: Yep, but usually the scheduling was more or less fair to lead to troubles, but make sense to make deterministic - done.

azat:
>Could you please squash these three commits? 

Sure, done

spredolac (approved):
lgtm.  If CI is all green we'll merge it. cc: @guangli-dai

PR #2992, opened 2026-09-01
Fix edata_before_get address arithmetic

yuxuanchen1997:
edata_before_get computes the page immediately preceding an extent. When an extent starts at PAGE, that address is legitimately NULL and the caller uses the NULL result to avoid an rtree lookup.

Subtracting PAGE from a byte pointer is undefined unless the result stays within the pointed-to array. Clang can therefore assume the result is non-NULL and optimize away the caller's check. Subtract the integer addresses instead, then convert the result back to a pointer, so the PAGE case reliably produces NULL.


guangli-dai:
@yuxuanchen1997 Can you check the static analysis error there?

PR #2996, opened 2026-09-09
Percpu arena map fix

spredolac:
Fix out-of-range arena selection under restricted CPU affinity

## Reproducing

  Four CPUs out of 176, four arenas, four threads allocating ~32 MB total:

```
  $ taskset -c 150-153 env MALLOC_CONF=percpu_arena:percpu,narenas:4 ./repro
    thread.arena = 151
    thread.arena = 152
    thread.arena = 152
    thread.arena = 153
  arenas.narenas = 5, stats.allocated = 102400
```

  Threads bind to arenas 151 through 153 when only five arenas exist.

  ## The allocations become invisible

Every stats and ctl path iterates to `narenas_total_get()`, so arenas above `manual_arena_base` are never visited. The run above holds ~32 MB live and `stats.allocated` reports 102 KB.(same blind spot applies to `malloc_stats_print()`, decay, and `arenas.*` introspection).

  After the fix, same command:

```
    thread.arena = 0
    thread.arena = 1
    thread.arena = 2
    thread.arena = 3
  arenas.narenas = 5, stats.allocated = 32894976

```
It can also result in collision with the huge arena  with `narenas:1` on CPU 1 the unclamped id is 1, which is `huge_arena_ind`, so  ordinary allocations are routed into the dedicated oversize arena:

 ```
 $ taskset -c 1 env MALLOC_CONF=percpu_arena:percpu,narenas:1 ./repro
  before: thread.arena = 1   (arenas.narenas = 2, so index 1 is the huge arena)
  after:  thread.arena = 0
```

Fix is to create a map of CPU-to-arena selection now goes through a table built once at boot.

### DIFF
--- a/include/jemalloc/internal/percpu_arena.h
+++ b/include/jemalloc/internal/percpu_arena.h
@@ -1,0 +1,89 @@
+#ifndef JEMALLOC_INTERNAL_PERCPU_ARENA_H
+#define JEMALLOC_INTERNAL_PERCPU_ARENA_H
+
+#include "jemalloc/internal/jemalloc_preamble.h"
+#include "jemalloc/internal/assert.h"
+#include "jemalloc/internal/jemalloc_internal_types.h"
+#include "jemalloc/internal/os/cpu.h"
+
+/******************************************************************************/
+/* TYPES */
+/******************************************************************************/
+
+typedef enum {
+	percpu_arena_mode_names_base = 0, /* Used for options processing. */
+
+	/*
+	 * *_uninit are used only during bootstrapping, and must correspond
+	 * to initialized variant plus percpu_arena_mode_enabled_base.
+	 */
+	percpu_arena_uninit = 0,
+	per_phycpu_arena_uninit = 1,
+
+	/* All non-disabled modes must come after percpu_arena_disabled. */
+	percpu_arena_disabled = 2,
+
+	percpu_arena_mode_names_limit = 3, /* Used for options processing. */
+	percpu_arena_mode_enabled_base = 3,
+
+	percpu_arena = 3,
+	per_phycpu_arena = 4 /* Hyper threads share arena. */
+} percpu_arena_mode_t;
+
+#define PERCPU_ARENA_ENABLED(m) ((m) >= percpu_arena_mode_enabled_base)
+#define PERCPU_ARENA_DEFAULT percpu_arena_disabled
+
+/******************************************************************************/
+/* EXTERNS */
+/******************************************************************************/
+
+extern percpu_arena_mode_t opt_percpu_arena;
+extern const char *const   percpu_arena_mode_names[];
+
+/******************************************************************************/
+/* INLINES */
+/******************************************************************************/
+
+JEMALLOC_ALWAYS_INLINE malloc_cpuid_t
+malloc_getcpu(void) {
+	assert(have_percpu_arena);
+	return (malloc_cpuid_t)os_cpu_current();
+}
+
+/* Return the chosen arena index based on current cpu. */
+JEMALLOC_ALWAYS_INLINE unsigned
+percpu_arena_choose(void) {
+	assert(have_percpu_arena && PERCPU_ARENA_ENABLED(opt_percpu_arena));
+
+	malloc_cpuid_t cpuid = malloc_getcpu();
+	assert(cpuid >= 0);
+
+	unsigned arena_ind;
+	if ((opt_percpu_arena == percpu_arena)
+	    || ((unsigned)cpuid < ncpus / 2)) {
+		arena_ind = cpuid;
+	} else {
+		assert(opt_percpu_arena == per_phycpu_arena);
+		/* Hyper threads on the same physical CPU share arena. */
+		arena_ind = cpuid - ncpus / 2;
+	}
+
+	return arena_ind;
+}
+
+/* Return the limit of percpu auto arena range, i.e. arenas[0...ind_limit). */
+JEMALLOC_ALWAYS_INLINE unsigned
+percpu_arena_ind_limit(percpu_arena_mode_t mode) {
+	assert(have_percpu_arena && PERCPU_ARENA_ENABLED(mode));
+	if (mode == per_phycpu_arena && ncpus > 1) {
+		if (ncpus % 2) {
+			/* This likely means a misconfig. */
+			return ncpus / 2 + 1;
+		}
+		return ncpus / 2;
+	} else {
+		return ncpus;
+	}
+}
+
+#endif /* JEMALLOC_INTERNAL_PERCPU_ARENA_H */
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

**Comment by spredolac** [comment]
This would be code for repro

```
/* cc repro.c -o repro -ljemalloc -lpthread; taskset -c 150-153 \
     env MALLOC_CONF=percpu_arena:percpu,narenas:4 ./repro                  */
  #include <pthread.h>
  #include <stdio.h>
  
  static void *
  worker(void *arg) {
        (void)arg;
        for (int i = 0; i < 2000; i++) {
                (void)malloc(4096);
        }
        unsigned ind;
        size_t   sz = sizeof(ind);
        mallctl("thread.arena", &ind, &sz, NULL, 0);
        printf("  thread.arena = %u\n", ind);
        return NULL;
  }

  int
  main(void) {
        pthread_t t[4];
        for (int i = 0; i < 4; i++) {
                pthread_create(&t[i], NULL, worker, NULL);
        }
        for (int i = 0; i < 4; i++) {
                pthread_join(t[i], NULL);
        }

        uint64_t epoch = 1;
        mallctl("epoch", NULL, NULL, &epoch, sizeof(epoch));

        unsigned narenas;
        size_t   allocated, sz;
        sz = sizeof(narenas);
        mallctl("arenas.narenas", &narenas, &sz, NULL, 0);
        sz = sizeof(allocated);
        mallctl("stats.allocated", &allocated, &sz, NULL, 0);
        printf("arenas.narenas = %u, stats.allocated = %zu\n", narenas,
            allocated);
        return 0;
  }

```

File: src/percpu_arena.c [hunk 1]
```diff
@@ -14,4 +16,104 @@ const char *const percpu_arena_mode_names[] = {
...
+	unsigned n = percpu_arena_ngroups_for(mode, ncpus_mapped);
+
+	/*
+	 * CPU ids we know nothing about keep a bounded fallback.  If the
+	 * process later moves to a CPU that was not allowed at boot, the index
+	 * stays valid even though it may not preserve the affinity rank
+	 * mapping.
+	 */
+	for (size_t c = 0; c < map_len; c++) {
+		map[c] = (uint16_t)(c % n);
```
guangli-dai, line 82: In a very rare case, if somehow n is 0, this would crash. So a question here is: what should we do if the query (ncpu or the affinity query) fails, should we just abort or silently transfer to a policy safe to run?

If to abort, release mode must also include the abort pass. If to silently transfer, all system queries should have a safe fallback to avoid %n cases like this one.

src/jemalloc_init.c
guangli-dai: this comment can use an update?

test/unit/percpu_arena_map.c
```diff
@@ -0,0 +1,148 @@
...
+	for (unsigned cpu = 0; cpu < ncpus_; cpu++) {
+		seen[map[cpu]] = true;
+	}
+	for (unsigned g = 0; g < ngroups; g++) {
+		expect_true(seen[g], "Arena %u is unreachable (%s, ncpus %u)", g,
+		    name, ncpus_);
+	}
+}
+
+TEST_BEGIN(test_map_matches_legacy) {
```
guangli-dai: `legacy` is only in terms of the current change, the naming is confusing. I would just call it something like `test_dense_cpu_mappings`. Same for the function naming above.

src/percpu_arena.c
spredolac: Added silent transfer to 1.

include/jemalloc/internal/percpu_arena.h
```diff
@@ -33,13 +34,59 @@ typedef enum {
 #define PERCPU_ARENA_ENABLED(m) ((m) >= percpu_arena_mode_enabled_base)
 #define PERCPU_ARENA_DEFAULT percpu_arena_disabled
 
+/*
+ * Size of the CPU id -> arena index table.  4096 entries because:
+ *   - MALLOCX_ARENA_LIMIT is 4095, so every CPU id that the legacy identity
+ *     mapping could have used as an arena index fits;
```
guangli-dai: Valid arena indices and possible CPU IDs are different. The 0xfff below is a valid reason, but this line is a bit hard to understand.

src/percpu_arena.c
```diff
@@ -14,4 +16,104 @@ const char *const percpu_arena_mode_names[] = {
...
+		return ncpus_mapped;
+	case per_phycpu_arena:
+		/*
+		 * Hyper threads on the same physical CPU share an arena.  An
+		 * odd count likely means a misconfig; round up so that the
+		 * unpaired CPU still has a group of its own.
+		 */
+		return ncpus_mapped > 1
+		    ? ncpus_mapped / 2 + ncpus_mapped % 2
+		    : ncpus_mapped;
```
guangli-dai, line 43: To cover corner cases, can we assert that the final `ngroups < MALLOCX_ARENA_LIMIT`?

include/jemalloc/internal/os/posix/cpu.h
```diff
@@ -24,17 +24,63 @@ os_cpu_ncpus(void) {
...
 	result = sysconf(_SC_NPROCESSORS_ONLN);
 #endif
-	return ((result == -1) ? 1 : (unsigned)result);
+	return result <= 0 ? 1 : (unsigned)result;
+}
+
+/*
+ * Write the ids of the CPUs this thread is allowed to run on into cpus, in
+ * ascending order, and return how many were written.  Returns 0 if the mask is
+ * unavailable or does not fit, in which case cpus is untouched.
```
guangli-dai: Currently if the mask does not fit, `cpus` is partly initialized and returns `max_cpus`. Fix the comment to reflect the truth?

include/jemalloc/internal/percpu_arena.h
```diff
@@ -54,36 +101,31 @@ malloc_getcpu(void) {
...
 	}
 
-	return arena_ind;
+	return percpu_arena_map[cpu];
 }
 
 /* Return the limit of percpu auto arena range, i.e. arenas[0...ind_limit). */
 JEMALLOC_ALWAYS_INLINE unsigned
 percpu_arena_ind_limit(percpu_arena_mode_t mode) {
 	assert(have_percpu_arena && PERCPU_ARENA_ENABLED(mode));
```
guangli-dai: We can remove `mode` and use the global `opt_percpu_arena` for the assertion? Suggesting so because the current usage always passes in `opt_percpu_arena` as `mode` and more importantly, it now does not rely on `mode` to determine the value anymore.

spredolac:
Feedback.  There are some other issues with per-cpu (`last_thd` makes things often wrong) and assumption about ids if `physcpu` option is chosen.  That work is beyond  the scope of this PR and I plan to address it separately.

guangli-dai (approved):

PR #3001, opened 2026-09-15
Fix aligned new oom retry

devnexen:


test/integration/cpp/failing_new.cpp
```diff
@@ -17,7 +17,56 @@ TEST_BEGIN(test_failing_alloc) {
...
+}
+
+TEST_BEGIN(test_failing_aligned_alloc_new_handler) {
+	/*
+	 * The top bit of size_t exceeds SC_LARGE_MAXCLASS on every platform,
+	 * so no allocation can satisfy this alignment and the retry after the
+	 * new_handler must fail too, rather than returning an unaligned
+	 * pointer.
+	 */
+	const std::size_t alignment = (SIZE_MAX >> 1) + 1;
```
guangli-dai: This emits a warning `equested alignment must be 4294967296 bytes or smaller; maximum alignment assumed [-Wbuiltin-assume-aligned-alignment]` for clang. Adding a `volatile` should suppress the warning. Can you add that?

guangli-dai (commented):
The fix (first commit) LGTM. To make the history clean (CI should not fail on any commits), can you squash the two test commits into one?

guangli-dai (approved):

PR #3002, opened 2026-09-15
Fix __int128 detection when cross-compiling

mablr:
Cross-compiling currently defaults `je_cv_int128` to `no`, disabling the optimized extent comparator even when the target compiler supports `__int128`. Replace `AC_RUN_IFELSE` with `AC_LINK_IFELSE`: the probe only initializes a value to zero and returns it, so compilation and linking are sufficient.

Tested locally with an ARM64-to-x86_64 macOS cross-build: detection changes from `no` to `yes`, and both static and shared allocator libraries build successfully.

Existing native CI tests do not cover this detection regression. I also prepared a small cross-configure assertion in an existing macOS CI entry, verified to fail before the fix and pass afterward. Happy to include that CI update if you consider it useful.

AI-assisted


spredolac:
Thanks, I will merge it as-is.

PR #3006, opened 2026-09-18
Remove last-thread shortcut from per-CPU arena selection

spredolac:
The shortcut produced stale arena choices for 10% and 37% of selections in two production workloads. The current CPU lookup took at most 7 cycles in a Linux microbenchmark.

Always read the current CPU and test migration between CPUs.

**Why do this?**

Cost has gone down since `sched_getcpu()` became rseq-backed, but more importantly high mistake rate will muddy the water and lead us to wrong conclusions during experimentation.


test/unit/percpu_arena_resume.c
```diff
@@ -74,8 +74,70 @@ TEST_BEGIN(test_thread_arena_resume_percpu) {
...
+
+	expect_false(os_cpu_set_affinity((int)cpus[0]),
+	    "Could not pin to first CPU");
+	unsigned first_arena = percpu_arena_choose();
+	void *first = mallocx(1024, flags);
+	expect_ptr_not_null(first, "Unexpected mallocx() failure");
+
+	unsigned actual;
+	size_t sz = sizeof(actual);
+	expect_d_eq(mallctl("thread.arena", (void *)&actual, &sz, NULL, 0), 0,
```
guangli-dai, line 90: Is the goal to make sure pointer first come from the previously set arena? Then we should use [`arenas.lookup`](https://jemalloc.net/jemalloc.3.html#arenas.lookup). `thread.arena` calls `percpu_arena_choose()` itself as well.

guangli-dai (commented):
Any local benchmark on how much time we are spending more on CPU query? Just want to have a record for the number if it is not neutral.

spredolac:
Single-socket 72-core Arm Neoverse V2 (3.4 GHz, 1 NUMA node):` sched_getcpu() ≤2.14 ns (~7 cycles) `per call, totaling `~0.0044%` of combined CPU across the two measured (major) services.  This is direct cost. Overall, cost is obviously dominated by service changing shape now when correct arena is chosen.

guangli-dai (approved):

PR #3008, opened 2026-09-21
Optimize fastpath for free, rtree metadata decoding, mallocx and dallocx

GermanAizek:
I have old server motherboard CPU 2xE5-2699v3 (Haswell 2014). Please retest on new and very old cpus.
Bench results aren’t very volatile because I’ve tested many iterations.
image
image


src/jemalloc.c
```diff
@@ -1880,6 +1907,12 @@ JEMALLOC_EXPORT void JEMALLOC_NOTHROW
 je_dallocx(void *ptr, int flags) {
 	LOG("core.dallocx.entry", "ptr: %p, flags: %d", ptr, flags);
 
+	if (likely(flags == 0)) {
+		je_free_impl(ptr);
+		LOG("core.dallocx.exit", "");
+		return;
+	}
+
```
spredolac, line 1915: This block should be below the asserts on lines 1916-1917

Otherwise, it is a change of behavior in debug builds.

src/jemalloc.c
```diff
@@ -1406,16 +1406,45 @@ JEMALLOC_EXPORT JEMALLOC_ALLOCATOR JEMALLOC_RESTRICT_RETURN smallocx_return_t
...
+	sopts.oom_string = "<jemalloc>: Error in mallocx(): out of memory\n";
+
+	dopts.result = &ret;
+	dopts.num_items = 1;
+	dopts.item_size = size;
+
+	imalloc(&sopts, &dopts);
+
+	return ret;
+}
```
spredolac, line 1430: I originally noticed duplication between `malloc_default` and `mallocx_default`, but decided that some inline that handles subtle differences between them (assert_nonempy_alloc, set_errno, and oom_string) would be harder to read.  However, I think what would be neat is not to duplicate this option init in both mallocx body and here as it is exactly the same. So, if you make

```c
JEMALLOC_ALWAYS_INLINE void
mallocx_default_opts_init(static_opts_t *sopts, dynamic_opts_t *dopts, void **result,
    size_t size) {
	static_opts_init(sopts);
	dynamic_opts_init(dopts);

	sopts->assert_nonempty_alloc = true;
	sopts->null_out_result_on_error = true;
	sopts->oom_string = "<jemalloc>: Error in mallocx(): out of memory\n";

	dopts->result = result;
	dopts->num_items = 1;
	dopts->item_size = size;
}
```

`mallocx_default()` becomes clean:
```c
void          *ret;
static_opts_t  sopts;
dynamic_opts_t dopts;

mallocx_default_opts_init(&sopts, &dopts, &ret, size);

imalloc(&sopts, &dopts);
return ret;
```

and 

and in `mallocx` below you have
```c
...
mallocx_default_opts_init(&sopts, &dopts, &ret, size);
dopts.alignment = MALLOCX_ALIGN_GET(flags);
dopts.zero = MALLOCX_ZERO_GET(flags);
dopts.tcache_ind = mallocx_tcache_get(flags);
dopts.arena_ind = mallocx_arena_get(flags);
...
```

include/jemalloc/internal/rtree.h
```diff
@@ -270,6 +285,30 @@ rtree_leaf_elm_read(
 #endif
 }
 
+JEMALLOC_ALWAYS_INLINE rtree_metadata_t
```
spredolac, line 288: The rtree change is a good idea as fast deallocation path can avoid reading the edata ptr, but the metadata decode logic is now duplicated between the normal read path and the metadata-only (the patch now has one copy for the full leaf read and another copy for the metadata-only read). Since this encoding is subtle and has compact/non-compact variants, I would prefer one decode helper used by both paths.
 
The helper should preserve the original input width for each preprocessor branch(the non-compact path should keep decoding from `unsigned`, not widen the helper argument to `uintptr_t`, otherwise the generated code can change even if the semantics are the same.  You can check that assembly is the same before and after for both compact and non-compact cases.


For example:

```c
#ifdef RTREE_LEAF_COMPACT
typedef uintptr_t rtree_leaf_elm_bits_t;
#else
typedef unsigned rtree_leaf_elm_bits_t;
#endif

JEMALLOC_ALWAYS_INLINE rtree_metadata_t
rtree_leaf_elm_metadata_decode(uintptr_t bits) {
	rtree_metadata_t metadata;
#ifdef RTREE_LEAF_COMPACT
	metadata.szind = bits >> LG_VADDR;
#else
	metadata.szind = bits >> (RTREE_LEAF_STATE_SHIFT + RTREE_LEAF_STATE_WIDTH);
#endif
	metadata.slab = (bool)(bits & 1);
	metadata.is_head = (bool)(bits & (1 << 1));
	uintptr_t state_bits = (bits & RTREE_LEAF_STATE_MASK)
	    >> RTREE_LEAF_STATE_SHIFT;
	assert(state_bits <= extent_state_max);
	metadata.state = (extent_state_t)state_bits;
	return metadata;
}
```

Then both `rtree_leaf_elm_bits_decode()` and the metadata-only read can use that one decode routine.

include/jemalloc/internal/rtree.h
spredolac, line 219: This can be then

`contents.metadata = rtree_leaf_elm_metadata_decode(bits);`

spredolac (commented):
Hey thanks for improving this!  There are few things to make a bit neater and I mentioned them inline.

src/jemalloc.c
GermanAizek: Fix it in https://github.com/jemalloc/jemalloc/pull/3008/commits/0dad24539c4da1aeda77d0066a4a52d24b2287bf

src/jemalloc.c
GermanAizek: Fix it in https://github.com/jemalloc/jemalloc/pull/3008/commits/0dad24539c4da1aeda77d0066a4a52d24b2287bf

include/jemalloc/internal/rtree.h
GermanAizek: Fix it in https://github.com/jemalloc/jemalloc/pull/3008/commits/0dad24539c4da1aeda77d0066a4a52d24b2287bf

include/jemalloc/internal/rtree.h
GermanAizek: Fix it in https://github.com/jemalloc/jemalloc/pull/3008/commits/0dad24539c4da1aeda77d0066a4a52d24b2287bf

GermanAizek:
@spredolac thanks for code review. When you have time, check new changes I made—I might have forgotten something.

spredolac:
@GermanAizek , static analysis is failing now on some path(s).  Need to make that green before merging

GermanAizek:
> @GermanAizek , static analysis is failing now on some path(s). Need to make that green before merging

Fix it in https://github.com/jemalloc/jemalloc/pull/3008/commits/930d271e5250e67743e695485a7af70049f83c21 @spredolac

